PR #843, opened 2017-11-29
Back out MP span caching and use new lib that handles it

saolsen:
PR for lib changes is here https://github.com/artsy/datadog-tracer-js/pull/1


package.json
```diff
@@ -46,7 +45,7 @@
     "bluebird": "^3.3.4",
     "body-parser": "^1.15.2",
     "cors": "^2.7.1",
-    "datadog-tracer": "^0.3.4",
+    "datadog-tracer": "github:saolsen/datadog-tracer-js#2a566437913cda63359e5098595c5fc85fe004aa",
```
orta: does this require transpiling to be used?

lib/tracer.js
```diff
@@ -100,7 +78,7 @@ export function middleware(req, res, next) {
   }
 
   res.locals.span = span // eslint-disable-line no-param-reassign
-  res.locals.finishedSpans = [] // eslint-disable-line no-param-reassign
+  // res.locals.finishedSpans = [] // eslint-disable-line no-param-reassign
```
alloy: This can go.

saolsen:
Ok, this should be good to go.

alloy:
👍

PR #849, opened 2017-12-05
[Sale] Add live_url_if_open to Sale query

damassi:
- [x] Update staging with `PREDICTION_ENDPOINT` env var
- [x] Update production with `PREDICTION_ENDPOINT` env var

screen shot 2017-12-04 at 7 23 06 pm

```
{
  sale(id:"eriks-moction") {
    is_live_open
    live_url_if_open
    id
  }
}
```


damassi:
Good call, will make that update 👍

PR #858, opened 2017-12-13
Add the image field to SuggestedGene type

yuki24:
Right now the `SuggestedGene` type is a different type from the `Gene` type. The Follow Gene step in the onboarding flow needs to deal with both the types and adding the same fields will allow Relay to respect required fields like `name` and `image`.

Alternatively, we could just use the `Gene` type anywhere rather than maintaining two similar types. Let us know if this is something we should be doing.


orta:
Took a look at this and paired on converting the `SuggestedGene` to be a `Gene`, as this is going to be how it will work in the future via an actual API. 👍

PR #859, opened 2017-12-13
Remove changelog

orta:
Our weekly updates in slack is basically enough to cover what changed

screen shot 2017-12-13 at 5 06 18 pm

If we want to get serious on this, we can do but for now it's not being used.


alloy:
Danke bitte viele malen 👍

PR #860, opened 2017-12-14
[Library] Add @artsy/express-reloadable

damassi:
Adds [@artsy/express-reloadable](https://github.com/artsy/express-reloadable) to the mix and slightly restructures project around an `/src` folder similar to our other JS-based apps. 

mp-express


damassi:
@orta - I put out a call to arms for someone else to handle this but the only respondee was @erikdstock, but he got pulled onto other work.

index.js
```diff
@@ -137,7 +64,7 @@ xapp.init(
       console.warn("[FEATURE] Enabling schema stitching") // eslint-disable-line
       mergeSchemas().then(startApp)
```
alloy: It looks like this won’t end up using the merged schema, no?

Procfile.dev
```diff
@@ -1,2 +1,2 @@
-web: node_modules/.bin/babel-watch --inspect index.js
+web: node_modules/.bin/babel-node --inspect index.js
```
alloy: \o/

You can also remove the `babel-watch` dependency btw.

index.js
damassi: I'm not entirely sure how merge shemas works internally, but looking closely [#master index.js](https://github.com/artsy/metaphysics/blob/master/index.js) nothing has quite changed as it doesn't accept any parameters. If `enableSchemaStitching` is `true` it should execute, and in the same way as it did before.

damassi: Oh! actually i take that back. I see that this is chained from a promise. Will update.

alloy: Aye, thanks 👍

index.js
```diff
@@ -125,19 +53,25 @@ xapp.on("error", err => {
...
+  secret: GRAVITY_SECRET,
+}
+
+xapp.init(xappConfig, () => {
+  config.GRAVITY_XAPP_TOKEN = xapp.token
+
+  if (enableSchemaStitching) {
+    mergeSchemas()
+      .then(schema => {
+        config.SCHEMA = schema
```
damassi: Due to our need to fetch / merge schemas before we boot MP this had to remain in the root index file and the cleanest way to pass this value over to `src/index` was to set it on config; `mountAndReload` doesn't accept arguments yet but could be something we could update in the future. Open to suggestions as this feels kinda meh...

src/index.js
```diff
@@ -0,0 +1,79 @@
...
+    if (moment.tz.zone(timezone)) {
+      defaultTimezone = timezone
+    }
+
+    const loaders = createLoaders(accessToken, userID, requestID)
+    // Share with e.g. the Convection ApolloLink in mergedSchema.
+    res.locals.dataLoaders = loaders // eslint-disable-line no-param-reassign
+
+    return {
+      schema: config.SCHEMA || localSchema,
```
damassi: And this is where we read it

index.js
alloy: Ah but we do need to reload the schema, because that’s where you would usually make most changes, no? I do think the fetching/merging of schemas could be done after boot, though.

damassi: Ah, if fetching / merging can happen after then all good 👍 - updated.

orta:
When I first run this it works fine, if I go edit a file ( I [changed this line to a string](https://github.com/damassi/metaphysics/blob/master/schema/services/convection.js#L18) and from then on I get `Cannot GET /` as a response. E.g.

```sh
❯ curl 'http://localhost:5000/' -H 'x-user-id: 4ec92ccb56ae88000100038e' -H 'Origin: electron://graphiql-app' -H 'Accept-Encoding: gzip, deflate' -H 'Accept-Language: en-US' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_2) AppleWebKit/537.36 (KHTML, like Gecko) GraphiQL/0.6.1 Chrome/53.0.2785.143 Electron/1.4.13 Safari/537.36' -H 'content-type: application/json' -H 'Accept: */*' -H 'Connection: keep-alive'  --data-binary '{"query":"{\n  services {\n    convection {\n      geminiTemplateKey\n    }\n  }\n}\n","variables":null,"operationName":null}' --compressed
{"data":{"services":{"convection":{"geminiTemplateKey":"convection-staging"}}},"extensions":{"requests":{},"requestID":"3ee6bef0-e0ed-11e7-999c-c9699d465508"}}⏎
```

Now I edit the file:

```sh
❯ curl 'http://localhost:5000/' -H 'x-user-id: 4ec92ccb56ae88000100038e' -H 'Origin: electron://graphiql-app' -H 'Accept-Encoding: gzip, deflate' -H 'Accept-Language: en-US' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_2) AppleWebKit/537.36 (KHTML, like Gecko) GraphiQL/0.6.1 Chrome/53.0.2785.143 Electron/1.4.13 Safari/537.36' -H 'content-type: application/json' -H 'Accept: */*' -H 'Connection: keep-alive'  --data-binary '{"query":"{\n  services {\n    convection {\n      geminiTemplateKey\n    }\n  }\n}\n","variables":null,"operationName":null}' --compressed
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot POST /</pre>
</body>
</html>

~
```

PR #862, opened 2017-12-14
[Sale] Add display_timely_at field

damassi:
Adds a new field to `Sale`, `display_timely` and updates `Sale` and `SaleArtwork` schemas to use dataLoaders rather than raw `gravity` call. 

screen shot 2017-12-14 at 9 21 30 am

```
{
  sales(size: 100 sort:TIMELY_AT_NAME_DESC) {
    id,
    display_timely_at
  }
}
```

### DIFF
--- a/lib/loaders/api/index.js
+++ b/lib/loaders/api/index.js
@@ -10,6 +10,15 @@ import { apiLoaderWithAuthenticationFactory } from "lib/loaders/api/loader_with_
 import { apiLoaderWithoutAuthenticationFactory } from "lib/loaders/api/loader_without_authentication_factory"
 
 export default requestIDs => ({
+  // Unauthenticated loaders
+
+  /**
+   * The Diffusion loaders produced by this factory _will_ cache all responses to memcache.
+   */
+  diffusionLoaderWithoutAuthenticationFactory: apiLoaderWithoutAuthenticationFactory(diffusion, "diffusion", {
+    requestIDs,
+  }),
+
   /**
    * The Gravity loaders produced by this factory _will_ cache all responses to memcache.
    *
@@ -23,6 +32,18 @@ export default requestIDs => ({
    * Do **not** use it for authenticated requests!
    */
   positronLoaderWithoutAuthenticationFactory: apiLoaderWithoutAuthenticationFactory(positron, "positron", {
+    requestIDs,
+  }),
+
+  // Authenticated loaders
+
+  /**
+   * The Convection loaders produced by this factory _will_ cache responses for the duration of query execution but do
+   * **not** cache to memcache.
+   *
+   * Use this for authenticated requests.
+   */
+  convectionLoaderWithAuthenticationFactory: apiLoaderWithAuthenticationFactory(convection, "convection", {
     requestIDs,
   }),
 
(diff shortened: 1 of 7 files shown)


schema/sale/index.js
```diff
@@ -150,6 +150,41 @@ const SaleType = new GraphQLObjectType({
       description: {
         type: GraphQLString,
       },
+      display_timely_at: {
+        type: GraphQLString,
+        resolve: ({ is_live_open, live_start_at, start_at, end_at }) => {
+          if (end_at < moment()) {
```
orta: maybe micro-optimization: is it worth taking `moment()` out and making it `const now = moment()` the functions is used a bunch

lib/loaders/api/index.js
```diff
@@ -10,6 +10,15 @@ import { apiLoaderWithAuthenticationFactory } from "lib/loaders/api/loader_with_
 import { apiLoaderWithoutAuthenticationFactory } from "lib/loaders/api/loader_without_authentication_factory"
 
 export default requestIDs => ({
+  // Unauthenticated loaders
```
damassi: Grouped by auth and unauth and alpha

schema/sale_artwork.js
```diff
@@ -40,10 +39,15 @@ const SaleArtworkType = new GraphQLObjectType({
...
-              key: sale.increment_strategy,
+        resolve: (
+          { minimum_next_bid_cents, sale_id },
+          options,
+          request,
+          { rootValue: { incrementsLoader, saleLoader } }
+        ) => {
+          return saleLoader(sale_id).then(sale => {
+            return incrementsLoader({
+              key: sale.incremenet_strategy,
```
alloy: This doesn’t appear to be spelled correctly.

alloy: So there must also be a missing/faulty test.

alloy:
Like @orta mentioned on an earlier commit (I think), I think it would make sense to call `moment()` as few times as possible and instead assign it to e.g. a `now` const.

schema/sale_artwork.js
damassi: Yeah I'm kind of unsure how this didn't fail, lots of tests around increments. Let me look into it

damassi:
@alloy / @orta - so i looked into the `moment()` thing but was receiving strange results and digging deeper it seems like its due to moment's [mutable data structure](https://www.google.com/search?q=google+moment.js+mutability&oq=google+moment.js+mutability&aqs=chrome..69i57j69i64.3690j0j1&sourceid=chrome&ie=UTF-8). I've run into this before in subtle ways so to keep things clean I'm just going to stick with new constructors for comparisons. `moment().clone()` would work but that kind of defeats the point.

schema/__tests__/sale_artwork.test.js
```diff
@@ -65,69 +62,94 @@ describe("SaleArtwork type", () => {
...
+        saleLoader: () => {
+          return Promise.resolve({
+            minimum_next_bid_cents: 2400000000,
+            increment_strategy: "default",
+          })
+        },
+        incrementsLoader: sale => {
+          return Promise.resolve([
+            {
+              key: sale.increment_strategy,
```
damassi: @alloy - Made the code a little safer in two spots. This brings up a subtle point that with our new ease in testing via simple plug-in functions that resolve values its easy to miss concrete expectations.

schema/sale_artwork.js
```diff
@@ -40,9 +39,18 @@ const SaleArtworkType = new GraphQLObjectType({
...
+        resolve: (
+          { minimum_next_bid_cents, sale_id },
+          options,
+          request,
+          { rootValue: { incrementsLoader, saleLoader } }
+        ) => {
+          return saleLoader(sale_id).then(sale => {
+            if (!sale.increment_strategy) {
+              return Promise.reject("schema/sale_artwork - Missing increment strategy")
+            }
```
damassi: And also tightened it up here

alloy:
Mmm, I see, that’s unfortunate. Ok, this otherwise looks good to me 👍

PR #868, opened 2017-12-19
Move has_message to @GET conversations

ashkan18:
from @GET conversations/:id

### DIFF
--- a/lib/loaders/loaders_with_authentication/impulse.js
+++ b/lib/loaders/loaders_with_authentication/impulse.js
@@ -20,7 +20,7 @@ export default (accessToken, userID, requestIDs) => {
     conversationsLoader: impulseLoader("conversations", {
       from_id: userID,
       from_type: "User",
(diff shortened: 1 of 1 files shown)


alloy:
If possible it would be nice to record a few more details around this change here. I assume we were just sending a param to the wrong endpoint? What does the param do?

orta:
Re: #867 - I applied it to the individual endpoint instead of the index - that flag stops pre-radiation conversations being passed to metaphysics

ashkan18:
Conversations created pre-radiation don't have `initial_message` and also don't have `radiation_conversation_id` in CMS we show them in inbox but when we click on them, since technically there are no messages, we show them the inquiry with a message below it, saying this conversation pre-dates artsy's conversation (or something like that) and they can't take any action on these conversations (no reply, dismiss or anything).

alloy:
Gotcha, thanks!

PR #875, opened 2017-12-26
Add support for wallaby js, and remove flow completely

orta:
The flow stuff is just because it was time. 

This adds support for [wallabyjs](https://wallabyjs.com) which I freakin' love. I use it it both [Danger JS](https://github.com/danger/danger-js/pull/75) and [Peril](https://github.com/danger/peril/pull/178). I have a PR in the works [for Emission](https://github.com/artsy/emission/tree/wallaby-js) but it's safe to say that the Emission toolchain is more complicated and requires a bit more thought.

It's like my extension but on steroids


schema/gene.js
```diff
@@ -173,7 +165,7 @@ export const GeneType = new GraphQLObjectType({
   },
 })
 
-const Gene: GraphQLFieldConfig<GeneType, *> = {
```
damassi: 👍 - i've been wondering about this

damassi:
LGTM 👍

alloy:
LGTM2 👌 

I’ll have to try it soon!

orta:
Yeah, it affects none of our current build tools as-is. The work is all in a confined config file - my rough notes during implementation are https://github.com/artsy/artsy.github.io/issues/406

I probably won't be trying to push it Artsy wide, as you already need to be proficient in Jest to  start feeling comfortable here. 

I think you can try it for 30 minutes for free, or something like that, you can install it as a vscode extension

orta:
This'll need a rebase, will handle a nd merge

PR #876, opened 2017-12-27
Fix build after changing file tree.

alloy:
The Heroku build step was putting the `src/{config,index}.js` files inside `build/src/` but others were stored in `build/{lib,schema}` and the root `index.js` file was completely missing.


alloy:
Going to self-merge, because I think this guy writes stellar code ✨

orta:
:shipit:

PR #880, opened 2017-12-31
Improve the dataloader docs

orta:
Took a stab at improving the dataloader docs with a diagram


docs/dataloaders.md
```diff
@@ -1,25 +1,60 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+See how in the above query we ask for `# GET artist/:id` twice, in the popular artists' name, and then later in the
+artwork's artist - using the dataloader pattern that call would only happen once. This also means that multiple API
+calls for the same artist when looping through artworks be coalesced into the request.
```
alloy: Caching and batching aren’t related to each other, so the “this also means” part of the last sentence isn’t really necessarily true and could confuse a newcomer to dataloader. Batching probably needs its own entry.

docs/dataloaders.md
```diff
@@ -1,25 +1,60 @@
...
-* The `api` object: for example [`/lib/apis/gravity.js`][api_grav] - it is a wrapper around [fetch][fetch] that customizes
-  the request to the service. The params tend to differ depending on the authentication method for that server.
+* The `api` object: for example [`/lib/apis/gravity.js`][api_grav] - it is a wrapper around [fetch][fetch] that
+  customizes the request to the service. The params tend to differ depending on the authentication method for that
+  server.
 
-* The `loader factory` - there are three loader factories: two provide caching, which should be your default and
-  one that provides the same API but for specific one-off problems. They all end up exposing the same API, so your
-  tests should be the same.
+* The `loader factory` - there are three loader factories: two provide caching, which should be your default and one
```
alloy: Re “two provide caching”, there should probably be some clarification given with regards to ‘per request caching’ and ‘request spanning caching with e.g. memcache’.

alloy:
Bump

orta:
OK, gave it another shot

alloy:
Looks great 👍

alloy:
Oops, needs a rebase :/

orta:
👍

PR #895, opened 2018-01-09
@damassi => adds promoted_sale relation to sale model

sweir27:
This is a follow-up to https://github.com/artsy/gravity/pull/11495.


damassi:
LGTM 👍

PR #899, opened 2018-01-11
[FilterArtworks] Omit `medium` filter when `null`.

alloy:
Fixes https://github.com/artsy/collector-experience/issues/893

I’ve got a much larger PR coming in with refactoring and tests.


orta:
👍

PR #902, opened 2018-01-19
[Prettier] Set column width to 80 and format code

damassi:
Updating per RFC [outcome](https://github.com/artsy/potential/issues/127)


src/schema/artist/index.js
```diff
@@ -544,7 +672,12 @@ export const ArtistType = new GraphQLObjectType({
           },
           sort: SaleSorts,
         },
-        resolve: ({ id }, options, _request, { rootValue: { relatedSalesLoader } }) =>
+        resolve: (
+          { id },
+          options,
+          _request,
+          { rootValue: { relatedSalesLoader } }
```
damassi: I agree this kind of formatting is annoying but soon we should sweep over and refactor everything so that the loader is pulled from the first slot rather than the last, allowing us to eliminate all of the additional arguments.

orta:
Yeah, after looking at this PR I'm definitely sure that I prefer 120. A non-trivial amount of these changes make the code read worse IMO. I'm not gonna go through picking them out as we came to a consensus though, and that means something.

damassi:
At the very least if at some point people are like "this is just terrible" its absolutely trivial to update to a new setting.

damassi:
So that the changes aren't so dramatic, I'm going to revert this PR and then submit a new one with updated Prettier config but not reformat entire code-base. That way, going forward, only files that are touched will apply default prettier settings. This will give us time to ease into things rather than making a dramatic change that touches the whole git history.

src/lib/loaders/api/index.js
```diff
@@ -17,33 +17,49 @@ export default requestIDs => ({
...
-  deltaLoaderWithoutAuthenticationFactory: apiLoaderWithoutAuthenticationFactory(delta, "delta", {
-    requestIDs,
-  }),
+  deltaLoaderWithoutAuthenticationFactory: apiLoaderWithoutAuthenticationFactory(
+    delta,
+    "delta",
+    {
+      requestIDs,
+    }
+  ),
```
damassi: @alloy, @orta - Applying Prettier over the codebase made these previous one-liners jump out, as well as where they're used in the code. 

This is debatable, but I think one way to clean up some of our loader use throughout MP is to refactor `apiLoaderWithoutAuthenticationFactory` (and other such factories) into a function that takes a single configuration object, rather than a function that takes multiple unnamed parameters. 

For example:
```javascript
export const apiLoaderWithoutAuthenticationFactory = (api, { 
  apiName,
  globalAPIOptions = {}
}) => {
```
And then later:
```javascript
export default requestIDs => ({
  gravityLoaderWithoutAuthenticationFactory: apiLoaderWithoutAuthenticationFactory(delta, {
    apiName: "delta",
    globalAPIOptions: {
      requestIDs,
    }
  }),
```
Which further down the tree in standard usage would look like:
```javascript
similarGenesLoader: gravityLoader(id => `gene/${id}/similar`, {
  apiOptions: {
    headers: true
  }
})
```
rather than 
```javascript
similarGenesLoader: gravityLoader(
  id => `gene/${id}/similar`,
  {},
  { headers: true }
),
```

At a glance unnamed parameters make API use less informative, and also leaves a lot of empty data around as sometimes the last parameter is needed but the intermediate two aren't. I've found that Relay often suffers from this as well. 

When everything is on one line this isn't as noticable, but now that we're at 80 column width it most definitely reduces readability.

src/lib/loaders/loaders_with_authentication/gravity.js
```diff
@@ -41,10 +65,26 @@ export default (accessToken, userID, requestIDs) => {
...
+    suggestedArtistsLoader: gravityLoader(
+      "me/suggested/artists",
+      {},
+      { headers: true }
+    ),
+    updateCollectorProfileLoader: gravityLoader(
+      "me/collector_profile",
+      {},
+      { method: "PUT" }
+    ),
```
damassi: Another example of what I mean from above. Could be something like 
```javascript
updateCollectorProfileLoader: gravityLoader("me/collector_profile", {
  apiOptions: {
    method: "PUT"
  }
})
```
Which would clear up some ambiguity and make the code look nicer.

PR #907, opened 2018-01-24
WIP: metaphysics on k8s

saolsen:
metaphysics on k8s, WIP DONTMERGE

### DIFF
--- a/hokusai/common.yml
+++ b/hokusai/common.yml
@@ -1,0 +1,7 @@
+---
+version: '2'
+services:
+  metaphysics:
+    build:
+      context: ../
+    env_file: ../.env
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -1,0 +1,51 @@
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: metaphysics-web
+spec:
+  replicas: 2
+  strategy:
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 0
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: metaphysics
+        layer: application
+        component: web
+      name: metaphysics-web
+      namespace: default
+    spec:
+      containers:
+        - name: metaphysics-web
+          envFrom:
+          - configMapRef:
+              name: metaphysics-environment
+          image: 585031190124.dkr.ecr.us-east-1.amazonaws.com/metaphysics:staging
+          imagePullPolicy: Always
+          ports:
+          - containerPort: 3000
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: metaphysics
+    layer: application
+    component: web
+  name: metaphysics-web
+  namespace: default
+spec:
+  ports:
+  - port: 80
+    targetPort: 3000
+    protocol: TCP
+  selector:
+    app: metaphysics
+    layer: application
+    component: web
+  sessionAffinity: None
+  type: LoadBalancer
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+FROM node:8.4.0
+
+# Set up deploy user and working directory
+RUN adduser --disabled-password --gecos '' deploy
+RUN mkdir -p /app
+
+RUN npm install -g yarn@1.2.1
+
+# Set up node_modules
+WORKDIR /tmp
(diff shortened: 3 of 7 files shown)


hokusai/common.yml
```diff
@@ -0,0 +1,7 @@
+---
+version: '2'
+services:
+  metaphysics:
+    build:
+      context: ../
+    env_file: ../.env
```
izakp: This will be included in builds as `hokusai build` `hokusai push` effectively do a `docker-compose build -f ./hokusai/common.yml` to define the build context.  I am wary about local env sneaking into prod containers

hokusai/staging.yml
```diff
@@ -0,0 +1,51 @@
...
+          - configMapRef:
+              name: metaphysics-environment
+          image: 585031190124.dkr.ecr.us-east-1.amazonaws.com/metaphysics:staging
+          imagePullPolicy: Always
+          ports:
+          - containerPort: 3000
+---
+apiVersion: v1
+kind: Service
+metadata:
```
izakp: Suppose we will need our SSL wildcard certs here as well...

hokusai/common.yml
izakp: Use of `.env` files is how the frontend team likes to do their workflow.  Docker compose can obviously support either but I am opposed to having `staging.env` / `production.env` files checked into 1Password as a way to distribute configuration for live environments, which should only live on the server as env vars.   @alloy any ideas for a middle way here?

alloy: Checking those into 1Password is what’s done today, but that also leads to outdated copies and things seemingly breaking out of the blue.

What would be great is if we would be able to replicate production/staging env files easily, e.g. `hokusai config --env staging > .env`

Doable/crazy?

damassi: @jonallured has a nice little ruby script that does something similar: https://github.com/jonallured/copy_env/blob/master/copy_env

jonallured: For context, I made `copy_env` so that I could improve setup scripts:

https://github.com/artsy/bearden/blob/master/bin/setup
https://github.com/artsy/rosalind/blob/master/bin/setup
https://github.com/artsy/radiation/blob/master/bin/setup
https://github.com/artsy/impulse/blob/master/bin/setup
https://github.com/artsy/vibrations/blob/master/bin/setup

Hopefully it helps new devs get setup on Partner Success projects! But I would also be super happy if it could help here too!!

izakp: @alloy I think a little crazy ;)  Given the recent problems with copying live env locally https://github.com/artsy/post_mortems/issues/28 I want to get away from having any live env in local files. cc @jonallured @damassi

izakp: In any case the command would be `hokusai staging env get > .env` bit I would exercise extreme caution.  An alternative is to have the `.env` file stored in S3 somewhere and pulled in via the setup script.  Will iterate on this in our sprint https://artsyproduct.atlassian.net/browse/PLATFORM-18

jonallured: Another idea I had but haven't pursued here is checking these files into an encrypted Keybase repo. My thought was that as a team we could have our collected .env files live there and then when you pull latest, it would update on your file system with a hook. Doesn't really address getting config from staging or whatever, but I thought I'd mention it! 🤗

joeyAghion: Where does the need "to replicate production/staging env files easily" come from?

I'm a big fan of defaulting in-app config to something that's reasonable for development. Ideally this means project check-outs "just work" (though I can see certain work-arounds being necessary for open-source projects). Staging and production are very different and the authoritative source for their config should be the `ConfigMap` stored by the cluster. Why have these copied into 1P--where it will just become out-of-date and untrustworthy--at all?

damassi: The only value that I'm regularly switching out is Gravity's API, to toggle between prod and staging

izakp: @jonallured this is how I have put sensitive information in git repos before https://github.com/AGWA/git-crypt but I like the keybase idea, and then a setup script could pull it in - again I would not pull in staging env locally but have the development .env somewhere we can access programmatically.  I had looked at https://support.1password.com/command-line/ but it doesn't seem that friendly ;)

@joeyAghion yes the files in 1Pass are very much out of date and it's now unclear how some env vars are even used - wherever we store it I am working through what is actually necessary and will address in another PR from my fork

PR #915, opened 2018-01-29
[Messaging] Adds a param for ignoreBlankMessages to the messages connection

orta:
*BUT* does not do anything with it

Unblocks: https://github.com/artsy/emission/pull/968


alloy:
Thanks!

PR #918, opened 2018-01-30
Remove Ignore bulk messages

ashkan18:
# Problem
We still want to return these messages from Radiation/Impulse and need to make sure we render them properly on client side.

# Solution
Remove unused/unimplemented param.

cc: @sarahscott


orta:
👍

PR #923, opened 2018-02-01
Add is_auction to sale_artworks filter query

ashkan18:
# Change
Gravity's `sales_artworks` endpoint has been updated to include `is_auction` flag to filter lots only in auction. We need to update MP to include that param.

# Weirdness?
I _think_ there is only one args `filterSaleArtworksArgs` is defined for accessing both `filters` and `sales_artworks` endpoints in Gravity, I added `is_auction` which currently would only be applicable to `sales_artworks` and can be ignored for `filter` endpoint, hope thats ok.

### DIFF
--- a/src/schema/__tests__/sale_artworks.test.js
+++ b/src/schema/__tests__/sale_artworks.test.js
@@ -10,7 +10,7 @@ describe("Sale Artworks", () => {
     })
   }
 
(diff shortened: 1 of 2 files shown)


orta:
Yeah, looking at https://github.com/artsy/emission/pull/977 is seems like this is correctly re-used across a few places. So I'm 👍

PR #927, opened 2018-02-14
[STI-36] Add is_online_exclusive flag to partner show

ashkan18:
# Problem
We want to be able to distinguish between `online_exclusive` shows and shows with actual locations.

# Solution
Add new `boolean` field to `partnerShow` which based on `location`, `fair_location` and `partner_city` of the PartnerShow returns flag showing if it's `has_location` or not.

### DIFF
--- a/src/schema/__tests__/partner_show.test.js
+++ b/src/schema/__tests__/partner_show.test.js
@@ -35,6 +35,73 @@ describe("PartnerShow type", () => {
     PartnerShow.__ResetDependency__("total")
   })
 
+  it("include false is_online_exclusive flag for shows with location", () => {
+    showData.location = "test location"
+    const query = `
+      {
+        partner_show(id: "new-museum-1-2015-triennial-surround-audience") {
+          is_online_exclusive
+        }
+      }
+    `
+    return runQuery(query).then(data => {
+      expect(data).toEqual({
+        partner_show: {
+          is_online_exclusive: false,
+        },
+      })
+    })
+  })
+  it("include false is_online_exclusive flag for shows with fair_location", () => {
+    showData.fair_location = "test location"
+    const query = `
+      {
+        partner_show(id: "new-museum-1-2015-triennial-surround-audience") {
+          is_online_exclusive
+        }
+      }
+    `
+    return runQuery(query).then(data => {
+      expect(data).toEqual({
+        partner_show: {
+          is_online_exclusive: false,
+        },
+      })
+    })
+  })
+  it("include false is_online_exclusive flag for shows with partner_city", () => {
+    showData.partner_city = "test location"
+    const query = `
+      {
+        partner_show(id: "new-museum-1-2015-triennial-surround-audience") {
+          is_online_exclusive
+        }
+      }
+    `
+    return runQuery(query).then(data => {
+      expect(data).toEqual({
+        partner_show: {
+          is_online_exclusive: false,
+        },
+      })
+    })
+  })
+  it("include true is_online_exclusive flag for shows without any location", () => {
+    const query = `
+      {
+        partner_show(id: "new-museum-1-2015-triennial-surround-audience") {
+          is_online_exclusive
+        }
+      }
+    `
+    return runQuery(query).then(data => {
+      expect(data).toEqual({
+        partner_show: {
+          is_online_exclusive: true,
+        },
+      })
+    })
+  })
   it("doesn't return a show that’s not displayable", () => {
     showData.displayable = false
     const query = `
--- a/src/schema/__tests__/partner_show.test.js
+++ b/src/schema/__tests__/partner_show.test.js
@@ -35,69 +35,69 @@ describe("PartnerShow type", () => {
     PartnerShow.__ResetDependency__("total")
   })
 
-  it("include false is_online_exclusive flag for shows with location", () => {
+  it("include true has_location flag for shows with location", () => {
     showData.location = "test location"
     const query = `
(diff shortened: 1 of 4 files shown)


src/schema/__tests__/partner_show.test.js
```diff
@@ -35,6 +35,73 @@ describe("PartnerShow type", () => {
...
+    return runQuery(query).then(data => {
+      expect(data).toEqual({
+        partner_show: {
+          is_online_exclusive: false,
+        },
+      })
+    })
+  })
+  it("include false is_online_exclusive flag for shows with fair_location", () => {
+    showData.fair_location = "test location"
```
oxaudo: In CMS I opted for checking for Fair in general as well as is_reference boolean + partner_city. I think in our current thinking we only talking about regular partner_shows that can be online only. Like I'm wondering if we somehow will end up with the show that has a fair but fair did not have a location that got stored into fair_location - can this show be considered online_exclusive?

src/schema/partner_show.js
```diff
@@ -182,6 +182,13 @@ const PartnerShowType = new GraphQLObjectType({
         return gravity(`partner_show/${id}/images`, options).then(Image.resolve)
       },
     },
+    is_online_exclusive: {
```
joeyAghion: Is this standard naming style for custom graphql properties? I could do without the `is_` and would expect `onlineExclusive`, but :shrug:.

ashkan18: i kind of followed the pattern from other flags like https://github.com/artsy/metaphysics/pull/927/files#diff-787a1e14fd2b358d2ab504851b2d878bL185 bellow it.

src/schema/partner_show.js
```diff
@@ -182,6 +182,13 @@ const PartnerShowType = new GraphQLObjectType({
...
+      type: GraphQLBoolean,
+      description: "Flag showing if show is online exclusive or not.",
+      resolve: ({ location, fair_location, partner_location }) => {
+        return !location && !fair_location && !partner_location
```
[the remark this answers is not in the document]
ashkan18: im actually not sure where exactly its coming from but it comes in from somewhere! https://github.com/artsy/metaphysics/pull/927/files#diff-787a1e14fd2b358d2ab504851b2d878bL212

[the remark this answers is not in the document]
joeyAghion: But where does _that_ come from? Asking because this might also want to consider the `partner_city` field in the API.

src/schema/partner_show.js
joeyAghion: Related question: We'll probably need to support a new search flag so clients can indicate to elasticsearch that they want to include only shows with real physical locations (e.g., artists' exhibition history). This overlaps with that meaning, so we might want to use the same language. What do you think about something like `has_location: [true|false]` instead of "is online exclusive." The former is closer to the way the data is structured, while the latter is closer to how the shows are presented.

mzikherman:
So, `PartnerShow` is actually deprecated in favor of `Show`, (since there were some breaking changes around reference shows/galaxy partners a while back).

Would you mind terribly moving this new field over to `Show`? I think the show page in Force you are adding this in support of can be trivially updated to query `Show` vs `PartnerShow` (I can totally help you with that as well).

src/schema/partner_show.js
mzikherman: Yea, prefixing boolean fields with `is` _is_ some sort of GraphQL convention (will have to dig up where I saw that).

We are totally inconsistent and haven't settled on 🐫 vs 🐍 case, I would be 👍adding new fields as :camel: ?

src/schema/partner_show.js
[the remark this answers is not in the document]
mzikherman: These are fields being destructured from the first argument to this resolver, which is the object itself.

So, in this case, it is the result of https://github.com/artsy/metaphysics/blob/a5e6c01fa0251279ace2dbb64eeab9f9ef817358/src/schema/partner_show.js#L270-L274

aka the gravity public JSON for a `PartnerShow`.

It doesn't look like `partner_location` is included in that JSON FYI....I'm definitely out of context on this project, but I _believe_ a show's location can be set from the partner location (or something like that...), so maybe we don't need to check `partner_location`? Either way, I don't actually think `partner_location` is what you want, as it won't actually exist in the returned JSON for a show (and thus not be present here).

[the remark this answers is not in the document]
joeyAghion: Thanks. Looks like `PartnerShow#partner_location` becomes `location` in the JSON, which leads to the question: what is the _other_ `partner_location` above?

src/schema/partner_show.js
ashkan18: i kind of wanted to have `is_online_exclusive` as its own flag to abstract the logic out of clients, so lets say if tomorrow we decided `online_exclusive` means no location and no date, we'd need to only change the logic in here. 
Good point on the filtering side though, i think that would be an argument probably with the same name?

src/schema/partner_show.js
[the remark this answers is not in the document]
ashkan18: yes, just double checked with Gravity' `partner_location` becomes `location` in the json and there is no `partner_location` I added `partner_city`. Adding check for `partner_city` too.

ashkan18:
oooh didn't know about that @mzikherman I pretty much copy pasted that same flag/logic into `Show` as well, i kept the code in `partnerShow` in case we need to have them in synch in meantime.

PR #930, opened 2018-02-16
[test] Fix unhandled promise and the assertion.

alloy:
* Need to await
* Had to use this way to assert, as suggested here:
  https://github.com/facebook/jest/issues/3601#issuecomment-333094824


damassi:
👍

PR #936, opened 2018-02-19
[dev] Fixes redis line of setup script

sarahscott:
Small fix so the setup script doesn't get stuck checking for a redis installation.


izakp:
Also a note that `yarn build:lib` that calls the Babel compiler isn't anywhere in the setup script and has bitten me before.  It's unclear when this needs to be called.

orta:
Running the babel compiler during the user setup script isn't needed, as during dev the JS is compiled at runtime, the compilation process is only needed for distribution.

PR #939, opened 2018-02-21
@orta => Add ids to sales

eessex:
First Metaphysics PR 🎈 

Following up on pairing yesterday -- adds ids type to sales schema, and updates resolver to use `id` rather than `ids` (to stay in accordance with Gravity).

One thing I noticed is that this file is not yet tested, and many of the plural endpoints are not.  Tried a few of the testing models laid out in the schema folder but kept getting the error `Error: getaddrinfo ENOTFOUND api.artsy.test api.artsy.test:443` -- am I just missing an env var, or maybe this is another pairing opportunity?

### DIFF
--- a/src/schema/sales.js
+++ b/src/schema/sales.js
@@ -1,8 +1,9 @@
 // @ts-check
+import { clone } from "lodash"
 import gravity from "lib/loaders/legacy/gravity"
 import Sale from "./sale/index"
 import SaleSorts from "./sale/sorts"
-import { GraphQLList, GraphQLInt, GraphQLBoolean } from "graphql"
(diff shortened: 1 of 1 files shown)


orta:
Yep this looks 👍

orta:
The testing issue is likely because it's not using a gravity dataloader, I'm happy to merge this PR, and then we can pair on switching it over separately?

orta:
( [docs on dataloaders](https://github.com/artsy/metaphysics/blob/master/docs/dataloaders.md) )

eessex:
Docs on dataloaders are useful, probably a good idea to implement these when we pair!

PR #940, opened 2018-02-21
[STI-15] Filter out shows without location in exhibition highlights query

ashkan18:
# Problem
We want to filter out online exclusive shows from artist's exhibition highlights.
https://artsyproduct.atlassian.net/browse/STI-15

# Solution
In `exhibition_highlights` resolver, add `has_location: true` to make sure we filter shows without location.

# Depends on STI-39
Gravity change to support filter by `has_location`.
https://artsyproduct.atlassian.net/browse/STI-39

### DIFF
--- a/src/schema/artist/__tests__/index.test.js
+++ b/src/schema/artist/__tests__/index.test.js
@@ -143,7 +143,7 @@ describe("Artist type", () => {
     })
   })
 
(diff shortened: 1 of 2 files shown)


ashkan18:
Gravity dependency is deployed and currently running related migration, we should be 👍 to go here

PR #942, opened 2018-02-23
@alloy => Can return artists by array of ids

eessex:
Adds `ids` param to `artists` schema, to return artists matching an array of passed ids.

### DIFF
--- a/src/schema/__tests__/artists.test.js
+++ b/src/schema/__tests__/artists.test.js
@@ -26,4 +26,27 @@ describe("Artists", () => {
     const { artists } = await runQuery(query, { artistsLoader })
     expect(artists).toEqual([{ name: "Han Myung-Ok" }])
   })
+
+  it("returns a list of artists matching array of ids", async () => {
+    const artistsLoader = ({ ids }) => {
+      if (ids) {
+        return Promise.resolve([
+          {
+            _id: "52c721e5b202a3edf1000072",
+            name: "Han Myung-Ok",
+          },
+        ])
+      }
+      throw new Error("Unexpected invocation")
+    }
+    const query = gql`
+      {
+        artists(ids: ["52c721e5b202a3edf1000072"]) {
+          _id
+        }
+      }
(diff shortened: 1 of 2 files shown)


alloy:
Implementation looks good. The test could be a little improved, but otherwise 👍

src/schema/__tests__/artists.test.js
```diff
@@ -26,4 +26,26 @@ describe("Artists", () => {
     const { artists } = await runQuery(query, { artistsLoader })
     expect(artists).toEqual([{ name: "Han Myung-Ok" }])
   })
+
+  it("returns a list of artists matching array of ids", async () => {
+    const artistsLoader = ({ ids }) => {
+      if (ids) {
+        return Promise.resolve([
+          {
+            name: "Han Myung-Ok",
```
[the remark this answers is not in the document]
eessex: Makes sense to me!  Updated now.

src/schema/__tests__/artists.test.js
```diff
@@ -26,4 +26,27 @@ describe("Artists", () => {
...
+      if (ids) {
+        return Promise.resolve([
+          {
+            _id: "52c721e5b202a3edf1000072",
+            name: "Han Myung-Ok",
+          },
+        ])
+      }
+      throw new Error("Unexpected invocation")
+    }
```
alloy: I was actually thinking of _using_ the passed in `ids`, so something like:

```js
const artistsLoader = ({ ids }) => Promise.resolve(ids.map(_id => ({ _id })))
```

That way you can rest assured that the data you put in influences the data that comes out.

eessex: Oh I see yes-- thanks for clarifying this test pattern

eessex: This is updated now, do you have other changes before this is ready to merge?

alloy:
Sorry for the delay, looking good!

PR #948, opened 2018-02-27
@damassi => compress responses

sweir27:
This PR adds gzip compression via the handy [compression](https://github.com/expressjs/compression) package. Part of @yuki24 and I's sprint to improve front-end performance.

We have been using webpagetest.org to benchmark performance of the major force pages (artsy.net, artsy.net/collect, and artsy.net/artist/....). It included this recommendation for how we can improve:
screen shot 2018-02-27 at 12 03 51 pm


alloy:
Sweet!

Please also test with Eigen 🙏🏽

PR #950, opened 2018-02-27
@mzikherman => Search fairs by [:ids]

eessex:
Very similar to https://github.com/artsy/metaphysics/pull/949, adds `:ids` param to Fairs endpoint.


eessex:
Gravity is deployed and this is functional now

PR #955, opened 2018-02-28
[AC-4] Add attribution_class related fields to artwork

oxaudo:
https://artsyproduct.atlassian.net/browse/AC-4 requests us to display classification on artwork page. We have multiple strings that need to be associated with each attribution_class value.

Seems like so far the convention was to just hardcode relevant string into files here. In this case all of this looks pretty ugly and repetitive. Definitely open to refactor suggestions here. But so far just gets the values across...


src/schema/artwork/index.js
```diff
@@ -122,6 +122,88 @@ export const artworkFields = () => {
     category: {
       type: GraphQLString,
     },
+    attribution_class: {
+      type: GraphQLString,
+      description: "Attribution class as classification name",
+      resolve: artwork => {
```
mzikherman: #minor but the signature can be written as `resolve: ({ attribution_class }) => {` and you can drop the `artwork.` part in the function. Up to you!

src/schema/artwork/index.js
```diff
@@ -122,6 +122,88 @@ export const artworkFields = () => {
...
+        if (artwork.attribution_class) {
+          return capitalizeFirstCharacter(artwork.attribution_class)
+        }
+      },
+    },
+    attribution_class_hover: {
```
mzikherman: I'm thinking the naming here should maybe drop the `_hover` part (since that's a function of the current UI), and maybe be `attribution_class_info` or something?

mzikherman:
One thought looking at the proposed schema, is that we don't have to be limited to mimic-ing the Gravity API response (ie- flat schema/long-ish and potentially repetitive field names).

How about something like:

```
artwork(id: "andy-warhol-skull") {
  attribution {
    class
    info
    description(type: "short")
    description(type: "long")
  }
}
```

You see where I'm going with this, we can be flexible with the schema to group related items together, etc. - take a look and see what you think!

src/schema/artwork/index.js
```diff
@@ -122,6 +122,88 @@ export const artworkFields = () => {
...
+      type: GraphQLString,
+      description: "Descriptive sentance used as companion for attribution class name display",
+      resolve: artwork => {
+        switch (artwork.attribution_class) {
+          case "editioned multiple": return "High quantity editions, without direct artist involvement."
```
mzikherman: It might be cleaner to have a JSON file here with the info, similar to the spec above.

We already do that for a few things, here's an example file: https://github.com/artsy/metaphysics/blob/master/src/lib/auction_records_trusted.json

Then you can just `require(...)` it.

src/schema/artwork/__tests__/artwork.test.js
```diff
@@ -1,4 +1,5 @@
 import { assign } from "lodash"
+import _ from "lodash"
```
anandaroop: Could be combined with line above — `import { assign, keys }` …

src/schema/artwork/index.js
```diff
@@ -122,6 +122,88 @@ export const artworkFields = () => {
...
+      description: "Attribution class as classification name",
+      resolve: artwork => {
+        if (artwork.attribution_class) {
+          return capitalizeFirstCharacter(artwork.attribution_class)
+        }
+      },
+    },
+    attribution_class_hover: {
+      type: GraphQLString,
+      description: "Descriptive sentance used as companion for attribution class name display",
```
anandaroop: #minor "sentance" &rarr; "sentence", but also maybe it's just a "phrase" not a full sentence

src/schema/artwork/index.js
anandaroop: Agree, maybe `hint` so that it's not tied to a specific kind of device?

oxaudo:
Thanks @mzikherman for recs here.

@anandaroop and myself thought about having this attribution as one object here as opposed to combination of different attribution related fields as you've suggested above. The issue with this (maybe) is that in our case different attributes will be used on different pages. Artwork page will need short and long description and bricks on the rest of the pages will use name and hover. My thinking was that it might be better to only get the attributes needed if they defined as separate attributes.
Not sure it's enough benefit rather than keeping all attribution related stuff in one obj.

src/schema/artwork/index.js
oxaudo: 👍

anandaroop:
Yeah @oxaudo I think we were not thinking "graphql-y" enough when we had that earlier chat. The beauty of it is you only request the fields you need for your client's response, anything from 

```
artwork(id: "cow-55") {
  attribution_class { 
    name
  }
}
```

all the way up to the full enchilada that Matt wrote out above.

oxaudo:
@mzikherman, @anandaroop - Attempted to refactor this addressing your comments above. Probably still off track in places, but bear with me:) It's my first pr here...

src/schema/types/attribution_class.js
```diff
@@ -0,0 +1,33 @@
+import {
```
mzikherman: We're a bit inconsistent on where we put these, would you mind moving to `src/schema/artwork` folder? That way attribution and all those other 'partials' are colocated with the `index` which uses all of them.

src/schema/types/attribution_class.js
```diff
@@ -0,0 +1,33 @@
...
+  GraphQLObjectType,
+  GraphQLString,
+  GraphQLNonNull,
+} from "graphql"
+
+const AttributionClass = new GraphQLObjectType({
+  name: "AttributionClass",
+  description: "Collection of fileds that describe attribution class",
```
mzikherman: Typo: **fields**

src/schema/artwork/__tests__/artwork.test.js
```diff
@@ -937,4 +938,96 @@ describe("Artwork type", () => {
...
+                name,
+              }
+            }
+          }
+        `
+        return runQuery(query, rootValue).then(data => {
+          expect(data).toEqual({
+            artwork: {
+              attribution_class: {
+                name: attributionClasses[attributionClass].name,
```
mzikherman: In general I think it's preferable to have a more explicit assertion in a spec, as inadvertent `nil` or `undefined` values creeping in can frequently cause tests to still pass (in this style).

Also, I love how each class is tested, but that feels a bit much, maybe we can just have one explicit test for a specific attribution type?

oxaudo:
@mzikherman, @anandaroop - Ok, third time is a charm(hopefully)

src/lib/attribution_classes.js
```diff
@@ -0,0 +1,66 @@
...
+    name: "Reproduction",
+    info: "Reproduction authorized by artist’s studio or estate.",
+    short_description: "This work is a reproduction.",
+    long_description: [
+      "Reproduction of an original work authorized by artist’s studio or estate.",
+      "The artist was not directly involved in production.",
+    ].join(" "),
+  },
+  "editioned multiple": {
+    id: "editioned multiple",
```
anandaroop: This id has a space but the one on [L10](https://github.com/artsy/metaphysics/pull/955/files#diff-bd6d53f9c5aff2f8fd7f40cf98028d60R10) has an underscore. Intentional?

Do these need to have parity [with these](https://github.com/artsy/gravity/blob/master/app/models/domain/attribution.rb#L8-L14)?

(Same observation for some other id's in this file)

src/schema/artwork/attribution_class.js
```diff
@@ -0,0 +1,33 @@
...
+const AttributionClass = new GraphQLObjectType({
+  name: "AttributionClass",
+  description: "Collection of fields that describe attribution class",
+  fields: {
+    id: {
+      type: new GraphQLNonNull(GraphQLString),
+    },
+    name: {
+      type: GraphQLString,
+      description: "Shortest form of arttribution class display",
```
anandaroop: arttribution &rarr; attribution

src/schema/artwork/attribution_class.js
```diff
@@ -0,0 +1,33 @@
...
+    },
+    info: {
+      type: GraphQLString,
+      description: "Descriptive phrase used as companion for attribution class name display",
+    },
+    short_description: {
+      type: GraphQLString,
+      description: "Longer version of arttribution class display",
```
anandaroop: Same

src/schema/artwork/index.js
```diff
@@ -29,6 +29,9 @@ import {
   GraphQLList,
   GraphQLInt,
 } from "graphql"
+import AttributionClass from "schema/artwork/attribution_class"
+// Mapping of attribution_class ids to AttributionClass values
+import attributionClasses from "../../lib/attribution_classes.js"
```
anandaroop: Not sure what the right pattern for filenames is here — I know there's [active talk about 🐍 vs 🐫 right now](https://artsy.slack.com/archives/C1HH3KNJG/p1519931489000527).

oxaudo: Oh... I did not even realize that there are both versions here in the project:( Just looked at the one above and copied the convention....

src/lib/attribution_classes.js
oxaudo: That was not intentional. Technically we do not even need this Id here, but I though it might be nice to get is as part of the object as something to refer back to in gravity if we need to. Corrected.

src/schema/artwork/index.js
oxaudo: Though reading the discussion you linked above those file should be camelCased, no? Did I interpret it correctly?

mzikherman: Yea 👍 on 🐫 !!

mzikherman:
Looks great!! 😎 

Aw yea, we got another committer in the house!

giphy 1

anandaroop:
Congrats on the Metaphysics PR - nice work!

PR #975, opened 2018-03-16
@mzikherman => Users endpoint/schema

eessex:
- Adds an endpoint for Users to query by [:ids]
- Adds gravityLoader for Users

Wasn't completely sure that I needed to use the authenticated loader here, but it seemed like the right idea for user info.  We are passing in the user access token in positron's fetch, so I'm assuming that will be enough to get what we need.  (When I run the query on the local version, I get `"Not authorized"` error.

### DIFF
--- a/src/lib/loaders/loaders_with_authentication/gravity.js
+++ b/src/lib/loaders/loaders_with_authentication/gravity.js
@@ -128,9 +128,6 @@ export default (accessToken, userID, opts) => {
       {},
       { method: "DELETE" }
     ),
-    usersLoader: gravityLoader(
-      "users",
-      {}
-    ),
+    usersLoader: gravityLoader("users"),
   }
 }
--- a/src/lib/loaders/loaders_with_authentication/gravity.js
+++ b/src/lib/loaders/loaders_with_authentication/gravity.js
@@ -130,8 +130,7 @@ export default (accessToken, userID, opts) => {
     ),
     usersLoader: gravityLoader(
       "users",
-      {},
-      { headers: true }
+      {}
     ),
   }
 }
--- a/src/lib/loaders/loaders_with_authentication/gravity.js
+++ b/src/lib/loaders/loaders_with_authentication/gravity.js
@@ -128,5 +128,10 @@ export default (accessToken, userID, opts) => {
       {},
       { method: "DELETE" }
     ),
+    usersLoader: gravityLoader(
+      "users",
+      {},
+      { headers: true }
+    ),
   }
 }
--- a/src/lib/loaders/loaders_without_authentication/gravity.js
+++ b/src/lib/loaders/loaders_without_authentication/gravity.js
@@ -98,6 +98,5 @@ export default opts => {
     ),
     tagLoader: gravityLoader(id => `tag/${id}`),
     trendingArtistsLoader: gravityLoader("artists/trending"),
-    usersLoader: gravityLoader("users"),
   }
 }
--- a/src/lib/loaders/loaders_without_authentication/gravity.js
+++ b/src/lib/loaders/loaders_without_authentication/gravity.js
@@ -98,5 +98,6 @@ export default opts => {
     ),
     tagLoader: gravityLoader(id => `tag/${id}`),
     trendingArtistsLoader: gravityLoader("artists/trending"),
+    usersLoader: gravityLoader("users"),
   }
 }
--- a/src/schema/__tests__/users.test.js
+++ b/src/schema/__tests__/users.test.js
@@ -1,0 +1,24 @@
+import { runAuthenticatedQuery } from "test/utils"
+import gql from "test/gql"
+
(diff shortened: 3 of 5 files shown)


src/lib/loaders/loaders_without_authentication/gravity.js
```diff
@@ -98,5 +98,6 @@ export default opts => {
     ),
     tagLoader: gravityLoader(id => `tag/${id}`),
     trendingArtistsLoader: gravityLoader("artists/trending"),
+    usersLoader: gravityLoader("users"),
```
mzikherman: Nah, you can get rid of this one, we just want the authenticated one (which only exists in a GraphQL request if you pass along the `X-Access-Token`with your request).

src/lib/loaders/loaders_with_authentication/gravity.js
```diff
@@ -128,5 +128,10 @@ export default (accessToken, userID, opts) => {
       {},
       { method: "DELETE" }
     ),
+    usersLoader: gravityLoader(
+      "users",
+      {},
+      { headers: true }
```
mzikherman: This is useful when you want to resolve with body and headers. You might want to resolve this way if you are paginating and need total count (accessible in the header, for example). Otherwise you may not need this.

src/schema/users.js
```diff
@@ -0,0 +1,16 @@
...
+const Users = {
+  type: new GraphQLList(UserType),
+  description: "A list of Users",
+  args: {
+    ids: {
+      type: new GraphQLList(GraphQLString),
+    },
+  },
+  resolve: (root, options, request, { rootValue: { usersLoader } }) =>
+    usersLoader(options),
```
mzikherman: In the context of an authenticated request, this would resolve with an object containing `body` and `headers`.

So I _think_ (the way you have it setup with `headers: true`), is you'd want `usersLoader(options).then(({ body }) => body)`

src/schema/__tests__/users.test.js
```diff
@@ -0,0 +1,24 @@
+import { runAuthenticatedQuery } from "test/utils"
+import gql from "test/gql"
+
+describe("Users", () => {
+  it("returns a list of users matching array of ids", async () => {
+    const usersLoader = ({ ids }) => {
+      if (ids) {
+        return Promise.resolve(
```
mzikherman: This stub should be `body: ids.map(...`.

src/schema/users.js
eessex: Ah ok I don't actually need the headers and have removed them.  I wonder how I can test this locally, is there a tactic to pass my x-access-token to the browser?

xtina-starr: @eessex not sure if you have [graphiQL](https://github.com/graphql/graphiql) but it allows you to set the `x-access-token` for testing locally.

mzikherman: Yea exactly using Chrome to point to GraphQL may not be sufficient b/c I'm not sure how you can pass in headers.

definitely recommend using GraphiQL! It lets you easily pass in your header (which you can access by logging in to staging.artsy.net, and in the JS console looking at `sd.CURRENT_USER.accessToken`).

src/lib/loaders/loaders_with_authentication/gravity.js
```diff
@@ -128,5 +128,9 @@ export default (accessToken, userID, opts) => {
       {},
       { method: "DELETE" }
     ),
+    usersLoader: gravityLoader(
+      "users",
+      {}
```
mzikherman: Ah, in this case you don't even need this `{}`. It was just a 'placeholder' for if you needed to pass an option in to the last param.

src/schema/users.js
xtina-starr: Yeah that's probably the source of your ` "Not authorized"` error. Let me know if you're still having trouble with that. Happy to walk you through setting it up.

eessex: Yeah I am using graphiql and do have the access token, but I don't know where to set this data -- don't see a place on the UI, do I need a browser plugin or something?

xtina-starr: In the right hand corner there is an edit HTTP Headers button.

screen shot 2018-03-16 at 12 47 34 pm

And then you can edit the value.

drawing

src/schema/__tests__/users.test.js
```diff
@@ -0,0 +1,24 @@
+import { runAuthenticatedQuery } from "test/utils"
+import gql from "test/gql"
+
+describe("Users", () => {
+  it("returns a list of users matching array of ids", async () => {
+    const usersLoader = ({ ids }) => {
```
mzikherman: `({ id })` to match the new arg signature.

mzikherman:
Looks great! Tiny ❌ b/c of the new `id` arg.

xtina-starr:
👍

PR #984, opened 2018-03-28
#migration require pod execution on foreground nodes and autoscale deployment

izakp:
Use the [Horizontal Pod Autoscaler](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) to scale metaphysics-web in response to load.

Also use the "hard" `requiredDuringSchedulingIgnoredDuringExecution` `nodeAffinity` selector to ensure these pods are always scheduled on `tier=foreground` nodes, and if resource is unavailable, prompt the [cluster autoscaler](https://github.com/artsy/substance/pull/77) to scale up this node group.

## Migration

cc @alloy @saolsen

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -4,7 +4,6 @@ kind: Deployment
 metadata:
   name: metaphysics-web
 spec:
-  replicas: 2
   strategy:
     rollingUpdate:
       maxSurge: 1
@@ -29,6 +28,8 @@ spec:
         ports:
         - containerPort: 3000
         resources:
+          requests:
+            cpu: "1"
           limits:
             memory: "768Mi"
       affinity:
@@ -41,6 +42,23 @@ spec:
                 operator: In
                 values:
                 - foreground
+                - foreground
+
+---
+apiVersion: autoscaling/v1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: metaphysics-web
+  namespace: default
+spec:
+  scaleTargetRef:
+    apiVersion: extensions/v1beta1
+    kind: Deployment
+    name: metaphysics-web
+  minReplicas: 2
+  maxReplicas: 5
+  targetCPUUtilizationPercentage: 80
+
 ---
 apiVersion: v1
 kind: Service
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -29,7 +29,7 @@ spec:
         - containerPort: 3000
         resources:
           requests:
-            cpu: "1"
+            cpu: "0.7"
           limits:
             memory: "768Mi"
       affinity:
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -34,14 +34,12 @@ spec:
             memory: "768Mi"
       affinity:
         nodeAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - weight: 1
-            preference:
-              matchExpressions:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
               - key: tier
                 operator: In
                 values:
-                - foreground
                 - foreground
 
 ---
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -4,7 +4,6 @@ kind: Deployment
 metadata:
   name: metaphysics-web
 spec:
-  replicas: 1
   strategy:
     rollingUpdate:
       maxSurge: 1
@@ -29,6 +28,8 @@ spec:
         ports:
         - containerPort: 3000
         resources:
+          requests:
+            cpu: "0.5"
           limits:
             memory: "512Mi"
       affinity:
@@ -41,6 +42,22 @@ spec:
                 operator: In
                 values:
                 - foreground
+
+---
+apiVersion: autoscaling/v1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: metaphysics-web
+  namespace: default
+spec:
+  scaleTargetRef:
+    apiVersion: extensions/v1beta1
+    kind: Deployment
+    name: metaphysics-web
+  minReplicas: 1
+  maxReplicas: 3
+  targetCPUUtilizationPercentage: 80
+
 ---
 apiVersion: v1
 kind: Service
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -34,10 +34,9 @@ spec:
             memory: "512Mi"
       affinity:
         nodeAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - weight: 1
-            preference:
-              matchExpressions:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
               - key: tier
                 operator: In
                 values:


alloy:
Does this only scale by CPU load or also memory pressure?

izakp:
@alloy it scales by amount of CPU used / CPU requested - if this is over `targetCPUUtilizationPercentage` it scales out

izakp:
As I was load-testing:

```
isacpetruzzi@Artsy-Isac metaphysics (master) $ kubectl get horizontalpodautoscalers
NAME                 REFERENCE                       TARGETS    MINPODS   MAXPODS   REPLICAS   AGE
metaphysics-web      Deployment/metaphysics-web      7% / 80%   1         3         1          19h
isacpetruzzi@Artsy-Isac metaphysics (master) $ kubectl describe hpa metaphysics-web
Name:							metaphysics-web
Namespace:						default
Labels:							<none>
Annotations:						kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"autoscaling/v1","kind":"HorizontalPodAutoscaler","metadata":{"annotations":{},"name":"metaphysics-web","namespace":"default"},"spec":{"m...
CreationTimestamp:					Tue, 27 Mar 2018 18:19:00 +0200
Reference:						Deployment/metaphysics-web
Metrics:						( current / target )
  resource cpu on pods  (as a percentage of request):	7% (7m) / 80%
Min replicas:						1
Max replicas:						3
Events:
  FirstSeen	LastSeen	Count	From				SubObjectPath	Type		Reason			Message
  ---------	--------	-----	----				-------------	--------	------			-------
  19h		46m		35	horizontal-pod-autoscaler			Normal		SuccessfulRescale	New size: 2; reason: cpu resource utilization (percentage of request) above target
  19h		40m		37	horizontal-pod-autoscaler			Normal		SuccessfulRescale	New size: 1; reason: All metrics below target
```

hokusai/production.yml
```diff
@@ -29,18 +28,35 @@ spec:
         ports:
         - containerPort: 3000
         resources:
+          requests:
+            cpu: "1"
```
izakp: Just for some more context, this means "allocate 1 core to a Metaphysics pod". See https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-cpu

Compared to the running pods on production, it seems each metaphysics pod is using ~0.6 CPU so this gives us a lot of overhead.

screen shot 2018-03-28 at 5 03 51 pm

However, we could set this closer to what we actually see as "normal" operation - i.e. set it to `0.6` and scale horizontally to spread load across more replicas.

izakp:
Set CPU requests to `0.7` (cores) for production - matching what we now see running (see outdated comment above).

alloy:
Like @mzikherman mentioned, I’d also expect memory to be more important as an indicator to scale, as it needs to hold all the gathered data from backend services in memory until it has been able to complete a GraphQL request.

izakp:
@alloy that's not generally the way to go about scaling deployments, on k8s or elsewhere as memory is generally not correlated with request throughput - CPU / load averages are (k8s does not support scaling pods by memory).  Scaling on memory is a bit of an anti-pattern in my experience.  So despite  the particularities of GraphQL resolvers, I don't think it necessarily means memory footprint _increases linearly_ with request throughput.  In my testing, load does increase linearly with throughput as is generally the behavior you would expect for any application.

Concerning memory, we just need to make sure it has enough headroom.  The trick is to set the parameters correctly which comes with observing the behavior of the application in production and tuning the requests / limits and scaling thresholds so we know how much request load a given container is expected to handle within given memory limits.

izakp:
@alloy that said, there is this design proposal for the Horizontal Pod Autoscaler v2 https://github.com/kubernetes/community/blob/master/contributors/design-proposals/autoscaling/hpa-v2.md

izakp:
Also AFAICT Kubernetes' [CPU utilization metrics](https://github.com/kubernetes/heapster/blob/master/docs/storage-schema.md) include Docker's CPU.system [metrics](https://www.datadoghq.com/blog/how-to-monitor-docker-resource-metrics/#cpu), so this CPU utilization calculation is inclusive of I/O wait

izakp:
Merged and applied.  I can see Metaphysics pods scaling out in response to reflection scraping the site.

alloy:
I’d like to hear more about _why_ it would be an anti-pattern, rather than just accepting that at face value. Memory has definitely been _the_ issue with MP and it seems to me a very reasonable bottleneck in the case of a facade service like MP.

izakp:
@alloy I call it an anti-pattern because this is a API deployment behind a load balancer, and as such the goal of scaling out (adding more replicas to a deployment) is to be able to handle an increase in request throughput.  In all other deployment strategies I have seen / worked with, API application instances are always scaled as a function of load.  Here's why:

Consider the request path:
- A request comes into the load balancer.  The load balancer applies [certain methods](http://nginx.org/en/docs/http/load_balancing.html#nginx_load_balancing_methods) (similar in HAProxy / ELB) to try and keep the load spread evenly across all servers /  backend application instances
- The request is proxied to a given application instance
- The request is processed and in doing so it produces load on the server

If a server is overloaded / cannot complete a request, the load balancer [removes the server](http://nginx.org/en/docs/http/load_balancing.html#nginx_load_balancing_health_checks) from its pool and waits until it is healthy again to again proxy requests to it.  That said, the backend server should respond with appropriate HTTP status codes that indicate if it is under load and unable to process the request.

To summarize, this is the general deployment pattern for a web-facing API.  And it is designed to spread load over a group of backends in response to request throughput.  As you can see, memory doesn't come into play here - it is assumed each backend server has at least enough memory to not crash.

As a counterexample, consider this pod-memory [autoscaler script](https://github.com/powerupcloud/kubernetes-1/blob/master/memory-based-autoscaling.sh) - it's particular to a Java deployment and discovering the "Total Heap Capacity Allocated" so whereas the JVM will allocate a certain amount of heap memory and make use of that the Node VM behaves quite a bit differently, and can grow unbounded.  I would be at a loss applying this pattern to Metaphysics because whereas _all requests_ produce load on the servers, and so follow the general pattern described above (inclusive of `iowait`) only some requests (and likely the 95th percentile requests taking > 1 second to complete) need to hold a lot of data in-memory waiting on multiple backend services to complete their request.  Thus we would be scaling on only the outliers of the bulk of requests.  In any case, this scaling pattern would be particular to Metaphysics' implementation of its GraphQL API and schema - a very tight coupling between application behavior and orchestration which I would be wary of - the schema might change and we would have to re-instrument our memory profiling and change the scaling pattern in response to schema changes.

Whew!  Okay that was a lot but I hope it helps!

alloy:
Thanks a lot for taking the time to write this up! 🙏 

> What we were seeing before was a fixed number of application instances trying to handle all load as particular to NodeJS behavior as it accepts all requests - puts them on its stack and memory grows unbounded 

Right, fixed amount of dynos was definitely not a good environment for MP and its stack, so being able to even respond to load is an infinite amount of times better than hosting on Heroku 👍 

> think of load based scaling as "horizontal" - i.e. adding more application instances in response to load. In contrast, "vertical" scaling is what is referred to as increasing the resources (memory particularly) of a backend server […] as the resource requirements of the application […] grow.

This may have been the source of some confusion. If the term ‘load’ only applies to number of requests or CPU usage, then that’s not the metric I meant to use. So without fussing to long about the semantics of the overloaded (pun intended) use of the word ‘load’, let me try to rephrase:

In my mind, the number of requests/CPU usage is only **correlated** to the pressure on the server and the need for more instances. The **causation** is the memory needed whilst a query is being resolved, which can be very different based on the queries being performed. i.e. imagine one instance needing to serve 10 home page requests vs another instance needing to serve 10 single artwork requests. The former is going to use much more memory than the latter, but CPU usage will most likely be similar and in terms of number of requests there’s no difference.

If you’re sure that this setup will still be able to address that scenario, then it’s probably just a matter of my inexperience with maintaining such pools of servers and load balancing. In which case, let’s leave it at that and see if that indeed turns out to be the case.

alloy:
Btw, another thing to keep in mind is that while requests might have finished serving, the process may still have a bunch of queued work to do that also retains memory (things like error reporting to Sentry, performance metrics to datadog). In those cases, the process kinda needs a pause so it can flush that work.

izakp:
> Btw, another thing to keep in mind is that while requests might have finished serving, the process may still have a bunch of queued work to do that also retains memory (things like error reporting to Sentry, performance metrics to datadog). In those cases, the process kinda needs a pause so it can flush that work.

Yes I can see evidence of that in a recent restart.  No increase in throughput but unbounded memory growth.

izakp:
Thanks for the chat @alloy (and explaining some finer details about Node's memory management) - I think we may be able to achieve "benching" containers at the k8s service level by modifying the container or service's [label selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) - a POC watchdog daemon will be necessary as k8s services are protocol-agnostic and just do packet routing so a 503 from Metaphysics won't be sufficient.  This is uncharted territory, which isn't in the end a bad thing ^_^

izakp:
I think it's going to be helpful to think about normal operation versus failover scenarios here.  I should note that my comment above...

> In the end, load-based autoscaling creates more replicas and scales out in response to load - each application instance taking on a proportion of that load. This in turn means that we should not see any one Metaphysics container processing a disproportionately high percentage of requests, and thus we should see each application instance's memory footprint stay within reasonable limits.

... is an overly-optimistic "normal" application behavior.  I think what we've converged on here is how we want to deal with orchestration when the application goes out-of-bounds in some way i.e. a failover scenario.  Right now k8s just kills the pods when they hit memory limits which is not ideal.  Whether this "timeout" or "beching" approach can compliment horizontal autoscaling based on load and/or memory or merits a reimplementation of scaling rules from the ground up TBD.

PR #987, opened 2018-04-02
Update to v3 of graphql-tools pre-release

freiksenet:
v3 should have a fix for interface and union fragments with fragment
spreads

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "graphql": "^0.13.2",
     "graphql-depth-limit": "^1.1.0",
     "graphql-relay": "^0.5.4",
-    "graphql-tools": "^3.0.0-alpha.11",
+    "graphql-tools": "^3.0.0-alpha.12",
     "graphql-type-json": "^0.1.4",
     "i": "^0.3.5",
     "jwt-simple": "^0.5.0",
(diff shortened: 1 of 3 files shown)


mzikherman:
Pulling this branch down and I was getting some errors in the console from usage of the new API in `graphql-tools`, but a test example I had of the original bug (spread fragment on an interface in a union), is working!

```
{
  artwork(id:"banksy-di-faced-tenner-21") {
    id
    context {
      ... on Node {
        __id
        __typename
      }
    }
  }
}
```

This [query](https://metaphysics-staging.artsy.net/?query=%7B%0A%20%20artwork(id%3A%22banksy-di-faced-tenner-21%22)%20%7B%0A%20%20%20%20id%0A%20%20%20%20context%20%7B%0A%20%20%20%20%20%20...%20on%20Node%20%7B%0A%20%20%20%20%20%20%20%20__id%0A%20%20%20%20%20%20%20%20__typename%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D), currently working since stitching is disabled, returns the error we were seeing when enabled: `Cannot return null for non-nullable field ArtworkContextAuction.__id.`.

Awesome!!!  🍾

freiksenet:
Yes, seems there is another bug that causes this to fail or some edge case
of previous. I only got credentials to run the server in the evening, I'll
investigate further today.

freiksenet:
This should fix this now.

freiksenet:
This is ready now. I wonder if it's possible to force running tests at mergedSchema mode, so I could add test here.

PR #989, opened 2018-04-04
Add tests to merged schema regression and bump graphql-tools version

freiksenet:


src/lib/__tests__/mergeSchemas.js
```diff
@@ -91,3 +92,53 @@ describe("convection link", () => {
     })
   })
 })
+
+describe("stiched schema regressions", () => {
+  it.only("union in interface fragment issue", async () => {
```
alloy: Probably shouldn’t remain `it.only`

src/test/utils.js
```diff
@@ -41,3 +42,29 @@ export const runAuthenticatedQuery = (query, rootValue = {}) => {
...
+) => {
+  if (!mergedSchema) {
+    mergedSchema = await mergeSchemas()
+  }
+  return graphql(mergedSchema, query, rootValue, {}).then(result => {
+    if (result.errors) {
+      const error = result.errors[0]
+      throw error.originalError || error
+    } else {
+      return Promise.resolve(result.data)
```
alloy: Off-topic: I’m wondering why we’re boxing the result data in a resolved promise here and above, that should be done automatically when returning data from a `then`, no?

src/lib/__tests__/mergeSchemas.js
```diff
@@ -91,3 +92,53 @@ describe("convection link", () => {
...
+              __typename
+            }
+          }
+        }
+      }
+    `,
+      {
+        artworkLoader: async () => artworkResponse,
+        salesLoader: async () => sales,
+        relatedFairsLoader: async () => ({}),
```
alloy: I haven’t been using the `async` keyword in anger yet; this is equal to `() => Promise.resolve({})` ?

src/lib/__tests__/mergeSchemas.js
```diff
@@ -91,3 +92,53 @@ describe("convection link", () => {
...
 })
+
+describe("stiched schema regressions", () => {
+  it.only("union in interface fragment issue", async () => {
+    const artworkResponse = {
+      id: "banksy-di-faced-tenner-21",
+      sale_ids: ["foo-foo"],
+    }
+
+    const sales = [
```
alloy: Nit: maybe call this `salesResponse` to be consistent with the one above it?

src/lib/__tests__/mergeSchemas.js
freiksenet: Oops. :)

src/test/utils.js
freiksenet: Indeed it's superficial. I just copied :)

src/lib/__tests__/mergeSchemas.js
freiksenet: Yes, exactly.

src/lib/__tests__/mergeSchemas.js
freiksenet: Ok.

PR #992, opened 2018-04-08
Fix foreman

orta:
Node Foreman hasn't been updated in a few years, and it looks like Heroku have opted to start shipping their own with a few fixes.


alloy:
Yeah, good stuff!

PR #1011, opened 2018-04-19
@joeyAghion => allow reserve to be passed when updating an order

sweir27:
Per your comment on my previous PR!

We need to be able to pass `reserve: true` when we update an order.


joeyAghion:
👍

PR #1020, opened 2018-04-25
@yuki24 => return whether or not I have qualified credit cards

sweir27:
Don't merge this until https://github.com/artsy/gravity/pull/11689 is deployed to production (I'll do that in the morning)!-- but looking for comments.

This PR adds a the field `has_qualified_credit_cards` on `me`, so we will know whether or not to show the credit card form when you attempt to confirm a bid.


sweir27:
Gravity-side is deployed, so this is good to go.

yuki24:
LGTM!

PR #1023, opened 2018-04-26
Query Bidder Position by ID

sepans:
Fixes: https://artsyproduct.atlassian.net/browse/PURCHASE-98

Makes it possible to send queries like
```
query {
  me {
   bidder_position(id: "5ae1df168b3b8141bfc32e5d") {
    processed_at
   }
  }
 }
```

and check to see if the bidding position is successfully created.

### DIFF
--- a/src/lib/loaders/loaders_with_authentication/gravity.js
+++ b/src/lib/loaders/loaders_with_authentication/gravity.js
@@ -72,7 +72,7 @@ export default (accessToken, userID, opts) => {
     meLoader: gravityLoader("me"),
     meBiddersLoader: gravityLoader("me/bidders"),
     meBidderPositionsLoader: gravityLoader("me/bidder_positions"),
-    bidderPositionStatus: gravityLoader(
+    bidderPositionLoader: gravityLoader(
       ({ id }) => `me/bidder_position/${id}/`,
       {},
       { headers: true }
--- a/src/lib/loaders/loaders_with_authentication/gravity.js
+++ b/src/lib/loaders/loaders_with_authentication/gravity.js
@@ -72,7 +72,7 @@ export default (accessToken, userID, opts) => {
     meLoader: gravityLoader("me"),
     meBiddersLoader: gravityLoader("me/bidders"),
     meBidderPositionsLoader: gravityLoader("me/bidder_positions"),
-    bidderPositionLoader: gravityLoader(
+    meBidderPositionLoader: gravityLoader(
       ({ id }) => `me/bidder_position/${id}/`,
       {},
       { headers: true }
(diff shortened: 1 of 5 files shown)


sweir27:
🎉 An initial thought is that we might want to nest the `status` under `bidder_position`-- so the query would look like:
```
query {
  me {
    bidder_position(id: "12345") {
      status
    }
  }
}
```

that way we can extend the model to include other properties that are returned from that endpoint without having to add new ones.

joeyAghion:
Maybe this was just an example, but note that bidder positions will always have a `created_at` set (including in the initial response to the API call that POSTs them). There's a `processed_at` field that will become set after the bid has been processed, which you might want instead.

sepans:
@sweir27 makes sense. [bidder_position.js](https://github.com/artsy/metaphysics/blob/master/src/schema/bidder_position.js) already exist but it is one level up in the directory structure so I can rename everything (including the filename) to be consistent and match `bidder_position`.

@joeyAghion Yes thats what I meant. Thanks for pointing that out. Updated the query above to avoid confusion.

src/lib/loaders/loaders_with_authentication/gravity.js
```diff
@@ -72,6 +72,11 @@ export default (accessToken, userID, opts) => {
     meLoader: gravityLoader("me"),
     meBiddersLoader: gravityLoader("me/bidders"),
     meBidderPositionsLoader: gravityLoader("me/bidder_positions"),
+    bidderPositionLoader: gravityLoader(
```
sweir27: Since there is a corresponding `bidder_positions_endpoint` in gravity that is _not_ scoped to `me`, maybe we want to name this `meBidderPositionLoader` (different from the one above because it's singular)

src/schema/me/__tests__/bidder_position.test.js
```diff
@@ -0,0 +1,44 @@
+import { runAuthenticatedQuery } from "test/utils"
+
+describe("BidderPosition", () => {
+  it("returns processed_at", () => {
+    const rootValue = {
+      bidderPositionLoader: sinon.stub().returns(
```
sweir27: You shouldn't need the stub here since you can just return a function like `bidderPositionLoader: () => Promise.resolve...`

src/schema/me/bidder_position.js
```diff
@@ -0,0 +1,21 @@
+import { GraphQLNonNull, GraphQLString } from "graphql"
+import BidderPositionType from "schema/bidder_position"
+
+export const BidderPosition = {
+  type: BidderPositionType.type,
+  description: "Returns the bidder position status",
```
sweir27: This can be updated now to something like `Returns a single bidder position`

sweir27:
Few minor comments but this is looking 👍 !

sweir27:
👍 👍

PR #1031, opened 2018-05-02
Handle Low Bid Gravity Error

sepans:
### Problem
Gravity returns an [error](https://github.com/artsy/gravity/blob/cd8f9c29648f5b6e18048b8db32d419644a9b47a/app/models/domain/bidder_position.rb#L162) when bid is not high enough before processing the bid. When emission creates a bidder position and gets such an error it [treats](https://github.com/artsy/emission/blob/master/src/lib/metaphysics.ts#L40) MP error as 'Server-side error' which is too severe for just a bid that is not high enough. ([more context](https://github.com/artsy/emission/pull/1025))

### Solution
Instead of throwing an error, the mutation catches the error and adds the error message to the empty `BidderPosition`. And since `id` is required it returns `"BID_NOT_HIGH_ENOUGH"` as id 🙈

A more elegant solution @orta suggested is to return an Error message object instead of `BidderPosition` [here](https://github.com/artsy/metaphysics/blob/master/src/schema/me/bidder_position_mutation.js#L23) using type union. I tried something like:
```
const BidderPositionErrorType = new GraphQLObjectType({
  name: "BidderPositionError",
  fields: () => ({
    message: {
      type: GraphQLString,
    },
  }),
})

// is this object needed or I can use BidderPositionErrorType directly below?
const BidderPositionError = {
  type: BidderPositionErrorType,
  description: "An BidderPosition",
}

const BidderPositionResultType = new GraphQLUnionType({
  name: "BidderPositionResult",
  types: [BidderPositionError.type, BidderPosition.type],
  resolveType(value) {
    if (value.errors) {
      return BidderPositionError.type;
    }
    return BidderPosition.type;
  },
});

// is this object needed or I can use BidderPositionResultType directly below?
const BidderPositionResult = {
  type: BidderPositionResultType,
  description: "An BidderPosition result object",
}
```
For the union type and using that as `outputField` 
```
  outputFields: {
    result: {
      type: BidderPositionResult.type,
      resolve: (position) => position,
    },
  },
```

It more or less worked although I wasn't sure if it was the right way of doing it and @sweir27 and I where not sure how much value handling the error message this way and dealing with its complexity here and also on emission side will bring. (well this discussion was before I realized `id` is required and cannot be null and I had to pass something like `"BID_NOT_HIGH_ENOUGH"` as id)


src/schema/bidder_position.js
```diff
@@ -119,6 +119,10 @@ const BidderPositionType = new GraphQLObjectType({
       type: GraphQLInt,
       deprecationReason: "Favor `suggested_next_bid`",
     },
+    message: {
+      type: GraphQLString,
```
sweir27: Just pinged @briansw about clarifying how we want to display the error messages, but I think we should handle mapping the gravity errors --> user-friendly copy _here_ so we don't have to replicate in clients.

src/schema/me/bidder_position_mutation.js
```diff
@@ -40,5 +40,17 @@ export const BidderPositionMutation = mutationWithClientMutationId({
       return new Error("You need to be signed in to perform this action")
     }
     return createBidderPositionLoader({ sale_id, artwork_id, max_bid_amount_cents })
+      .then(p => p)
```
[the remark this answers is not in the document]
sepans: I just wanted to process the error in the `catch` block of the promise. So the `then` should just return the object. Thought it was needed. Isn't it?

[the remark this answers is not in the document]
ashfurrow: To be honest I'm not 100% sure. I think it should work without it since `createBidderPositionLoader` returns a promise anyway?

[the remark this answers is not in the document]
sepans: 👍 I will give it a try.

src/schema/bidder_position.js
```diff
@@ -119,6 +119,14 @@ const BidderPositionType = new GraphQLObjectType({
       type: GraphQLInt,
       deprecationReason: "Favor `suggested_next_bid`",
     },
+    message_header: {
```
sweir27: I don't think you need these anymore, right?

src/schema/me/bidder_position_mutation.js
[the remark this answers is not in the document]
sepans: It worked without `.then(p => p)`, but ended up adding some code there. 😄

src/schema/me/bidder_position_mutation.js
```diff
@@ -1,10 +1,70 @@
...
+    status: {
+      type: new GraphQLNonNull(GraphQLString),
+    },
+    message_header: {
+      type: GraphQLString,
+    },
+    message_description: {
+      type: GraphQLString,
+    },
+    position: {
```
sweir27: Oh nice solution to include the bidder position here so you don't have to make up an ID!

orta:
From reading the PR body - this is an elegant solution

src/schema/me/__tests__/bidder_position_mutation.test.js
```diff
@@ -1,33 +1,122 @@
...
+      return runAuthenticatedQuery(query, rootValue).then(data => {
+        expect(data.createBidderPosition.result.position).toBeNull()
+        expect(data.createBidderPosition.result.message_header).toEqual("Bid not placed")
+        expect(data.createBidderPosition.result.message_description).toEqual(
+          `Your bid can’t be placed at this time.\
+ Please contact support@artsy.net for more information.`)
+      })
     })
   })
 })
```
orta: These are great tests at describing the work done

orta:
WRT links, if they're markdown then it's probably fine IMO to call it ``message_md` and then both native and web can parse and present it, we have a bunch of ways for displaying markdown in emission already

src/schema/me/bidder_position_mutation.js
```diff
@@ -1,10 +1,82 @@
...
+    id: "ERROR_SALE_CLOSED",
+    gravity_key: "Sale Closed to Bids",
+    header: "Lot closed",
+    description: "Sorry, your bid wasn’t received before the lot closed.",
+  },
+  {
+    id: "ERROR_LIVE_BIDDING_STARTED",
+    gravity_key: "Live Bidding has Started",
+    header: "Live bidding has started",
+    description: `Sorry, your bid wasn’t received before live bidding started.\
```
sweir27: Do we need both `description` and `description_md`?

src/schema/me/bidder_position_mutation.js
```diff
@@ -40,5 +112,22 @@ export const BidderPositionMutation = mutationWithClientMutationId({
...
     }
     return createBidderPositionLoader({ sale_id, artwork_id, max_bid_amount_cents })
+      .then(p => { return { status: "SUCCESS", position: p } })
+      .catch(e => {
+        const errorSplit = e.message.split(" - ")
+        const errorObject = errorSplit.length > 1 ? JSON.parse(e.message.split(" - ")[1]) : null
+        if (errorObject && errorObject.type === "param_error") {
+          const error = biddingErrors.find(d => errorObject.message.trim().startsWith(d.gravity_key)) ||
+            errorObject.ERROR_UNKNOWN
+          const liveAuctionUrl = `${PREDICTION_ENDPOINT}/${sale_id}`
```
sweir27: A note that when generating the URL to a live auction, we generally [care whether or not you are logged in](https://github.com/artsy/force/blob/master/src/utils/domain/auctions/urls.js#L11-L19), but I think in this case this URL is fine since you can't get to the bid screen unless you're logged in.

src/schema/me/bidder_position_mutation.js
sepans: Not for the current screens. Thought it would be useful in future for displaying plain messages but that might never happen. 🤷‍♂️

src/schema/me/bidder_position_mutation.js
sepans: Not sure but I think we throw error when request is not coming from a logged in user with 
```
 if (!accessToken) {
      return new Error("You need to be signed in to perform this action")
    }
```
above.

sweir27: You're right! Definitely covered then 👍

src/schema/me/bidder_position_mutation.js
sweir27: Maybe we can remove for now then? Just one extra place to care about typos etc.

src/schema/me/__tests__/bidder_position_mutation.test.js
```diff
@@ -1,33 +1,126 @@
...
-    return runAuthenticatedQuery(query, rootValue).then(data => {
-      expect(data.createBidderPosition.position.suggested_next_bid_cents).toEqual(110000)
+  describe("success", () => {
+    it("creates a bidder position", async () => {
+      const rootValue = {
+        createBidderPositionLoader: sinon.stub().returns(
+          Promise.resolve(createBidderPosition)
+        ),
+      }
+      return runAuthenticatedQuery(query, rootValue).then(data => {
```
acjay: Since you're in an `async` function already, it would be better to do:

```
  describe("success", () => {
    it("creates a bidder position", async () => {
      const rootValue = {
        createBidderPositionLoader: sinon.stub().returns(
          Promise.resolve(createBidderPosition)
        ),
      }
      const data = runAuthenticatedQuery(query, rootValue)
      expect(data.createBidderPosition.result.position.suggested_next_bid_cents).toEqual(110000)
      expect(data.createBidderPosition.result.message_header).toBeNull()
      expect(data.createBidderPosition.result.message_description_md).toBeNull()
    })
  })
```

In a world with `async`/`await`, I'd almost never use `.then` explicitly. The only time I can think of where it's the way to go is if you need two paths of execution to happen concurrently after a promise resolves, which should be really rare.

sweir27:
Awesome work on this, @sepans ! Especially handling all of the new requirements/edge cases 😓.

PR #1039, opened 2018-05-08
[dev] set dummy vals in .env.example

l2succes:
Allows us to copy `.env.example` to `.env` and run `yarn dump-schema` with no errors


eessex:
Thanks for following up on this and helping debug!

PR #1043, opened 2018-05-12
Add eslint-config-prettier (fixes #705)

### LINKED ISSUE #705: Add eslint-plugin-prettier to remove rules related to prettier
https://github.com/prettier/eslint-plugin-prettier

zephraph:
The plugin is configured and conflicting rules have been removed. It errors out though due to the current version of eslint being too far out of date.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,15 +6,14 @@
     "jasmine": true,
     "es6": true
   },
-  "extends": "airbnb",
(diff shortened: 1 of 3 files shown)


orta:
Hrm yeah, if we can get eslint working with typescript then MP could be the first place where we start to move back to eslint (this is the only prod app we care about using using it) instead of migrating to tslint

PR #1047, opened 2018-05-16
bidder_position returns status/message 

sepans:
`bidder_position_mutation` returns status and error message when an error occurs at the time of placing the bidder position. After that the client (emission) polls MP to check if the bidder position is processed and then if it is active and met reserve.
 
This PR adds `status` and `message_header` and `message_description_md` to `bidder_position`  query returns similar to `bidder_position_mutation` to make the flow consistent on the client, also move the logic of checking bidder position status to MP.

**Note:** `bidder_position` (https://github.com/artsy/metaphysics/pull/1023) query is pretty new and hasn't been used on production yet so it is okay to change its return type.

sample query:
```
{
  me {
    bidder_position(id: "5af9fce35acaa60020dbeeed") {
      status
      position {
        processed_at
        is_active
        sale_artwork {
          reserve_status
        }
      }
    }
  }
}
```

response:
```
{
  "data": {
    "me": {
      "bidder_position": {
        "status": "SUCCESS",
        "position": {
          "processed_at": "2018-05-14T21:17:24+00:00",
          "is_active": true,
          "sale_artwork": {
            "reserve_status": "no_reserve"
          }
        }
      }
    }
  },
  "extensions": {
    "requests": {
      "gravity": {
        "requests": {
          "me/bidder_position/5af9fce35acaa60020dbeeed/": {
            "time": "0s 205.429ms",
            "cache": false
          },
          "sale_artwork/5af9f7bf02ac643526028809": {
            "time": "0s 14.294ms",
            "cache": true
          }
        }
      }
    },
    "requestID": "cec839b0-5959-11e8-a39a-7ba93b93d692"
  }
}
```

If you also have suggestion about name/location of the files let me know.

### DIFF
--- a/src/schema/me/bidder_position.js
+++ b/src/schema/me/bidder_position.js
@@ -1,8 +1,15 @@
 import { GraphQLNonNull, GraphQLString } from "graphql"
-import BidderPositionType from "schema/bidder_position"
+import { BiddingMessages } from "./bidder_position_messages"
+import { BidderPositionResultType } from "../types/bidder_position_result"
+
+const ANY_RESERVE_MET_STATUSES = ["no_reserve", "reserve_met"]
+
+const anyReserveMet = (position) => {
+  return ANY_RESERVE_MET_STATUSES.indexOf(position.sale_artwork.reserve_status) > -1
+}
 
 export const BidderPosition = {
-  type: BidderPositionType.type,
+  type: BidderPositionResultType,
   description: "Returns a single bidder position",
   args: {
     id: {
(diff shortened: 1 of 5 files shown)


src/schema/me/bidder_position.js
```diff
@@ -17,5 +24,35 @@ export const BidderPosition = {
   ) =>
     meBidderPositionLoader({
       id,
-    }).then(response => response.body),
+    }).then(response => {
+      const position = response.body
+      let status
+      if (anyReserveMet(position) && position.processed_at && position.active) {
+        return {
+          status: "SUCCESS",
```
sweir27: The naming here is a little confusing to me-- even though we show the error-looking symbol on the emission views, I think _all_ cases where the bidder position is processed are technically "successful", and there's only really one true error case.

Kind of hard to describe-- but maybe something like this maps a little better?
```js
if (position.processed_at) {
  if (!position.active) {
    status = 'OUTBID'
  } else if (anyReserveMet(position)) {
    status = 'WINNING'
  } else {
    status = 'RESERVE_NOT_MET'
  }
} else if (!position.processed_at) {
  status = 'PENDING'
} else {
  status = 'ERROR'
}
```

src/schema/me/bidder_position_messages.js
```diff
@@ -0,0 +1,45 @@
...
+    gravity_key: "Please enter a bid higher than",
+    header: "Your bid wasn't high enough",
+    description_md: () => `Another bidder placed a higher max bid or the same max bid before you did.  \
+ Bid again to take the lead.`,
+  },
+  {
+    id: "ERROR_RESERVE_NOT_MET",
+    gravity_key: "Please enter a bid higher than",
+    header: "Your bid wasn't high enough",
+    description_md: () => `Another bidder placed a higher max bid or the same max bid before you did.  \
```
sweir27: I'm not sure if this error message is correct

PR #1048, opened 2018-05-17
capitalisation fix #minor

cavvia:
Updates titles of new homepage modules

### DIFF
--- a/src/schema/home/title.js
+++ b/src/schema/home/title.js
@@ -32,9 +32,9 @@ const moduleTitle = {
     return artistLoader(params.related_artist_id).then(artist => artist && artist.name)
   },
   saved_works: () => "Recently saved",
-  similar_to_saved_works: () => "Similar to Recently Saved",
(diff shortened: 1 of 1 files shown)


src/schema/home/title.js
```diff
@@ -32,9 +32,9 @@ const moduleTitle = {
     return artistLoader(params.related_artist_id).then(artist => artist && artist.name)
   },
   saved_works: () => "Recently saved",
-  similar_to_saved_works: () => "Similar to Recently Saved",
+  similar_to_saved_works: () => "Similar to works you've saved",
   recently_viewed_works: () => "Recently viewed",
-  similar_to_recently_viewed: () => "Similar to Recently Viewed",
+  similar_to_recently_viewed: () => "Similar to works you've viewed",
```
alloy: What do you think about using a pretty apostrophe `’`?

cavvia: oh good point - it does print at a large font size on the homepage so totally worth it. will follow up.

PR #1050, opened 2018-05-22
[debug] Log query after N sec with `LOG_QUERY_DETAILS_THRESHOLD=N`

alloy:
This can be used to get full reproduction details of long running
queries. Do not set this too low by default, as it will flood the
logs.


alloy:
By default this will be disabled, but maybe post this incident we want to leave it turned on with a high threshold, just in case we ever miss data in DataDog to repro again?

cavvia:
Great - can update our k8s env with `LOG_QUERY_DETAILS_THRESHOLD` env var - what do you think is a good level? 60 seconds? or maybe 80 to begin with. Also, are these secs or millisecs?

alloy:
These are seconds. I think that 80 is probably a good default, but right now we may want to temporarily go lower? I was thinking 30, but that may already trigger too often, maybe 50?

cavvia:
Yes we don't want to spam the logs... let's try 50. I've updated k8s using `hokusai <production/staging> env set`

PR #1051, opened 2018-05-22
[debug] Add `warn` log level and make some logging less severe.

alloy:
We want the `error` logs to be enabled in production and staging, so it should really only log unexpected/severe situations.


alloy:
@cavvia I’ll self-merge this to get into staging asap.

PR #1054, opened 2018-05-22
log slow cache fetches

alloy:
Supersedes #1053.


src/lib/cache.js
```diff
@@ -67,8 +72,12 @@ export default {
   get: key => {
     return new Promise((resolve, reject) => {
       if (isNull(client)) return reject(new Error("Cache client is `null`"))
-
+      const start = Date.now()
       client.get(key, (err, data) => {
+        const time = Date.now() - start
+        if (time > CACHE_QUERY_LOGGING_THRESHOLD_MS) {
+          error(`Slow Cache#Get: ${time}ms`)
```
alloy: @cavvia Not sure why you went from `warn` to `info`, but I made it into `error`.

cavvia: :+1:

PR #1058, opened 2018-05-23
[debug] Make sure we log errors comming from `redis.get()`

alloy:
The docs aren’t entirely clear on whether or not the global event
listener on the client will always log errors or truly only when
individual commands do not have callback handlers attached. However,
this code in the library does seem to suggest that the latter is the
case: https://github.com/NodeRedis/node_redis/blob/9bb1454ca183f4132dbec8d12d010f28303cdebd/lib/utils.js#L87-L93

### DIFF
--- a/src/lib/cache.js
+++ b/src/lib/cache.js
@@ -102,9 +102,14 @@ export default {
         if (time > CACHE_QUERY_LOGGING_THRESHOLD_MS) {
           error(`Slow Cache#Get: ${time}ms key:${key}`)
         }
-        if (err) return reject(err)
-        if (data) return resolve(JSON.parse(data))
-        reject(new Error("cache#get did not return `data`"))
+        if (err) {
+          error(err)
+          reject(err)
+        } else if (data) {
+          resolve(JSON.parse(data))
+        } else {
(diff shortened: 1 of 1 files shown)


cavvia:
Looks good to me - I wonder if connection errors should also get logged? I don't see connections dropping out and in much on the redis side, but as a precaution perhaps.

alloy:
I _think_ those should get logged by the global handler [here](https://github.com/artsy/metaphysics/blob/a9f43277ca51037b33c6673b1e29b58e77449aa3/src/lib/cache.js#L63), but with the caveat that there is [this](https://github.com/artsy/metaphysics/blob/a9f43277ca51037b33c6673b1e29b58e77449aa3/src/lib/cache.js#L36-L40).

joeyAghion:
When you say "make sure we _log_ errors," will these only be inspectable in logs?

cavvia:
Good point, I'm not sure what gets sent to sentry here... I was talking about logs as a start.

alloy:
Log to stdout/papertrail. A follow-up TODO for when we have Sentry again should be to post anything that gets logged with the error logger to sentry indeed.

PR #1059, opened 2018-05-23
Create external operator role in causality_jwt

williardx:
https://artsyproduct.atlassian.net/browse/SELL-530

- Adds a new `external_operator` role to the Causality JWT. Checks if aspiring `OPERATOR` is first an admin; if not, checks whether or not they qualify to be an `external_operator`. External operators have access to the partner managing the sale. 
- Adds `mePartnersLoader` that fetches the partners the user currently has access to. Depends on https://github.com/artsy/gravity/pull/11736.
- Adds access to partner object in sale schema so that we can fetch the sale partner's ID in Prediction.

### DIFF
--- a/src/schema/causality_jwt.js
+++ b/src/schema/causality_jwt.js
@@ -19,7 +19,6 @@ export default {
         values: {
           PARTICIPANT: { value: "PARTICIPANT" },
           OPERATOR: { value: "OPERATOR" },
-          EXTERNAL_OPERATOR: { value: "EXTERNAL_OPERATOR" },
         },
       }),
       description: "",
@@ -91,14 +90,28 @@ export default {
           HMAC_SECRET
         )
       })
-    } else if (options.role === "EXTERNAL_OPERATOR") {
+    // Operator role if logged in as an admin or if user has access to partner
+    } else if (options.role === "OPERATOR") {
       return Promise.all([saleLoader(options.sale_id), meLoader()]).then(
         ([sale, me]) => {
+          if (me.type === "Admin") {
+            return jwt.encode(
+              {
+                aud: "auctions",
+                role: "operator",
+                userId: me._id,
+                saleId: sale._id,
+                bidderId: me.paddle_number,
+                iat: new Date().getTime(),
+              },
+              HMAC_SECRET
+            )
+          }
           return mePartnersLoader({ "partner_ids[]": sale.partner._id }).then(
             mePartners => {
               // Check if current user has access to partner running the sale
               if (!isExternalOperatorAuthorized(sale, mePartners)) {
-                throw new Error("Unauthorized to be operator for this sale")
+                throw new Error("Unauthorized to be operator")
               }
               return jwt.encode(
                 {
@@ -112,26 +125,6 @@ export default {
                 HMAC_SECRET
               )
             }
-          )
-        }
-      )
-      // Operator role if logged in as an admin
-    } else if (options.role === "OPERATOR") {
-      return Promise.all([saleLoader(options.sale_id), meLoader()]).then(
-        ([sale, me]) => {
-          if (me.type !== "Admin") {
-            throw new Error("Unauthorized to be operator")
-          }
-          return jwt.encode(
-            {
-              aud: "auctions",
-              role: "operator",
-              userId: me._id,
-              saleId: sale._id,
-              bidderId: me.paddle_number,
-              iat: new Date().getTime(),
-            },
-            HMAC_SECRET
           )
         }
       )
--- a/src/schema/causality_jwt.js
+++ b/src/schema/causality_jwt.js
@@ -1,13 +1,8 @@
 import jwt from "jwt-simple"
 import { GraphQLString, GraphQLNonNull, GraphQLEnumType } from "graphql"
 import config from "config"
-import { includes } from "lodash"
 
 const { HMAC_SECRET } = config
-
-const isExternalOperatorAuthorized = (sale, mePartners) => {
-  return includes(mePartners.map(p => p._id), sale.partner._id)
-}
 
 export default {
   type: GraphQLString,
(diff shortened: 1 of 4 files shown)


src/schema/sale/index.js
```diff
@@ -293,6 +294,10 @@ export const SaleType = new GraphQLObjectType({
           }
         },
       },
+      partner: {
+        type: Partner.type,
+        resolve: ({ partner }) => partner,
```
sweir27: This will only return in the `partner` data that gravity sends back in the `sale` JSON (which may not include all of the fields we need). You probably want to actually load the partner here, similar to what we do for the `promoted_sale` below.

williardx: That works okay for me since all I need is the partner's `_id`. Do you think it's worth the extra request to fetch a more complete record of the partner?

starsirius: I think it's fine to start from here and expand when needed.

sweir27: Makes sense! Can we possibly add a comment? 😄 This is probably going to be fine for a while, but (as you bring up @williardx ) it's sort of an odd experience from a graphiQL UI or otherwise that suggests you can retrieve fields that won't exist.

src/schema/causality_jwt.js
```diff
@@ -77,24 +90,41 @@ export default {
...
-              userId: me._id,
-              saleId: sale._id,
-              bidderId: me.paddle_number,
-              iat: new Date().getTime(),
-            },
-            HMAC_SECRET
+          return mePartnersLoader({ "partner_ids[]": sale.partner._id }).then(
+            mePartners => {
+              // Check if current user has access to partner running the sale
+              if (!isExternalOperatorAuthorized(sale, mePartners)) {
```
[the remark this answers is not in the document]
starsirius: Oh this is a good point, @mzikherman ! Missed that the partner ids are already encoded in the JWT for non-admin users. Want to point out a few differences with 2 approaches:

* Looks like we currently encode [*all*](https://github.com/artsy/gravity/blob/e1d8f8c3a896e0a6573f846ecd7a317c1e8e636f/app/models/util/user_trust.rb#L33) partner ids in the Gravity JWT access token. This is unlikely to be a problem given that most of the case a user would only manage a few partners, but it may depend on future business logic (however, we do have one edge case where a user manages 77 partners 🙀 (some data below)). Hitting the `/me/partners?partner_ids[]=..` Gravity endpoint might not suffer from the same edge case.
* Using the `partner_ids` in the JWT might require the user to sign out and sign back in again after granting a user partner access. We may need to make sure last-minute sign out and sign in is avoided for new external operators before a sale starts, and this can be handled in the UI, i.e. Prediction.

Given the causality JWT is only fetched once, maybe the performance impact is low. However, I don't feel too strongly about either approach.

Lastly... leaderboard of users who have access to most partners (I think the outlier was an edge case that we gave the fair organizer access to all the partners at the Spring/Break):

```ruby
user_ids_by_partner_access = AccessControl.where(property_type: 'Partner').pluck(:user_ids)
partner_access_count_by_user_id = user_ids_by_partner_access.flatten.reduce(Hash.new(0)) { |h, user_id| h[user_id.to_s] += 1; h }
top10_users = partner_access_count_by_user_id.sort_by(&:last).reverse.take(10)
top10_users.each { |t| printf "%-30s %2d\n", User.find(t[0]).email, t[1] }

# sales@springbreakartshow.com   77
# presse@smb.spk-berlin.de        9
# haleyrosecohen@gmail.com        5
# jchart92@gmail.com              4
# ashkan.nasseri@gmail.com        4
# julia@art.sy                    4
# mebo@mebobeijing.com            3
# madeleine.boucher@gmail.com     3
# street.art.zoombie@gmail.com    3
# anandaroop.roy@gmail.com        3
```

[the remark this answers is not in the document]
williardx: Good catch @mzikherman! That sounds like a better overall approach, though I agree with @starsirius that it's best if users don't encounter a scenario where they need to sign out and sign back in. We could combine the two approaches: First try the JWT, and if they don't already have access then go and fetch from Gravity to see if anything changed.

The sign out/sign in flow is also an issue for Prediction as well (https://github.com/artsy/prediction/pull/551), so looks like I'll need to rethink that!

src/schema/sale/index.js
williardx: Will do!

src/schema/causality_jwt.js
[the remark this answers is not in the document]
williardx: Yeah, that's a good question. Is this a problem across other Artsy apps, e.g. Volt?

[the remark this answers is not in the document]
starsirius: Volt's implementation predates the addition of the `partner_ids` in Gravity JWT, so it's still [hitting the `/me/partners` endpoint](https://github.com/artsy/kinetic/blob/3afcb258e911f6159c16c1020ff41f04c289c140/app/controllers/kinetic/application_controller.rb#L32) and serializing partner data in the session.

Usually partners have to sign out and sign back in to refresh the session, but sometimes we are able to do it smartly in the background, usually to [re-gain access](https://github.com/artsy/volt/blob/5eebe6703b13fbefd56b8683a4c19226743c9aee/app/controllers/application_controller.rb#L42-L43). For cases where user's access to the partner is removed, in Volt we might have similar issues (but we are able to rely on other indicators like partner subscription state that [updates periodically](https://github.com/artsy/kinetic/blob/3afcb258e911f6159c16c1020ff41f04c289c140/app/controllers/kinetic/application_controller.rb#L138)). I feel that Volt's use cases can be a bit more tolerant and the Operator's use cases may need to be more strict.

williardx:
After some more discussion about the right authentication strategy we decided not to rely on the Artsy JWT for partner authentication. The JWT could be out of date which could create security issues or unpleasant auth flows. Metaphysics will check with Gravity every time for the most up to date information when a user requests an external operator JWT. On the Prediction side, if we see a `null` JWT after attempting to access operator then we redirect back to that auction's sale page. (See https://github.com/artsy/prediction/pull/551.)

As it stands, this PR is good to go on my end -- does anyone have more questions/comments before we can merge?

starsirius:
This looks good to me!! 💯 Let's rebase on top of master and merge? Thanks, @williardx!!

PR #1061, opened 2018-05-28
Enable crunching with X-Crunch header

zephraph:
This allows enabling the crunching of a response (#1042) via an `X-Crunch` header. It'll just be a little bit easier to integrate into some of the projects this way. 

In [emission](https://github.com/artsy/emission/blob/6d3c3496e407207f625dcbbb3946a7a69a05c48f/src/lib/metaphysics.ts#L10), the current approach would require us to process the URL to safely add the query parameter. Enabling it via a header would just be a bit cleaner. 

Hindsight.

### DIFF
--- a/src/lib/__tests__/crunchInterceptor.test.js
+++ b/src/lib/__tests__/crunchInterceptor.test.js
@@ -32,6 +32,17 @@ describe("crunchInterceptor", () => {
       })
   })
 
+  it("should crunch the result when header is present", () => {
+    return request(app(crunchInterceptor))
+      .get("/?query={greeting}")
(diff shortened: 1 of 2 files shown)


zephraph:
See https://github.com/artsy/emission/pull/1068 for emission implementation.

zephraph:
When responses are crunched the response now has an `X-Crunch` header as well. This header gives the clients something to hook onto to know if the incoming response should be uncrunched.

zephraph:
It's not strictly necessary, but it also doesn't really hurt. It's probably easier to use graphiql with the query param. It's just another options for any clients that are connecting.

PR #1063, opened 2018-05-29
ESLint version bump + improvements

mbilokonsky:
This PR does a few things:

1. It increments the version of  ESLint as well as some related modules.
2. It adds a promise plugin to ESLint so that we can ensure safety around certain kinds of promise-based expressions
3. It updates the AirBNB linting rules to their most recent state.
4. It applies this updated/upgraded ruleset against the MP codebase. I autofixed everywhere possible and I chose to ignore those rules that generated hundreds or thousands of errors and could not be autofixed, except in cases where the rule was specifically important to the task of adding support for promise linting.


src/lib/__tests__/total.test.js
```diff
@@ -4,14 +4,11 @@ describe("total", () => {
...
-    return totalViaLoader(loader, {}, { size: 10, extra_option: 1 }).then(
-      total => {
-        expect(loader.mock.calls[0][0].size).toBe(0)
-        expect(loader.mock.calls[0][0].total_count).toBe(true)
-        expect(loader.mock.calls[0][0].extra_option).toBe(1)
-        expect(total).toBe(50)
-      }
-    )
-  })
+  it("loads the path and passes in the token", () =>
```
damassi: Is prettier stripping out the `{` from `it("loads the path and passes in the token", () =>`? If not, implicit returns are hard to maintain across multiple lines and should be avoided.

orta: I think we added eslint-prettier after this, that said - this test could probably just be an async test, would be easier to read too

mbilokonsky: So, I tweaked the formatting without changing the structure, if this helps? I find the multiline implicit returns really readable if they're indented in a way that reflects that flow.

If we don't want to do multiline implicit returns then we'll need a custom lint rule for that, probably, rather than trying to catch them one-at-a-time in a big PR like this?

damassi: @mbilokonsky - I think its a good rule to of thumb to try to avoid non-automated structural changes during refactors. Would you mind re-adding brackets? Missing brackets tend to introduce fragility over the long-term. (We can develop custom rules when necessary but that's for a different PR.)

src/lib/__tests__/total.test.js
```diff
@@ -5,10 +5,11 @@ describe("total", () => {
...
-      expect(loader.mock.calls[0][0].extra_option).toBe(1)
-      expect(total).toBe(50)
-    }))
+    totalViaLoader(loader, {}, { size: 10, extra_option: 1 })
+      .then(total => {
+        expect(loader.mock.calls[0][0].size).toBe(0)
+        expect(loader.mock.calls[0][0].total_count).toBe(true)
+        expect(loader.mock.calls[0][0].extra_option).toBe(1)
+        expect(total).toBe(50)
+      }))
```
damassi: Just confirming that Prettier is installed in your editor, and you have it set to format on save?

mbilokonsky: I'm using the default .vscode config files per project, I haven't set up prettier independently. Prettier does run on a pre-commit hook though, so it should be a non-issue?

src/lib/__tests__/total.test.js
mbilokonsky: but this was an automated change, driven by the listing rules and (probably?) implemented via `lint --fix`. Does that change your request? 

Even just scanning the files changed in this PR shows me half a dozen similar changes in the first few files. If we as an institution care about implicit vs explicit returns I'm happy to add a lint rule around that, but if it's personal preference then I'd rather have that conversation in a subsequent PR? So, sort of the opposite of what you're saying, I guess?

damassi: Ah, I thought this was a manual change 👍. Would you mind turning off that rule so that the structure remains the same, with brackets, and rerun `lint --fix`?

src/lib/__tests__/total.test.js
damassi: When booting VSCode it should read the recommended plugin list and if you install them then the default `.vscode` settings will be picked up in your editor, including fix on save (eslint) and format on save (prettier).

src/lib/__tests__/total.test.js
mbilokonsky: So, I tried that and it left me with a new 500+ linting errors and I'm not willing to make that change right now.

This is the "Our ESLint toolchain was brought into the present with some rough attempts at sensible defaults" PR. 

I don't want to suggest that your preference is invalid here, but I do suggest that stuff like this be moved to subsequent "I want a different rule" PRs that we can evaluate and apply on a case-by-case basis.

Is that reasonable?

src/lib/__tests__/total.test.js
mbilokonsky: I think prettier is at war with ESLint in spite of my attempts to reconcile them. Yesterday, running `yarn prettier-project` worked beautifully with no conflicts. Today when I run it it generates 800 new lint issues. I'm not sure what's happening.

src/lib/__tests__/total.test.js
damassi: There were brackets before, and now there are not brackets. Since code is more stable with them over time, and there's existing precident, and additionally since this change was automated (meaning it can be undone easily), I don't think this is that controversial, or even in need of discussion. 

I can pull down your branch and take a look if you'd like?

src/lib/__tests__/total.test.js
damassi: If you ping @orta he will help get your environment in sync.

src/lib/loaders/__tests__/api.test.js
```diff
@@ -43,15 +41,14 @@ describe("API loaders", () => {
       })
     })
 
-    it("caches the response for the lifetime of the loader", () => {
-      return Promise.all([loader(), loader()]).then(responses => {
+    it("caches the response for the lifetime of the loader", () =>
```
damassi: Here's another good example about taking brackets away -- if someone wanted to log or create a variable before running promise.all they would have to adjust the structure of the block before doing so, and add in two addition brackets at the top and the bottom.

mbilokonsky:
Merged upstream master again, also reconciled some prettier/eslint issues.

Current status: prettier is happy, ESlint is warning against some nested promises which we should probably fix but which would be really easy to screw up, so let's save that for a subsequent PR?

image

mbilokonsky:
Finally, here's the set of lint rules that I landed on. Note: about half of these were already in place, and the rest were my attempts to land on sensible defaults. I don't claim to speak for everyone! Definitely we should agree on what linting rules we're enforcing!

But I do think that it makes sense to have those conversations in subsequent PRs, because this one is growing uncontrollably and it'd be great to merge it sooner than later.

image

src/lib/loaders/__tests__/api.test.js
mbilokonsky: I definitely understand your preference. I tend to find myself on the flip-side of that opinion, where implicit returns are shorter and more readable, and there's an easy-at-a-glance way to tell if a function is just returning something or if it's doing more work than that. 

Implicitly returning functions carry a certain semantic value that allows me to offload an amount of cognitive load when reading code. That matters a lot to me, but it's just a preference.

I'm happy to be overruled if we have a company policy around this, but don't necessarily feel the need to privilege one preference over another, right?

orta:
Can we ditch any of the prettier owned rules (like the 120, semis etc) please

.vscode/launch.json
```diff
@@ -5,12 +5,17 @@
...
+      "port": 9229,
       "address": "localhost",
       "stopOnEntry": false,
       "runtimeExecutable": null,
-      "runtimeArgs": ["--inspect-brk", "./node_modules/.bin/jest", "-i"],
+      "runtimeArgs": [
+        "--inspect-brk",
+        "./node_modules/.bin/jest",
+        "-i",
+        "/Users/mykola/Work/metaphysics/src/schema/home/__tests__/home_page_artwork_module.test.js"
```
orta: we should remove this

package.json
```diff
@@ -59,6 +60,8 @@
     "dataloader": "^1.3.0",
     "dd-trace": "^0.2.0",
     "debug": "^2.2.0",
+    "deep-equal": "^1.0.1",
+    "dotenv": "^5.0.1",
     "express": "^4.13.3",
     "express-force-ssl": "^0.3.0",
```
orta: the idea that eslint can figure this out is impressive

scripts/dump-remote-schema.js
```diff
@@ -17,21 +17,29 @@ const httpConvectionLink = createHttpLink({
...
+  .then(schema =>
+    fs.writeFileSync(
+      path.join(destination, "convection.graphql"),
+      printSchema(schema, { commentDescriptions: true })
+    )
   )
-})
+  .catch(e => {
+    throw e
+  })
```
orta: feels a bit weird to just pass the throw up from top level down and then back, but we're planning on removing this script eventually anyway

src/integration/index.js
```diff
@@ -93,13 +89,14 @@ describe("Integration specs", () => {
...
+    it("is in sync with production", () =>
+      Promise.all([
         staging(query, { id: "cindy-sherman-untitled" }),
         production(query, { id: "cindy-sherman-untitled" }),
-      ]).then(([stagingResponse, productionResponse]) => {
+      ]).then(([stagingResponse, productionResponse]) =>
         deepEqual(stagingResponse, productionResponse).should.be.true()
-      })
+      )).catch(e => {
+      throw e
```
orta: also weird

mbilokonsky:
Hey so I added the `arrow-body-style` rule to always require brackets for the body of an arrow function, as per @damassi's preference. Turns out once I got the prettier conflicts resolved this was trivial to auto-apply, though I did have to do one pass against the tests manually since they're generally excluded from linting.

So, the current state of the build should be in keeping with what has been requested from everyone. I plan to open an RFC to question the merits of always using braces - as noted above, I do have a different opinion but am willing to concede to any institutional decisions. 

But, upon reflection, if the standard convention on this project has historically been to always use braces then there's no reason for this PR to change that.

damassi:
> added "arrow-body-style": ["error", "always"] to .eslintrc.json. This makes implicit returns invalid. 

I don't think implicit returns should be invalid as they're useful and very common, only that the previous behavior in Metaphysics be maintained. 

Would it be difficult to step backwards a few commits before the `arrow-body-style` rule was initially applied, so that the code looked as it did before (meaning, disable the rule completely)? We don't want to intruduce things like 

```
const ids = items.map(item => { return item.id }`
```

as this kind of code is unnecessarily verbose and uncommon.

damassi:
> Finally, here's the set of lint rules that I landed on. Note: about half of these were already in place, and the rest were my attempts to land on sensible defaults.

Can you list out the rules you added to the top-level description? There's no summary for reviewers to review. It's hard to see from the screenshot / diff exactly what has changed, and since this applies to the entirety of Metaphysics' codebase as well as the overall flow for each dev working in MP, it's worth proposing first what should be added, and then afterwards working in the agreed upon changes.

mbilokonsky:
Right, so the list of rules is here: https://github.com/artsy/metaphysics/pull/1063/files#diff-df39304d828831c44a2b9f38cd45289c

When you look at the diff you can see that (1) most of the rules we had before were just AirBNB plus a few styling overrides. The only one we kept was camelCase, the others were all removed as a part of the move to prettier for formatting concerns., and (2) almost all of the new rules are me disabling additional new behavior that came from the AirBNB version bump.

All this PR does is upgrade our dependencies to a more recent version of AirBNB, and adding a rule to anticipate and disable the default value that AirBNB added for `arrow-body-style` was not something that I anticipated having to do. It's one of the auto fixable rules, which I did not disable.

So to roll back the changes and not have any rule at all around arrow-body-style would require a completely new PR since the AirBNB upgrade driving this change was my very first commit.

Personally, any style rule that isn't explicitly enforced by prettier or a linter feels to me like it's not actually a rule. I understand that you feel an attachment to a convention that we 'use implicit return unless it's a place where I don't want to use implicit return', but unless there's some way to enforce that behavior via linting rules I'm not sure how to give you what you want. We can't auto-apply the absence of a rule.

The only way to do it would be to discard this PR, create a new one with the same linting rules as we have now PLUS a rule that disables AirBNB from having an opinion about `arrow-body-style`, do all of the fixes and reconciliation work and then merge that one. But note that we haven't added a rule to enforce the model you want, all we've done is avoided having a rule around this at all - which brings us back to, is it really a rule if it can't be enforced?

To me the solution is really straightforward - let's use the AirBNB defaults and yeah, sometimes we'll have functions that use implicit return. From the RFC thread it sounds like folks are mostly fine with that. The good news is that prettier doesn't 'fix' this because prettier has no rules around it, so your concerns about developers having their console.log-enabling code disappear on save are I think unfounded? They'd have to run `yarn lint --fix` for this to be autocorrected, which normally doesn't run until commit.

Anyway man I appreciate your feedback and understand your concerns. I also disagree with your conclusions and priorities. If there was a way for me to give you what you're asking for without redoing this entire PR I would but that's not how this has shaken out. So now the question is, is it worth redoing the entire PR?

damassi:
To restate -- this task was about upgrading ESLint and exploring Metaphysics. It wasn't about introducing new conventions. Conventions, which are opinions, are meant to be RFC'd. The `arrow-body-style` rule that you introduced changes our codebase substantially, and that kind of thing should be discussed. When the new AirBnb rules were introduced that changed the structure of Metaphysics there should have been a hard stop, since there are long-term cultural code conventions at play. 

Is it possible to upgrade ESlint, and leave the current rule-set alone? Again, linting is opinionated and is something to decide upon as a team in an effort to reduce rogue decision making. Just because something was introduced incorrectly a few commits back does not mean that it has to proceed, and can't be undone. I'm more than happy to take a look or to help you work through changes.

mbilokonsky:
I understand. I apologize, I did not know about the RFC process when I
started work in this ticket. I have subsequently done everything you’ve
asked - I’ve undone this change to the fullest extent possible without
starting over, and there is an open RFC where this is under discussion.

I’m not emotionally invested in this, I didn’t (until I asked for one
yesterday) have a jira ticket for it, I just did it because you asked me to
add promise rules and this is the path that that work took.

If your view is “This PR is incorrect” then I guess reject it and we can
all reallocate our time. You could RFC exactly the upgrade path that you
want and someone can PR a version that you’re happy with. But I’d probably
find myself in that RFC arguing for this exact configuration, given that it
passes all the rules and all the tests work and just by merging it we gain
new safety.

Personally I think there’s utility in this PR and you’re bike-shedding, but
I’m new here and I don’t want to tell anyone else how to be. Every
organization has its own rules and norms and I’m still learning Artsy’s.
I’m sorry if this has caused you any stress.

damassi:
@mbilokonsky - Yes, I believe it would be good to close this one and open a new with just the ESLint upgrade (plus associated linter changes to keep Metaphysics as it currently is). For changes that effect workflows or structure we should stick to the formal channels.

mbilokonsky:
So I'm attempting to do a new PR that just bumps ESLint and, after digging down and resolving the minimum viable changes required to our other dependencies that were required for `yarn lint` to run without a stack trace I get a ton of errors about how all of my rule definitions are out of date because things changed between version 3 and version 16. I'm currently stepping through major versions one at a time to see if one exists that'll do what you'd like.

mbilokonsky:
Hey everyone I'm going to close this PR, upon reflection I see that I made a big mistake by trying to both upgrade our versions for everything and add the new rule. I assumed that because we were using AirBNB rules v1.x it would be correct to move us to AirBNB rules v4.x, and so I assumed that this big change was helpful or necessary.

I'll be submitting a new PR today that just upgrades ESLint in isolation while preserving the exact set of rules in the old AirBNB ruleset, as much as is humanly possible. There may be places where this doesn't work, because a lot has changed across both ESLint and AirBNB-Config. Stand by and sorry if I got salty at all at your feedback, @damassi, I had a skewed perspective on what I was doing.

Cheers!

PR #1069, opened 2018-05-30
is_price_range should return true for single editioned artwork with price range

zephraph:
Fixes [EV-110](https://artsyproduct.atlassian.net/browse/EV-110)

### DIFF
--- a/src/schema/artwork/__tests__/artwork.test.js
+++ b/src/schema/artwork/__tests__/artwork.test.js
@@ -904,7 +904,20 @@ describe("Artwork type", () => {
       })
     })
 
-    it("returns false if artwork price is a range with single edition.", () => {
+    it("returns false if artwork price with single edition is not a range.", () => {
+      artwork.price = "$200"
+      artwork.edition_sets = [{}]
+      return runQuery(query, rootValue).then(data => {
+        expect(data).toEqual({
+          artwork: {
+            id: "richard-prince-untitled-portrait",
+            is_price_range: false,
+          },
+        })
+      })
+    })
+
+    it("returns true if artwork price with single edition is a range.", () => {
       artwork.price = "$200 - $300"
       artwork.edition_sets = [{}]
       return runQuery(query, rootValue).then(data => {
@@ -917,7 +930,7 @@ describe("Artwork type", () => {
       })
     })
 
-    it("returns false if artwork price is a range with multiple editions.", () => {
+    it("returns false if artwork price with multiple editions is a range.", () => {
       artwork.price = "$200 - $300"
       artwork.edition_sets = [{}, {}]
       return runQuery(query, rootValue).then(data => {
--- a/src/schema/artwork/__tests__/artwork.test.js
+++ b/src/schema/artwork/__tests__/artwork.test.js
@@ -904,9 +904,22 @@ describe("Artwork type", () => {
       })
     })
 
-    it("returns false if artwork price is a range with multiple editions.", () => {
+    it("returns false if artwork price is a range with single edition.", () => {
       artwork.price = "$200 - $300"
       artwork.edition_sets = [{}]
+      return runQuery(query, rootValue).then(data => {
+        expect(data).toEqual({
+          artwork: {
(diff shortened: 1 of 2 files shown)


alloy:
Code change looks good 👍 The CI failure is legit, though, and I’d also like to see a new test for the added logic.

src/schema/artwork/index.js
```diff
@@ -450,14 +454,15 @@ export const artworkFields = () => {
     is_price_range: {
       type: GraphQLBoolean,
       resolve: ({ price, edition_sets }) =>
-        has_price_range(price) && !has_multiple_editions(edition_sets), // eslint-disable-line max-len
+        has_price_range(price) &&
+        (has_single_edition(edition_sets) || !has_editions(edition_sets)), // eslint-disable-line max-len
```
oxaudo: This works but I think it might have been a bit more readable if we kept `has_multiple_editions` and made it check for more than 1 edition.

Something like:
```
const has_editions = edition_sets => {
   return edition_sets && edition_sets.length > 0
 }

const has_single_edition = edition_sets => {
  return has_editions && edition_sets.length > 1
}
```

And then here the code would not change at all and the change that you did below still valid. Just another way to do the same thing. I personally find combination of `&&`, `||` and `!` pretty confusing to read.

zephraph: Oh! Nice, yeah, good catch.

src/schema/artwork/__tests__/artwork.test.js
```diff
@@ -904,9 +904,22 @@ describe("Artwork type", () => {
       })
     })
 
-    it("returns false if artwork price is a range with multiple editions.", () => {
+    it("returns false if artwork price is a range with single edition.", () => {
```
alloy: It’s not returning `false`. The description could also maybe be updated to something like `it returns that an edition set with a single edition has no price range` to better describe the intent.

alloy: Well, that copy is not the right one for this test, so that goes to show how the intent isn’t made entirely clear at the moment 😄

src/schema/artwork/index.js
```diff
@@ -450,14 +454,14 @@ export const artworkFields = () => {
     is_price_range: {
       type: GraphQLBoolean,
       resolve: ({ price, edition_sets }) =>
-        has_price_range(price) && !has_multiple_editions(edition_sets), // eslint-disable-line max-len
```
oxaudo: Added bonus of removing lint comment! 👍

src/schema/artwork/__tests__/artwork.test.js
oxaudo: I think there should be 2 test for the artwork with one edition set. 
Something like: `Returns true for artwork with one edition set that is set as range`
and `Returns false for artwork with one edition with exact price`
I guess there is a potential 3rd case where we could test that it `Returns false for artwork with one edition set and no price info`

oxaudo:
One other thing to note and I might be off here, but I believe we have this convention to put the person that you want to decide when to merge this PR into `Assignee` field. I think there were some discussions about it on slack and there is some Notion document that is trying to formalize it (https://www.notion.so/artsy/Code-review-merge-proposal-9e405e955154472e8a701b8ac2a35d01). I personally look at this field to decide if I should be the one to press `merge` button. I think @alloy might have a more definitive opinion here though:)

alloy:
Yeah, for now at least that’s still the convention. The `@alloy` in the subject is a legacy thing that some still do out of habit, I suppose.

zephraph:
If that's a legacy pattern would it be worth looking into adding a danger rule to help phase it out and suggest the accepted approach?

alloy:
It certainly could be, but it may well be that some groups still like to also use the handle in subject approach and in general we don’t try to dictate too much. We make danger rules for repetitive PR feedback, whereas this one doesn’t happen too often afaik.

Maybe @orta has something to add to this?

alloy:
Feel free to merge when it’s a good time and staging is not possibly about to be promoted to production.

PR #1075, opened 2018-06-04
Bumped all versions, disabled all rules that were not already satisfied.

mbilokonsky:
A few things to note: this bumps us up to the most recent `eslint`, which we need in order to add some rules in a subsequent PR around promises.

Once we update `eslint`, we also have to update `eslint-config-airbnb`, `babel-eslint` and a bunch of other dependencies.

This is all fine and good except with the AirBNB rules - it turns out that we were like 12 major versions behind and the current AirBNB rules are much more strict than the ones we've had enabled. 

So for the purposes of this PR I have turned off every linting rule that is not already satisfied. Allowing our linting rules to change just because AirBNB updated theirs is not something we should just do, right?

One caveat I will throw out there, though: at what point does it make sense to stop using AirBNB's linting rules if our `.eslintrc` disabled so many? 

(Also one particular note: I had to disable the `quotes` rule because I guess it didn't used to have support for backticks. This is the one place where we have lost parity with existing rules in this upgrade, but fortunately it's easy enough to autofix. I would personally advocate for a rule that auto converts all strings to template strings using backticks, but that's a separate conversation we can have).

So, this PR is a staging ground. Once this gets merged we can have conversations about specific rules and we can add rules around promises and async/await. Thanks to @damassi in particular for helping me to understand why my initial approach to this PR was incorrect!


orta:
nice 👍 

I'll look at the peril issue, I had that this morning

damassi:
@orta - thoughts on just going with [`react-app`](https://github.com/damassi/react-styled-components-for-designers/blob/master/.eslintrc#L2)? (That or standard.) This seems to consolidate things nicely so we don't have to dance around airbnb.

orta:
Yeah, sounds nice and simple ( their opinions on linting is much closer to my own )  can do that i na separate PR 👍

mbilokonsky:
Would one of you mind merging this, @orta @damassi?

damassi:
@mbilokonsky - i can merge; did you want to handle orta's responsive above re `react-app`?

mbilokonsky:
Yeah, I'll do a separate PR that pulls out `airbnb` and pulls in `react-app`. Unless you think that an RFC to figure out exactly which rules we care about should come first? Whatever change I make next is going to touch a lot of files, I think, right? I don't want to end up in the same loop we had last time.

Is there clear direction here to just turn on the `react-app` rules with their defaults, or do we have preferences beyond that?

damassi:
I'll defer to @orta on this but IMO CRA is used widely enough that everything is likely tilted towards the absolute middle of the road / safe.

orta:
IMO, switch it, if it's tonnes of changes we can talk about it. But if it's less than ~20 source code changes, we just call it as an acceptable bit of churn for some simplicity.

mbilokonsky:
lol react-app requires an older version of eslint and is not compatible with this PR. :-D Workarounds include skipping a few rules, but ol' Dan Abramov says not to do that.

So if this gets us there with a few version decrements then is that fine? Or should I look at the other one you mentioned instead, standard?

damassi:
I'm ok with locking version numbers to whatever `react-app` uses since that seems pretty safe 👍

PR #1079, opened 2018-06-05
Move from `airbnb` to `react-app` eslint rules

mbilokonsky:
As discussed with @orta and @damassi in [this PR](https://github.com/artsy/metaphysics/pull/1075) I've gone ahead and replaced the AirBNB linter rules with the react-app rules. I've left almost all rules on their default settings, but I did disable the ones you see in .eslintrc.json. The import rule is disabled because there are a few places where we intentionally reorder our imports. The others are warnings that are sufficiently plentiful to be annoying but don't actually break the build so I've left them off for now - if someone is looking for low-hanging fruit in the future they can look there.

Two things to note:
1. I've manually disabled the `promise/always-return` rule anywhere it was flagging problems in our tests. It's common enough to invoke promise chains for their side effects in tests that I didn't feel like it was appropriate use of time to fix the hundreds of failures we were getting, but I did want to make sure this rule ran against production code. Having the `/* eslint-disable promises/always-return */` at the top of our tests is ugly but this was the closest compromise I could come up with, and @damassi [agrees with the reasoning](https://artsy.slack.com/archives/C02BC3HEJ/p1528237967000072). I grouped all of the test disabling into [a single commit](https://github.com/artsy/metaphysics/pull/1079/commits/aeb506d9fd4ac5e7443d0887142e3963294ea89e).

2. There was one place where, after making all of the rules pass, I got a failing test. Specifically, the coverImage resolver in `src/schema/show.js` was crapping out in our unit tests because `body` came back undefined. I'm not sure how or why my changes would have introduced an `undefined` where one had not previously existed, and I'd appreciate getting some additional eyes on it. Regardless, [I stuck an `if (!body) { return undefined; }` at the top of the function block](https://github.com/mbilokonsky/metaphysics/blob/e80e39b4259f93c8e824ea2b8c8613bf024ac090/src/schema/show.js#L180-L182) and now tests all pass again. This makes me mildly nervous.


mbilokonsky:
Ok, some more information on the test that started failing. When you just run jest I guess you don't see warnings, but when I run jest through the VSCode debugger I get the following output *on the master branch*:

image

This suggests to me that the test has always been bad, in that the initial setup in the mock does not correctly pass an object with a `body` property; it's just that the test isn't making any assertions about error handling.

I'm still not 100% clear on why the test arbitrarily stopped passing, but I am now confident that the fix for the test was correct.

scripts/dump-schema.js
```diff
@@ -13,20 +13,19 @@ if (destination === undefined || !fs.existsSync(destination)) {
...
+graphql(schema, introspectionQuery)
+  .then(result => {
+    return fs.writeFileSync(
       path.join(destination, "schema.json"),
       JSON.stringify(result, null, 2)
     )
-  },
-  error => {
+  })
+  .catch(error => {
```
damassi: nice 👍

src/lib/loaders/api/loader_without_authentication_factory.js
```diff
@@ -41,7 +41,7 @@ export const apiLoaderWithoutAuthenticationFactory = (
             const clock = timer(key)
             clock.start()
             return new Promise((resolve, reject) => {
-              cache.get(key).then(
+              return cache.get(key).then(
```
damassi: It's weird that the rule requires a return because promises are pretty nice for managing async flow independent of data being passed back up. I wonder what the reasoning is?

damassi: Ah https://github.com/xjamundx/eslint-plugin-promise#rules

mbilokonsky: We had surprisingly few issues with most of those rules, so go us! The nested promises and callbacks-in-promises are fair warnings that, if we fix them, should make the code a bit more reasonable.

PR #1084, opened 2018-06-07
@acjay: Lower production memory limit to kill runaway pods earlier

joeyAghion:
We've seen pods get into a bad state, with high latency and high memory usage. As a mitigating measure, we [discussed](https://artsy.slack.com/archives/CA8SANW3W/p1528376895000157) lowering the memory limit, in order to kill those runaway pods earlier in their degradation.

This brings the `production.yml` in line with `staging.yml`. Both of these were part of the original transition (https://github.com/artsy/metaphysics/pull/968). This has been applied already (via `hokusai production update`), but is worth monitoring for a bit.

CC @izakp

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -30,10 +30,10 @@ spec:
         resources:
           requests:
             cpu: "0.5"
-            memory: "512Mi"
+            memory: "256Mi"
           limits:
             cpu: "1"
-            memory: "768Mi"
+            memory: "512Mi"
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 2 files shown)


acjay:
Sounds good to me. Does this hit a sweet spot between the normal and runaway regimes? 

Also, seems there are merge conflicts.

joeyAghion:
This _seemed_ to hit a sweet spot, in that when everything is healthy the pods hover in the 200-300 MB range. However in certain periods (perhaps when a backing service latency increases slightly or when there's a high volume of large queries) the pods either can't start or immediately exhaust the prescribed limits. I've reverted to 512/768 MB memory request/limit and updated staging's config to match.

I do wonder if there's some in-between sweet spot but it seems like we need to monitor things during a spike to confirm or get meaningful results.

acjay:
Sigh. Oh well.  Maybe there's a way to have duration-over-threshold as a trigger? In any case, you've got my 👍. Reassigning to you to merge when you're happy.

PR #1085, opened 2018-06-07
Rename external_operator to externalOperator

williardx:
Causality expects a JWT with `externalOperator` instead of `external_operator`, so we're renaming to match what Causality expects.


starsirius:
Thanks for the quick fix! 👍

PR #1088, opened 2018-06-08
#migration scale up replicas

izakp:
We've been running between 5 and 6 pods since the downtime the other week, which led to the cache investigation.  The spec was still set to the old min replicas and this is definitely insufficient.

Let's throw more resource at it.  We have the capacity.

# Migration

`hokusai staging update && hokusai production update`

cc @joeyAghion @alloy @mzikherman

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -6,7 +6,7 @@ metadata:
 spec:
   strategy:
     rollingUpdate:
-      maxSurge: 1
+      maxSurge: 2
       maxUnavailable: 0
     type: RollingUpdate
   template:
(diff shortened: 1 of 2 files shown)


hokusai/production.yml
```diff
@@ -6,7 +6,7 @@ metadata:
 spec:
   strategy:
     rollingUpdate:
-      maxSurge: 1
+      maxSurge: 2
```
izakp: i.e. start new pods in batches of 2 - this should reduce the time it takes for deploys / restarts to complete

sweir27:
So many pods! Thanks for this. Feel free to merge when you're ready.

PR #1089, opened 2018-06-08
add metrics connection and expose on /metrics

izakp:
Let's get some internal metrics!

We can hit `/metrics` via a [kubernetes probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/) and take action on it, i.e. either take the pod out of service or restart it.

(We should still implement a health check endpoint to indicate if the application is fully bootstrapped and ready to serve traffic.)  cc @joeyAghion


joeyAghion:
Needs a rebase.

Nice! Looking at the [example metrics](https://github.com/idanto/express-node-metrics) this does seem like it could be illuminating. @alloy agree?

My only question is that it might be helpful to accept a `reset=true` parameter at the new endpoint to be able to narrow the time window and decrease the noise.

izakp:
@joeyAghion Why rebase?  No merge conflicts... I wouldn't want to reset the process metrics as we want to see how it grows through the length of the process' lifetime.

izakp:
Ah I guess there is a lot in the lockfile after doing a `yarn install` - how do y'all generally commit updates to Yarn's lockfile @alloy ?

alloy:
Sweet!

Just don’t look behind the curtain’s of npm/yarn, is kinda how you have to deal with it 🤷‍♂️

PR #1092, opened 2018-06-11
Only collect and expose process metrics

izakp:
Collecting express metrics as middleware was creating a new route for each GraphQL query.  Not sure why the query params were collected - the express [getRoute helper](https://github.com/idanto/express-node-metrics/blob/master/src/middleware.js#L32) is called in the middleware module of the metrics library.

In any case I am hesitant to throw this out on production - let's just start with basic Node and GC metrics before re-enabling this (or possibly just writing our own middleware function).

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,9 +37,8 @@ const app = express()
 app.use(compression())
 
 if (enableMetrics) {
-  app.use(expressMetrics.middleware)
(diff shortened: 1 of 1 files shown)


alloy:
Looks good, merge on green 👍

PR #1093, opened 2018-06-11
Add field for returning whether user has credit cards at all

sweir27:
We tracked the bug this morning to the fact that we were checking for `has_qualified_credit_cards` in one place, but only checking whether the user had credit cards at all (qualified or not) in another place.

This adds a field for us to check whether or not a user has a credit card on file at all.


sweir27:
Got the thumbs up so going to merge to unblock the force PR!

PR #1094, opened 2018-06-11
hotfix - logging redis client info

izakp:
move redis client info to verbose - drop logging full command queue as it's too noisy

cc @alloy @mbilokonsky


izakp:
merge on green

PR #1097, opened 2018-06-12
WIP: MP endpoint to confirm/reject/finalize orders

sepans:
MP endpoint to confirm/reject/finalize orders


jonallured:
With artsy/gravity#11773 merged, let's rebase this on top of master and see about getting it merged too - sound good?

src/schema/index.js
```diff
@@ -157,8 +160,11 @@ const schema = new GraphQLSchema({
       updateCollectorProfile: UpdateCollectorProfile,
       updateMyUserProfile: UpdateMyUserProfileMutation,
       updateConversation: UpdateConversationMutation,
-      updateOrder: UpdateOrderMutation,
+      confirmOrder: ConfirmOrderMutation,
+      finalizeOrder: FinalizeOrderMutation,
+      rejectOrder: RejectOrderMutation,
       submitOrder: SubmitOrderMutation,
+      updateOrder: UpdateOrderMutation,
```
ashkan18: with these more specific mutations, do we still need `updateOrderMutation`?

sepans: Most likely not. `updateOrder` and `submitOrder` were there from another project for checkouts in emission which didn't get pushed to production. After finalizing MP API we can remove/refactor them.

PR here: https://github.com/artsy/metaphysics/pull/1009

sweir27: This isn't being used in production, so you can feel free to remove.

sweir27: (actually let me verify-- I know it's not being _consumed_ but I'm not sure if its presence here will break anything? I feel like this is auto-generated https://github.com/artsy/reaction/blob/f765cb4a46e3f68c62de243ad193f68e69582741/data/schema.graphql#L5609)

src/lib/loaders/loaders_with_authentication/gravity.js
```diff
@@ -138,6 +138,21 @@ export default (accessToken, userID, opts) => {
     ),
     updateMeLoader: gravityLoader("me", {}, { method: "PUT" }),
     orderLoader: gravityLoader(id => `order/${id}`, {}, { headers: true }),
+    confirmOrderLoader: gravityLoader(
```
sweir27: should we rename to `approveOrderLoader` for consistency?

src/schema/order/finalize_order_mutation.js
```diff
@@ -0,0 +1,32 @@
+import { GraphQLString, GraphQLNonNull } from "graphql"
+import { OrderType } from "schema/me/order"
+import { mutationWithClientMutationId } from "graphql-relay"
+
+export default mutationWithClientMutationId({
+  name: "FinalizeOrderMutation",
+  description: "Conforms an order",
```
sweir27: #minor this should say "Finalizes" order?

src/schema/order/reject_order_mutation.js
```diff
@@ -0,0 +1,32 @@
+import { GraphQLString, GraphQLNonNull } from "graphql"
+import { OrderType } from "schema/me/order"
+import { mutationWithClientMutationId } from "graphql-relay"
+
+export default mutationWithClientMutationId({
+  name: "RejectOrderMutation",
+  description: "Conforms an order",
```
sweir27: ditto-- "Rejects an order"

src/schema/order/confirm_order_mutation.js
```diff
@@ -0,0 +1,32 @@
+import { GraphQLString, GraphQLNonNull } from "graphql"
+import { OrderType } from "schema/me/order"
+import { mutationWithClientMutationId } from "graphql-relay"
+
+export default mutationWithClientMutationId({
+  name: "ConfirmOrderMutation",
+  description: "Conforms an order",
```
sweir27: ditto to previous rename comment-- should this be "ApproveOrderMutation"? In any case "Conforms" --> "Confirms" 😄

src/lib/loaders/loaders_with_authentication/gravity.js
```diff
@@ -138,6 +138,21 @@ export default (accessToken, userID, opts) => {
...
+      id => `order/${id}`,
+      { state: "approved" },
+      { method: "PUT" }
+    ),
+    finalizeOrderLoader: gravityLoader(
```
sweir27: @ashkan18 does this map how stress would receive these mutations? If we're talking directly to gravity we could have the three separate mutations but a single loader here that passes in different params from the mutation itself.

src/lib/loaders/loaders_with_authentication/gravity.js
sepans: Yes, I was confused about what the state name going to be. In some discussion/docs it was `confirm` but in stress and gravity it is `approved`. Are we going with `approve`? cc: @jonallured @ashkan18

src/lib/loaders/loaders_with_authentication/gravity.js
ashkan18: Stress also has separate mutations for each state, Currently:
- `CreateOrder`
- `SubmitOrder`
- `ApproveOrder`
- `RejectOrder`
and
- `FinalizeOrder`
Lets decide on these names, i can change names on stress mutations too but lets finalize the names internally. Do these names make sense? Is it easy to make changes here? I can easily change names in Stress too if we want other names.

src/lib/loaders/loaders_with_authentication/gravity.js
ashkan18: added some comments about this below :)

src/schema/index.js
sepans: Yes I think so. Emission also has a similar file that is the output of MP `dump-schema` copied using `sync-schema` [here](https://github.com/artsy/reaction/blob/master/package.json#L43).

PR #1099, opened 2018-06-13
hotfix - use tcp health check to bypass force-ssl middleware

izakp:
Express [force-ssl middleware](https://www.npmjs.com/package/express-force-ssl) is causing the health check route to return a 301 redirect when hit via the Kubernetes readiness probe.

The middleware doesn't appear to allow you to disable for certain routes.  Until we fix use a tcp probe on port 3000 for a readiness check.

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -35,9 +35,8 @@ spec:
             cpu: "1"
             memory: "768Mi"
         readinessProbe:
-          httpGet:
+          tcpSocket:
             port: 3000
-            path: /health
           initialDelaySeconds: 5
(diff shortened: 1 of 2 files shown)


alloy:
It looks like we could enable `trustXFPHeader` if our load balancer strips those out when clients were to include it?

izakp:
Ah. `trustXFPHeader` is already enabled.  How it works is the ELB adds the header `X-FORWARDED-PROTO:https` when the client -> load balancer connection is over SSL (and SSL terminates at the load balancer).  But good find - we can hack the health check by setting this header.

PR #928, opened 2018-02-15
[cache] Switch from memcache to redis to get rid of memjs.

alloy:
Closes #869.

This switches us back to Redis to fix the issue of the memos client getting into a state where it can’t reconnect to the service anymore. Hopefully this also reduces some back pressure during high load times, but at the very least remove noise that makes it harder to diagnose performance issues.

Of the various available Heroku addons I for now chose openredis, as they are the cheapest and I think we already use that in other apps (?).

We currently pay $300/5GB for memcache, the available options with openredis are $200/3.4GB or $400/5.6GB. I guess we’ll have to shell out the extra $100 to get at a minimum the same amount we have now?

Unfortunately openredis does not have a free option that we could use in staging, the smallest is $8. Would you rather have we pay a little to use the same service-provider in staging or use one of the other redis addons that provide a free option?

PR #1100, opened 2018-06-13
#migration switch cache to memcached driver

izakp:
We've gathered sufficient evidence that cache timeouts:
- Occur in bursts in a single process
- Once triggered have a cumulative/cascading effect on further cache requests

Redis drivers use a pipelining feature to batch queries and reduce network round-trip-time, but forces the Redis server to [queue the replies](https://redis.io/topics/pipelining).  It appears that in a given resolver we may enqueue 100s or even 1000s of cache hits, which get pipelined together, and when we start timing out cache requests, it does not de-queue them / remove them from the pipeline.

The Redis drivers https://github.com/NodeRedis/node_redis and https://github.com/luin/ioredis do _not_ implement connection pooling as they see it as unnecessary.  However, it seems we have exhausted the throughput of a single socket as a cache connection, given that Redis' internal slowlog shows slowest operations at only ~70ms

The [initial switch to Redis](https://github.com/artsy/metaphysics/pull/928) was motivated by the fact that the `memjs` driver was particular to the Heroku "Memcachier" addon and did not handle server reconnections successfully.

The https://github.com/3rd-Eden/memcached driver is better recommended (and although it doesn't appear to be actively maintained it seems to be the only mature driver available), supports connection pooling, and handles server disconnections / reconnections well (at least in terms of my local testing).

### Migration

Will need to create an Elasticache Memcached instance for both staging and production, then set the `MEMCACHED_URL` env var and remove the `REDIS_URL` env var before deploying.


alloy:
> It appears that in a given resolver we may enqueue 100s or even 1000s of cache hits

This surprises me. Do you mean a single query as opposed to a single resolver? If not, could you please point me to the data for this, because that seems very excessive and something we should look into.

index.js
```diff
@@ -66,11 +66,11 @@ function bootApp() {
   })
 
   app.get('/health', (req, res) => {
-    if (cache.isAvailable()) {
-      res.status(200).end()
-    } else {
-      res.status(503).end()
-    }
+    cache.isAvailable().then((stats) => {
```
alloy: Do the `stats` hold any data that would influence the outcome or is any response instead of an error considered good enough?

src/lib/cache.js
```diff
@@ -168,13 +125,23 @@ export default {
...
-    return client.ping()
+    return new Promise((resolve, reject) => {
+      client.stats((err, resp) => {
+        if (err) {
+          error(err)
+          reject(err)
+        } else {
+          if (resp){
+            resolve(resp)
+          }
```
alloy: This looks like there’s the potential for the promise never resolving or rejecting. That’s not a problem for the promise, but seeing as the caller isn’t handling this possibility in any way it could mean there’s HTTP health requests that never get responded to.

izakp:
@alloy re:

> It appears that in a given resolver we may enqueue 100s or even 1000s of cache hits

Fair - those numbers are just spitballing. Assuming these bursts of queries are scoped to the resolver  is an inadequate assessment - it could well be manifest at the query level that includes multiple resolvers.

It's in the log archives now, but searching Papertrail for `commandQueue` I saw the queue (waiting on replies from Redis) backed up to 186 requests at one point https://artsy.slack.com/files/U0QRCNUAV/FB67XSJAK/screen_shot_2018-06-11_at_1.24.03_pm.png

See https://artsy.slack.com/archives/C04DU9H0L/p1528735987000445

We need to drill down into these metrics to find out more:
- https://app.datadoghq.com/apm/service/metaphysics.graphql-query/graphql.query
- https://app.datadoghq.com/apm/service/metaphysics.graphql-resolver/graphql.resolver
- https://app.datadoghq.com/apm/service/metaphysics.redis/redis.command

index.js
izakp: Just checking connectivity with this as there is no command analogous to Redis' `PING` see https://docs.oracle.com/cd/E17952_01/mysql-5.6-en/ha-memcached-stats.html but I don't think any of that is relevant to a service-level check

src/lib/cache.js
izakp: Will remove the nested `if` - it's redundant if either `err` or `resp` is always provided, but true that it doesn't guard against the edge case where neither is.

izakp:
@saolsen will this need any updates for the datadog tracer to start measuring memcached?

izakp:
review app up at https://metaphysics-staging-blue.artsy.net/

izakp:
Force staging configured to use https://metaphysics-staging-blue.artsy.net/

izakp:
Deleted review app

izakp:
Provisioned the memcached clusters in https://github.com/artsy/infrastructure/pull/89 and configured `MEMCACHED_URL` for staging and production so OK to merge.

acjay:
Maybe a silly question, but would it be ideal to be able to switch cache back-ends at least somewhat on the fly?

izakp:
@acjay 

> would it be ideal to be able to switch cache back-ends at least somewhat on the fly

As most of this has to do with the client implementation (connection pooling vs pipelining) I don't think we'd gain much here by straddling two backends - we can always revert (will keep the Redis servers around for a little bit while we assess memcached performance on prod)

izakp:
No Memcached datadog integration - will have to use https://datadog.github.io/dd-trace-js/#manual-instrumentation__anchor

izakp:
Thanks for the cache tracing @alloy !

alloy:
LGTM 👍 

Merge on green.

PR #1101, opened 2018-06-13
Fix Http health checks

izakp:
Set `X-FORWARDED-PROTO:https` header to bypass redirect to https on the health check

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -38,6 +38,9 @@ spec:
           httpGet:
             port: 3000
             path: /health
+            httpHeaders:
+            - name: X-FORWARDED-PROTO
+              value: https
           initialDelaySeconds: 5
           periodSeconds: 5
       affinity:
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -35,8 +35,9 @@ spec:
             cpu: "1"
             memory: "768Mi"
         readinessProbe:
-          tcpSocket:
+          httpGet:
             port: 3000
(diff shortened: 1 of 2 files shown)


alloy:
Just to be sure, the ELB will strip out the header if clients in front of it were to include it? We only want to allow https bypass internally.

izakp:
The ELB always writes (i.e. overwrites) this header, so if spoofed by the client (say https over an http connection) the backend won't see it as such.

```
$ curl -v http://metaphysics-staging.artsy.net -H 'X-Forwarded-Proto:https'
* Rebuilt URL to: http://metaphysics-staging.artsy.net/
*   Trying 54.164.93.13...
* TCP_NODELAY set
* Connected to metaphysics-staging.artsy.net (54.164.93.13) port 80 (#0)
> GET / HTTP/1.1
> Host: metaphysics-staging.artsy.net
> User-Agent: curl/7.54.0
> Accept: */*
> X-Forwarded-Proto:https
>
< HTTP/1.1 301 Moved Permanently
< Content-Type: text/plain; charset=utf-8
< Date: Wed, 13 Jun 2018 09:56:32 GMT
< Location: https://metaphysics-staging.artsy.net/
< Vary: Accept, Accept-Encoding
< X-Powered-By: Express
< Content-Length: 72
< Connection: keep-alive
```

alloy:
Ah right right, and seeing as the MP pods aren’t directly accessible it won’t matter. Makes sense :+1:

PR #1104, opened 2018-06-15
#experimental split cache client to reader and writer clients

izakp:
Making an experiment to split cache reads and writes onto two separate clients, each with their own connection pools - in this way ensure that no `SET` or `DEL` commands can be interleaved with `SET` commands.

cc @acjay @alloy

### DIFF
--- a/src/lib/__tests__/cache.test.js
+++ b/src/lib/__tests__/cache.test.js
@@ -1,6 +1,6 @@
 /* eslint-disable promise/always-return */
 import zlib from 'zlib'
-import cache, { client } from "lib/cache"
+import cache, { writeClient } from "lib/cache"
 
 describe("Cache", () => {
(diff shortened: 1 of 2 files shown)


alloy:
Interesting! Is there an intrinsic problem you’re thinking of with interleaving them?

izakp:
I want to rule out the possibility that our cache read latency is at all compounded by `GET` commands queued up behind `SET` commands (as we do an async `SET` for every `GET`).  Will instrument the number of slow cache reads / sec first.

Marking this as #experimental - will revert if it proves no performance benefit

izakp:
See https://github.com/artsy/metaphysics/pull/1105

izakp:
No observed change in cache latency or read timeouts.  Reverting in https://github.com/artsy/metaphysics/pull/1106

PR #1105, opened 2018-06-15
add statsd lib - counter and timer for cache slowlog

izakp:
Set `STATSD_HOST=datadog` in staging and production

cc @alloy @acjay

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "graphql-relay": "^0.5.4",
     "graphql-tools": "^3.0.0",
     "graphql-type-json": "^0.1.4",
+    "hot-shots": "^5.6.1",
     "i": "^0.3.5",
     "invariant": "^2.2.4",
     "jwt-simple": "^0.5.0",
--- a/src/config.js
+++ b/src/config.js
@@ -55,6 +55,8 @@ const {
   RESIZING_SERVICE,
   RESOLVER_TIMEOUT_MS,
   SENTRY_PRIVATE_DSN,
+  STATSD_HOST,
+  STATSD_PORT,
 } = process.env
 
 const mustHave = {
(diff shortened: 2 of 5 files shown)


joeyAghion:
+1 to incorporating statsd, but don't we already have essentially this metric in the full latency histogram?

https://cl.ly/2G2Z0r0x2L3E

izakp:
I want to see a count / second of values over out timeout threshold.  The data is essentially there but I don't think we can transform it to reflect timeouts / second

izakp:
Creating an alert averages values over its monitoring window (and you can't make a query like `SELECT * from cache.hits where cache.duration > 200`)

izakp:
Gonna merge this in so I can verify stats are getting through to datadog

izakp:
Following up on the switch to Memcached with connection pooling in https://github.com/artsy/metaphysics/pull/1100

Currently Metaphysics is running with a connection pool of 2 Memcached connections per-process (the driver [scales connections automatically](https://github.com/3rd-Eden/jackpot/blob/master/index.js#L269) based on write throughput), and I've observed no significant change in overall cache query duration or number of slow queries / minute [when scaling out overall connections](https://github.com/artsy/metaphysics/compare/master...izakp:fea-client-pooling) or routing GET / SET requests [to different connection pools](https://github.com/artsy/metaphysics/pull/1106).

I think this invalidates my conclusion that a single connection bottleneck or Redis server-side queuing was really the culprit.  Assuming we don't have a network bottleneck or significant server-side response queueing, my only other idea is that latency is being introduced in the event loop when all cache and http requests return.

(Still learning about what gets executed where and how in NodeJS) but would all Promises resolving / callbacks from cache / http I/O streams be executed on the event loop, once async I/O completes and data is ready on the stream buffer?

If this seems likely, the next step would be to instrument the connection's [responseBuffer](https://github.com/3rd-Eden/memcached/blob/master/lib/memcached.js#L185) as a gauge but as it's scoped to a private var in the context of a factory, I think it will take some surgery, or even forking the memcached library.

@mbilokonsky feel free to jump in if you want to take a crack at this / have other ideas

alloy:
> would all Promises resolving / callbacks from cache / http I/O streams be executed on the event loop, once async I/O completes and data is ready on the stream buffer?

Yes.

PR #1107, opened 2018-06-18
Replace express metrics with appmetrics to statsd

izakp:
Replacing node express metrics module with https://github.com/RuntimeTools/appmetrics as it:
- Appears to report eventloop latency in a more meaningful way
- Emits events that we can hook into to push metrics to statsd
- Is more robust (AFAICT) / fully-featured and developed by IBM

Push eventloop latency to statsd 

cc @cavvia


izakp:
Want to get this out so I can add more instrumentation

src/lib/stats.js
```diff
@@ -1,22 +1,38 @@
...
+  const appmetrics = require("appmetrics")
+  appmetrics.configure({
+    mqtt: 'off'
+  })
+  const monitoring = appmetrics.monitor()
+
+  monitoring.on('eventloop', (eventloopMetrics) => {
+    statsClient.timing('eventloop.latency.min', eventloopMetrics.latency.min)
+    statsClient.timing('eventloop.latency.max', eventloopMetrics.latency.max)
+    statsClient.timing('eventloop.latency.avg', eventloopMetrics.latency.avg)
```
joeyAghion: Usually statsd takes care of tracking the min/max/avg and just expects to receive timing data in ms. I'm curious if/why these are required as separate metrics here.

izakp: Good point - we're seeing min/max/avg of max, min/max/avg of min and min/max/avg of avg

Should probably use a Gauge here https://docs.datadoghq.com/developers/dogstatsd/data_types/#gauges as it is designed to track values _over time_ and doesn't take calculate min/max and percentiles, etc.

joeyAghion: A gauge might also be imperfect, since we'll be reporting this metric every few seconds from multiple concurrent processes (so any visualization of the gauge metrics will be mixed up.

izakp: @joeyAghion note the metrics are tagged with `hostname: os.hostname()` which on k8s resolves to the pod name so metrics can be segmented by pod

izakp: The only tradeoff with a Gauge is that if you send multiple metrics to dogstatsd during its flush period (of 10s) it will not be averaged, but de-duped.  From https://docs.datadoghq.com/developers/metrics/gauges/

> Each value in the stored timeseries is the last gauge value submitted for that metric during the statsd flush period.

izakp: I think timers are actually fine here as the max value seems to be the most useful in performance monitoring.  Also in terms of granularity, I would rather that dogstatsd computes the stats over each metric it receives in its 5-second flush interval rather than drop all but the last one.

PR #1108, opened 2018-06-18
async JSON parsing and stringifying

izakp:
Looking into what operations get processed on the event loop versus the worker pool - [this article](https://nodejs.org/en/docs/guides/dont-block-the-event-loop/) pointed out that the native `JSON` module is synchronous.  Given how many cache queries and API responses we are serializing/deserializing in a single request, this may well be a source of latency on the event loop.

Depends on https://github.com/artsy/metaphysics/pull/1107 to measure the event loop latency before and after this is deployed

cc @cavvia


alloy:
Nice 👌 Feel free to merge when you think is a good time.

mbilokonsky:
Ah, I see the latency measurement will allow us to time it. LGTM I'll merge on green!

src/lib/cache.js
```diff
@@ -21,12 +22,14 @@ const VerboseEvents = ["issue", "failure", "reconnecting", "reconnect", "remove"
...
 const deflateP = (dataz) => {
   return new Promise((resolve, reject) =>
-    zlib.deflate(JSON.stringify(dataz), (er, deflatedData) => {
-      if (er) {
-        error(er)
-      } else {
-        resolve(deflatedData)
-      }
+    yj.stringifyAsync(dataz, (err, data) => {
+      zlib.deflate(JSON.stringify(dataz), (er, deflatedData) => {
```
acjay: If we are actually CPU-bound, is it possible that compression could make things worse?

acjay: Hmm, I see from the article in the PR description that `zlib` runs on a worker thread. So maybe not an issue from an event loop blocking perspective, but still, would add load if our worker threads are also CPU-bound.

mbilokonsky: That load has to happen somewhere, though, right? Moving it to workers can only help, I think?

acjay: Sorry, this wasn't pertaining to using `yj`, but to the recently introduced compression.

mbilokonsky: ooooh, I see what you mean. That's a good point, if we're CPU bound compression would make it worse not better!

PR #1109, opened 2018-06-18
Feature flag cache compression

izakp:
Feature flag cache compression - may want to compare metrics (especially slow cache reads) with it on or off

Disable cache key compression by setting `CACHE_COMPRESSION_DISABLED=true`


joeyAghion:
I still think it would be simpler if the true/false (`0`/`1`) value of this flag was used as a component in the cache key. It could be applied transparently by the client, with callers left unchanged. That way we wouldn't have to clear or swap the back-ends.

alloy:
If we want the ability to disable/enable to remain long term, then I definitely agree with @joeyAghion

izakp:
Yea agreed @joeyAghion - we may want to disable then re-enable it.

izakp:
@joeyAghion as far as implementation goes...

> true/false (0/1) value of this flag was used as a component in the cache key

Keys are generally URLs of endpoints.  Would you advise creating a composite key using some delimeter, i.e. `{URL}--{flag}`?  Any other ideas for a non-url-encodable character we can use as a delimiter?

alloy:
@izakp Switch them around maybe? `{flag}:{URL}`

izakp:
Also I manually ran tests with `CACHE_COMPRESSION_ENABLED` set to `true` and then `false` but how would I run the cache test suites twice while overriding this constant?

alloy:
@izakp Seeing as `CACHE_COMPRESSION_ENABED` is coming out the `config` module, rather than directly from `process.env`, I think it may be possible to use jest module mocking to mock the `config` import.

izakp:
@alloy I can't find a way to mock the config or cache modules with Jest and actually set the constant `cache.cacheCompressionEnabled` so that the cache module uses it as such and turns compression on/off - it only mocks in the scope of the test suite.  Following examples in https://stackoverflow.com/questions/42977961/how-to-mock-an-exported-const-in-jest

izakp:
@alloy okay figured out how to mock it in the config module - now testing both enabled and disabled.

izakp:
Set `CACHE_COMPRESSION_ENABED=true` in staging and production as it now defaults to false

alloy:
Yeah I guess this is fine too, as most people won’t be touching code that would chang this and so it’s not strictly necessary to run this test during dev with eg `yarn jest —watch`

alloy:
Needs an update btw. Merge on green

src/lib/cache.js
```diff
@@ -19,6 +20,16 @@ const isTest = NODE_ENV === "test"
 
 const VerboseEvents = ["issue", "failure", "reconnecting", "reconnect", "remove"]
 
+const cacheKeyDelimiter = "::"
+
+export const cacheKey = (key) => {
+  if (CACHE_COMPRESSION_ENABLED) {
+    return "1" + cacheKeyDelimiter + key
```
izakp: @joeyAghion @alloy should I change this line to `return key` so as to preserve the existing compressed cache keys?  Otherwise we are going to effectively clear the cache when this is deployed.

izakp:
Backwards compatibility - preserve existing compressed cache and flip to `CACHE_DISABLED` env var

src/lib/__tests__/cache.test.js
```diff
@@ -1,8 +1,89 @@
...
+        })
+      })
+    })
+  })
+})
+
+
+describe("Cache with compression disabled", () => {
+  config.CACHE_COMPRESSION_DISABLED = false
+  expect(config.CACHE_COMPRESSION_DISABLED).toBe(false)
```
alloy: Does this test serve a specific purpose?

src/lib/cache.js
```diff
@@ -110,25 +125,45 @@ function _set(key, data) {
...
+        cacheKey(key),
+        payload,
+        CACHE_LIFETIME_IN_SECONDS,
+        err => {
+          if (err) error(err)
+        }
+      )
+    }).catch(err => {
+      error(err)
+    })
```
alloy: FYI Because the `error` function takes a single `err` parameter, you don’t even really need to create this extra closure and could just do `.catch(error)`.

src/lib/cache.js
alloy: Good call 👍

alloy:
Still needs an update.

src/lib/__tests__/cache.test.js
izakp: Just checking that the config module was mocked properly

src/lib/cache.js
izakp: Righto

src/lib/__tests__/cache.test.js
alloy: Ok, fair enough. Could you maybe add a comment about that (as it’s so out of the ordinary of other tests) and add an empty line below it so the following description isn’t harder to read due to the clutter?

alloy:
Merge on green 👍

PR #1110, opened 2018-06-18
#WIP Proposal to change the downstream caching strategy #doNotMerge

izakp:
I want to propose a change to our current cache setup because as it currently stands:
- We make API requests to Gravity on both cache hits and misses
- If cache keys exist, they are only refreshed _after_ the first read
- Key expiration is set to 30 days

These parameters have the effect that for up to 30 days we will serve stale data on the first `GET` of a cached endpoint for which any of its resolved domain objects have been updated.

The motivation for this caching strategy was to return data as quickly as possible to the client.  However, as I understand it:
- The performance of a Metaphysics query is going to be at least the worst time of _any_ cache request or HTTP request to a downstream service, as all Promises created by data loaders need to resolve (probably oversimplifying this?  @acjay and I looked into this a bit but please clarify if incorrect)
- We made a design decision to make a single, long-running GraphQL query, with no upper bound on query execution, in lieu of multiple RESTful queries to performant endpoints

Given these two facets, I wonder if the tradeoff in data integrity is worth the blanket approach to caching and what the performance impact to GraphQL queries would be when only caching the responses from high-throughput endpoints, with a low TTL.  This change would:
- Guarantee stale data only remains in the cache for a period of 5 minutes
- Drop async cache refreshes on each cache hit and reduce overhead load across our backends
- Reduce total number of cache queries executed in callbacks / resolved Promises and decrease load on Metaphysics itself

Opening this up here for general discussion... I'd propose setting up an A/B test on a review-app stack and load testing.  Sample queries (or a way to programmatically generate them to feed into Jmeter / another load testing tool would be helpful here as if they did not include mutations, we could run them against production Grav and they would give us a finer-grained view of performance impact across types of queries


acjay:
> The performance of a Metaphysics query is going to be at least the worst time of any cache request or HTTP request to a downstream service

Strictly speaking, the performance of an MP query is dominated by the critical path through all its resolvers. 

But I think that it's possible that [batching on some data loaders](https://github.com/artsy/metaphysics/blob/master/src/lib/loaders/api/loader_with_authentication_factory.js#L66) means that there are ways some resolvers can impact others that aren't in their descendant tree, making it more likely for 2 bad queries to compound and potentially delaying work that could otherwise be scheduled more quickly.

> We made a design decision to make a single, long-running GraphQL query, with no upper bound on query execution, in lieu of multiple RESTful queries to performant endpoints

To be clear, there's no fundamental reason this type of usage of GraphQL should be worse than REST for server-side rendering. The problem is we aren't generally availing ourselves of any means of recovering from a poorly performing resolver, for example, by allowing part of a query to fail. And also that we're backing our resolvers with *a lot* of complexity in the form of REST requests that result in a lot of work.

acjay:
It occurs to me that most of our individual pages don't get that much traffic, so we might actually have pretty high hit rates on old data, which would mean potentially a significant performance hit to long-tail content. Of course, if particular pages go days between hits, then any edit would mean that an individual visitor would be pretty likely to see the stale version.

I'm somewhat concerned that this doesn't quite solve the staleness problem, because at least the old way causes frequently refreshes of data. Seems like we're trading one source of staleness (first person to ask for recently changed data) for another (definitely no refresh until the key times out).

I don't suppose it could be possible to A/B test a change like this in real production?

Also, it occurs to me that if we keep today's scheme, perhaps we should actually be measuring the frequency with which we have served data from cache that differs from the fetch to refresh the cache.

mbilokonsky:
Caching only makes sense when and where it solves a problem. The strategy that we've got in place now may be the best possible solution to the problem it was meant to solve, but we should have an RFC or similar discussion with metrics that allow us to know for certain (a) what we're caching, (b) why we're caching it and (c) what the tradeoffs are. That should all be available at-a-glance somewhere, even if only in a PR.

I feel a little bit unqualified to have an opinion on this one because it seems so obvious that it's wrong that I feel like I must be missing some context. From what I can tell, we're always guaranteed to have fast stale data or slow fresh data. I understand the rationale for having fast stale data, because it's important to serve a response quickly, but I share the above concerns about this putting us kind of into a worst-of-all-possible-worlds state. 30 days seems way to long to cache stuff that's rarely looked up.

My sort of naive opinion at this moment is that it may behoove us to remove the cache entirely, and then add a new cache that solves specific problems as we identify and document them.

src/lib/loaders/api/loader_without_authentication_factory.js
```diff
@@ -57,30 +56,7 @@ export const apiLoaderWithoutAuthenticationFactory = (
...
-                        .catch(err => {
-                          if (err.statusCode === 404) {
-                            // Unpublished
-                            cache.delete(key)
-                          }
-                        })
-                    },
-                    { requestThrottleMs: apiOptions.requestThrottleMs }
-                  )
+                  return data
```
mbilokonsky: this is so much cleaner. The only thing I'd do differently is rip the whole cache out, to start - see me larger comment. 

I feel like premature caching is the root of the hardest and most damaging bugs I've worked on.

acjay:
I hear ya....but I'm _pretty_ sure removing the cache would be a huge problem. Then an MP request would only complete when the many many HTTP requests it makes to Gravity all return their data. In the time before MP, _all_ of these HTTP requests still more or less happened; they just came directly from client-side apps. We used convoluted client-side rendering strategies to try to hide all that latency. But now, we can usually render full pages server-side in acceptable amounts of time. (When everything is healthy, of course.)

Don't get me wrong, I agree that the unorthodox caching strategy is not ideal. But just to make sure we don't get distracted -- it doesn't seem like load on systems behind MP is a problem, and AFAIK, staleness isn't having a major business impact. So the effect here would be to (probably) reduce the number of cache `set` operations. Do we have a strong hunch that this is a major source of issues?

izakp:
Thanks for the thoughts!  Yes definitely just opened this PR to try and approach the problem from another angle / the standard approach to HTTP caching.  It's definitely not an easy problem - a good hit rate means converging on a sweet spot between fast, acceptably-stale data for the highest throughput endpoints and acceptably-slow fresh data for the most infrequently accessed.

But as @acjay puts it, let's not get distracted by but marinate on this problem.  We've been running the cache this way for so long and it's working (though I did ask gallery liaisons here if they ever get complaints from galleries that they publish an artwork in CMS but don't see it reflected live on the site and it seems to be a common complaint).

izakp:
Meant to link this here as an example of a standard approach to [HTTP caching with Varnish](https://www.varnish-software.com/wiki/start/httpcaching_basics.html).

izakp:
More food for thought - I totally forgot how you _can_ invalidate the caches for your HTTP API endpoints with Fastly CDN.  They have a feature called [surrogate keys](https://robots.thoughtbot.com/a-guide-to-caching-your-rails-application-with-fastly), which essentially lets you selectively invalidate some of your cache without completely purging it.

Your API sets the `Surrogate Key` header in responses along with `Cache Control` headers.  The surrogate key(s) themselves refer to all domain objects included in that response.

> Think of the Surrogate-Key header as a map that tells Fastly where each item in your system appears. When that item changes, Fastly goes to each location on the map and refreshes the content so that it is up to date.

So for example, say Gravity updates `Artwork#abc123` and we want to invalidate _all_ API endpoints that include that Artwork, it would do so programmatically in an `after_save` callback - instructing the caching layer to invalidate / refresh all keys in its cache that map to the surrogate key `Artwork#abc123`

Et voilà!  A full cache that's so fresh and so clean clean!

Varnish also took this feature onboard: https://github.com/varnish/varnish-modules/blob/master/docs/vmod_xkey.rst

cavvia:
I'm in favour of experimenting with switching off the cache and seeing how that impacts [average client page load times](https://app.datadoghq.com/monitors/4982211), to get a sense for how critical this cache is. As @mbilokonsky says, premature caching is to be avoided at all costs. In this case I imagine it was born out of necessity - but we should still confirm that.

I'm also in favour of trying out this more orthodox caching strategy. We may be executing a large amount of redundant SET operations. We make around 220/s, but it's difficult to estimate what portion of these are redundant (fall out of cache without ever being accessed). Our hit rate is currently healthy however (94%), and that could be impacted by this change. That is an argument against switching as it indicates a fairly efficient cache from a retrieval standpoint.

The current strategy is oriented towards fast first load times on infrequently accessed (long tail) content, while a traditional cache like this is optimized for frequently accessed data (ie. popular artist/artwork pages). My hypothesis is that more than 50% of our search traffic is on high popularity artists where this cache could still perform quite well, although 5 minutes is admittedly pretty aggressive. That estimate is based on the kind of frequency distribution we see for artist follows and searches on the site, which is roughly exponential.

The hypothesis for this PR is that contention on the socket, caused by high throughput, is currently causing some latency. The introduction of gzip compression implied network I/O itself may not be the bottleneck, as it seemed to have no impact. But the volume of JSON parsing and other callbacks associated with cache throughput could still be reduced considerably by this change, which may have a performance impact. 

I think the kind of HTTP caching @isac is referring to is tried and tested and could actually be a better solution than maintaining an MP cache in the long run (unless we want to remove the REST layer entirely).

mbilokonsky:
One other point: I don't suggest that we ship to production with no cache - I'm just saying that as a part of this unit of work we should consider disabling it and then re-enabling tactically to fix pain points as they emerge and get isolated and identified.

acjay:
Judging by recent convos, it kind of seems like we're on the search for whatever can give us more potential insight. So, may as well try and see what it does?

joeyAghion:
In my opinion we should merge this, or maybe a version of this that allows us to revert to the current behavior via config. Anything that can be as stale as the _last request_ (from up to a month ago) can also be 5 (or 10!) minutes stale.

That said, I'm positive that our current behavior has simply hidden staleness problems. When partners or users scratch their heads about stale data on a page, they refresh, see the correct data populated, and move on. (I've run into this a bunch personally.) Either way, I think explicitly allowing up to a few minutes of staleness for general purpose requests is a reasonable default, as long as we also make it easy to bypass caching altogether, which we do. We can address any specific problems that surface or allow callers to customize TTL in targeted ways.

Regarding the more general discussion, I agree that it could be illuminating to operate Metaphysics in a non-caching mode, at the very least to establish a baseline for what we're trying to improve by caching.

cavvia:
@izakp Since you brought this up yesterday in retro - if you're concerned about front-end developer buy-in, you could present it as a proposal at the next front-end practice meeting.

izakp:
> Regarding the more general discussion, I agree that it could be illuminating to operate Metaphysics in a non-caching mode, at the very least to establish a baseline for what we're trying to improve by caching.

Yea I think this would be a good idea to do first.  Will open another PR for feature flagging the cache completely, so as to establish a baseline.

orta:
I don't think this strictly needs front-end buy in, in a sense, most FE devs don't know or need to know too much about the caching implementation unless it's directly affecting them ( for example https://github.com/artsy/metaphysics/pull/1149#discussion_r199830467 )

Maybe I'm an outlier, but I've always considered MP's caching to be pretty easy to reason about (for un-authenticated requests, we cache any resolver that goes to gravity on a first in first out basis, keyed to API + params) but I can see a simpler model is definitely worth a shot given some of the details from @cavvia above

We should get this in and on an ENV var and tested

sweir27:
(I think I'm mostly caught up on this discussion) but noting that this _does_ provide weird behavior for clients right now. In emission, we re-fetch the increments on a sale artwork immediately after placing a bid. Because the results are cached on the initial fetch, we'd have to re-fetch _twice_ to get the most up-to-date result and show the correct screen.

orta:
Noting Sarah's point, regardless of the caching changes, we should also add a way for a client to request an uncached response

izakp:
Closing this in favor of https://github.com/artsy/metaphysics/pull/1174

PR #1114, opened 2018-06-18
dont batch data loaders that dont fetch in batch

acjay:
I think we only want batching to be enabled on our data loaders when we can make the underlying data fetch as a batch. Otherwise, slowest response dominates the batch. Asking for a bunch of review just because I'm not sure whether the batching was intentional for reasons I'm overlooking.

### DIFF
--- a/src/lib/loaders/api/loader_with_authentication_factory.ts
+++ b/src/lib/loaders/api/loader_with_authentication_factory.ts
@@ -61,7 +61,7 @@ export const apiLoaderWithAuthenticationFactory = (
           )
         },
         {
-          batch: true,
+          batch: false,
           cache: true,
         }
       )


izakp:
@acjay didn't we look at this before and infer that `maxBatchSize` needed to be passed as a parameter for batching to occur?  https://github.com/facebook/dataloader/blob/master/src/index.js#L233-L236

acjay:
Ahhh, yes, you're right. I forgot about that.

Although, I wasn't 100% convinced `batch: true` has _no_ effect, because it [appears in a separate code path](https://github.com/facebook/dataloader/blob/master/src/index.js#L102). And then it also seemed from some of our traces that some of our long-running resolvers were actually correlated. But maybe that's due to requests to Gravity that actually are requests to endpoints that return collections.

alloy:
@acjay @izakp What’s the conclusion of that discussion?

acjay:
Ah yeah, forgot about this. I'd suggest we do make the change, just because the loader generated by the factory is explicitly fetching keys one-by-one. It's definitely not the smoking gun I thought it might be, but I do think it's causing an extra deferral to happen, which may be delaying the start of a fetch longer than necessary.

alloy:
Cool. I’ll leave it to @izakp to determine what time to merge, though, so he knows what changes aere deployed when.

orta:
https://github.com/artsy/metaphysics/pull/1193 was somewhat related to that

PR #1116, opened 2018-06-19
Revert "Merge pull request #1108 from izakp/fea-async-json"

izakp:
This reverts commit 9a1d8bdcc4a1fa494fa8c3f88203b7afb534eb89, reversing
changes made to e331386e91fee444057fe7efa9a789cdfab5da9b.

See https://github.com/artsy/metaphysics/pull/1108

Deploying this to production caused a spike in CPU / event loop latency and memory usage

cc @alloy @mbilokonsky @orta @acjay

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,8 +96,7 @@
     "request": "^2.67.0",
     "source-map-support": "^0.4.3",
     "url-join": "^4.0.0",
-    "uuid": "^3.1.0",
-    "yieldable-json": "1.0.1"
+    "uuid": "^3.1.0"
   },
   "resolutions": {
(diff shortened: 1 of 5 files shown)


acjay:
Oh wow, interesting. Perhaps due to decrease in throughput from all the switching between in - progress deserialization? That would explain higher memory usage, too. But increased event loop latency is a big surprise to me. Great experiment though!

acjay:
Unless latency is measuring time in the task queue. That, I might expect to increase.

izakp:
@cavvia pointed out this async json is an implementation that runs on the event loop, not the UV thread pool << sigh >>

We should definitely take another crack at it

PR #1119, opened 2018-06-19
Resolver timeout improvements

alloy:
This PR does 2 things:

* Allows us to disable the timeouts completely on boot by setting the env variable to `0`
* No longer adds timeouts to resolvers that don’t return pending promises, which would have made those resolvers needlessly (extra) async.

Lesson learned, always using async/await as a convenience rather than checking if a return value is a promise can do more harm.


src/config.js
```diff
@@ -141,7 +141,7 @@ export default {
   REQUEST_THROTTLE_MS: Number(REQUEST_THROTTLE_MS) || 5000,
   REQUEST_TIMEOUT_MS: Number(REQUEST_TIMEOUT_MS) || 5000,
   RESIZING_SERVICE,
-  RESOLVER_TIMEOUT_MS: Number(RESOLVER_TIMEOUT_MS) || 5000,
+  RESOLVER_TIMEOUT_MS: Number(RESOLVER_TIMEOUT_MS) || 0,
```
alloy: Disabled by default from now on.

joeyAghion:
Mind describing the problem this solves, for the rest of us?

Also, will production continue to apply a timeout? Staging should match, if so.

alloy:
Need to amend the graphql-middleware docs https://github.com/prismagraphql/graphql-middleware/pull/14#issuecomment-398443252

src/test/helper.js
```diff
@@ -11,6 +11,9 @@ if (fs.existsSync(".env.test")) {
 import sinon from "sinon"
 global.sinon = sinon
 
+import Bluebird from "bluebird"
+global.Promise = Bluebird
```
mbilokonsky: Wait how come we're using Bluebird for testing and not native promises?

src/lib/graphqlTimeoutMiddleware.ts
```diff
@@ -50,18 +71,16 @@ export const graphqlTimeoutMiddleware = (defaultTimeoutInMS: number) => {
...
-      new Promise(async (resolve, reject) => {
-        try {
-          const result = await middlewareResolver(parent, args, context, info)
-          resolve(result)
-        }
-        catch (error) {
-          reject(error)
-        }
-        finally {
+      resolverResult.then(
```
mbilokonsky: This is dense so I want to make sure I understand it: rather than resolving the middleware in a promise race, you do it above and keep a reference to `resolverResult`. If nothing is waiting on resolverResult (I'm still fuzzy on what exactly `isPendingPromise` is doing?) then go ahead and just return it.

However, if it has been `then`'d and something is waiting for it, race it against a promise that autorejects within a given timeout window. 

This is how we add timeouts to all resolvers, right, if the timeout value is > 0?

I'm going to try to understand the `isPendingPromise` a bit better before, but this looks clever and good.

src/test/helper.js
alloy: We do use Bluebird in the app, but [it’s setup](https://github.com/artsy/metaphysics/blob/4968b86e19bd38acb823d2cacdbf5bcd1b7688ba/index.js#L21) in a module that’s not loaded in tests.

src/lib/graphqlTimeoutMiddleware.ts
alloy: Yup, you got it 👌

src/index.js
```diff
@@ -86,7 +86,10 @@ async function startApp() {
     }
   }
 
-  schema = applyGraphQLMiddleware(schema, graphqlTimeoutMiddleware(RESOLVER_TIMEOUT_MS))
+  if (RESOLVER_TIMEOUT_MS > 0) {
+    console.warn("[FEATURE] Enabling resolver timeouts")
+    schema = applyGraphQLMiddleware(schema, graphqlTimeoutMiddleware(RESOLVER_TIMEOUT_MS))
```
izakp: If I understand this correctly - if setting `RESOLVER_TIMEOUT_MS=0`, `graphqlTimeoutMiddleware` will use a `timeoutForField` - is this set on the schema?

src/lib/graphqlTimeoutMiddleware.ts
```diff
@@ -39,6 +55,11 @@ export const graphqlTimeoutMiddleware = (defaultTimeoutInMS: number) => {
     context,
     info
   ) => {
+    const resolverResult = middlewareResolver(parent, args, context, info)
+    // Only add a timeout if there’s async work in progress.
+    if (!isPendingPromise(resolverResult)) {
+      return resolverResult
```
izakp: 👍 I suppose we should test this change first then set the `RESOLVER_TIMEOUT_MS ` down to 0

src/index.js
alloy: If `RESOLVER_TIMEOUT_MS` is set to `0` then the timeout middleware doesn't get applied to the schema at all.

alloy:
@izakp I’ll leave it up to you to merge when you think is a good time.

src/index.js
izakp: Ach. Duh.

izakp:
Errors on deploy:
```
OST /? 500 115.142ms 108.54.246.210 "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Safari/605.1.15"
POST /? 500 93.677ms 108.54.246.210 "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Safari/605.1.15"
POST /? 500 13.684ms 108.54.246.210 "Mozilla/5.0 (Macintos
```

PR #1128, opened 2018-06-21
Inject response body to HTTPError whenever possible

yuki24:
We [construct an error message](https://github.com/artsy/metaphysics/blob/597405d0678e941948275c006cb256acb343eccd/src/lib/apis/fetch.js#L32-L35) based on the request URI and the response body and assign it to the error object. This is most useful when debugging in a client like GraphiQL, but sometimes we [split the error message back](https://github.com/artsy/metaphysics/blob/597405d0678e941948275c006cb256acb343eccd/src/schema/me/bidder_position_mutation.js#L50-L51) within Metaphysics to generate a more helpful error message for the end-user. We should not have to do this as you already have the response in place.

This PR updates the `fetch` function to assign the response body to the `HTTPError` object so we will gain direct access to the body.

### DIFF
--- a/src/lib/apis/fetch.js
+++ b/src/lib/apis/fetch.js
@@ -33,7 +33,7 @@ export default (url, options = {}) => {
           get(response, "request.uri.href"),
           response.body,
         ]).join(" - ")
-        return reject(new HTTPError(message, response.statusCode))
+        return reject(new HTTPError(message, response.statusCode, response.body))
       }
 
       try {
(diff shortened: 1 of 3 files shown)


src/lib/apis/fetch.js
```diff
@@ -33,7 +33,7 @@ export default (url, options = {}) => {
           get(response, "request.uri.href"),
           response.body,
```
alloy: We still also include it here, is that still necessary?

PR #1131, opened 2018-06-25
Fea custom metrics

izakp:
- Adds memory and GC metrics measured by Node's VM
- Only send metrics in production environments - change tagging to match `pod_name` rather than `hostname` so as to match metrics collected by the kubernetes integrations

### DIFF
--- a/src/lib/stats.js
+++ b/src/lib/stats.js
@@ -7,6 +7,7 @@ import { error } from "./loggers"
 const { NODE_ENV, ENABLE_METRICS, STATSD_HOST, STATSD_PORT } = config
 
 const isTest = NODE_ENV === "test"
+const isDev = NODE_ENV === "development"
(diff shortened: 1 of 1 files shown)


src/lib/stats.js
```diff
@@ -7,6 +7,7 @@ import { error } from "./loggers"
 const { NODE_ENV, ENABLE_METRICS, STATSD_HOST, STATSD_PORT } = config
 
 const isTest = NODE_ENV === "test"
+const isDev = NODE_ENV === "development"
```
alloy: I don’t know if we really set this env, do you? In any case, it looks like functional-wise these end up being used the same way, so how about replacing these both with `production` and just check in all the places if this is _not_ production?

izakp: Ah yes in staging `NODE_ENV=production` as well

PR #1132, opened 2018-06-25
Properly format error description for BidderPositionResult with '\n'

yuki24:
Now that we have https://github.com/artsy/emission/pull/1083 in and we should update the error descriptions for `BidderPositionResult`. I confirmed the `\n` syntax adds a line break properly. Let me know if there's any feedback!

screen shot 2018-06-25 at 10 13 17 am


sepans:
Thanks @yuki24 added merge on green

PR #1134, opened 2018-06-26
raise resource limits

izakp:
Let's see how it behaves with a 1Gi ceiling

Also removing CPU limit in production as we have more threads running under the `UV_THREADPOOL`

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -30,10 +30,9 @@ spec:
         resources:
           requests:
             cpu: "0.5"
-            memory: "512Mi"
-          limits:
-            cpu: "1"
             memory: "768Mi"
+          limits:
+            memory: "1Gi"
         readinessProbe:
           httpGet:
             port: 3000
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -32,8 +32,8 @@ spec:
             cpu: "0.2"
             memory: "512Mi"
           limits:
-            cpu: "0.5"
-            memory: "768Mi"
+            cpu: "1"
+            memory: "1Gi"
         readinessProbe:
           httpGet:
             port: 3000


alloy:
The change seems to do what the PR mentions, so 👍

But I’m not sure if/how I should be considering the new limits, in general 1gig seems like an ok amount, I guess.

> Also removing CPU limit in production as we have more child processes running under the `UV_THREADPOOL`

Can you explain or point me to existing doc/chat about what this means?

izakp:
@alloy see https://artsy.slack.com/archives/C04DU9H0L/p1529055939000275

Also this doc https://docs.google.com/document/d/1L293YbDVtxFRL1sIrbgSCEVDm0mXpwBTI5OffQ3LvkE

cavvia:
One question is if the raised CPU limits will impact other apps running on the same kube nodes.

alloy:
@izakp Awesome, thanks 👍  Great discussion!

izakp:
> One question is if the raised CPU limits will impact other apps running on the same kube nodes

I was operating under the assumption that Node _shouldn't_ use more than 1 CPU as a single process.

Ach so my mistake - we don't have "more child processes running" as I say above.  The UV threadpool is implemented in a C library `libuv`, uses native POSIX threading http://docs.libuv.org/en/latest/threadpool.html and I'm not certain that this will make use of more than 1 CPU but I want to make sure by removing the limit.

I can't get a clear answer to this but reading around https://github.com/joyent/libuv/issues/649

izakp:
I'm going to close this and re-open to just raise the memory limits for now - we can approach the CPU limits in a separate PR

izakp:
Only raise the memory limits here: https://github.com/artsy/metaphysics/pull/1135

PR #1143, opened 2018-06-27
Ecommerce Mutations for Users

sepans:
MP mutation endpoints that call express mutations to [create](https://github.com/artsy/exchange#create-an-order) or [submit](https://github.com/artsy/exchange#submit-an-order) an order.

I am not completely familiar with MP filename conventions so I wasn't sure if these mutations need to be under `/me/` path since they are triggered by a user from force, or better grouped with other e-commerce endpoints where it is now?

#### Sample queries:

* create
```
mutation {
  createOrder(input: {
      partnerId: "111",
      userId: "111",
      currencyCode: "usd",
      lineItems: [{
        artworkId: "111",
        quantity: 1
        priceCents: 300000
      }]
    }) {
      result {
        order {
          id
          code
          currencyCode
          state
          partner {
            id
            name
          }
          user {
            id
            email
          }
          lineItems {
            edges {
              node {
                artwork {
                  id
                  title
                }
              }
          }
        }
      }
    }
  }
``` 

* submit
```
mutation {
  submitOrder(input: {
      orderId: "111",
      creditCardId: "111",
    }) {
      order {
        id
        code
        currencyCode
        state
        partner {
          id
          name
        }
        user {
          id
          email
        }
        lineItems {
          edges {
            node {
              artwork {
                id
                title
              }
            }
          }
        }
      }
    }
  }
```

### DIFF
--- a/src/schema/ecommerce/create_order_mutation.js
+++ b/src/schema/ecommerce/create_order_mutation.js
@@ -1,0 +1,110 @@
+import {
+  GraphQLInputObjectType,
+  GraphQLNonNull,
+  GraphQLString,
+  GraphQLList,
+  GraphQLInt,
+  graphql,
+} from "graphql"
+import { OrderType } from "schema/ecommerce/types/order"
+import { mutationWithClientMutationId } from "graphql-relay"
+
+const LineItemInputType = new GraphQLInputObjectType({
+  name: "LineItemInput",
+  fields: {
+    artworkId: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: "ID of artwork",
+    },
+    quantity: {
+      type: new GraphQLNonNull(GraphQLInt),
+      description: "quantity of artwork",
+    },
+    priceCents: {
+      type: new GraphQLNonNull(GraphQLInt),
+      description: "Price in cents",
+    },
+  },
+})
+
+const CreateOrderInputType = new GraphQLInputObjectType({
+  name: "CreateOrderInput",
+  fields: {
+    userId: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: "ID of user submitting the order",
+    },
+    partnerId: {
+      type: GraphQLString,
+      description: "ID of partner representing artwork",
+    },
+    currencyCode: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: "Currency code",
+    },
+    lineItems: {
+      type: new GraphQLList(LineItemInputType),
+      description: "Line items in the order",
+    },
+  },
+})
+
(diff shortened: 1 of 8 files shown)


src/schema/ecommerce/submit_order_mutation.js
```diff
@@ -0,0 +1,79 @@
...
+            updatedAt
+            createdAt
+            lineItems{
+              edges{
+                node{
+                  id
+                  priceCents
+                  artworkId
+                  editionSetId
+                  quantity
```
ashkan18: lets add `createdAt` here too

src/schema/ecommerce/submit_order_mutation.js
```diff
@@ -0,0 +1,79 @@
...
+                }
+              }
+            }
+          }
+          errors
```
ashkan18: i think errors have `message` and `path`? i may be missing more stuff too but maybe we can get those here too. or what happens if we don't specify any fields, does it get everything?

src/schema/ecommerce/create_order_mutation.js
```diff
@@ -0,0 +1,110 @@
...
+  GraphQLNonNull,
+  GraphQLString,
+  GraphQLList,
+  GraphQLInt,
+  graphql,
+} from "graphql"
+import { OrderType } from "schema/ecommerce/types/order"
+import { mutationWithClientMutationId } from "graphql-relay"
+
+const LineItemInputType = new GraphQLInputObjectType({
```
ashkan18: im not too familiar with file structure here either, what do you think about moving input types to their own folder and have this file only define the mutation by importing these input types?

sepans: I thought about it and decided to move InputTypes to the Types folder when they are reused in other queries/mutations. I will move them as I add more mutations.

src/schema/ecommerce/submit_order_mutation.js
sepans: 👍  added `errors`. I have `errors: [String!]!` in *exchange.graphql*. Has it been changed?

src/schema/ecommerce/types/order.js
```diff
@@ -60,7 +60,7 @@ export const OrderType = new GraphQLObjectType({
         _args,
         _context,
         { rootValue: { userByIDLoader } }
-      ) => userByIDLoader(userId),
+      ) => (userId ? userByIDLoader(userId) : null),
```
sepans: I didn't understand why `userByIDLoader` from `partnerLoader` but with Order being null MP was making a request to gravity with `undifined` as ID and getting a gravity error.

ashkan18:
🕺

PR #1146, opened 2018-06-29
[WIP] Add user fields to sale and artwork bidding status

alloy:
Depends on https://github.com/artsy/gravity/pull/11796

This example still uses the incorrect `Artwork.sale` field which assumes that there’s only ever 1 live sale, but the `Artwork. bidderStatus` schema addition in this PR will correctly return a list, which for now the client should just select the first entry from.

```graphql
{
  artwork(id: "yohei-imamura-self-portrait") {
    sale {
      registrationStatus {
        qualified_for_bidding
      }
    }
    bidderStatus(live: true) {
      is_highest_bidder
    }
  }
}
```

```json
{
  "data": {
    "artwork": {
      "sale": {
        "registrationStatus": {
          "qualified_for_bidding": true
        }
      },
      "bidderStatus": [
        {
          "is_highest_bidder": true
        }
      ]
    }
  }
}
```

### DIFF
--- a/src/__tests__/integration.test.ts
+++ b/src/__tests__/integration.test.ts
@@ -32,6 +32,6 @@ it("It can make a request against the schema", async () => {
       `,
     })
 
-  expect(response.body).toEqual({ data: { artist: { name: "Mr Bank" } } })
+  expect(response.body.data).toEqual({ artist: { name: "Mr Bank" } })
   expect(response.statusCode).toBe(200)
 })
--- a/src/schema/artwork/index.js
+++ b/src/schema/artwork/index.js
@@ -33,6 +33,7 @@ import {
 import AttributionClass from "schema/artwork/attributionClass"
 // Mapping of attribution_class ids to AttributionClass values
 import attributionClasses from "../../lib/attributionClasses.js"
+import { LotStandingType } from "../me/lot_standing"
 
 const is_inquireable = ({ inquireable, acquireable }) => {
   return inquireable && !acquireable
@@ -122,6 +123,24 @@ export const artworkFields = () => {
     },
     availability: {
       type: GraphQLString,
+    },
+    bidderStatus: {
+      type: new GraphQLList(new GraphQLNonNull(LotStandingType)),
+      args: {
+        live: {
+          type: GraphQLBoolean,
+          defaultValue: null,
+        },
+      },
+      resolve: (
+        { id },
+        { live },
+        _request,
+        { rootValue: { lotStandingLoader } }
+      ) => {
+        if (!lotStandingLoader) return null
+        return lotStandingLoader({ artwork_id: id, live })
+      },
     },
     can_share_image: {
       type: GraphQLBoolean,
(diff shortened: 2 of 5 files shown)


src/schema/sale/index.js
```diff
@@ -319,6 +320,19 @@ export const SaleType = new GraphQLObjectType({
...
+        type: Bidder.type,
+        description: "A registration for this sale",
+        resolve: (
+          { id },
+          _args,
+          _request,
+          { rootValue: { meBiddersLoader } }
+        ) => {
+          if (!meBiddersLoader) return null
+          return meBiddersLoader({ sale_id: id }).then(([bidder]) => bidder)
```
alloy: I’m assuming that there can only be one item in this list when scoping to a sale?

src/__tests__/integration.test.ts
```diff
@@ -32,6 +32,6 @@ it("It can make a request against the schema", async () => {
       `,
     })
 
-  expect(response.body).toEqual({ data: { artist: { name: "Mr Bank" } } })
+  expect(response.body.data).toEqual({ artist: { name: "Mr Bank" } })
```
alloy: @orta On a clean setup of the MP repo I was getting the debug extensions included as well, which wouldn’t match this, so I changed the test to only consider the `data` entry.

alloy:
Going to self-merge this so I can continue on to the components that consume this.

Danger, merge on green 🙏

src/__tests__/integration.test.ts
orta: Yeah,  I think that makes sense, this happened with someone else too - so I prefer this ^ implementation idea

src/schema/sale/index.js
alloy: Thanks 👍

sweir27:
👍 cool. Will try to update emission to consistently use `registrationStatus` before we ship.

src/schema/artwork/index.js
```diff
@@ -123,6 +124,24 @@ export const artworkFields = () => {
     availability: {
       type: GraphQLString,
     },
+    bidderStatus: {
```
sweir27: Actually question-- why use the `bidderStatus` nomenclature here instead of `lotStanding`?

oxaudo: I think it's just me inventing terms in reaction static comp. The bidder stats is what we wanted to use there - I named it that in absence of knowing real terms for all things auctions. If it sounds off - totally up for changing it. In my mind `BidderStatus` showed that we are getting the data for a given user standings on a particular lot. Where `lotStanding` does not signal that it's a specific user data.

alloy: My naive mind agrees with @oxaudo, but we’re not using this in production yet, so we could still change if you think that’s appropriate?

sweir27: I might tend towards `lotStanding` since it's what we use elsewhere (`bidderStatus` can be misleading as to me it would refer to the status of my `bidder` meaning, am I qualified or not to bid on this auction).

alloy: Got it. @oxaudo do you want to take this?

oxaudo: Sure. Will update a bit later. Thanks @sweir27 for flagging... The only thing I think it'll be cool to know from the name that it's a current user info. Mayne it can be `currentUserLotStanding` - or it's too verbose and lotStanding applies that it's a current user specific info?

In my mind just `lotStanding` will apply that this field contains the most current info from all the bidders where in this case it's specifically scoped to the user on this sale artwork information. @sweir27 ?

alloy: Thus far we have been referring to the ‘current user’ as ‘me’ in MP, so (shudder) it should probably be `myLotStandings`

PR #1148, opened 2018-07-02
Update exchange.graphql

ashkan18:
# Change
Exchange was updated to add two new timestamps on order in https://github.com/artsy/exchange/pull/51, this PR follows the update and exposes those in MP.

cc: @sepans

PR #1151, opened 2018-07-02
Adds approve/reject/finalize order mutations for partner users

sepans:
Adds mutations to approve/reject/finalize orders, to be called from CMS.

### DIFF
--- a/src/schema/ecommerce/approve_order_mutation.js
+++ b/src/schema/ecommerce/approve_order_mutation.js
@@ -40,6 +40,8 @@ export const ApproveOrderMutation = mutationWithClientMutationId({
             transactionFeeCents
             updatedAt
             createdAt
+            stateUpdatedAt
+            stateExpiresAt
             lineItems{
               edges{
                 node{
--- a/src/schema/ecommerce/approve_order_mutation.js
+++ b/src/schema/ecommerce/approve_order_mutation.js
@@ -1,0 +1,69 @@
+import { graphql } from "graphql"
+import { OrderReturnType } from "schema/ecommerce/types/order_return"
+import { OrderMutationInputType } from "schema/ecommerce/types/order_mutation_input"
+import { mutationWithClientMutationId } from "graphql-relay"
+
+export const ApproveOrderMutation = mutationWithClientMutationId({
+  name: "ApproveOrder",
+  decription: "Approvess an order with payment",
+  inputFields: OrderMutationInputType.getFields(),
+  outputFields: {
+    result: {
+      type: OrderReturnType,
+      resolve: order => order,
+    },
+  },
+  mutateAndGetPayload: (
+    { orderId },
+    context,
+    { rootValue: { accessToken, exchangeSchema } }
+  ) => {
+    if (!accessToken) {
+      return new Error("You need to be signed in to perform this action")
+    }
+    const mutation = `
+      mutation approveOrder($orderId: ID!) {
+        ecommerce_approveOrder(input: {
+          id: $orderId,
+        }) {
+          order {
+           id
+            code
+            currencyCode
+            state
+            partnerId
+            userId
+            itemsTotalCents
+            shippingTotalCents
+            taxTotalCents
+            commissionFeeCents
+            transactionFeeCents
+            updatedAt
+            createdAt
+            lineItems{
+              edges{
+                node{
+                  id
+                  priceCents
+                  artworkId
+                  editionSetId
+                  quantity
+                }
+              }
+            }
+          }
+          errors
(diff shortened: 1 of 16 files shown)


src/schema/ecommerce/approve_order_mutation.js
```diff
@@ -0,0 +1,69 @@
...
+            code
+            currencyCode
+            state
+            partnerId
+            userId
+            itemsTotalCents
+            shippingTotalCents
+            taxTotalCents
+            commissionFeeCents
+            transactionFeeCents
```
ashkan18: lets add `stateUpdatedAt` and `stateExpiresAt` here too, they were added in https://github.com/artsy/metaphysics/pull/1148

src/schema/ecommerce/approve_order_mutation.js
```diff
@@ -0,0 +1,69 @@
...
+            }
+          }
+          errors
+        }
+      }
+    `
+    return graphql(exchangeSchema, mutation, null, context, {
+      orderId,
+    }).then(result => {
+      const { order, errors } = result.data.ecommerce_approveOrder
```
ashkan18: should we do that `throw` thing we did for queries here too? mainly because the details of `errors` here would be wrong, it basically have exchange stack trace and etc.

src/schema/ecommerce/finalize_order_mutation.js
```diff
@@ -0,0 +1,71 @@
...
+            code
+            currencyCode
+            state
+            partnerId
+            userId
+            itemsTotalCents
+            shippingTotalCents
+            taxTotalCents
+            commissionFeeCents
+            transactionFeeCents
```
ashkan18: same here for new state timestamps :)

src/schema/ecommerce/types/order_mutation_input.js
```diff
@@ -0,0 +1,11 @@
+import { GraphQLInputObjectType, GraphQLNonNull, GraphQLString } from "graphql"
+
+export const OrderMutationInputType = new GraphQLInputObjectType({
+  name: "SubmitOrderInput",
```
ashkan18: should this be `OrderInput`?

src/schema/ecommerce/approve_order_mutation.js
sepans: I tried `throwing` an error doing
```
}).then(result => {
  const { order, errors } = result.data.ecommerce_createOrder
  if (errors) {
    throw Error(errors)
  }
  return {
    order,
  }
})
```
here are sample results:

1) currently:

screen shot 2018-07-05 at 11 58 41 am

2) with `throw`:

screen shot 2018-07-05 at 11 59 25 am

One one hand I like the 2 for being consistent with `order` and `orders` queries but the stacktrace and `location` are note useful at all for debugging so 1 is cleaner. Also 1 allows us to return kind of 'soft' errors to notify the client there were some errors but operation took place. So I am a little bit in favor of 1 but ok with 2. cc: @ashkan18  @jonallured

ashkan18: interesting... i feel like i kind of like/dont like both of them for different reason:
1) 
 - it's definitely cleaner
 - if we try to support lets say multiple mutation per request this is lot easier to find out which one had isssues
  - my problem with this one is the extra `result` field that we've added, i don't think thats `graphql` standard, is there a way we could ignore that, if we could i would def pick this option
2) 
  - this makes things easier for client where a client needs to always look into root level `errors` and see if anything is in there
  - the problem with ☝️ is, if we have multiple mutations/queries in one graphql call, it would be hard to find out which one caused the error, we'd need to get paths and match to mutations and etc.

In general im more in favor of 1 but id suggest removing `results` and have mutation just return order type and maybe we can add optional `errors` to `OrderType` if that allows us to remove this `result` type. what do you think?

jonallured: FWIW, I would just go with 1 and we can change our minds as we use it. Hard to tell now what I'll want next week when I start using this.

sepans: I don't know how it is possible to remove `result` from there but since this change should be applied to `createOrder` and `submitOrder` as well I would say lets keep it the way it is (1) and I will do a separate PR if we decided to do something else.

ashkan18: 👍 lets look into removing `result` in separate PR, i was thinking maybe we could update `OrderType` to have a list or `errors` but i may be wrong :)

ashkan18:
🙌

PR #1152, opened 2018-07-02
Fix null itemsTotalCents

ashkan18:
# Problem
`itemsTotalCents` was added in https://github.com/artsy/metaphysics/pull/1144 but it was not added to queries made to Exchange for fetching it, it was always null because of that.

# Solution
Add it :)

# Follow up
Add other price related fields, tax, shipping etc.

### DIFF
--- a/src/schema/ecommerce/order.js
+++ b/src/schema/ecommerce/order.js
@@ -35,7 +35,6 @@ export const Order = {
     return graphql(exchangeSchema, query, null, context, {
       id,
     }).then(result => {
-      console.log(result)
       if (result.errors) {
         throw Error(result.errors.map(d => d.message))
       }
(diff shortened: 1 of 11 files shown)


src/schema/ecommerce/order.js
```diff
@@ -34,6 +35,7 @@ export const Order = {
     return graphql(exchangeSchema, query, null, context, {
       id,
     }).then(result => {
+      console.log(result)
```
starsirius: 🙈

ashkan18: hehehe thats how all problems are found.....

jonallured:
Looks great, thank you!!

jonallured:
@ashkan18 

screen shot 2018-07-03 at 8 50 18 am

ashkan18:
❤️ 😍 ❤️

PR #1153, opened 2018-07-03
Add other price related fields

ashkan18:
Adding more Order price related fields, follow up on #1152 

cc: @jonallured


sepans:
LGTM-> merge on green.

PR #1160, opened 2018-07-05
Refactors mocking exchange

sepans:
#Trivial

Moved code for mocking exchange schema for tests into a helper function.


ashkan18:
nice! 😍

PR #1163, opened 2018-07-09
@mzikherman => Adds my_increments to sale_artwork

sweir27:
I'm opening this so we can start to have a discussion... _code talks_.

The problem I'm trying to solve is:
- in multiple places at Artsy (i.e. in emission and in force/reaction right now) we need to show users a dropdown list of increments that they can bid with. The logic for determining where to _start_ this increments list lives [here](https://github.com/artsy/force/blob/master/src/desktop/apps/artwork/components/auction/templates/bid.jade#L3-L4) right now, but as I'm working on adding it to emission it would live there as well.

This proposal moves the logic upstream to metaphysics, where it could potentially be re-used by force/reaction (I'm not sure if this has been implemented yet for the new artwork page, but presumably it would have to be).

Things I'm not sold on:
- `my_increments` isn't great naming, and I know we're always talking about camelCase.
- It's totally possible the client will need information about the lot standings as well (solo) and so technically they'd be fetching that twice. I guess with data loaders that shouldn't be an issue provided they're in the same query.
- I know @alloy was moving some things up to the `artwork` scope, but by the time we're querying for things like bid increments we must be scoping down to `sale_artwork` eventually. Increments make more sense to be here I think.

cc @alloy since this might affect some of the artwork page work.

### DIFF
--- a/src/schema/sale_artwork.js
+++ b/src/schema/sale_artwork.js
@@ -185,17 +185,42 @@ const SaleArtworkType = new GraphQLObjectType({
       },
       increments: {
         type: new GraphQLList(BidIncrementsFormatted),
+        args: {
+          useMyMaxBid: {
+            type: GraphQLBoolean,
+            description:
+              "Whether or not to start the increments at the user's latest bid",
+          },
+        },
         resolve: async (
-          { minimum_next_bid_cents, sale_id, symbol },
-          options,
+          { _id, minimum_next_bid_cents, sale_id, symbol },
+          { useMyMaxBid },
           request,
-          { rootValue: { incrementsLoader, saleLoader } }
+          { rootValue: { incrementsLoader, lotStandingLoader, saleLoader } }
         ) => {
+          let minimumNextBid
+          minimumNextBid = minimum_next_bid_cents
+
+          if (useMyMaxBid) {
+            const myLotStandings = await lotStandingLoader({
+              sale_artwork_id: _id,
+            })
+
+            const myCurrentMax = get(
+              myLotStandings,
+              "0.max_position.max_bid_amount_cents"
+            )
+
+            minimumNextBid = myCurrentMax
+              ? maxBy([myCurrentMax, minimum_next_bid_cents])
+              : minimum_next_bid_cents
+          }
+
           return bid_increments_calculator({
             sale_id,
             saleLoader,
             incrementsLoader,
-            minimum_next_bid_cents,
+            minimum_next_bid_cents: minimumNextBid,
           }).then(bid_increments => {
             return bid_increments.map(increment => {
               return {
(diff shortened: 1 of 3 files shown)


src/schema/sale_artwork.js
```diff
@@ -272,6 +261,42 @@ const SaleArtworkType = new GraphQLObjectType({
...
+            "0.max_position.max_bid_amount_cents"
+          )
+
+          const myMinimumNextBid = myCurrentMax
+            ? maxBy([myCurrentMax, minimum_next_bid_cents])
+            : minimum_next_bid_cents
+
+          return bid_increments_calculator({
+            sale_id,
+            saleLoader,
```
sweir27: Also, it won't let me comment so far above but when we fetch information about the sale artwork (specifically, the `minimum_next_bid_cents`), we'll want to grab that _ignoring_ metaphysics' cache. I don't think there's a way to specify this right now and I believe it's going to come up in the FE meeting tomorrow. (so not blocking, just a note)

mzikherman: Can you describe more the caching you expect? Maybe this should use an auth'ed loader (or the 'one off' factory type like for system time?).

Alternatively, we could enhance the 'without authentication' factory code to accept `{ cache: false }` and have it skip that.

I am curious about the flow of events however that you want this to be uncached from Metaphysics, but, presumably cached in Gravity? (If it's cached in gravity per user then we can just use an auth'ed loader here. If it's not cached per user I'm confused why you specifically want to skip MP's cache).

src/schema/sale_artwork.js
```diff
@@ -272,6 +261,42 @@ const SaleArtworkType = new GraphQLObjectType({
...
+            incrementsLoader,
+            minimum_next_bid_cents: myMinimumNextBid,
+          }).then(bid_increments => {
+            return bid_increments.map(increment => {
+              return {
+                cents: increment,
+                display: formatMoney(increment / 100, { symbol, precision: 0 }),
```
mzikherman: This `/ 100` feels like it will break when we do a JPY auction, is there a `subunit_to_unit` (or the equivalent) value around?

mzikherman: There is [this JSON](https://github.com/artsy/metaphysics/blob/e882ea9a3e447d2afc2b13c40198e0f4af55207d/src/lib/currency_codes.json) file where you can look up a currency code and get that value if you think it's a good idea to switch to that.

src/schema/sale_artwork.js
```diff
@@ -272,6 +261,42 @@ const SaleArtworkType = new GraphQLObjectType({
         type: GraphQLFloat,
         deprecationReason: "Favor `minimum_next_bid`",
       },
+      my_increments: {
```
mzikherman: I don't have a problem with this naming, esp since there's already an `increments` schema here. Maybe that should accept an arg `useMyMaxBid: true/false` to perform this more nuanced behavior and keep things contained to that field?

oxaudo:
This would def be useful for the new artwork page. I did not get to put it in - but we do have designs for it...

I’ll take a look at the code later but wanted to comment on usefulness right away:)

src/schema/sale_artwork.js
sweir27: The flow of events (where it goes wrong) is that during the bidding flow, after you place a bid you may have been immediately outbid and will have to place another one. At that point, you can click back to the "select my max bid" screen which will refresh the increments and _should_ show you the correct next bid.

At that point, the sale artwork's cache will have been invalidated in gravity, but the first request to MP will return a stale value before it's updated. So we'll have the wrong `minimum_next_bid_cents` from which to compare. This behavior will become better when we're adding the call to lot standings, but it's still not ideal.

I think your suggestion to enhance the factory is the way to go (but I won't do that in this PR).

src/schema/sale_artwork.js
sweir27: Haha, this is what I get for copy-pasta-ing the code from above 😆

src/schema/sale_artwork.js
sweir27: The `useMyMaxBid` param is a good idea too. Let me spike out what that would look like.

src/schema/sale_artwork.js
```diff
@@ -184,27 +181,46 @@ const SaleArtworkType = new GraphQLObjectType({
...
-          { minimum_next_bid_cents, sale_id, symbol },
-          options,
+          { _id, minimum_next_bid_cents, sale_id, symbol },
+          { useMyMaxBid },
           request,
-          { rootValue: { incrementsLoader, saleLoader } }
+          { rootValue: { incrementsLoader, lotStandingLoader, saleLoader } }
         ) => {
+          let minimumNextBid
+          minimumNextBid = minimum_next_bid_cents
```
sweir27: Made that update-- it's a lot less repeated code now 👏

src/schema/sale_artwork.js
```diff
@@ -184,27 +181,46 @@ const SaleArtworkType = new GraphQLObjectType({
...
+
+          if (useMyMaxBid === true) {
```
mzikherman: http://trueistrue.com/

mzikherman:
👏 

merge on green

PR #1166, opened 2018-07-11
Updates logic around increments when useMyMaxBid is true

sweir27:
While testing this out again, I noticed two minor things:
- If you attempt to pass `useMyMaxBid: true` but aren't logged in, we'd get an error from `lotStandingLoader` not existing. I updated this to check if that's there-- if not we return the generic increments.
- I think we can actually update the logic to consider your `leading` status (leading means you're the highest bidder on the lot-- I don't way "winning" since that has other implications around the reserve). More explanation below...

Current logic:
- If you have no bids on the lot, we return increments beginning at the `minimum_next_bid_cents`
- If you have a bid on the lot but aren't _leading_, we return increments beginning at the `minumum_next_bid_cents`. In this case, your max bid must be lower than the `minimum_next_bid` or else you would be the leader.
- If you have a bid on the lot and you _are_ leading, we return increments beginning at _your_ asking price. This will be one increment above your current maximum position on the lot.

### DIFF
--- a/src/schema/__tests__/sale_artwork.test.js
+++ b/src/schema/__tests__/sale_artwork.test.js
@@ -364,9 +364,14 @@ describe("SaleArtwork type", () => {
       ])
     })
 
-    it("returns increments from the most recent bid if greater than the minimum_next_bid_cents", async () => {
+    it("returns increments from the most recent bid if you are leading", async () => {
       const lotStandingLoader = () => {
-        return [{ max_position: { max_bid_amount_cents: 390000 } }]
+        return [
+          {
+            max_position: { max_bid_amount_cents: 390000 },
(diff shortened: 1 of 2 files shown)


src/schema/__tests__/sale_artwork.test.js
```diff
@@ -394,12 +395,21 @@ describe("SaleArtwork type", () => {
           cents: 420000,
           display: "€4,200",
         },
+        {
```
sweir27: This was the main logic change. I'm trying to make it so that when a client pulls `increments(useMyMaxBid: true)`, they can by default show the first item in the list. If we didn't do this, they'd have to still have some custom logic for when the increments are calculated based on your max bid (and to show the second item in the array in that case).

PR #1168, opened 2018-07-11
sample number of active handles and requests and send to custom metrics

izakp:
[Folks say](https://github.com/nodejs/node/issues/1128) that you should watch node's `process._getActiveHandles()` and `process._getActiveRequests()` as this can give an indication of "pending work" queued throughout the event loop's internal phases.


RaisinTen:
FYI, these APIs have been doc-deprecated and there's a new API, so I have sent a PR to use that instead - https://github.com/artsy/metaphysics/pull/4664

PR #1170, opened 2018-07-13
Adds danger and fails about adding new app JS files when they should be ts files

orta:
Will need a corresponding PR to artsy/peril-systems to have peril running this later.


dangerfile.ts
```diff
@@ -0,0 +1,11 @@
+import { danger, warn } from "danger"
+
+// Rule: encourage all new files to be TypeScript
+const jsAppFiles = danger.git.created_files.filter(
+  f => f.startsWith("src/") && f.endsWith(".js")
```
zephraph: ❤️

orta:
Danger will now fail the build if you push with *new* JS files

screen shot 2018-07-16 at 12 51 12 pm

package.json
```diff
@@ -25,6 +25,7 @@
     "build:fixtures": "cp src/lib/*.json build/src/lib/",
     "build:remote-schemas": "cp src/data/*.graphql build/src/data/",
     "precommit": "lint-staged",
+    "prepush": "yarn danger local",
```
mzikherman: i'm always confused as to why we have to duplicate things in these hooks that happen in CI.

It already takes 45 seconds for `git commit` and 45 seconds for `git push` (on my old computer, in Reaction). I thus always run with `--no-verify`.

Can we just include this as an opt-in and _not_ added to everyones toolchain/hook? Or if we do want to add it to everyone's hooks, can we RFC and discuss?

orta: You should upgrade your computer - we support that every 3 years. 45 seconds is totally unreasonable, maybe something's up with reaction?

I'd *much* rather get an error before I push or commit rather than whenever I remember to go and look at the PR much later.

orta: I guess I find it a little tough to grok that you'd want blocking feedback on your code like that later, and decoupled from when you're actually doing the work. By the RFC, do you mean we should have an RFC for all of the pre-existing hooks too?

mzikherman: > By the RFC, do you mean we should have an RFC for all of the pre-existing hooks too?

Oh not at all! I just thought we were trying to do RFCs for changes like this, like we have in Peril, that Prettier line length change, etc. I'm gonna keep rolling `--no-verify` so you can add a `sleep(100)` here if you want :)))

> I guess I find it a little tough to grok that you'd want blocking feedback on your code like that later, and decoupled from when you're actually doing the work.

My computer is so slow I've always had CI run things and I got used to checking there :(

orta: I'd recommend [setting](https://github.com/typicode/husky/blob/master/bin/install.js#L15) `HUSKY_SKIP_INSTALL`  to true in your computers ENV, then you'll not get them anywhere.

Though you'd need to run `rm .git/hooks/*` in existing clones to remove the hooks entirely, then you'll never be annoyed with them in the guture 👍

zephraph: Given that there's an option to disable the hooks, I'm 👍 with this change. @mzikherman does the ability to disable the hooks alleviate some of your concerns?

mzikherman: Yea, for sure!

izakp: Just came across this as I was unsure why the pre-push hooks suddenly appeared for me, (I started also using `--no-verify`) and had to do some digging around to find when these hooks were committed - I want to echo @mzikherman in that features optimized for local dev workflow, while they may be valuable to a given individual, do not  _a priori_ extend to any other individual.  i.e. I think these statement expresses a personal preference:

> I'd much rather get an error before I push or commit rather than whenever I remember to go and look at the PR later.

> I guess I find it a little tough to grok that you'd want blocking feedback on your code like that later, and decoupled from when you're actually doing the work.

I sometimes prefer to use CI to offload the work async and/or run tests in a Docker container if I am testing a change that has the possibility to work differently in a local setup than it does in a containerized Linux environment.

@orta can you add documentation about the husky hooks and using `HUSKY_SKIP_INSTALL` to the project's README?  And in the future try and surface these changes within the organization, whether it is by adding more Artsy-org engineers to PRs like these and/or by means of the RFC process?

orta: I'm not sure it really provides enough value for adding extra docs to remove it, I'll just the remove the githooks 👍

orta: Ah, this PR didn't add the githooks, just a quick danger lint before pushing - I'll just remove this

PR #1174, opened 2018-07-16
disable api loader caching with CACHE_ENABLED env var

izakp:
Disable the API loader cache by setting `CACHE_ENABLED=false`

Todo: Establish a baseline of response times with the cache disabled.


src/lib/loaders/api/loader_without_authentication_factory.ts
```diff
@@ -39,76 +39,101 @@ export const apiLoaderWithoutAuthenticationFactory = (
...
+                          globalAPIOptions.requestIDs.requestID,
+                          apiName,
+                          key,
+                          { time, cache: false }
+                        )
+                      })
+                      .catch(err => {
+                        warn(key, err)
+                        reject(err)
+                      })
```
alloy: Can you move this into its own function and share it with the code above that handles a cache miss?

src/lib/cache.js
```diff
@@ -19,6 +20,8 @@ const isTest = NODE_ENV === "test"
 
 const VerboseEvents = ["issue", "failure", "reconnecting", "reconnect", "remove"]
 
+export const cacheEnabled = CACHE_ENABLED === "true"
```
alloy: It isn’t idiomatic to have camelCase constants. How about defining `CACHE_ENABLED` as a boolean in the `config` module instead of defining `cacheEnabled` at all?

src/lib/loaders/api/loader_without_authentication_factory.ts
izakp: That would be a bit DRYer but not sure it's worth it because of all the scoped variables inside the call to `api` - it's all in the context of the Promise and needs access to `api`,  `key`, `apiOptions`, `clock`, etc

src/lib/cache.js
izakp: Going to switch it to `CACHE_DISABLED` in that case as I want it to be by default, enabled

src/lib/loaders/api/loader_without_authentication_factory.ts
alloy: Good point.

PR #1176, opened 2018-07-16
Switch to the native promise in node

orta:
There's a possibility that we're using features in bluebird that aren't available in the native implementation (and without types, we don't know) so I'm doing a run through - FWIW, tests all pass, including a single integration test.


izakp:
👍 let's give it a go

ermik:
What's the reasoning behind this? Just that 'native works now' or something more complex?

PR #1177, opened 2018-07-16
Allows requesting a heapdump of metaphysics to the current working dir

orta:
screen shot 2018-07-16 at 4 23 35 pm

This will allow placing requesting a heapdump by `ps | grep metaph` to grab the process id, then sending `kill -USR2 [pid]`. It will place a file like `heapdump-410951910.530807.heapsnapshot` in the same folder as you're currently in, which can be shipped to s3 or whatever independently.

Happy debate whether it should live behind flag etc - running the import is what allows node to listen out for the OS signal, but this is the code needed.


orta:
Interesting, looks like it's not building the dependency in docker, will look tomorrow

orta:
Today is tomorrow. Ish. 

Anyway, I figured it out, and got it green. 👍

src/index.js
```diff
@@ -53,6 +54,10 @@ const enableRequestLogging = ENABLE_REQUEST_LOGGING === "true"
 const logQueryDetailsThreshold =
   LOG_QUERY_DETAILS_THRESHOLD && parseInt(LOG_QUERY_DETAILS_THRESHOLD, 10) // null by default
 
+if (NODE_ENV !== "test") {
```
izakp: Perhaps add a boolean config value such as `config.HEAP_DUMPS_ENABLED` and only import if it is enabled?  Would be good to isolate this if it has the potential to impact runtime performance if imported but unused.

orta:
Sure thing, this is now scoped behind an environment flag 👍

izakp:
👍

PR #1181, opened 2018-07-17
Revert "Switch to the native promise in node"

orta:
Reverts artsy/metaphysics#1176 - looks like performance is better with bluebird


joeyAghion:
Picture:

alloy:
Very interesting learning, nonetheless!

PR #1182, opened 2018-07-17
fail health check and shut down server on SIGTERM

izakp:
Following the boilerplate in https://github.com/serby/express-graceful-shutdown and the recommendations in https://blog.risingstack.com/graceful-shutdown-node-js-kubernetes/

Sending `SIGTERM` to the node process will:
- Shut down the server gracefully with `.close()` - stops accepting new connections
- Cause the health check to fail, so Kubernetes will take the pod out of service

If the server does not shut down completely (i.e. some long-standing connection remains open and `close()` does not complete, it will be [eventually terminated](https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html) with a `SIGKILL`


joeyAghion:
Nice! I'm sort of curious: what's express's default behavior upon receiving a `SIGTERM`?

Looks like the [default grace period](https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html) between `SIGTERM` and `SIGKILL` is 30 seconds, which seems fine for our purposes.

izakp:
> what's express's default behavior upon receiving a SIGTERM

Nothing - express doesn't set up any signal handling when it creates a node `server`

orta:
Agree, merge on green

PR #1184, opened 2018-07-18
hotfix - update DD_TRACER_SERVICE_NAME default and use in statsd lib

izakp:
DD_TRACER_SERVICE_NAME is set to 'metaphysics' in staging and prod - also use in statsd lib 

Need this to isolate the metrics from a test deployment with cache disabled


alloy:
Makes sense 👍

PR #1186, opened 2018-07-18
Workaround DNS issues - use default DNS resolvers inherited from the host machine

izakp:
Use the pod config `dnsPolicy: Default` to bypass internal DNS - this directly inherits `/etc/resolv.conf` from the host machine, and in our case in AWS it is set to:

```
nameserver 10.0.0.2
search ec2.internal
```

This will completely bypass the Kubernetes DNS service and use the AWS internal DNS resolvers - the pod will not be able to resolve other internal service names like `gravity-web` and `datadog`, but it will work around the known [ndots problem](https://rsmitty.github.io/KubeDNS-Tweaks/) and avoid redundant DNS queries for `api.artsy.net.default.svc.cluster.local` / `api.artsy.net.svc.cluster.local` / `api.artsy.net.ec2.internal` - without `ndots` specified, only domains without _any_ dots are treated as un-fully-qualified and the search domains are applied.

We do need to talk to Datadog running in k8s so set `DD_TRACER_HOSTNAME` and `STATSD_HOST` from the node name.

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -20,6 +20,15 @@ spec:
     spec:
       containers:
       - name: metaphysics-web
+        env:
+        - name: DD_TRACER_HOSTNAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: STATSD_HOST
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
         envFrom:
         - configMapRef:
             name: metaphysics-environment
@@ -43,6 +52,7 @@ spec:
               value: https
           initialDelaySeconds: 5
           periodSeconds: 5
+      dnsPolicy: Default
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -20,6 +20,15 @@ spec:
     spec:
       containers:
       - name: metaphysics-web
+        env:
+        - name: DD_TRACER_HOSTNAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: STATSD_HOST
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
         envFrom:
         - configMapRef:
             name: metaphysics-environment
@@ -43,6 +52,7 @@ spec:
               value: https
           initialDelaySeconds: 5
           periodSeconds: 5
+      dnsPolicy: Default
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:


joeyAghion:
Gotta ask: what's the default `dnsPolicy` if not `Default`?

I'm curious to see how this behaves. It would be troubling to have to configure _all_ deployments to work this way, but the data should be interesting.

izakp:
@joeyAghion it is `ClusterFirst` if unspecified.  See https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/

Yea I agree that I don't want to have to configure _all_ deployments like this - until we can upgrade to Kubernetes >= 1.9 though, where we can customize the `/etc/resolv.conf` file it's going to remain problematic

izakp:
I haven't had much luck with either patching resolver config for languages that use their own resolver implementations (i.e. for ruby you can use `resolv-replace` to switch to a pure-Ruby DNS resolver, or modifying flags for languages that use system calls to resolve DNS (for example NodeJS uses the [c-ares library](https://nodejs.org/api/dns.html) and should be controllable by setting the `RES_OPTIONS` env var)

alloy:
Wow. Is that 20s latency the lookup time?

Btw didn’t we also at some point determine that node wouldn’t cache lookups and thus quickly compound an issue like this?

izakp:
@alloy that's 20s latency lookup time for a domain that does not exist (ie. it should get `NXDOMAIN` back from the kube-dns system for something like `api.artsy.net.ec2.internal`...) But yes when we first moved MP over to k8s we used an Alpine image, and ran into [this issue](https://github.com/nodejs/docker-node/issues/602) and we determined Node does not have a dns cache - you suggested `dnsmasq` and that is provided by the kube-dns system and AFAICT working for valid domains like `api.artsy.net` - but good point in that if we pluck MP out of the Kubernetes DNS resolvers, it should run `dnsmasq` then locally as well.

alloy:
Makes sense 👍

Btw, it’s not clear to me if this situation also applied to our cache instances or if those were all internal domains that resolved quickly?

izakp:
> Btw, it’s not clear to me if this situation also applied to our cache instances or if those were all internal domains that resolved quickly?

So to be clear the problem addressed here by sidestepping Kubernetes DNS addresses is the fact that Kubernetes writes `ndots:5` into all containers' `/etc/resolv.conf` - presumably because they were designing for in-cluster services and not really expecting a combination of internal and external services.

From https://linux.die.net/man/5/resolv.conf - "Resolver queries having fewer than ndots dots (default is 1) in them will be attempted using each component of the search path in turn until a match is found."

Fortunately our Memcached ElasticCache names come in the form of `cachename.uvuuuk.0001.use1.cache.amazonaws.com` which has 5 dots so the search paths aren't used like above.

izakp:
Seems to be 👍 just using AWS nameservers https://app.datadoghq.com/apm/service/metaphysics-green.graphql-query/graphql.query?start=1532014602887&end=1532015991000&env=production&paused=true

alloy:
Got it! I think when we figured out that Node does not cache ns lookups it wasn’t really related to the incidents anyways, just mostly a performance observation. This still holds, so in the future I think it would make sense to re-investigate this.

PR #1178, opened 2018-07-16
Renamed the one_off_loader to uncached_loader, and sets saleArtworkRootLoader to be one

orta:
1. Makes the loader look like the rest of the loaders as a consumer
2. Changes `saleArtworkRootLoader` to be uncached
3. WIP because I want to sit down with you and make sure it's 100% - the system time API still works

PR #1189, opened 2018-07-20
Uses uncached gravity loader for artworks' sale artworks

ashfurrow:
Sale artworks have bid increments which change often, so we don't want to cache them. https://github.com/artsy/metaphysics/pull/1178 changed the sale artwork _root_ loader to not use the cache, but sale artworks accessed through the artwork schema were still being cached. Consider the following two ways to access the same information:

```graphql
{
  artwork(id:"5b4de57ef0676200017a0b15")  {
    sale_artwork {
      minimum_next_bid {
        display
      }
    }
  }
  
  sale_artwork(id:"5b4de57ff0676200017a0b1c") {
    minimum_next_bid {
      display
    }
  }
}
```

Even though they operate on the same model, they use different caching strategies. This PR would make them both uncached.

This is a fix for [PURCHASE-297](
https://artsyproduct.atlassian.net/browse/PURCHASE-297).


sweir27:
Seems pretty straightforward! 🎉

PR #1191, opened 2018-07-23
Catch up with Exchange CreateOrder changes

ashkan18:
# Change
We simplified order creation mutation in Exchange and in new mutation (with different name) we only accept `artworkId`, `editionSetId` and `quantity`, last two being optional.

# Depends on
https://github.com/artsy/exchange/pull/77


ashkan18:
@ashfurrow yes, basically catch up with Exchange schema changes.

ashkan18:
thanks all, merging, this means that buy now button on the test page we had we'll be broken now till we switch to this new mutation.

PR #1192, opened 2018-07-23
hotfix - use dumb-init to proxy signals correctly to node

izakp:
https://github.com/artsy/metaphysics/pull/1182 is not working correctly on k8s because k8s sends a `SIGTERM` to PID 1 in the container on termination, which is by default `sh -c` which does not proxy signals to the process(es) it spawns.  `dumb-init` is designed to fix this problem as well as others arising from processes run as PID 1.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,11 @@ FROM node:8.4.0
 RUN adduser --disabled-password --gecos '' deploy
 RUN mkdir -p /app
 
+# Set up dumb-init
+ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.2.0_amd64 /usr/local/bin/dumb-init
+RUN chown deploy:deploy /usr/local/bin/dumb-init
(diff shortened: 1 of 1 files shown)


izakp:
More here https://engineeringblog.yelp.com/2016/01/dumb-init-an-init-for-docker.html

izakp:
Even this ain't working as expected...

```
Received signal SIGTERM, shutting down
POST / 200 282.459ms 54.208.226.164 "node-superagent/2.3.0"
POST / 200 526.878ms 54.208.226.164 "node-superagent/2.3.0"
POST / 200 739.592ms 54.208.226.164 "node-superagent/2.3.0"
POST / 200 9.348ms 54.208.226.164 "node-superagent/2.3.0"
POST / 200 179.367ms 54.208.226.164 "node-superagent/2.3.0"
POST / 200 413.746ms 54.208.226.164 "node-superagent/2.3.0"
```

I suppose we do need to close keep-alive connections, perhaps those from the ELB are preventing full shutdown.

PR #1194, opened 2018-07-24
hotfix - destroy keepalive HTTP connections on shutdown

izakp:
Follows up on https://github.com/artsy/metaphysics/pull/1192 and https://github.com/artsy/metaphysics/pull/1182 - keepalive HTTP connections need to be destroyed otherwise server will not shut down.  Use https://www.npmjs.com/package/http-shutdown to destroy keepalive connections.

Also a fix in Dockerfile `CMD` syntax

cc @alloy @orta

### DIFF
--- a/index.js
+++ b/index.js
@@ -85,9 +85,9 @@ function bootApp() {
     app.use(require("./src").default)
   }
 
-  server = app.listen(port, () =>
+  server = require('http-shutdown')(app.listen(port, () =>
     info(`[Metaphysics] Listening on http://localhost:${port}`)
-  )
+  ))
 }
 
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -85,9 +85,9 @@ function bootApp() {
     app.use(require("./src").default)
   }
 
-  server = app.listen(port, () =>
+  server = require('http-shutdown')(app.listen(port, () =>
```
alloy: FYI Normal imports should just go at the top of the file, require should only be used in special cases (such as hot reloading code).

alloy:
Yet another great learning! 👌

PR #1197, opened 2018-07-25
#migration test cache compression in metaphysics-web-green deployment

izakp:
This sets up a "green" production deployment to test against the current "blue" deployment with cache compression disabled (all env vars hardcoded in the kubernetes yaml spec override any pulled in from the config map (set with `hokusai [staging|prouduction] env`)

The deployment will create 2 replicas which will be targeted by the metaphysics service along with the existing replicas, and so will receive a proportional amount of the traffic (around 8% with the current metaphysics scaling parameters)

We will be able to compare metrics taken from this sample set to existing metrics as they will come up as separate "metaphysics-green" services set by `DD_TRACER_SERVICE_NAME`

# Migration

Run `hokusai production update` to create the new deployment

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -63,6 +63,78 @@ spec:
                 values:
                 - foreground
 
+
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: metaphysics-web-green
+spec:
+  replicas: 2
+  strategy:
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 0
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: metaphysics
+        layer: application
+        component: web
+      name: metaphysics-web-green
+      namespace: default
+    spec:
+      containers:
+      - name: metaphysics-web-green
+        env:
+        - name: DD_TRACER_SERVICE_NAME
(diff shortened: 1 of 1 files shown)


alloy:
I’m still learning about these configs, but nothing wrong jumps out so, merge on green 👍

hokusai/production.yml
```diff
@@ -63,6 +63,78 @@ spec:
...
+      labels:
+        app: metaphysics
+        layer: application
+        component: web
+      name: metaphysics-web-green
+      namespace: default
+    spec:
+      containers:
+      - name: metaphysics-web-green
+        env:
```
izakp: These environment variables override anything pulled in via `envFrom` below...

hokusai/production.yml
```diff
@@ -63,6 +63,78 @@ spec:
                 values:
                 - foreground
 
+
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: metaphysics-web-green
```
izakp: Besides this name change, setting `replicas: 2` and the env overrides below, this is identical to the existing metaphysics-web deployment

hokusai/production.yml
alloy: Makes sense 👍

PR #1198, opened 2018-07-26
Resource / Autoscaling tweaks - introduce cpu-hog label

izakp:
- Bump memory request for better scheduling
- Unbounded CPU limit (shouldn't ever really go over 1 as node is single-threaded so no need to cap it)
- Add `temperament: cpu-hog` label and an [anti-affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity) scheduling rule for this label - Metaphysics pods will "repel" one another as well as any other pods with the `temperament: cpu-hog` label applied - this will mean that no two Metaphysics pods should ever be scheduled on the same host, and no Metaphysics pod should be scheduled on the same host of any [gravity-web pod](https://github.com/artsy/gravity/pull/11829) as it also uses this label.

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -146,7 +146,7 @@ spec:
     name: metaphysics-web
   minReplicas: 10
   maxReplicas: 20
-  targetCPUUtilizationPercentage: 70
+  targetCPUUtilizationPercentage: 80
 
 ---
 apiVersion: v1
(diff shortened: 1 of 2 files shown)


alloy:
Interesting!

One off topic note, Node isn’t really single-threaded, the context that the JS code is evaluated in is. V8 presumably runs multiple threads, all Node Native modules that do IO work run in different threads, etc.

izakp:
> V8 presumably runs multiple threads, all Node Native modules that do IO work run in different threads, etc.

Yes indeed there's the libuv threadpool to consider

PR #1199, opened 2018-07-26
Fix cents issues

ashkan18:
# Problem
When we are mapping Exchange's price related fields we used `amount` helper which would return formatted price strings BUT we kept calling them `itemTotalCents` and etc. which is pretty confusing.

# Solution
For each price field expose two fields from MP, one the actual `Int` cent value and another formatted one without `Cents` in their name. Basically now we have

```
itemsTotalCents
shippingTotalCents
taxTotalCents
commissionFeeCents
transactionFeeCents
buyerTotalCents
sellerTotalCents
itemsTotal
shippingTotal
taxTotal
commissionFee
transactionFee
buyerTotal
sellerTotal
```

cc: @sepans @jonallured this impacts your current setup when you show prices, you may need to switch to non `Cents` fields.


src/schema/ecommerce/types/order_line_item.js
```diff
@@ -27,6 +27,10 @@ export const OrderLineItemType = new GraphQLObjectType({
       description: "Edition set on the artwork",
       resolve: ({ editionSetId }) => editionSetId,
     },
+    priceCents: {
+      type: GraphQLInt,
+      description: "Price in cents",
+    },
```
joeyAghion: Unclear if this is the price of 1 item or the subtotal of `quantity` items. The description could clarify.

orta:
Yeah, this is totally fine from a stitching perspective. We would merge in the original schema to get in the "XYZcents" then we would stitch in the resolvers we have here. 👍

joeyAghion:
This works for me! I'm fearful that merging may break dependent apps that need to be demo-ed. If that's all clear, feel free to merge.

orta:
This looks whole incremental, so shouldn't pose a problem

ashkan18:
👍 merging!

starsirius:
Nice! I always have this question of how much data presentation Metaphysics should be responsible for, but the price stuff seems to be pretty standard across consuming clients.

PR #1204, opened 2018-07-31
FulfillAtOnce Mutation

ashkan18:
# Change
We added `fulfillAtOnce` mutation in https://github.com/artsy/exchange/pull/89 this PR removes old `finalizeOrder` and adds mapping `fulfillAtOnce`.

Also changed `sampleOrder` to be a function so we can conditionally add `lineItems` and `fulfillments`.


src/test/fixtures/results/sample_order.js
```diff
@@ -1,4 +1,4 @@
-export default {
+const defaultResponse = {
```
sepans: You could also keep this as is and add fulfillments to the js object in the test where it is needed, instead of having the below function, if you thing it is simpler.

ashkan18: i thought about it but then we had to merge the `fulfillments` somewhere in nested hash, right?

ashkan18:
merging so i can try this on staging 🙌

PR #1209, opened 2018-08-07
Returns non-exchange order errors to client.

sepans:
In cases where running a mutation causes an error which is not coming from exchange (like authentication errors from gravity)
the error is in `result.errors` and not in `result.data.<query/mutation-name>.errors` so 
```
const { order, errors } = result.data.<query/mutation-name>
```
 throws a javascript runtime error in metaphysics and client receives the ambiguous error message: *Cannot match against 'undefined' or 'null'.*

This change saves developers debugging time for now, while we are figuring out the best pattern to handle graphql errors for order mutations.

(cc @orta)


orta:
Yeah, agree, this makes sense for now yep 👍

sweir27:
🙏 thank you!

PR #1212, opened 2018-08-13
#migration add goreplay sidecar container and forward scraped traffic

izakp:
Depends on https://github.com/artsy/substance/pull/103 - Do not merge this PR until it is

Run [goreplay](https://goreplay.org/) as a sidecar container with Metaphysics (containers in the same Pod share a network interface as if they were on the same host) - scrape traffic and forward to `goreplay.stg.artsy.systems` / `goreplay.prd.artsy.systems` for later reply.

Enabling / disabling traffic scraping and forwarding can be done by commenting out these changes, then running `hokusai [staging|production] update` (can't think of a good way enable / disable with an env var as running another container or not alongside MP is the cleanest deployment strategy)

# Migration

After merging, sync your local `master` branch then run:

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -52,6 +52,14 @@ spec:
               value: https
           initialDelaySeconds: 5
           periodSeconds: 5
+      - name: metaphysics-goreplay
+        image: artsy/goreplay:latest
+        args: ["/usr/local/bin/goreplay", "--input-raw", ":3000", "--output-tcp", "goreplay.prd.artsy.systems:80"]
+        securityContext:
+          capabilities:
+            add:
+            - NET_RAW
+            - NET_ADMIN
       dnsPolicy: Default
       affinity:
         nodeAffinity:
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -52,6 +52,14 @@ spec:
               value: https
           initialDelaySeconds: 5
           periodSeconds: 5
+      - name: metaphysics-goreplay
+        image: artsy/goreplay:latest
+        args: ["/usr/local/bin/goreplay", "--input-raw", ":3000", "--output-tcp", "goreplay.stg.artsy.systems:80"]
+        securityContext:
+          capabilities:
+            add:
+            - NET_RAW
+            - NET_ADMIN
       dnsPolicy: Default
       affinity:
         nodeAffinity:


alloy:
What a glorious day! ✨ 

I’m not seeing anything obviously wrong in this PR, but other than that it’s hard for me to judge it. I’ll await others’ review to see if there are any issues that need to be addressed (e.g. security concerns).

izakp:
> if there are any issues that need to be addressed (e.g. security concerns)

Jaaaa, this effectively introduces a special container with special privileges `CAP_NET_RAW` and `CAP_NET_ADMIN` as defined here: https://linux.die.net/man/7/capabilities

The container is a potential attack surface but it's not exposed anywhere so there's not additional port as an attack surface - @bhoggard wdyt?

izakp:
@alloy rolled out on staging with `hokusai staging update` - now collecting logs - see https://github.com/artsy/substance/pull/103 for instructions on how to view / retrieve them.

alloy:
Ace! Looking forward to grab some juicy logs in the next few days.

Btw how is it on staging when this PR hasn’t been merged yet?

izakp:
I ran `hokusai staging update` locally against these changes (which doesn't change the image tag deployed, only the k8s config - we intentionally left the image ref decoupled from the rest of the configuration in `./hokusai/staging.yml` / `./hokusai/production.yml` - note that these files contain a reference to the image tag `staging` / `production` respectively, which serves as a pointer to the currently deployed Git-SHA1 tag in `hokusai registry images`) - perhaps I should be explicit in that this needs a migration after merge.

joeyAghion:
@alloy @izakp we ended up needing to revert these .yml changes (https://github.com/artsy/metaphysics/pull/1218) and `update` in staging and production because `hokusai` wouldn't let us deploy otherwise (due to [this guard](https://github.com/artsy/hokusai/blame/51460c875981b7b27feae5b95227a2215c3dc5ae/hokusai/services/deployment.py#L125-L126)). Sorry if this impacts your work. I'm not sure of the right work-around, or maybe running multiple containers requires a hokusai update.

izakp:
Thanks @joeyAghion - yea I didn't think about this guard.  I'll refactor so we either a) run the same container but with different processes, b) run `goreplay` as a service within the main Metaphysics container

PR #1216, opened 2018-08-14
Update access token retrieval instructions

ds300:
The instructions for getting a local GraphiQL app setup were tripping me up and @alloy suggested I do this process instead, and that we should update the README to match.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,17 +91,8 @@ We recommend the [graphiql.app](https://github.com/skevy/graphiql-app) client fo
 
 You will need to set up headers with both:
 
-* `x-access-token` - Get your client ID and secret, use this command to generate
-  a token.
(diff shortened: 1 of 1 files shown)


alloy:
Sweet ✨ 

Merge on green.

orta:
We should also consider switching to GraphQL-playground too, that lets you set those vars in the site - then you don't need the app at all. I use it with peril: https://staging-api.peril.systems/api/graphiql?query=%7B%0A%20%20me%0A%7D

screen shot 2018-08-14 at 12 14 40 pm

alloy:
@orta is it extensible? I’d love to surface persisted queries.

orta:
We can pass in a set of `Tabs` - which could include all persisted queries: https://github.com/prismagraphql/graphql-playground#properties

Maybe we could also make it so that tabs aren't reset each time too

PR #1219, opened 2018-08-15
#migration add goreplay sidecar container and forward scraped traffic... again

izakp:
Follow-up to https://github.com/artsy/metaphysics/pull/1212

Install Goreplay inline and use the same image ref to run the sidecar container, working around the Hokusai guard that all containers in a Pod need to reference the same image tag.

As before, I'd let this sit on staging for a bit and check that everything is flowing correctly and it doesn't step on the toes of the normal application operation / ci pipeline before promoting to production and running the migration.

# Migration

After merging, sync your local `master` branch then run:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,11 @@ ADD https://github.com/buger/goreplay/releases/download/v0.16.1/gor_0.16.1_x64.t
 RUN tar xfvz /tmp/goreplay.tar.gz
 RUN mv goreplay /usr/local/bin/
 RUN rm -f /tmp/goreplay.tar.gz
+RUN addgroup gor
+RUN addgroup deploy gor
+RUN chgrp gor /usr/local/bin/goreplay
+RUN chmod 0750 /usr/local/bin/goreplay
+RUN setcap "cap_net_raw,cap_net_admin+eip" /usr/local/bin/goreplay
 
 RUN npm install -g yarn@1.0.1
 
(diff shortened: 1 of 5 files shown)


izakp:
Enable by setting env vars GOREPLAY_PORT and GOREPLAY_OUTPUT - disabled by leaving unset.

alloy:
Unlike #1212, this has changes to the `Dockerfile`, do these changes also not need to be in the master branch for us to be able to safely deploy staging?

izakp:
> do these changes also not need to be in the master branch for us to be able to safely deploy staging

Yup - changes in the `Dockerfile` need to be built into a new image - so will need to merge / build / deploy first then run `hokusai [staging|production] update`

alloy:
Ah ok, cool. I think I misunderstood when you said that you were going to let this sit for a while in staging, I thought you meant to not merge the PR like previously.

Let’s do this then 👍

izakp:
Ah yes I meant to merge but not run `hokusai production update` for a bit - but then I added the `run-goreplay.sh` script to configure / enable / disable as well.

Examples:

Scrape from port 3000 and log to stdout

```
hokusai production env set GOREPLAY_PORT=3000 GOREPLAY_OUTPUT=stdout
hokusai production refresh
```

Scrape from port 3000 and forward to Goreplay aggregator service at `goreplay.prd.artsy.systems:80`:

```
hokusai production env set GOREPLAY_PORT=3000 GOREPLAY_OUTPUT="tcp goreplay.prd.artsy.systems:80"
hokusai production refresh
```

Disable Goreplay:

```
hokusai production env unset GOREPLAY_PORT GOREPLAY_OUTPUT
hokusai production refresh
```

PR #1221, opened 2018-08-16
Fea take heapsnapshots

izakp:
Take heapsnapshots and upload to s3 with https://docs.minio.io/docs/minio-client-complete-guide

Includes an update to dumb-init

Invoke `./scripts/create-heapsnapshot.sh` with env vars "AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_BUCKET and AWS_BUCKET_PATH"

Execute on a running pod - i.e. for pod `metaphysics-web-2612115710-645fg` (which you would find via `hokusai [staging|production] status`), and passing in your own AWS credentials to upload to the `artsy-data` bucket, `metaphysics/heapsnapshots/staging` path, you would run:

```
kubectl exec metaphysics-web-2612115710-645fg -- bash -c "AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY AWS_BUCKET=artsy-data AWS_BUCKET_PATH=metaphysics/heapsnapshots/staging ./scripts/create-heapsnapshot.sh"
```

Snapshots ( sometimes 2 identical files are created...  not sure why cc @orta ) are uploaded with a timestamp and the hostname (in Kubernetes' case, the pod name prepended to the snapshot name) for later inspection.

To start with let's manually trigger the script at times of low and high memory consumption and take a look at the comparison of snapshots for a given pod over its lifetime.  In the future we may want to trigger this automatically, i.e. when a pod is terminated it can take a final snapshot using the `preStop` [lifecycle hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/).

### DIFF
--- a/scripts/create-heapsnapshot.sh
+++ b/scripts/create-heapsnapshot.sh
@@ -1,0 +1,27 @@
+#! /bin/bash
+
+if [[ -z $AWS_ACCESS_KEY_ID || -z $AWS_SECRET_ACCESS_KEY || -z $AWS_BUCKET || -z $AWS_BUCKET_PATH ]]; then
+  echo "Unset require env vars AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_BUCKET and AWS_BUCKET_PATH"
+  exit 1
+fi
+
+# Configure mc
+/usr/local/bin/mc config host add s3 https://s3.amazonaws.com $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY
+
+# create heapsnapshot - dumb-init proxies signal to children
+kill -USR2 1
+
+# find heapsnapshots
+for heapsnapshot in $(ls -l | grep heapsnapshot | awk '{print $9}')
+do
+  # wait for heapsnapshot placeholder to be replaced with full heapdump
+  while [ ! -s $heapsnapshot ]; do sleep 1; done
+  echo "Created $heapsnapshot"
+
+  # Upload heapsnpshot to S3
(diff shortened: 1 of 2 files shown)


izakp:
Remember to set `ENABLE_HEAPDUMPS=true` so Metaphysics loads the required library and listens for the `USR2` signal as per https://github.com/artsy/metaphysics/pull/1177

scripts/create-heapsnapshot.sh
```diff
@@ -0,0 +1,29 @@
+#! /bin/bash
+
+# Check required env vars
+if [[ -z $AWS_ACCESS_KEY_ID || -z $AWS_SECRET_ACCESS_KEY || -z $AWS_BUCKET || -z $AWS_BUCKET_PATH ]]; then
+  echo "Unset required env vars AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_BUCKET and AWS_BUCKET_PATH"
```
alloy: > ~~Un~~set

?

alloy:
One minor comment, otherwise looks splendid 👍

scripts/create-heapsnapshot.sh
izakp: Aha yea makes more sense as an error message

izakp:
I'm going to make a `metaphysics` IAM user with S3 access to `artsy-data` and set env vars persistently, so we can just run `kubectl exec {pod_name} -- bash -c "./scripts/create-heapsnapshot.sh"`

PR #1222, opened 2018-08-16
[dev] Make sure app is fully started on initial request.

alloy:
Previously the app would start launching when the initial request comes
in, but not yet be finished launching, which would lead to the inital
request failing. This was especially painful when that initial request
was an introspection query being performed by e.g. GraphiQL and then
leading to the app not knowing the schema.

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,15 +64,18 @@ function bootApp() {
       .end()
   })
 
-  app.get('/health', (req, res) => {
+  app.get("/health", (req, res) => {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -64,15 +64,18 @@ function bootApp() {
       .end()
```
alloy: These are just prettier changes

damassi:
LGTM 👍

Merge on green

PR #1227, opened 2018-08-17
Add shippingName and catch up with Exchange schema updates

ashkan18:
# Change
Catch up with Exchange schema updates for adding `shippingName` to `Order` and `setShipping` mutation.


src/test/fixtures/exchange/orders.json
```diff
@@ -9,6 +9,7 @@
         "partnerId": "111",
         "userId": "111",
         "fulfillmentType": "SHIP",
+        "shippingName": "Dr Collector",
```
ashfurrow: 💯

PR #1229, opened 2018-08-20
@ashkan18 an18 => update schema to allow null shipping names

sweir27:
I'm not sure what all the other formatting is all about :(


ashkan18:
i noticed in https://github.com/artsy/metaphysics/blob/master/src/schema/ecommerce/set_order_shipping_mutation.ts#L14 all fields are optional, this should probably reflect Exchange's definition for required/non-required so we get better error messages and etc. This was my bad in not adding those here, at the moment all fields are optional.

PR #1230, opened 2018-08-20
add data for Artwork condition box in force

oxaudo:
This still needs tests but PRing it here early to get some feedback.

The UI we need to create in force is like that:
https://app.zeplin.io/project/5ac6790d7f57100518f9ab44/screen/5b606192307c780395c2b26e

My thinking is that the best way to accomplish that is to return an object that stores label/details for each row and then metaphysics is going to be responsible for setting those properly. UI in force will just look if the object is there - display it - otherwise skip the row.

### DIFF
--- a/src/schema/artwork/index.js
+++ b/src/schema/artwork/index.js
@@ -35,6 +35,7 @@ import AttributionClass from "schema/artwork/attributionClass"
 import attributionClasses from "../../lib/attributionClasses"
 import { LotStandingType } from "../me/lot_standing"
 import { amount } from "schema/fields/money"
+import { capitalizeFirstCharacter } from "lib/helpers"
 
 const is_inquireable = ({ inquireable, acquireable }) => {
   return inquireable && !acquireable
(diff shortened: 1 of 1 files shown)


src/schema/artwork/index.js
```diff
@@ -705,6 +706,70 @@ export const artworkFields = () => {
...
+    signatureInfo: {
+      type: ArtworkInfoRowType,
+      resolve: ({
+        signature,
+        signed_by_artist,
+        stamped_by_artist_estate,
+        sticker_label,
+        signed_other,
+      }) => {
+        var detailsParts = []
```
starsirius: #minor. Can use `let` instead.

src/schema/artwork/index.js
```diff
@@ -705,6 +706,70 @@ export const artworkFields = () => {
...
+          detailsParts.push("hand-signed by artist")
+        }
+        if (stamped_by_artist_estate) {
+          detailsParts.push("stamped by artist's estate")
+        }
+        if (sticker_label) {
+          detailsParts.push("sticker label")
+        }
+        if (signature && signature.length > 0) {
+          detailsParts.push(signature)
```
starsirius: If I understand correctly, in the design, the description (free text signature) is after a period, not joined by ", ".

src/schema/artwork/index.js
```diff
@@ -705,6 +706,70 @@ export const artworkFields = () => {
...
     },
+    framed: {
+      type: ArtworkInfoRowType,
+      resolve: ({ framed }) => {
+        if (!framed) {
+          return null
+        }
+        return {
+          label: "Framed",
+          details: "",
```
starsirius: #minor. It's boolean, so maybe we can simply return `true` or `false`, if we want to differentiate unset value. However, it's a required field in CMS for BNMO.

oxaudo: My understanding is that in Metaphysics we want to have a value that can be directly displayed in clients. So as far as clients are concerned - they either get a string to display or get null. The implementation as boolean in gravity is kinda masked here.

That is seems to be the direction we are going lately - just get metaphysics to return what clients will use without any additional checks. Let me know if I'm misunderstanding something here @starsirius.

src/schema/artwork/index.js
oxaudo: Ha... Not sure. Missed that part. Let me check with @briansw

src/schema/artwork/index.js
starsirius: Yeah I agree with the response can be client-driven, and I simply wondered if this would be too specific to a single client, i.e. Force. Can be an ongoing discussion!

PR #1231, opened 2018-08-20
Changes the mutation method for creating an order to be based on the new mutation response types

orta:
Pairs with https://github.com/artsy/exchange/pull/93

Switches the metaphysics representation of exchange to use the new mutation response type.


src/schema/__tests__/ecommerce/create_order_with_artwork_mutation.test.js
```diff
@@ -11,54 +12,62 @@ describe("Create Order Mutation", () => {
...
       },
+      OrderOrFailureUnion: {
+        __resolveType(obj, context, info) {
+          if (obj.order) {
+            return "OrderWithMutationSuccess"
+          } else if (obj.error) {
+            return "OrderWithMutationFailure"
+          }
+        },
+      },
```
orta: this took a bit of thinking to figure out our reported error, docs for [it are here](https://www.apollographql.com/docs/graphql-tools/resolvers#Unions-and-interfaces)

src/schema/__tests__/ecommerce/create_order_with_artwork_mutation.test.js
```diff
@@ -11,54 +12,62 @@ describe("Create Order Mutation", () => {
...
     }
 
     rootValue = mockxchange(resolvers)
   })
+
   it("creates order and returns it", () => {
-    const mutation = `
+    const mutation = gql`
```
orta: `gql` is a no-op template string function, but it tricks prettier/vscode into formatting our graphqls correctly

src/schema/ecommerce/create_order_with_artwork_mutation.ts
```diff
@@ -0,0 +1,125 @@
...
+        ecommerce_createOrderWithArtwork(
+          input: {
+            artworkId: $artworkId
+            editionSetId: $editionSetId
+            quantity: $quantity
+          }
+        ) {
+          orderOrError {
+            ... on EcommerceOrderWithMutationSuccess {
+              order {
```
sweir27: Is this because the actual union type is defined in Exchange? It's sort of frustrating to have to list out all of the fields here :( (just one more thing that needs to be updated when we add a new one).

PR #1232, opened 2018-08-21
set X-Forwarded-For header on downstream requests

izakp:
Set the `X-Forwarded-For` [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#Directives) on downstream requests to represent a proxy chain to downstream services.

If the header exists, append the Client IP to the header from `req.connection.remoteAddress` - if it does not exist, simply set it to the Client IP.

### DIFF
--- a/src/lib/requestIDs.js
+++ b/src/lib/requestIDs.js
@@ -1,10 +1,19 @@
 import uuid from "uuid/v1"
 
-export function headers({ requestID, traceId, parentSpanId }) {
+export function headers({ requestID, traceId, parentSpanId, xForwardedFor }) {
   return {
     "X-Request-Id": requestID,
     "x-datadog-trace-id": traceId,
     "x-datadog-parent-id": parentSpanId,
+    "X-Forwarded-For": xForwardedFor,
+  }
+}
+
+function resolveProxies(req) {
+  if (req.headers["x-forwarded-for"]) {
+    return req.headers["x-forwarded-for"] + ", " + req.connection.remoteAddress
+  } else {
+    return req.connection.remoteAddress
   }
 }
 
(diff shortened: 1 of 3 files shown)


joeyAghion:
* How is the `requestIDs` module used? Would this behavior be used for all upstream requests from Metaphysics, just requests to Gravity, or something else?
* Might this cause Gravity's rate-limiting to kick in where it previously didn't? This may be fine, or the other conditions (such as excluding Artsy's own clients) may still prevent that, but I just want to be aware.
* Do you know if these `X-Forwarded-For` IPs make it into logs?

alloy:
> How is the `requestIDs` module used? Would this behavior be used for all upstream requests from Metaphysics, just requests to Gravity, or something else?

All those clients that decide to include it, which at the time of writing is Gravity (both REST and GraphQL requests) and all the stitching clients.

alloy:
Code looks good 👍  I’ll defer to @joeyAghion to merge re his questions.

src/lib/requestIDs.js
```diff
@@ -1,23 +1,35 @@
...
     "X-Request-Id": requestID,
     "x-datadog-trace-id": traceId,
     "x-datadog-parent-id": parentSpanId,
+    "X-Forwarded-For": xForwardedFor,
+  }
+}
+
+function resolveProxies(req) {
+  if (req.headers["x-forwarded-for"]) {
+    return req.headers["x-forwarded-for"] + ", " + req.connection.remoteAddress
```
alloy: FYI MP uses interpolation throughout the codebase:

```js
`${req.headers["x-forwarded-for"]}, ${req.connection.remoteAddress}`
```

izakp:
@joeyAghion 

> Might this cause Gravity's rate-limiting to kick in where it previously didn't?

I think we will still ignore all of our own clients, so as long as the [xapp token is present throttling will still be ignored](https://github.com/artsy/gravity/blob/87194599bd04910797a67aa4224b8022388517f0/config/initializers/rack_attack.rb#L61).

> Do you know if these X-Forwarded-For IPs make it into logs

It depends on how we configure express and lookup the IPs in our codebase - for example, if we want `req.ip` [here](https://github.com/artsy/metaphysics/blob/master/src/lib/checkForProblematicArtistQuery.ts#L11) to resolve to the actual client IP we need to set `app.set('trust proxy', true)` as per http://expressjs.com/en/guide/behind-proxies.html - 

```
If true, the client’s IP address is understood as the left-most entry in the X-Forwarded-* header.

If false, the app is understood as directly facing the Internet and the client’s IP address is derived from req.connection.remoteAddress. This is the default setting.
```

I suppose we want to enable this as we do so in Force

PR #1234, opened 2018-08-21
adding specs

oxaudo:
For what is implemented here: https://github.com/artsy/metaphysics/pull/1230

Not sure it's the final version of signature part but that is how I understand it now:)


oxaudo:
@starsirius - based on discussions in slack seems like another 'not_signed' attribute will be added in gravity. Will need to handle it here as well.

src/schema/artwork/__tests__/artwork.test.js
```diff
@@ -1194,4 +1194,201 @@ describe("Artwork type", () => {
...
+        }
+      }
+    `
+    it("is null when framed is null", () => {
+      artwork.framed = null
+      return runQuery(query, rootValue).then(data => {
+        expect(data).toEqual({ artwork: { framed: null } })
+      })
+    })
+    it("is null when framed is false", () => {
```
starsirius: Had a minor comment in https://github.com/artsy/metaphysics/pull/1230/files#r212498908 here, too.

src/schema/artwork/__tests__/artwork.test.js
```diff
@@ -1194,4 +1194,201 @@ describe("Artwork type", () => {
...
+      return runQuery(query, rootValue).then(data => {
+        expect(data).toEqual({ artwork: { framed: null } })
+      })
+    })
+    it("is set to proper object when framed is true", () => {
+      artwork.framed = true
+      return runQuery(query, rootValue).then(data => {
+        expect(data).toEqual({
+          artwork: {
+            framed: { label: "Framed", details: null },
```
starsirius: Is the details here supposed to be `true`?

src/schema/artwork/__tests__/artwork.test.js
```diff
@@ -1194,4 +1194,201 @@ describe("Artwork type", () => {
...
+        artwork(id: "richard-prince-untitled-portrait") {
+          certificateOfAuthenticity {
+            label,
+            details
+          }
+        }
+      }
+    `
+    it("is null when certificate_of_authenticity is null", () => {
+      artwork.framed = null
```
starsirius: Should be `artwork. certificateOfAuthenticity = null`

src/schema/artwork/__tests__/artwork.test.js
```diff
@@ -1194,4 +1194,201 @@ describe("Artwork type", () => {
...
+      })
+    })
+    it("is set to proper object when certificate_of_authenticity is true", () => {
+      artwork.certificate_of_authenticity = true
+      return runQuery(query, rootValue).then(data => {
+        expect(data).toEqual({
+          artwork: {
+            certificateOfAuthenticity: {
+              label: "Certificate of authenticity",
+              details: null,
```
starsirius: Ditto. Is the details here supposed to be ""?

src/schema/artwork/__tests__/artwork.test.js
oxaudo: My thinking here that it's null since it's never available for Framed right now but wanted to put this into the same format as other fields in checklist.

So in my mind distinction is - it's null when it's never available for now and it's "" when it's just not entered in UI.

Let me know if it sounds bogus. I can definitely put "" in all the cases @starsirius

starsirius: I was confused by the test here when I referred to the [implementation](https://github.com/artsy/metaphysics/pull/1230/files#diff-249d94a23065d61e259bcfa48241ce6dR717) in previous PR, and wondered why the test passed. Now I see it's actually changed in this PR [below](https://github.com/artsy/metaphysics/pull/1234/files#diff-249d94a23065d61e259bcfa48241ce6dR717). 👍 

Re: "never available for now", I assumed you meant BNMO hasn't been released on artsy.net. I was under the impression we are going to guard that with a lab feature on the frontend, instead of on MP, but I could be misunderstanding the situation here. Happy to discuss further!

PR #1237, opened 2018-08-23
Renamed gql to have .ts extension, moved to lib.

ashfurrow:
This should fix our deploys.


alloy:
Can you add some context as to how it broke deploys?

alloy:
Ahh, I see now https://github.com/artsy/metaphysics/pull/1231/files#diff-f9d4e6854b72b3605765cc6521551415R11. So we probably are indeed skipping test files in the build, but production runtime code was trying to import it and thus failing.

PR #1239, opened 2018-08-24
Update shipping mutation, use common order fields for tests

ashkan18:
# Changes
Update `setOrderShipping` mutation to match latest Exchange schema where we pass shipping as a separate hash.
Move common order fields to a separate file so we can share and we don't have to change n files every time we add/update fields.


src/schema/__tests__/ecommerce/order_fields.ts
```diff
@@ -0,0 +1,135 @@
+import gql from "lib/gql"
+
+export const OrderBuyerFields = gql`
```
alloy: Once we have GraphQL intellisense setup in MP at some point, this won’t really work as it’s not valid GraphQL syntax (does the syntax highlighter work?). You can fix that by making these inline fragments:

```gql
... on Order {
  id
  code
  # etc
}
```

alloy:
Looks good to me 👍 

My one comment isn’t a blocker right now, so feel free to merge as-is and add the fragment syntax later or do it now.

alloy:
Ace 👌

PR #1244, opened 2018-08-27
Buyer/Seller is the new User/Partner

ashkan18:
# Problem
`User` and `Partner` are very specific terms and not clear when it comes to ecommerce platform, we decided to [rename](https://github.com/artsy/exchange/pull/121) these fields in Exchange. This PR matches those changes by:
- Adding `OrderPartyUnionType`
- Switch all the places we use `user` and `partner`

# Update Notes
## CMS Side
Make sure to update your `orders` query. Replace `partnerId` argument to `sellerId` and `sellerType` with seller type being `Partner` (we can make this an enum too later.

## CMS and Collectors
in your queries replaces
```
userId
partnerId
 ```
to 
```graphql
seller {
    ...on Partner {
      id
      name
    }
    ... on User {
      id
      email
    }
  }
buyer {
    ... on User {
      id
      email
    }
    ...on Partner {
      id
      name
    }
  }
```

Ideally this gets deployed around the same time as https://github.com/artsy/exchange/pull/121

### DIFF
--- a/src/data/exchange.graphql
+++ b/src/data/exchange.graphql
@@ -167,6 +167,7 @@ type MutationError {
 
 # An Order
 type Order {
+  buyer: OrderPartyUnion!
   buyerTotalCents: Int
   code: String!
   commissionFeeCents: Int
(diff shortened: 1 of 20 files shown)


src/data/exchange.graphql
```diff
@@ -167,6 +167,7 @@ type MutationError {
 
 # An Order
 type Order {
+  buyer: OrderPartyUnion!
```
jonallured: `OrderPartyUnion!` sounds so fun!! 🎉 🍰 🎈

ashkan18:
merging along with exchange changes being merged now.

PR #1247, opened 2018-08-28
New style mutation for 5 more ecommerce mutations

ashkan18:
Follow up with https://github.com/artsy/exchange/pull/133 and updated 5 more mutations to new style.

# Update Note
Update your mutation responses to use 
```graphql
orderOrError {
  ... on OrderWithMutationSuccess {
    order {
      <order fields you want>
    }
  }
    error {
      description
    }
  }
}
```
instead of current
```graphql
result {
  order {
     <order fields you want>
  }
  errors 
}
```


src/schema/ecommerce/approve_order_mutation.ts
```diff
@@ -0,0 +1,80 @@
...
+                description
+              }
+            }
+          }
+        }
+      }
+    `
+    return graphql(exchangeSchema, mutation, null, context, {
+      orderId,
+    }).then(result => result.data!.ecommerce_approveOrder)
```
ds300: Before these callbacks were inspecting `result.errors` which is ignored now. Is that a potential source of protocol-level error messages that we should be propagating?

ashkan18: i _think_ that was for us to be able to read response in old format from Exchange and properly raise errors here, i think in this current case we are fine mainly since this call basically passes Exchanges response up and if there are errors it will be in the response returned from MP.

sepans: This is were I added those: https://github.com/artsy/metaphysics/pull/1209

I don't exactly remember the specific query I was running but I think it was something like confirming an order using authentication token from a wrong partner. @ashkan18 it would be great to give it a try because it seems this kind of errors bypass exchange error handling, unless that has been updated too.

Thanks @ds300 for catching it.

ashkan18: i _think_ we are good here now since this basically gets the response from exchange (either Order or an error) and returns it, before we had to see if there were errors and raise them to get MP style errors i think, but lets keep an eye on it. Im never 100% sure here :)

ds300: @alloy and I were debugging the issue described here: https://artsy.slack.com/archives/C1HH3KNJG/p1535634594000100 and discovered that these `results.errors` are indeed important. We were seeing a 401 error manifesting as a `cannot set property 'clientMutationId' of null` error because the `result.data.ecommerce_<foo>` properties are null when exchange rejected the request. So I'll make a PR to add the checks back in.

ashkan18: awesome! thanks! sorry for causing it, i guess here we'd need to check for `result.data.ecommerce_...orderOrError.error`?

ds300: That one gets propagated to the client, we'd need to check for `result.errors` here, just like it was doing before.

alloy: `result.data.ecommerce_...orderOrError.error` are user errors, other uncaught (runtime/protocol) errors thrown during execution will still go to the top-level errors.

PR #1253, opened 2018-08-30
Add totalCount to order connection …

ashkan18:
and also catch up with exchange's  new fields aka `lastApprovedAt` and `lastSubmittedAt`.


src/schema/ecommerce/types/order.ts
```diff
@@ -134,4 +136,10 @@ export const {
   edgeType: OrderEdge,
 } = connectionDefinitions({
   nodeType: OrderType,
+  connectionFields: {
+    totalCount: {
+      type: GraphQLInt,
+      resolve: ({ totalCount }) => totalCount,
```
orta: wanna humor me and see if we can remove this line?

ashkan18: looks like a very removable line... testing now.

orta:
👍

PR #1256, opened 2018-08-31
Removes the json file from the schema dumps

orta:
We used to need schema `.json` files for relay classic, but now we're on relay modern - we can just use the `.graphql` files.

This PR:

- Renames the script-y files to ts 
- Stops the JSON file from being dumped


alloy:
👍

Merge on green

PR #1259, opened 2018-08-31
Refactor the query check in the eigen queries

orta:
Moves the check for a query to where it's being used.

### DIFF
--- a/src/lib/nameOldEigenQueries.ts
+++ b/src/lib/nameOldEigenQueries.ts
@@ -3,7 +3,7 @@ import { RequestHandler } from "express"
 
 export const nameOldEigenQueries: RequestHandler = (req, _res, next) => {
   const agent = req.headers["user-agent"]
-  if (agent && agent.includes("Eigen") && req.body.query) {
+  if (req.body.query && agent && agent.includes("Eigen")) {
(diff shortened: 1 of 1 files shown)


src/lib/nameOldEigenQueries.ts
```diff
@@ -3,9 +3,9 @@ import { RequestHandler } from "express"
 
 export const nameOldEigenQueries: RequestHandler = (req, _res, next) => {
   const agent = req.headers["user-agent"]
-  if (agent && agent.includes("Eigen") && req.body.query) {
+  if (agent && agent.includes("Eigen")) {
     const { query } = req.body as { query: string }
-    if (!query.startsWith("query ")) {
+    if (query && !query.startsWith("query ")) {
```
alloy: I think this check should be the very first one in this function. There’s no need to do string contents checking on all incoming requests when we can already cut down many of them from the get go. So:

```ts
if (req.body.query && agent && agent.includes("Eigen") && req.body.query) {
  ...
}
```

alloy: It’s the little perf things that make up the whole

orta:
cool, merge on green 👍

PR #1260, opened 2018-08-31
Creates a script for running emission stored queries against instances of metaphysics

orta:
OK, so given the results of #1257 - I took a step back and wondered what other linting tools can we have that don't require putting stitching on staging. One of them is to run our production queries against different hosts of metaphysics.

This PR takes our stored queries, loops through them and calls them against an instance of MP

screen shot 2018-08-31 at 5 00 29 pm

- it handles looking up variables for a bunch of queries (I focused on root query renderers as they tend to have the complexity I'm interested in testing)
- it skips mutations
- it uses the same env var names as reaction, so you can C&P

It *does not* add to CI - that's feasible, I'm not too sure where the value lives there, but I'm open to ideas.

### DIFF
--- a/src/__tests__/runStoredQueryTests.ts
+++ b/src/__tests__/runStoredQueryTests.ts
@@ -1,0 +1,134 @@
+import { readFileSync } from "fs"
+import fetch from "lib/apis/fetch"
+import chalk from "chalk"
+const { USER_ACCESS_TOKEN: accessToken, USER_ID: userID } = process.env
+
+if (!accessToken && !userID) {
+  console.error(`
+  Cannot run queries without a user access token and user ID. You can get one by 
+  - 1. Signing in on staging.artsy.net
+  - 2. Open a dev console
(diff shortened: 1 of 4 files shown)


src/__tests__/runStoredQueryTests.ts
```diff
@@ -0,0 +1,134 @@
...
+  Dev,
+}
+
+const url = getURL(Env.Prod)
+
+const saleArtworkID =
+  "georg-jensen-beaded-slash-kugle-sterling-silver-cutlery-60"
+
+// We want to be able to run queries that require non-null variables
+// this is a map of a query name to their hard-coded variables.
```
alloy: Ah yes, so when we switch to a db for persisted queries we should also store example variables that we can use to test all these queries.

alloy: https://artsyproduct.atlassian.net/browse/PLATFORM-571

alloy:
Nice one! 👌

PR #1261, opened 2018-09-02
[tracer] Further improvements

alloy:
* Record user-agent of HTTP requests (https://github.com/DataDog/dd-trace-js/pull/266).
* Correctly send trace IDs over the wire to upstream services. (After figuring out we were doing the wrong thing by not ending IDs as big-endian 64-bit ints, I also found that dd-trace’s http plugin already [does this](https://github.com/DataDog/dd-trace-js/blob/aa607860ea6bb57f2f69fa7ee72ed3ce25abfbf6/src/plugins/http.js#L40) and we don’t need to do anything.)
* Fix double-counting of upstream HTTP requests (https://github.com/DataDog/dd-trace-js/pull/268).


alloy:
Going to self-merge and see results in staging asap.

Merge on green

izakp:
👍

PR #1264, opened 2018-09-03
resolve ipv6-formatted addresses to ipv4

izakp:
Seeing some remote ip addresses set to ipv6 notation in the proxy chain `X-Forwarded-For` headers rather than ipv4. See https://stackoverflow.com/questions/29411551/express-js-req-ip-is-returning-ffff127-0-0-1

Resolve ipv6 to ipv4 notation where appropriate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "hot-shots": "^5.6.1",
     "http-shutdown": "^1.2.0",
     "i": "^0.3.5",
+    "ip": "^1.1.5",
     "invariant": "^2.2.4",
     "jwt-simple": "^0.5.0",
     "lodash": "^4.3.0",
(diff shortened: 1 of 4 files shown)


alloy:
Is this potentially breaking us supporting ipv6 in any way?

izakp:
No it's just resolving ipv4 addresses in ipv6 notation to plain'ol ipv4 notation... See the SO answer...

> In IPv6, you are allowed to remove leading zeros, and then remove consecutive zeros, meaning ::ffff: actually translates to 0000:0000:ffff:0000, this address has been designated as the IPv4 to IPv6 subnet prefix, so any IPv6 processor will understand it's working with an IPv4 address and handle it accordingly.

alloy:
Ahh, gotcha, so it leaves actual ipv6 in tact 👍

PR #1267, opened 2018-09-04
Adds a logger to the HTTP links

orta:
This takes the trick I mentioned [in slack](https://artsy.slack.com/archives/C1HH3KNJG/p1536079718000100) and makes it apply to all HTTP links - this'll be useful in knowing what stitching is doing:

```GraphQL
{
  submission(id:3) {
    title
  }
}
```

```
> Made query to Convection:
>
query ($_v0_id: ID) {
  submission(id: $_v0_id) {
    title
  }
}
> Got Response:
> {"data":{"submission":{"title":"My artwork!!!"}}}
```

and the current exchange work, which uses multiple resolvers + http links under the hood. Either of you are fine to handle the PR.


src/lib/stitching/logLinkMiddleware.ts
```diff
@@ -0,0 +1,26 @@
+import { ApolloLink } from "apollo-link"
+import { print } from "graphql/language"
+
+const shouldLogLinkTraffic = !!process.env.LOG_HTTP_LINKS
```
ashkan18: nice! so the default here would be `false`, right?

orta: It's an optional env var, so the default is for it to not exist, which fails the not not

PR #1272, opened 2018-09-07
[COORDINATED MERGE] Move phoneNumber into Ship and setShipping input

ds300:
⚠️ Do not merge until we have all PRs related to this change in place and approved.

here it is working in metaphysics local (running against exchange from artsy/exchange#162):

image


ds300:
merge on green

PR #1276, opened 2018-09-13
[COORDINATED MERGE] update exchange errors api

ds300:
This is in response to artsy/exchange#172

tested locally: 

image


src/schema/__tests__/ecommerce/approve_order_mutation.test.js
```diff
@@ -32,7 +32,9 @@ describe("Approve Order Mutation", () => {
             }
             ... on OrderWithMutationFailure {
               error {
-                description
+                type
```
ashkan18: 🙌 I think we could make this whole 
```
error {
  type
  description
  data
}
``` 
a fragment so we could define it in one place and reuse it but we can do that later.

ashkan18:
oh just updated the comment to include example.

ashkan18:
merging....

Issue #1240, opened 2018-08-24
[Stitching] Type and root field prefixing


alloy:
We’ve already decided to prefix all types with a descriptive name (i.e. not just the service name, because that’s an implementation detail), but we should maybe do so for root fields and mutations as well.

E.g. right now
* we rename all types coming from exchange with `Ecommerce*` https://github.com/artsy/metaphysics/blob/aeadfb6155019ba0e200558060bde008b42707e7/src/lib/stitching/exchange/schema.ts#L22
* we rename root fields in the internally used exchange schema to `ecommerc_*`https://github.com/artsy/metaphysics/blob/aeadfb6155019ba0e200558060bde008b42707e7/src/lib/stitching/exchange/schema.ts#L24
* we don’t appear to do any renaming of mutations, but I also have no experience doing that so maybe I’m overlooking it?

Seeing as we don’t use stitching yet, we’re currently manually proxying these. However, the current exposed types/root fields/mutations all do _not_ have any of these prefixes, thus going to break the schema once we do switch over to a stitched version.

In short, the work that needs to be done is make sure that all types/root fields/mutations in our current paying setup have the same prefixes as would be the case when stitching. Additionally we should use camelCase, so root fields should become `ecommerceFoo` not `ecommerce_foo` (same applies to mutations).

ashkan18:
@orta do you remember why we added `ecomerce_` to root fields?

orta:
A mutation *is* a root field, I did a bit of fidgeting but this is the schema:

```graphql
"# Autogenerated input type of ApproveOrder
input EcommerceApproveOrderInput {
  # A unique identifier for the client performing the mutation.
  clientMutationId: String
  id: ID!
}

# Autogenerated return type of ApproveOrder
type EcommerceApproveOrderPayload {
  # A unique identifier for the client performing the mutation.
  clientMutationId: String

  # A union of success/failure
  orderOrError: EcommerceOrderOrFailureUnion!
}

# Autogenerated input type of CreateOrderWithArtwork
input EcommerceCreateOrderWithArtworkInput {
  # Artwork Id
  artworkId: String!

  # A unique identifier for the client performing the mutation.
  clientMutationId: String

  # EditionSet Id
  editionSetId: String

  # Number of items in the line item
  quantity: Int
}

# Autogenerated return type of CreateOrderWithArtwork
type EcommerceCreateOrderWithArtworkPayload {
  # A unique identifier for the client performing the mutation.
  clientMutationId: String

[...]

PR #1279, opened 2018-09-13
Adds tests to the exchange stitching notes

orta:
This is what the schema being merged in would look like.

Fixes https://github.com/artsy/metaphysics/issues/1240


orta:
Note: this doesn't stitch, and so can be merged regardless. I will look *now* at another PR that does stitching.

orta:
Correct - I think it maked sense to actually allow those input types to stay the same, then it's zero breaking changes for clients

orta:
merge on green

PR #1281, opened 2018-09-13
Converts stitching to be incremental

orta:
WIP - because red, but I ran out of time.


_schema.graphql
```diff
@@ -4269,29 +4269,6 @@ type Mutation {
     input: UpdateConversationMutationInput!
```
orta: the order in which these are defined has changed, so the text ordering has changed, should be the same values

src/lib/stitching/mergeSchemas.ts
```diff
@@ -2,52 +2,55 @@ import { mergeSchemas as _mergeSchemas } from "graphql-tools"
...
-  const anyMergedSchema = mergedSchema as any
-  anyMergedSchema.__allowedLegacyNames = ["__id"]
+  if (ENABLE_GRAVQL_ONLY_STITCHING) {
+    const gravitySchema = await executableGravitySchema()
+    schemas.push(gravitySchema)
+  }
 
-  return mergedSchema
-}
+  if (ENABLE_ORDER_STITCHING) {
```
ashkan18: should we call this Ecommerce instead of Order?

src/lib/stitching/mergeSchemas.ts
```diff
@@ -56,3 +59,39 @@ export const mergeSchemas = async () => {
 
   return mergedSchema
 }
+
+// The end goal:
```
ashkan18: do you want to keep these commented code around?

src/schema/index.ts
```diff
@@ -147,14 +148,48 @@ const Viewer = {
...
+
+if (!ENABLE_ORDER_STITCHING) {
+  // Deprecated
+  stitchedRootFields.order = Order
+  stitchedRootFields.orders = Orders
+
+  stitchedRootFields.ecommerceOrder = Order
+  stitchedRootFields.ecommerceOrders = Orders
+
+  // Deprecated
```
ashkan18: 🙌

src/lib/stitching/mergeSchemas.ts
orta: I do, I'm OK with this for now, I'll remove it once we hit some kind of partiy

orta:
> going to have both prefixed and not-prefixed mutations/types go through stitching right?

Right now metaphysics _without stitching exchange_ supports both, when stitched it supports only the prefixed version. I'm not flipping the exchange switch yet though, so this gives time to switch clients over.

PR #1282, opened 2018-09-14
bump memory limits in prod

izakp:
Been running this all week and it seems better - still getting OOMKilled but not as frequently

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -40,9 +40,9 @@ spec:
         resources:
           requests:
             cpu: 500m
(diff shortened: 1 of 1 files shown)


izakp:
No, I just made an estimation - I looked at some heapdumps and saw a lot more memory allocated to Datadog spans in open traces, but it seemed to be GC-ing properly, just looked like this change required some more headroom.   We could try pushing it even further.

alloy:
Gotcha, thanks for that update 👍

PR #1283, opened 2018-09-14
Adds published to the Artwork's schema

orta:
This converts the artwork to TypeScript, and removes the errors from the linters/compilers. The overall point was to add published.


src/schema/artwork/__tests__/artwork.test.js
```diff
@@ -792,6 +802,15 @@ describe("Artwork type", () => {
...
         })
 
+        it("is not hangable if the category is not applicable to wall display like installations", () => {
+          artwork.category = "installation"
+          artwork.width = 100
+          artwork.height = 100
+          return runQuery(query, rootValue).then(data => {
+            expect(data.artwork.is_hangable).toBe(false)
+          })
+        })
```
orta: I added this test to see whether the logic in `is_hangable` was doing what we thought it was, and it failed at first

src/schema/artwork/index.ts
```diff
@@ -780,7 +786,7 @@ export const ArtworkType = new GraphQLObjectType({
       cached,
     }
   },
-})
+} as any)
```
orta: These are rabbit holes that turn into black holes very quickly, so I as any'd it

PR #1288, opened 2018-09-20
[rateLimiter] Limit at 100 req / 15 min.

alloy:
Grabbed these numbers from the package’s README https://github.com/nfriedly/express-rate-limit#usage, which I assume are sensible values but open to suggestions.


src/lib/rateLimiter.ts
```diff
@@ -0,0 +1,22 @@
+import RateLimit from "express-rate-limit"
+import MemcachedStore from "rate-limit-memcached"
+import { client } from "./cache"
+import { Request } from "express"
+
+// Max 100 requests in 15 minutes from a single IP
+const max = 100
+const expiration = 15 * 60
```
joeyAghion: Can these be config values rather than hard-coded?

The max seems surprisingly low and the window surprisingly large (especially considering that requests from the same network will appear to us as a single IP, and requesting a single artsy.net page can generate a dozen or 2 direct requests to Metaphysics). Although, since this is scoped to just 1 query at the moment, I guess it might be enough to block our specific problem.

Finally, I wonder how we'll monitor the rate at which these `429` responses get generated. Can we easily tell from logs or ELB metrics? If not, we should record a custom metric or risk not knowing when this is occurring or when we might mistakenly be limiting legitimate requests.

alloy: Good point, made the values configurable 👍 

The papertrail logs do show status code and if we’d accidentally limit e.g. Force then we should in the traces see those HTTP requests having failed with a 429 status code.

orta:
Looking at the code, It will fall through to the normal force error handlers if the request to MP fails - which should pass though the error code etc

orta:
I added a script to verify it happens:

```
#!/bin/bash

for i in `seq 1 160`; do 
  curl 'https://metaphysics-staging.artsy.net/?' \
      -XPOST \
      -H 'Accept: application/json' \
      -H 'Content-Type: application/json' \
      -H 'Origin: https://metaphysics-staging.artsy.net' \
      -H 'Content-Length: 563' \
      -H 'Accept-Language: en-us' \
      -H 'Host: metaphysics-staging.artsy.net' \
      -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Safari/605.1.15' \
      -H 'Accept-Encoding: json' \
      -H 'Connection: keep-alive' \
      --data-binary '{"query":"query routes_OverviewQueryRendererQuery \n{\n  artist(id: \"glenn-brown\") {\n    artworks_connection(sort: ICONICITY_DESC, first:10) {\n      totalCount\n      pageCursors {\n        last{\n          page\n          cursor\n        }\n      }\n      edges {\n        node {\n          title\n          images {\n            image_versions\n            image_url\n          }\n        }\n      }\n    }\n  }\n}\n","variables":{"artworkID":"dabeiyuzhou-4th-genome-test-partner-wei-lai-fo-futurebudda"},"operationName":"routes_OverviewQueryRendererQuery"}'
 done
```

and it triggered:

screen shot 2018-09-20 at 3 29 20 pm

PR #1290, opened 2018-09-20
Remove un-resovled EditionSet from Order

ashkan18:
# Problem
Some apps where requesting edition set details https://github.com/artsy/volt/pull/3227 by doing 

```graohql
order{
  editionSet{
    id
   }
}
```
Which raises error.

# Cause
We had `editionSet` as a field with type `EditionSet.type` on an `Order` but we were not actually resolving it to an `EditionSet` and we were just returning `editionSetId`.

# Solution
Remove `editionSet` field and just expose `editionSetId`.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -4690,8 +4690,8 @@ type OrderLineItem {
   # Artwork that is being ordered
   artwork: Artwork
 
-  # Edition set on the artwork
-  editionSet: EditionSet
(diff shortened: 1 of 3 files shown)


orta:
merge on green

ashkan18:
checked Eigen/Force/Reaction and we are fine with this change 👍

PR #1291, opened 2018-09-20
Removes the newly created resolvers, and fixes the internal schema resolving for exchange 

orta:
OK, so this one *really* fixes the issues we were seeing. What happened is:

- Adding #1279 changed the exchange schema manipulations for fields
- This broke the *internal* graphql resolving for exchange
- I thought it broke the external resolving (thus #1289 which this reverts)

Now the internal schema (and its test fixtures) are always in sync.

### DIFF
--- a/src/lib/stitching/exchange/schema.ts
+++ b/src/lib/stitching/exchange/schema.ts
@@ -17,16 +17,18 @@ export const executableExchangeSchema = async () => {
   })
 
   // Return the new modified schema
-  return transformSchema(schema, [
-    // Apply a prefix to all the typenames
-    new RenameTypes(name => {
-      return `Ecommerce${name}`
-    }),
-    // Rename all the root fields to be camelCased
-    new RenameRootFields(
-      (_operation, name) =>
-        `ecommerce${name.charAt(0).toUpperCase() + name.slice(1)}`
-    ),
-  ])
+  return transformSchema(schema, transformsForExchange)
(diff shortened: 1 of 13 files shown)


src/lib/stitching/exchange/schema.ts
```diff
@@ -17,15 +17,18 @@ export const executableExchangeSchema = async () => {
...
-    new RenameRootFields(
-      (_operation, name) =>
-        `ecommerce${name.charAt(0).toUpperCase() + name.slice(1)}`
-    ),
-  ])
+  return transformSchema(schema, transformsForExchange)
+  // Note that changes in this will need to be
 }
+
+export const transformsForExchange = [
```
ashkan18: 🙌

src/test/fixtures/exchange/mockxchange.js
```diff
@@ -48,12 +49,7 @@ export const mockxchange = resolvers => {
   })
 
   // namespace schema similar to src/lib/stitching/exchange/schema.ts
-  const exchangeSchema = transformSchema(schema, [
-    new RenameTypes(name => {
-      return `Ecommerce${name}`
-    }),
-    new RenameRootFields((_operation, name) => `ecommerce_${name}`),
-  ])
+  const exchangeSchema = transformSchema(schema, transformsForExchange)
```
ashkan18: no more copy/paste

PR #1292, opened 2018-09-20
Adds credit card connection + deleteCreditCard mutation

sweir27:
JIRA reference: https://artsyproduct.atlassian.net/browse/PURCHASE-483

This PR adds metaphysics support for the user settings payment page, where we need to be able to:
- display a list of a user's credit cards
- allow a user to delete their credit cards

(Design reference)
image


src/schema/credit_card.js
```diff
@@ -3,8 +3,39 @@ import {
   GraphQLObjectType,
   GraphQLString,
   GraphQLNonNull,
+  GraphQLUnionType,
 } from "graphql"
 import { GravityIDFields } from "schema/object_identification"
+import { GravityMutationErrorType } from "lib/gravityErrorHandler"
+
+const CreditCardMutationSuccessType = new GraphQLObjectType({
```
sweir27: Just moved these out of the other file

src/lib/gravityErrorHandler.js
```diff
@@ -23,7 +23,16 @@ export const formatGravityError = error => {
...
+      const parsedError = JSON.parse(errorSplit[1])
+      if (parsedError.error) {
+        return {
+          type: "error",
+          message: parsedError.error,
+          detail: parsedError.text,
+        }
+      } else {
+        return { ...parsedError }
+      }
```
damassi: 👍

src/schema/me/index.js
```diff
@@ -80,8 +82,8 @@ const Me = new GraphQLObjectType({
         request,
         { rootValue: { meCreditCardsLoader } }
       ) => {
-        return meCreditCardsLoader().then(results => {
-          return results.length > 0
+        return meCreditCardsLoader().then(({ body }) => {
+          return body && body.length > 0
```
damassi: Curious, was this previously a bug? 

screen shot 2018-09-20 at 3 41 15 pm

damassi:
LGTM 👍 

merge on green

src/schema/me/index.js
sweir27: We used to have this loader with out `headers: true`, so it returned the body directly (so this would have been an empty array).

I updated it because I thought I needed the headers to handle the relay connection, but ended up not using it in the end. Could change back but I'm not sure there's an argument either way.

damassi: np! 👍

PR #1293, opened 2018-09-21
Update exchange schema

jonallured:
This PR gets MP back in sync with Exchange for the order sorts that I added over there with https://github.com/artsy/exchange/pull/186.


orta:
This has a failed snapshot tests BTW, I wonder if we should be putting this in a snapshot or just always make it output to the FS instead

PR #1294, opened 2018-09-21
add --git-remote origin to pipeline promote and deploy steps in readme

mbilokonsky:
ugh, prettier added a ton of style changes, I'll make comments under the actual changes I made. @orta @joeyAghion let me know if this is what you wanted, I haven't run this just added the documentation changes that Joey said we needed.

PR #1296, opened 2018-09-23
Update config.yml to include the remote

orta:
This basically copies how [gravity does it](https://github.com/artsy/gravity/blob/master/.circleci/config.yml#L46) - so at least staging deploys will end up with a note

For prod we'll need #1294


.circleci/config.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           command: hokusai configure --kubectl-version 1.6.3 --s3-bucket artsy-citadel --s3-key k8s/config --platform linux
       - run:
           name: Deploy
-          command: hokusai staging deploy $CIRCLE_SHA1
+          command: hokusai staging deploy $CIRCLE_SHA1 --git-remote origin
```
mbilokonsky: Do we want `origin` here or do we want `upstream`/`artsy`?

orta: It's running on CI, so it should be origin

orta:
We chatted about this IRL - merge on green

PR #1301, opened 2018-09-25
[rateLimiter] Trust clients based on having set a DataDog header.

alloy:
Because the DD libs already patch all outgoing HTTP requests to include
this, meaning we don’t need to go through all our code to find the
places where we make requests and update each of those.

Additionally we shouldn’t rely on these types of headers at all in the
long run, so using this for pragmatic reasons seems fine.

### DIFF
--- a/src/lib/__tests__/rateLimiter.test.ts
+++ b/src/lib/__tests__/rateLimiter.test.ts
@@ -8,15 +8,15 @@ describe("rateLimiter", () => {
         headers: {},
         body: {},
       }
-      expect(skip(req as Request)).toBeFalsy
+      expect(skip(req as Request)).toBeFalsy()
     })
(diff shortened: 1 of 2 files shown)


src/lib/__tests__/rateLimiter.test.ts
```diff
@@ -8,15 +8,15 @@ describe("rateLimiter", () => {
         headers: {},
         body: {},
       }
-      expect(skip(req as Request)).toBeFalsy
+      expect(skip(req as Request)).toBeFalsy()
```
alloy: These tests never ran… 🙈

orta: hah, too much ruby in your background dude

PR #1303, opened 2018-09-25
Add billing address fields to the CreditCard type

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-466

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -2406,6 +2406,24 @@ type CreditCard {
 
   # Credit card's expiration year
   expiration_year: Int!
+
+  # Billing address street1
+  street1: String!
+
+  # Billing address street2
+  street2: String
+
+  # Billing address city
+  city: String!
+
+  # Billing address state
+  state: String!
+
+  # Billing address country code
+  country: String!
+
+  # Billing address postal code
+  postal_code: String!
 }
 
 # A connection to a list of items.
--- a/src/schema/credit_card.js
+++ b/src/schema/credit_card.js
@@ -61,6 +61,30 @@ const CreditCardType = new GraphQLObjectType({
       type: GraphQLNonNull(GraphQLInt),
       description: "Credit card's expiration year",
     },
+    street1: {
+      type: GraphQLNonNull(GraphQLString),
+      description: "Billing address street1",
+    },
+    street2: {
+      type: GraphQLString,
+      description: "Billing address street2",
+    },
+    city: {
+      type: GraphQLNonNull(GraphQLString),
+      description: "Billing address city",
+    },
+    state: {
+      type: GraphQLNonNull(GraphQLString),
+      description: "Billing address state",
+    },
+    country: {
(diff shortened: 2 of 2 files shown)


_schema.graphql
```diff
@@ -4347,9 +4365,6 @@ type Mutation {
     input: CreateGeminiEntryForAssetInput!
   ): CreateGeminiEntryForAssetPayload
 
-  # Records an artwork view.
-  recordArtworkView(input: RecordArtworkViewInput!): RecordArtworkViewPayload
```
yuki24: I wasn't sure if this is intentional, but seems like this is gone now. 

cc @mzikherman

_schema.graphql
```diff
@@ -5787,16 +5802,6 @@ type Query {
...
 
-input RecordArtworkViewInput {
-  artwork_id: String!
-  clientMutationId: String
-}
-
-type RecordArtworkViewPayload {
-  artwork_id: String!
-  clientMutationId: String
-}
```
yuki24: Same here, not sure why these are gone.

src/schema/credit_card.js
```diff
@@ -61,6 +61,30 @@ const CreditCardType = new GraphQLObjectType({
...
+      description: "Billing address state",
+    },
+    country: {
+      type: GraphQLNonNull(GraphQLString), // TODO: We may make this type more strict by throwing ISO "ALPHA-2 Codes
+      description: "Billing address country code",
+    },
+    postal_code: {
+      type: GraphQLNonNull(GraphQLString),
+      description: "Billing address postal code",
+    },
```
yuki24: So this type basically just passes the specified fields from Gravity down to the client. I wonder if we should use more appropriate names or leave them as they are now.

alloy: Maybe grouping these under an `address` field would be nice? Maybe it can even share an ‘address’ type with the fulfilment types, which means you could theoretically re-use components as well.

alloy: This is by no means a requirement, just thinking out loud because you were musing about it too.

alloy:
I’ll re-assign to @orta for those `recordArtworkView` changes.

src/schema/credit_card.js
yuki24: Noticing that the name `state` is very confusing in React...

orta:
There aren't any `recordArtworkView` changes anymore, so I'm good for merging.

yuki24:
@sweir27 @orta I'll send a follow-up PP rather than reverting.

src/schema/credit_card.js
anandaroop: :ceiling-cat:

Heh, but a "federal state" is actually a [thing](https://en.wikipedia.org/wiki/Federation).

When "state" isn't good enough, the generic & jargony name for that kind of entity is "first-level administrative division" — which encompasses states, provinces, cantons, etc — and is often abbreviated in code as `admin1`.

alloy: TIL. Well stated!

PR #1304, opened 2018-09-25
Make credit card billing address fields nullable

yuki24:
A lot of credit cards in Gravity do not have these fields filled in and marking them as NonNull would make it impossible to query.


sweir27:
Gracias for the follow-up!

PR #1307, opened 2018-09-26
Convert more files to TypeScript + update babel

orta:
I had some easy thinking time on a computer last night so I converted more JS files to TS, and updated babel to move off the betas.


orta:
Oh cool, guess that fail was just an edge case on my setup

package.json
```diff
@@ -184,5 +182,10 @@
       "yarn prettier --write",
       "git add"
     ]
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged; yarn dump-schema _schema.graphql"
+    }
   }
```
orta: husky 1.0 asks that you move everything into this

src/lib/apis/fetch.ts
```diff
@@ -1,13 +1,15 @@
...
 
+// TODO: This `any` is a shame, but
+// the type seems to be a bit of a mix of the original
+// response and some faffing
+
 export default (url, options = {}) => {
-  return new Promise((resolve, reject) => {
-    const opts = clone(
+  return new Promise<any>((resolve, reject) => {
+    const opts: any = clone(
```
alloy: I actually have a local branch that changes this implementation completely and also fixes the typings, so I’ll try to get that up asap.

src/lib/loaders/loaders_with_authentication/tracked_entity.ts
```diff
@@ -11,10 +11,10 @@ import { map, find, extend } from "lodash"
  * @param {string} [entityKeyPath] an optional path to a nested entity
  */
 const trackedEntityLoaderFactory = (
-  dataLoader,
-  paramKey,
-  trackingKey,
-  entityKeyPath
+  dataLoader: any,
```
alloy: Don’t we know the type of this?

alloy:
Nice! 👍 

Only question is about that `any` type for a `dataLoader` param, otherwise feel free to merge.

src/lib/loaders/loaders_with_authentication/tracked_entity.ts
orta: Eh, somewhat - after some exploration it's a few levels deep inside the dataloader factories with inferred typing most of the way. Don't think we'll gain too much by having the /exact/ type, so I've tightened it to a func that returns a promise 👍

PR #1310, opened 2018-09-26
Expose shippingContinentalUsOnly flag on artwork

javamonn:
Exposing this flag for [PURCHASE-477](https://artsyproduct.atlassian.net/browse/PURCHASE-477). Currently, continental shipping only is also tested for when building the `shippingInfo` field and duplicate the logic there.


src/schema/artwork/index.ts
```diff
@@ -511,6 +511,17 @@ export const artworkFields = () => {
     pickup_available: { type: GraphQLBoolean },
     price: { type: GraphQLString },
     price_currency: { type: GraphQLString },
+    shippingContinentalUsOnly: {
```
sweir27: #minor but I wonder if `shipsToContinentalUSOnly` or even just `shipsToUSOnly` is clearer?

src/schema/artwork/__tests__/artwork.test.js
```diff
@@ -1232,6 +1232,51 @@ describe("Artwork type", () => {
...
+    const query = `
+      {
+        artwork(id: "richard-prince-untitled-portrait") {
+          shippingContinentalUsOnly
+        }
+      }
+    `
+    it("is true when domestic_shipping_fee_cents and international_shipping_fee_cents reflect continental only shipping", () => {
+      artwork.domestic_shipping_fee_cents = 1000
+      artwork.international_shipping_fee_cents = 0
```
sweir27: If `international_shipping_fee_cents` is 0, doesn't that mean that it has _free_ shipping? I would expect this case to only be true if it is `null`.

src/schema/artwork/index.ts
javamonn: I did a lot of waffling on what to name this field and I agree that `shipsToContinentalUSOnly` is clearer than what I currently have, I'll update.

src/schema/artwork/__tests__/artwork.test.js
javamonn: I'm not sure! I duplicated the logic that `shippingInfo` uses, which has the same check, i.e. checking that `international_shipping_fee_cents` is falsey, not just `null`: https://github.com/javamonn/metaphysics/blob/50ef4974728bcb5363539711c69a98ac1d25149d/src/schema/artwork/__tests__/artwork.test.js#L1186-L1196. Perhaps we should update both spots?

sweir27: Let's check in the #commerce-pilot-pde channel since there may be some nuance to the UI/modeling that I'm not aware of. But if it's supposed to just check for `null` then we should for sure update both!

javamonn:
Updated this PR after a [discussion in slack](https://artsy.slack.com/archives/CB110C6LE/p1538072221000100) so that `domestic_shipping_fee_cents` and `international_shipping_fee_cents` treat `null` as indicating non-availability and `0` as free shipping. Note that this also changes how `shippingInfo` constructs its display string.

src/schema/artwork/__tests__/artwork.test.js
```diff
@@ -1195,8 +1195,32 @@ describe("Artwork type", () => {
...
     })
 
-    it("is set to free domestic shipping when domestic_shipping_fee_cents is null and international_shipping_fee_cents is present", () => {
+    it("is set to international shipping only when its international_shipping_fee_cents is present and domestic_shipping_fee_cents is null", () => {
       artwork.domestic_shipping_fee_cents = null
+      artwork.international_shipping_fee_cents = 1000
+      return runQuery(query, rootValue).then(data => {
+        expect(data).toEqual({
+          artwork: {
+            shippingInfo: "Shipping: $10 rest of world only",
```
javamonn: This is awkward copy, and we don't enforce international shipping only in reaction - should this be ticketed, or treated as an edge case for now?

javamonn:
Updated to reflect that works must ship to the US i.e. there will currently not be a case where a work is international shipping only.

sweir27:
🙏 thanks for sticking through this!

PR #1312, opened 2018-09-27
Removes the extra query wrapping for eigne queries that are mutations

orta:
Removes the extra wrappers around mutations as a mutation cannot live inside a query.


sweir27:
Coool, thanks!

joeyAghion:
For the record: https://sentry.io/artsynet/metaphysics-production/issues/670436878/?query=is:unresolved

PR #1313, opened 2018-09-28
Update exchange.graphql and add confirmPickup mutation

ashkan18:
# Change
`confirmPickup` mutation was added in https://github.com/artsy/exchange/pull/201 This PR catches up with that and adds `confirmPickup` mutation to MP

Depends on https://github.com/artsy/exchange/pull/201


_schema.graphql
```diff
@@ -1032,7 +1032,7 @@ type ArtworkContextFair {
     section: String
 
     # Sorts for shows in a fair
-    sort: ShowSort = FEATURED_DESC
+    sort: ShowSort = null
```
ashkan18: not sure why these are here

orta: This is something to do with schema stitching - it's blocking me on 
https://github.com/artsy/metaphysics/pull/1311 too

orta: I'll fix it today and unblock this PR

ashkan18:
@orta i was expecting to see `confirmPickup` mutation in the danger message ☝️ , right?

_schema.graphql
```diff
@@ -4416,6 +4427,9 @@ type Mutation {
   # Fulfills an Order with one fulfillment by setting this fulfillment to all line items of this order
   fulfillOrderAtOnce(input: FulfillOrderAtOnceInput!): FulfillOrderAtOncePayload
 
+  # Confrims pickup for an ecommerce order
+  confirmPickup(input: ConfirmPickupInput!): ConfirmPickupPayload
```
orta: Please drop this one, I know it'll be inconsistent in clients (and I'd like to go and migrate everything to use `ecommerce` asap)

orta:
Thanks you

PR #1314, opened 2018-09-28
Add support for dumping schema as json

zephraph:
Adds the ability to dump the schema as a json as well as continuing support to dump the schema as SDL.


orta:
Cool, merge on green - this is so that Ruby clients and Reaction (until https://github.com/artsy/reaction/pull/1228 is ready) can request the JSON schema from metaphysics

PR #1315, opened 2018-09-28
Loosen node version requirements

zephraph:
Currently metaphysics is pinned to 8.11.3. Having the patch version pinned seems a little more restrictive than necessary. This opens it up to only pin to the minor version.


alloy:
Isn’t that to match the deployed version of MP? I don’t really see a problem with using a newer minor version locally, but would like to understand if there’s a practical benefit?

PR #1316, opened 2018-09-28
Run schema.json through prettier

zephraph:
Runs results through prettier for more consistent formatting


zephraph:
merge on green

PR #1321, opened 2018-10-03
[package] Update dd-trace to use our fork with connection tracing.

alloy:
Besides moving this patch upstream https://github.com/DataDog/dd-trace-js/pull/299, the cleanup to the connection establishment tracing is to use the same `service` name as the HTTP request that it belongs to.


orta:
👍

PR #1323, opened 2018-10-04
Add creditCardId to response coming back from Exchange

ashkan18:
# Problem
When getting `creditCard` info back after `setOrderPayment` mutation, we were getting `null` even when credit card was set properly on the order.

# Cause
We were missing getting `creditCardId` from the Exchange's response coming back.

# Solution
Well... add it!

### DIFF
--- a/src/schema/__tests__/ecommerce/set_order_payment_mutation.test.ts
+++ b/src/schema/__tests__/ecommerce/set_order_payment_mutation.test.ts
@@ -47,7 +47,7 @@ describe("Approve Order Mutation", () => {
 
     return runQuery(mutation, rootValue).then(data => {
       expect(data!.setOrderPayment.orderOrError.order).toEqual(
-        sampleOrder(true, false)
+        sampleOrder(true, false, true)
       )
     })
   })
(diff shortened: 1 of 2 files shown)


src/schema/__tests__/ecommerce/set_order_payment_mutation.test.ts
```diff
@@ -47,7 +47,7 @@ describe("Approve Order Mutation", () => {
 
     return runQuery(mutation, rootValue).then(data => {
       expect(data!.setOrderPayment.orderOrError.order).toEqual(
-        sampleOrder(true, false)
+        sampleOrder(true, false, true)
```
ashkan18: this last `true` makes sure `creditCard` data is back in response.

src/schema/ecommerce/set_order_payment_mutation.ts
```diff
@@ -54,26 +54,27 @@ export const SetOrderPaymentMutation = mutationWithClientMutationId({
             __typename
             ... on EcommerceOrderWithMutationSuccess {
               order {
-              id
-                code
-                currencyCode
-                state
+                id
```
ashkan18: this looks more complicated than what it is, i basically ordered fields and only added `creditCardId`.

PR #1326, opened 2018-10-11
Use artwork version when resolving orders

jonallured:
I'm working on this ticket:

https://artsyproduct.atlassian.net/browse/SELL-1031

And had a convo with @ashkan18 on Slack here:

https://artsy.slack.com/archives/CB110C6LE/p1539269193000100

What I'm attempting to do with this PR is have orders resolve their artwork using the particular version of the artwork at the time of the sale. This should ensure that any mutations on the artwork won't cause the order detail page in CMS to blow up.

I have no idea if I have done this correctly, so looking for direction, but this seemed like all I had to do!

### DIFF
--- a/src/schema/ecommerce/types/order_line_item.ts
+++ b/src/schema/ecommerce/types/order_line_item.ts
@@ -9,6 +9,7 @@ import Artwork from "schema/artwork"
 import { OrderFulfillmentConnection } from "./order_fulfillment"
 import { amount } from "schema/fields/money"
 import date from "schema/fields/date"
+import { ArtworkVersion } from "../../artwork_version"
 
 export const OrderLineItemType = new GraphQLObjectType({
   name: "OrderLineItem",
@@ -20,6 +21,19 @@ export const OrderLineItemType = new GraphQLObjectType({
     artwork: {
       type: Artwork.type,
       description: "Artwork that is being ordered",
+      deprecationReason:
+        "Prefer artworkVersion or your data can be out of date",
+      resolve: (
+        { artworkId },
+        _args,
+        _context,
+        { rootValue: { authenticatedArtworkLoader } }
+      ) => authenticatedArtworkLoader(artworkId),
+    },
+    artworkVersion: {
+      type: ArtworkVersion,
+      description: "Artwork version that is being ordered",
+
       resolve: (
         { artworkVersionId },
         _args,
--- a/src/schema/ecommerce/types/order_line_item.ts
+++ b/src/schema/ecommerce/types/order_line_item.ts
@@ -21,11 +21,11 @@ export const OrderLineItemType = new GraphQLObjectType({
       type: Artwork.type,
       description: "Artwork that is being ordered",
       resolve: (
-        { artworkId },
+        { artworkVersionId },
         _args,
         _context,
-        { rootValue: { authenticatedArtworkLoader } }
-      ) => authenticatedArtworkLoader(artworkId),
+        { rootValue: { authenticatedArtworkVersionLoader } }
+      ) => authenticatedArtworkVersionLoader(artworkVersionId),
     },
     editionSetId: {
       type: GraphQLString,
--- a/src/schema/ecommerce/types/order_line_item.ts
+++ b/src/schema/ecommerce/types/order_line_item.ts
@@ -21,8 +21,6 @@ export const OrderLineItemType = new GraphQLObjectType({
     artwork: {
       type: Artwork.type,
       description: "Artwork that is being ordered",
-      deprecationReason:
-        "Prefer artworkVersion or your data can be out of date",
       resolve: (
         { artworkId },
         _args,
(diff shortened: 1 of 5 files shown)


src/schema/ecommerce/types/order_line_item.ts
```diff
@@ -21,11 +21,11 @@ export const OrderLineItemType = new GraphQLObjectType({
       type: Artwork.type,
       description: "Artwork that is being ordered",
       resolve: (
-        { artworkId },
+        { artworkVersionId },
         _args,
         _context,
-        { rootValue: { authenticatedArtworkLoader } }
-      ) => authenticatedArtworkLoader(artworkId),
+        { rootValue: { authenticatedArtworkVersionLoader } }
```
ashkan18: this endpoint returns an `ArtworkVersion` (https://github.com/artsy/gravity/blob/master/app/api/v1/artwork_versions_endpoint.rb#L13) but its being resolved to an `Artwork` in https://github.com/artsy/metaphysics/pull/1326/files#diff-03ea9cde229e19c3db2c2640d865c49aL21, I think we can do two things,
- either have artwork version return actual artwork and resolve this with `artwork_version.artwork`
or
- add a new `ArtworkVersion` type which matches to pretty much https://github.com/artsy/gravity/blob/master/app/models/domain/artwork_version.rb#L2 and just return that

I think having `ArtworkVersion` as a new type makes more sense to me.

orta: Yeah, I wonder if this should return something like:

```graphql
type ArtworkVersion {
  artwork: Artwork
  version: ArtworkVersionMetadata
}

type Query {
  authenticatedArtworkVersionLoader(ID: ID): ArtworkVersion
}
```

And MP composes both the artwork and the version metadata  in 
https://github.com/artsy/gravity/blob/master/app/models/domain/artwork_version.rb#L2

So that you can just ignore the version metadata when you don't need it (which at a guess, is most of the time) ?

orta: That may break existing queries though, so unsure if that's a blocker.

Another option is to make a separate resolver that specifically claims to only return the versioned artwork, instead of an ArtworkVersion from grav?

joeyAghion: Because we're not sure how versions will get used across the platform in the future, I'd prefer a design that doesn't mingle the version object's data into artworks and instead treats it as a separate, first-class thing.

Most uses can continue to reference "artworks" (which implicitly refers to currently-published artwork data), while any uses that want versions would make that explicit, with similar (but not necessarily identical) fields available as on artworks.

orta: Yeah, after chatting with Ashkan - I agree, we should make a new type in MP for this

saolsen: Yeah, it should be it's own type. We can also add the current version, and a list of all versions as fields on the artwork type.

jonallured:
Does anyone understand the build failure here:

https://circleci.com/gh/artsy/metaphysics/3192

It appears that danger failed the build and is pointing to this comment:

https://github.com/artsy/metaphysics/pull/1326#issuecomment-430374573

And I think that comment is trying to say that there are changes to the schema that aren't checked in, but when I run this locally there is nothing to commit. I checked my .env as instructed and see both of those var set to true, so I'm not sure what I'm supposed to do here.

damassi:
The error pointing to `url-join` seems to indicate its a missing ENV var, or an invalid url string somewhere? Might be good to use a debugger and see what's being evaluated. See: https://github.com/GoogleChromeLabs/ndb and https://blog.codeship.com/debug-node-js-effectively-with-chrome-devtools.

jonallured:
Found it! It was the `KAWS_API_BASE` key!! Thanks for taking a look!!

src/schema/ecommerce/types/order_line_item.ts
```diff
@@ -20,13 +21,26 @@ export const OrderLineItemType = new GraphQLObjectType({
     artwork: {
       type: Artwork.type,
       description: "Artwork that is being ordered",
+      deprecationReason:
```
ashkan18: im not super sure about deprecating the `artwork` field since `artwork_id` is still a thing being returned from Exchange schema and there may be some usages on exposing it. I agree we should encourage people using `artworkVersion` though but I think `artwork` may still have its own use.
also think of the case where we want to know for example if the current version is the latest version, we'd have to fetch `artwork` and check latest version. This could use some helper field too `is_latest_version` type of thing but that can be added when needed.

jonallured: Hmm, good point! I'll take off the deprecation. 👍

jonallured:
@ashkan18 good call on the deprecation! Yeah, I'm happy to add tests, but not sure what makes sense here - what were you thinking?

jonallured:
I'm going to merge this so I can move forward on my story but would be happy to address any testing ideas in a follow up if that seems warranted.

PR #1327, opened 2018-10-11
Stitches KAWS

l2succes:
Depends on: https://github.com/artsy/kaws/pull/4

Adds support for the following query
```graphql
{
  marketingCollection(slug: "kaws-companions") {
    id
    title
    query {
      artist_ids
      artist_id
      gene_id
    }
    artworks {
      __id
      hits {
    	id
        title
      }
    }
  }
}
```
screen shot 2018-10-11 at 3 43 40 pm


src/lib/stitching/mergeSchemas.ts
```diff
@@ -48,6 +51,20 @@ export const incrementalMergeSchemas = (testConfig?: any) => {
     }
   }
 
+  if (ENABLE_KAWS_STITCHING) {
```
orta: TBH, you can probably skip this env var and always stitch 👍

l2succes: True, I'm just gonna leave it for now as I don't have KAWS setup on production yet

orta: Cool - your errors are beast the `.env.testing` doesn't have the URLs set

l2succes: Yeah, I noticed, just set the variable via hokusai but I may have to set a default 🤔

l2succes: Nevermind, I see what you're saying... just updated `.env.test`

PR #1328, opened 2018-10-12
Add oneTimeUse to CreateCreditCard mutation

ashkan18:
# Change
We introduced `one_time_use` as a new optional param in https://github.com/artsy/gravity/pull/11953 for creating new credit cards for a user.

This PR adds `oneTimeUse` as a new input param which then translates to `one_time_use` when calling Gravity.

*Question* I used camel case `oneTimeUse` for input param, let me know if that supposed to match the param in Gravity.


_schema.graphql
```diff
@@ -4104,72 +4103,6 @@ type LotStanding {
   sale_artwork: SaleArtwork
 }
 
-# Object representing a collection page
-type MarketingCollection {
```
ashkan18: not sure what happened here!! 🤔

sweir27: look at the danger failure below! :)

ashkan18: i did that and got the same thing

sweir27:
👍  merge on green

PR #1329, opened 2018-10-12
Remove Kaws from being behind an env var flag

orta:
As there's no reimplementation of Kaws in MP, this removes the need for all the trickery with env vars.


ashkan18:
thanks 🙌 merge on green

PR #1330, opened 2018-10-12
Dumping a schema first applies the example env vars, then your own

orta:
This means the defaults are always set for schema dumping.


ashkan18:
oh nice and easy! Thanks!

ashkan18:
merge on green

PR #1331, opened 2018-10-12
Add commissionRate to EcommerceOrder

javamonn:
Exposes `commissionRate` and `displayCommissionRate` on Order as implemented in https://github.com/artsy/exchange/pull/224.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -4718,6 +4718,12 @@ type Order {
     thousand: String = ","
   ): String
 
+  # Partner commission rate used to calculate commission fee
+  commissionRate: Float
+
+  # Partner commission rate formatted into percentage
+  displayCommissionRate: String
+
   # Buyer total in cents
   buyerTotalCents: Int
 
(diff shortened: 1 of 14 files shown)


_schema.graphql
```diff
@@ -2598,23 +2598,6 @@ type dimensions {
   cm: String
 }
 
-# Partner
-type DoNotUseThisPartner {
```
javamonn: From names it seems like these are safe to remove, but it's not clear to me what triggered their removal?

ashkan18: I just ran into exact same issue, you need to rebase with latest `master` and then do a `yarn dump-schema _schema.graphql` that should fix this.

javamonn:
Oof missed your review @ashkan18, I just unassigned you as I was wondering why it wasn't working correctly. I'll update according to those docs.

_schema.graphql
javamonn: I was following that conversation, and I am up to date against the latest master. These aren't Kaws types getting removed as far as I can tell, though I'm unsure why your PR didn't also remove these.

javamonn:
I ran `dump-schema` with `.env.example` and that fixed the diff issues noted above by Danger, I'll debug further what piece of my local env caused this.

ashkan18:
Awesome! thanks! lets do this

PR #1333, opened 2018-10-12
[KAWS] fixes artworks connection in extension schema

l2succes:
- Passes total aggregation into filter_artworks field in metaphysics
- Passes arguments through from artworks to filter_artworks


orta:
👍

PR #1336, opened 2018-10-16
Cap pagination results to 100

yuki24:
We've discussed this in many places (in standup, in slack, and even on Github https://github.com/artsy/reaction/pull/1412#discussion_r225246844), but seems like we just want to cap the results to 100 for the time being. This PR updates the `createPageCursors`  function to check the `totalPages` and replaces it with `100` if it exceeds more than that. This may affect pagination results globally, and I wonder if there should be a way of capping a more specific element. Open to suggestions. Let me know what you all think!

### DIFF
--- a/src/schema/fields/__tests__/pagination.test.js
+++ b/src/schema/fields/__tests__/pagination.test.js
@@ -113,4 +113,14 @@ describe("createPageCursors", () => {
     expect(around[0].isCurrent).toBe(true)
     expect(offsetFromCursor(around[0])).toBe("-1")
   })
+
+  it("caps the page number to 100", () => {
+    const size = 10
+    const totalPages = 101
+    const totalRecords = totalPages * size
+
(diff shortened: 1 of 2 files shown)


src/schema/fields/__tests__/pagination.test.js
```diff
@@ -113,4 +113,14 @@ describe("createPageCursors", () => {
     expect(around[0].isCurrent).toBe(true)
     expect(offsetFromCursor(around[0])).toBe("-1")
   })
+
+  it.only("caps the page number to 100", () => {
```
sweir27: extra `.only`?

yuki24: 😫

yuki24:
Cool! Merge on green

yuki24:
It's odd that Peril is not merging this when it's labeled as merge on green. Maybe it's crying watching the Github Universe keynote.

PR #1337, opened 2018-10-16
#migration schedule MP pods on api tier nodes

izakp:
Schedule MP pods on `tier=api` nodes (our [fixed capacity group](https://github.com/artsy/substance/blob/master/clusters/production/federation/kubernetes-production-orion.artsy.systems/spec.yml#L93) in k8s production) preferentially, [as we do in Gravity](https://github.com/artsy/gravity/blob/master/hokusai/production.yml#L57) 

## Migration

`hokusai production update`

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -72,7 +72,16 @@ spec:
               - key: tier
                 operator: In
                 values:
+                - api
                 - foreground
+          preferredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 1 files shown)


izakp:
This follows up on our production cluster upgrade - now running the latest generation `m5.xlarge` machines - see https://github.com/artsy/substance/pull/109

izakp:
Going to merge

PR #1339, opened 2018-10-17
remove Lewitt api integration.

oxaudo:
This was added to be able to generate acceptsCardPayments on partner. This filed was never released and now is completely removed from Force so cleaning up here.

Not really sure why there are so many changes in `_schema.graphql` file. I did not manually change it. 

The only one I was expecting the change there is removal of `acceptsCardPayments` not sure what the other changes are coming from:(


ashkan18:
Nice! Poor Lewitt-API has been through so many changes... thanks for cleanup, i think you need to do the thing mentioned in danger message ☝️ , the schema file changes include some stuff that shouldn't be there.

starsirius:
👼

alloy:
Oops, lost track of this one. My bad!

PR #1340, opened 2018-10-17
fix null internation shipping for works with free domestic only shipping

oxaudo:
Address https://artsyproduct.atlassian.net/browse/DISCO-395


oxaudo:
Should/could I merge @sweir27?

PR #1341, opened 2018-10-18
set dnsPolicy: ClusterFirst and dnsConfig option ndots: 1

izakp:
Another k8s update - reconfigure Metaphysics to use the in-cluster DNS service.

Update `dnsPolicy` and `dnsConfig` options - this will create an `/etc/resolv.conf` file with the following content, where `100.x.x.x` is the IP of the internal cluster  [kube-dns service](https://kubernetes.artsy.net/#!/service/kube-system/kube-dns?namespace=kube-system):

```
nameserver 100.x.x.x
search default.svc.cluster.local svc.cluster.local cluster.local ec2.internal
options ndots:1
```

With this configuration, all DNS queries will use the in-cluster DNS service, and any domain name with at least one `.` will be considered fully-qualified.

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -63,7 +63,11 @@ spec:
             add:
             - NET_RAW
             - NET_ADMIN
-      dnsPolicy: Default
+      dnsPolicy: ClusterFirst
+      dnsConfig:
+        options:
+        - name: ndots
+          value: "1"
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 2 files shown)


izakp:
Tested on staging and production and no knock-on latency effects or DNS errors observed.

alloy:
For posterity, can you share the benefit of doing this?

izakp:
Will follow up with a PR to test MP talking to Gravity directly via `gravity-web.default.svc.cluster.local` - the caveat is that it needs to talk http rather than https as SSL is terminated at the load balancer, and Gravity is configured to force https, so we will need a way to work around Rails' force_ssl configuration for in-cluster services.

PR #1342, opened 2018-10-18
Update kaws

l2succes:
Related: https://github.com/artsy/kaws/pull/17


_schema.graphql
```diff
@@ -4121,11 +4121,17 @@ type MarketingCollection {
   description: String
 
   # Background image for the header of the collection page
-  headerImage: MarketingImage
+  headerImage: String
```
orta: If this is the ID, it might be worth renaming to `headerImageID`?

PR #1343, opened 2018-10-18
Expand OrderSortMethodEnum to sort over state_expires_at

javamonn:
Exposes additional Order sorting options added in https://github.com/artsy/exchange/pull/229 as part of [SELL-1042](https://artsyproduct.atlassian.net/browse/SELL-1042).


ashkan18:
Nice! lets do this

PR #1345, opened 2018-10-19
Remove unused gravity loaders

ashkan18:
Removing order loader, these are not currently used and also wont be useable after https://github.com/artsy/gravity/pull/11968, we've switched to Exchange's order


sweir27:
Nice! merge on green

PR #1346, opened 2018-10-19
Adds `precariousField` factory to create fields that handle our error pattern

alloy:
This utility eases the definition of the error field pattern as outlined in [my blog post](http://artsy.github.io/blog/2018/10/19/where-art-thou-my-error/#Make.exceptions.first-class.citizens.of.your.schema).

See [the lib implementation file](https://github.com/artsy/metaphysics/blob/4ef49d7d7368894818b2afac724f3b1ef57a08a9/src/lib/precariousField.ts) for the documentation or [the tests](https://github.com/artsy/metaphysics/blob/4ef49d7d7368894818b2afac724f3b1ef57a08a9/src/lib/__tests__/precariousField.test.ts) for some usage examples.

In short, when using this utility, your field implementation can throw expected errors and they will result in `null` for the regular field (e.g. `show`) or an error type that implements the `Error` and `HTTPError` generic interfaces when using the error union field (e.g. `showOrError`).

⚠️ This PR changes the semantics of the existing `show` root field, which would previously throw a `GraphQLError` when a show is considered to be 404. Afaik we don’t use the `show` root field in Emission _yet_ (although @l2succes and @javamonn are starting to use it for local-discovery). So perhaps Force may need some slight different handling if it was checking the `errors` key in the response `{ data: { … } errors: [GraphQLError] }`. If it was already checking the `data.show` key-path for `null` then nothing needs to change.

----

A more concrete example (from this PR):

### show.js

```ts
export const showFields = precariousField({
  name: "show",
  errors: [HTTPErrorType],
  field: {
    type: ShowType,
    description: "A Show",
    args: {
      id: {
      },
    },
    resolve: (root, { id }, request, { rootValue: { showLoader } }) => {
      return showLoader(id).then(show => {
        if (!show.displayable && !show.is_reference) {
          throw new HTTPError("Show Not Found", 404)
        }
        return show
      })
    },
  },
})
```

### schema.js

```ts
import { showFields } from "./show"

const rootFields: any = {
  // ...
  ...showFields,
}
```

#### Nullable field

screen shot 2018-10-25 at 15 13 22

screen shot 2018-10-25 at 15 13 40

#### Error union field

screen shot 2018-10-25 at 15 12 50

screen shot 2018-10-25 at 15 12 40

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -2694,7 +2694,7 @@ type EndSalePayload {
 }
 
 interface Error {
-  message: String
+  message: String!
 }
 
 enum EventStatus {
@@ -4006,13 +4006,13 @@ type HomePageModulesParams {
 }
 
 interface HTTPError {
-  statusCode: Int
-  message: String
+  statusCode: Int!
+  message: String!
 }
 
 type HTTPErrorType implements HTTPError & Error {
-  message: String
-  statusCode: Int
+  message: String!
+  statusCode: Int!
 }
 
 type Image {
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -2693,6 +2693,10 @@ type EndSalePayload {
   clientMutationId: String
 }
 
+interface Error {
+  message: String
+}
+
 enum EventStatus {
   closed @deprecated(reason: "use capital enums")
   current @deprecated(reason: "use capital enums")
@@ -3999,6 +4003,16 @@ type HomePageModulesParams {
   medium: String
   price_range: String
   related_artist_id: ID
+}
+
+interface HTTPError {
+  statusCode: Int
+  message: String
+}
+
+type HTTPErrorType implements HTTPError & Error {
+  message: String
+  statusCode: Int
 }
 
 type Image {
@@ -5931,6 +5945,12 @@ type Query {
     # The slug or ID of the Show
     id: String!
   ): Show
+
+  # A Show
+  showOrError(
+    # The slug or ID of the Show
+    id: String!
+  ): ShowOrError
   status: Status
 
   # Fields related to internal systems.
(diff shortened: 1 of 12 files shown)


ashkan18:
Nice! this is pretty useful, in #1354 we ended up updating existing Union to also support our new approach which may work better for now and allow us to support old-new styles (and maybe deprecate old one), but maybe we can later look at it an update Exchange's mutations.

Also in #1354 ran into an interesting problem, we used to do 
```graphql
mutation {
  ecommerceCreateOrderWithArtwork(
    input: { artworkId: "111", editionSetId: "232", quantity: 1 }
  ) {
    orderOrError {
      ... on OrderWithMutationSuccess {
        order {
          id
          code
        }
      }
        error {
          type
          code
          data
        }
      }
    }
  }
}
```
in ☝️since `order` and `error` weren't in the same level, it was ok to have `code` in both of them with different definitions. 
Following new approach where we do
```graphql
mutation {
  ecommerceCreateOfferOrderWithArtworkMutation(
    input: { artworkId: "111", editionSetId: "232", quantity: 1 }
  ) {
    orderOrError {
      ... on Order {
        id
        code
      }
      ... on EcommerceError {
        type
        code
        data
      }
    }
  }
}
```
since Order's code and Error's code are totally different things, on is `String!` other `String`, we get 
```
Fields "code" conflict because they return conflicting types String and String!. Use different aliases on the fields to fetch both if this was intentional.
```
I can fix that with aliasing one of the codes 
```graphql
... on EcommerceError {
        type
        errorCode: code
        data
      }
```
but this is not ideal since clients now need to handle error codes with `code` or `errorCode`. Is there any solutions there? looks like this conflict checking is part of current spec https://github.com/graphql/graphql-js/issues/1361#issuecomment-393489320

it's not a big deal, but concerning a little since we need to be careful about this conflicts. Might be useful to add to the blogpost too.

alloy:
Ooh, very interesting 🤔 I’ll have a read tomorrow to see if I feel that actually makes sense in the spec and possibly request to restart that discussion.

sweir27:
Going to close this for now since it's been open for a while!

PR #1347, opened 2018-10-19
Update emission querymap JSON

orta:
Adds the root fairs query from our mob workshop


l2succes:
merge on green

l2succes:
@javamonn We'll need to do this for the Show detail view also

javamonn:
@l2succes Does the key of the query change as the query itself changes? It looks like a hash, but not sure if it's a persistent ID of some sort. Would we want to wait until our query stabilizes or does it make sense to go ahead and add it?

PR #1350, opened 2018-10-23
Adds width&height to artwork schema

ashfurrow:
This is required for [DISCO-438](https://artsyproduct.atlassian.net/browse/DISCO-438). Eigen requires dimensions for view-in-room and AR functionality. When we [recently changed](https://github.com/artsy/eigen/pull/2688) the artwork view to consume Metaphysics, we neglected to include the `width` and `height` in the artwork query; indeed, we couldn't have because those fields don't exist on the `Artwork` type in Metaphysics. This PR adds them.

Not sure about the `_schema.graphql` noise. 

/cc @orta


alloy:
This looks good in general, but am wondering if you think it makes sense to move width/height into a ‘dimensions’ object?

Re the noise, you probably don’t have [the stitching env vars](https://github.com/artsy/metaphysics/blob/5656ff0a9bb4df305f938226e8b0e2f69c5ab1e6/.env.example#L65-L66) set?

ashfurrow:
Gotcha, I'll look at the env vars. 

Agreed, and we _have_ a `dimensions` object, but there is _significant_ amounts of code in Eigen that relies on the dimensions as-is (from Gravity). The simplest fix for our bug is to expose those Gravity fields; I know that's not a satisfying answer, but the alternative is to refactor Eigen's VIR and AR code, plus anywhere else it gets used, for this bug fix. Let me know what you think.

ashfurrow:
Env vars were indeed the problem – thanks. I think I had disabled stitching locally a while ago for some reason.

Yeah, I mean, I'm fine with digging into Eigen, but since AR/VIR is totally disabled right now, I got direction from Joanna to prioritize this. My understanding was once we fixed this, we would submit a beta soon after. Can we balance that against the longterm health of the schema? Maybe mark the fields as deprecated and ticket Eigen work to move over to them?

alloy:
Is there no place in Eigen where receive the data and we could hoist them back up to the flat artwork payload? If you truly think that the Eigen changes for that would be too substantial, then I’m ok with immediately deprecating them, but in reality that probably means they’ll stick around for quite a while, so I’d prefer to avoid it entirely.

ashfurrow:
The issue is about data formatting; Eigen refers to width, height, and a unit while metaphysics has dinensions: { in, cm }. I’ll look into the refactor – I agree these would stick around for a long time and that Eigen changes are more desirable.

alloy:
Ah gotcha. Thanks and feel free to merge this if you consider the refactor too large 👍

ashfurrow:
Okay, tried to get this to work without changing the schema but decided to add the deprecated fields instead. `dimensions` returns a string computed by Gravity, like `"14 × 7 1/2 × 7 1/2 in"`, and I don't want to add string-parsing logic to Eigen (especially for edge cases like diameter, etc). Adding structured `width`, `height`, and `metric` makes sense, though I'd prefer they were in an object type. I looked into adding them to the `dimensions` field already used but didn't like how it muddied its function (especially as its type is shared within Metaphysics).

That's all to say, I think merging is the right path forward; I added the deprecation notices as discussed and will follow-up with an Eigen PR.

PR #1354, opened 2018-10-24
createOfferOrderWithArtworkMutation

ashkan18:
# Changes
Updated schema to match changes introduced in https://github.com/artsy/exchange/pull/243

Added new `createOfferOrderWithArtworkMutation` which maps to newly added Exchange mutation. 

This new mutation is slightly different than the existing ecommerce mutation where we start to support new flattened version of `OrderOrError` and removed nested `Failure` and `Successs` responses following http://artsy.github.io/blog/2018/10/19/where-art-thou-my-error/

We ended up, updating existing `OrderOrFailureUnion` to support for new style instead of adding a separate union, the benefit of this is, it may make later migrations to new style little easier.

# Conflict issue
Read the comment 👇 I added about the conflicted names issue I ran into with new style.


ashkan18:
@sweir27 I fixed some tests and added new one for our `ecommerceOffer...` mutation, few things to follow up
- [ ] Not sure why but had to alias the `code` in the `error` section, otherwise it would complain about a conflict in schema!
- [ ] Need to update snapshots for schema, tried to run `yarn jest -u` but that doesn't seem to update snapshots for some reason
- [ ] Seeing 👇 which seems there is something wrong with our resolvetype
```
 Type "OrderOrErrorUnion" is missing a "resolveType" resolver. Pass false into "resolverValidationOptions.requireResolversForResolveType" to disable this warning.

  ● Create Offer Order Mutation › creates offer order and returns it
```

ashkan18:
Alright managed to update the schema from another computer, might have been a weird yarn/jest/node/nvm/monitor/keyboard/mouse/timeofday issue but that part is fixed now and all tests should pass now.

As for the first item ☝️ looks like thats a new possible issue with our new approach, mentioned more details about it in https://github.com/artsy/metaphysics/pull/1346#issuecomment-433228880, but basically that issue is fixed by aliasing the error's code, this may be problematic on the client sides, but let me know what you think and if it's ok to alias Error's `code` with `errorCode: code`.

ashkan18:
@alloy wanted to get your thoughts on this. @sweir27 and I were talking last week and we were thinking maybe knowing the issue ☝️ , it may be easier and more stable to go with the same mutation response as other existing mutations with nested `order` and `error` for now to avoid the conflict issue:
```graphql
orderOrError{
   ... on Success{
      order {
      }
    ... on Failure {
      error {
      }
   }
```
FWIW, the new style, without nested `order` and `error` looks LOT cleaner on both Exchange and MP, so if there is a non-risky way to deal with conflict issue i'd vote for keeping it.

src/schema/__tests__/ecommerce/create_offer_order_with_artwork_mutation.test.ts
```diff
@@ -0,0 +1,50 @@
...
+            }
+          }
+        }
+      }
+    `
+
+    return runQuery(mutation, rootValue).then(data => {
+      expect(
+        data!.ecommerceCreateOfferOrderWithArtworkMutation.orderOrError
+      ).toEqual(sampleOrder(true, false))
```
sweir27: I find the tests that hide the output in a fixture to be a little magical for my taste.

But the real comment-- in this case, it would be good to assert both the "success" case and the "failure" cases. At this layer we could test for known errors from exchange and errors _reaching_ exchange at all (or a generic 500) since they have different behavior.

alloy:
@ashkan18 Ok, so in this case the issue is that the error’s `code` is non-nullable, whereas the order’s `code` field is nullable. I don’t really agree with the GraphQL spec on this and will bring this up, but that doesn’t solve our issue straight away. The only immediate workaround solutions I see are to:

1. make the error‘s `code` field nullable too, which should be backwards compatible, although I don’t know if it‘s backwards compatible to make a field nullable again in the future (as long as the iOS app doesn’t query for this that can all be fixed in an orchestrated client update, though)
2. rename the order’s `code` field to something like `trackingCode`, which also requires updates to the client code already using this
3. go with the union style used previously

I would personally prefer 2, because `code` seems a bit ambiguous anyways if it’s about the delivery tracking-code, but I get that it will be more work and I don’t have insight into wether or not there is the time to do that, so I’ll leave that up to y’all to decide.

ashkan18:
I can go eitherway, I personally like the simplicity of current approach and it makes more sense so if i have to pick i'd go with renaming `code` but in that case maybe it's easier to alias the error one since that'll be less impact on other places.  I don't know the impact of this possible alias on the front end though, does that complicate things? Basically we are saying, for this new style mutation, we want to alias `error.code` to `error.internalCode` or something @sweir27 is that an scary change?

sweir27:
Since this is still in flux a little bit (and potentially depends on an update to graphQL?), I think we should just keep the current way of doing things. Would rather follow up (which we'll have to do anyways for the other mutations) with a solution we're confident in.

sweir27:
@ashkan18 it's not scary, it's just confusing and hard to anticipate. It sounds like all of this can be a follow-up, but I'd rather not merge a solution we're not happy about as there's no rush on determining the ideal pattern (but there's a slight rush on getting the mutation out). 😄

sweir27:
Sorry about the odd git history-- in order to fix something I attempted a rebase.

src/schema/schema.ts
```diff
@@ -175,6 +176,7 @@ if (!ENABLE_ECOMMERCE_STITCHING) {
   stitchedMutations.submitOrder = SubmitOrderMutation
 
   stitchedMutations.ecommerceCreateOrderWithArtwork = CreateOrderWithArtworkMutation
+  stitchedMutations.ecommerceCreateOfferOrderWithArtwork = CreateOfferOrderWithArtworkMutation
```
sweir27: I noticed that this was previously named `ecommerceCreateOfferOrderWithArtworkMutation`. Looks like we're naming the others without the `Mutation` suffix which seems fine (and also shortens it slightly).

src/schema/__tests__/ecommerce/create_offer_order_with_artwork_mutation.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    rootValue = mockxchange(resolvers)
+
+    return runQuery(mutation, rootValue).then(data => {
+      expect(
+        data!.ecommerceCreateOfferOrderWithArtwork.orderOrError.order
+      ).toEqual(sampleOrder(true, false))
+    })
+  })
+
+  it("returns an error if there is one", () => {
```
ashkan18: 🙌

PR #1357, opened 2018-10-26
adjust horizontal pod autoscaler parameters for metaphysics-web 

izakp:
Seems like 70% target CPU utilization gets Metaphysics to run around 10-12 replicas during normal morning hours traffic, (9-12am EST today, 26th October) - let's see if taking down the floor of min replicas can scale back for overnight periods - 12 replicas probably too generous for all traffic hous

@dleve123 for some context - Metaphysics does a lot of waiting on downstream I/O and memory consumption tends to sawtooth , so we've kept it over-provisioned and set with a fixed memory limit, so it's getting OOMKilled quite frequently but keeps it HA

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -98,9 +98,9 @@ spec:
     apiVersion: extensions/v1beta1
     kind: Deployment
     name: metaphysics-web
-  minReplicas: 12
+  minReplicas: 7
   maxReplicas: 15
-  targetCPUUtilizationPercentage: 80
+  targetCPUUtilizationPercentage: 70
 
 ---
 apiVersion: v1


dleve123:
Joey - when you say applied to prod already, is there a separate process for deploying to prod? More than just kicking off a circle CI build?

So do we think there is a mem leak somewhere, leading to the OOM kills?

How do you know how many instances are running? Do you generally just poke around in the kubernetes dash, or is there some canonical chart that you look at?

joeyAghion:
I sneaked a peak at the metaphysics-web deployment in the dash [here](https://kubernetes.artsy.net/#!/deployment/default/metaphysics-web?namespace=default) and noted that it already reflected the min of `7` updated in this PR.

Updates to these *.yml configs are generally _not_ applied automatically as part of CI or deploys. This is usually a good thing, allowing the developer to make updates and evaluate results before committing the change. But it leads to some ambiguity about the state.

izakp:
Yeah it's a 🐔 and 🥚 problem sometimes when finding the right runtime parameters on production... generally I try and tune parameters like these via the dashboard / kubectl then commit what works.

dleve123:
> Updates to these *.yml configs are generally not applied automatically as part of CI or deploys. This is usually a good thing, allowing the developer to make updates and evaluate results before committing the change. But it leads to some ambiguity about the state.

How are these changes deployed then?

dleve123:
Makes sense, thanks.

PR #1356, opened 2018-10-26
Order's Mode

ashkan18:
# Change
in https://github.com/artsy/exchange/pull/243, we added `mode` to `Order` which can be `BUY` or `OFFER`.

This PR follows up with that change to be able to:
- Add `mode` as new a field in `Order`
- Add option for being able to filter by `mode`

https://artsyproduct.atlassian.net/browse/PURCHASE-611

# Sample Query
You can now do 
```graphql
query {
	ecommerceOrders(sellerId: "581b45e4cd530e658b000124", mode: BUY){
		totalCount
		edges{
			node{
				id
				mode
```

PR #1358, opened 2018-10-26
Cleanup OrderFields

ashkan18:
# Problem
We keep repeating list of order fields passed to Exchange in different queries/mutations. This makes adding a new field more involved Ex. #1356 

# Change
This PR consolidates fields in a very basic way (can be improved more later) by usage in `query_helper` and use those variables in different places instead of repeating them.

# Depends on
#1356 
This branch was built on top of #1356 , we need to first merge that and then this

# Possible follow up
Consolidate even more and have possibly a method that returns list based on input params, like 
```
orderFields(include_credit: true, include_fulfillments: false)
```


src/schema/ecommerce/query_helpers.ts
```diff
@@ -48,3 +48,96 @@ export const BuyerSellerFields = gql`
...
+        artworkId
+        editionSetId
+        quantity
+      }
+    }
+  }
+`
+
+export const SellerOrderFields = gql`
+  ${BuyerSellerFields}
```
sweir27: #minor but  given the pattern (`BuyerOrderFields`, `SellerOrderFields`), `BuyerSellerFields` is a little confusing. Maybe we can call it `ParticipantFields`?

ashkan18:
Addressed the comment, going to merge to get it ready for demo.

PR #1362, opened 2018-10-29
Add a root field for Local Discovery

anandaroop:
This PR creates an entry point for querying Local Discovery data. 

The goals here are:

- Create a root query that takes a Gravity-like city slug as its arg
- Resolve that slug to a known LD city
- Resolve local shows and fairs for that LD city
- Provide a choke point for LD shows, so that we can have a clear place to merge in non-partner data once that strategy is settled (currently this query only returns our regular partner-contributed show data)
- Learn some GraphQL finally 😝 

This is an example of how the querying is looking at the moment:

city

I've only tinkered around the edges of existing MP code before, and this is my first time adding something completely new — so I'm all ears about approach and best practices.


src/schema/local_discovery/city.ts
```diff
@@ -0,0 +1,127 @@
...
+      resolve: obj => {
+        return obj.lat
+      },
+    },
+    lng: {
+      type: GraphQLFloat,
+      resolve: obj => {
+        return obj.lng
+      },
+    },
```
alloy: Once you have such an object, its fields will probably never be null?

alloy: graphql-js has a default field resolver that will try to use the name of the field on the data being passed to its parent, so these don’t need a resolver at all.

src/schema/local_discovery/city.ts
```diff
@@ -0,0 +1,127 @@
...
+})
+
+const LocalDiscoveryCityType = new GraphQLObjectType({
+  name: "LocalDiscoveryCity",
+  fields: {
+    slug: {
+      type: GraphQLString,
+      resolve: obj => {
+        return obj.slug
+      },
```
alloy: Ditto, no resolver needed.

src/schema/local_discovery/city.ts
```diff
@@ -0,0 +1,127 @@
...
+    },
+    name: {
+      type: GraphQLString,
+      resolve: obj => {
+        return obj.name
+      },
```
alloy: Etcétera

src/schema/local_discovery/city.ts
```diff
@@ -0,0 +1,127 @@
...
+    },
+    coordinates: {
+      type: LatLngType,
+      resolve: obj => {
+        return obj.coordinates
+      },
```
alloy: Tiny nit: prefer single line without curly braces and implicit return.

src/schema/local_discovery/city.ts
```diff
@@ -0,0 +1,127 @@
...
+    },
+    shows: {
+      type: new GraphQLList(Show.type),
+      args: {
+        size: { type: GraphQLInt },
+        sort: PartnerShowSorts,
+        status: EventStatus,
+      },
+      resolve: (obj, args, _context, { rootValue: { showsLoader } }) => {
+        let gravityOptions = {
```
alloy: Prefer `const`

src/schema/local_discovery/city.ts
```diff
@@ -0,0 +1,127 @@
...
+          displayable: true,
+          near: `${obj.coordinates.lat},${obj.coordinates.lng}`,
+          max_distance: LOCAL_DISCOVERY_RADIUS_KM,
+        }
+
+        // here is where we could fetch non-partner shows and
+        // merge them in with partner shows.
+        //
+        // but how to deal with sorting across two disparate
+        // collections?
```
alloy: I know you know that we discussed this on the notion doc, but leaving here for posterity that we shouldn’t be sorting large paginated lists on MP, as it would require loading all data.

src/schema/local_discovery/city.ts
```diff
@@ -0,0 +1,127 @@
...
+  type: LocalDiscoveryCityType,
+  description: "A city-based entry point for local discovery",
+  args: {
+    slug: {
+      type: new GraphQLNonNull(GraphQLString),
+      description:
+        "A slug for the city, conforming to Gravity's city slug naming conventions",
+    },
+  },
+  resolve: (_obj, args, _context, _info) => {
```
alloy: No need to specify trailing parameters that you won’t use

src/schema/schema.ts
```diff
@@ -108,6 +109,7 @@ const rootFields: any = {
   gene_families: GeneFamilies,
   gene_family: GeneFamily,
   home_page: HomePage,
+  local_discovery_city: LocalDiscoveryCity,
```
alloy: We’ve been using camelCase for new fields

alloy: Not a big fan of this very specific name. Can this just be “city”?

src/schema/local_discovery/city.ts
anandaroop: Is that level of magic allowed outside of Rails :trollface: ?

src/schema/local_discovery/city.ts
anandaroop: Heh, I was using those verbose signatures while getting my head around how resolvers work. I'll take off the training wheels now.

src/schema/schema.ts
anandaroop: Do we also prefer camel for all filenames now? I was wondering about `city_data.json`

anandaroop: Glad you brought this up — I went with this naming for a couple of reasons:

- It seemed to me that the notion of _city_ being used here was not necessarily a general one, because of our need to handle non-Artsy shows in the local discovery use case

- I anticipated we might have a need for other local discovery-specific concepts, as we flesh out stubbed shows, follows etc

But if that's all premature I'm happy to go with the simpler naming.

src/schema/local_discovery/city.ts
alloy: Magic is the best

_schema.graphql
```diff
@@ -4084,6 +4092,11 @@ union Item = ArtistItem | ArtworkItem | FeaturedLinkItem | GeneItem
 # The `JSON` scalar type represents JSON values as specified by [ECMA-404](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf).
 scalar JSON
 
+type LatLng {
+  lat: Float
+  lng: Float
+}
```
orta: This already exists as "coordinates" (who made a lowercase type?!)

https://github.com/artsy/metaphysics/blob/5659ff0807cca37705520b47124d51f62c8de873/_schema.graphql#L2285-L2288

src/schema/local_discovery/city.ts
l2succes: Agreed, that should most likely happen in Gravity if that's where we end up sorting Non Partner models (sql join?). It would be way faster too

_schema.graphql
```diff
@@ -1966,6 +1966,14 @@ enum CategoryType {
   INSTITUTION
 }
 
+type City {
+  slug: String
```
l2succes: we'll probably want to add more metadata here as the requirements solidify.

e.g.
`
type City {
  image: Image
}
`
also should BMW shows go in their own field? We can answer these questions later though

src/schema/schema.ts
alloy: Good question. I think it could, but don’t recall some sort of standardization.

alloy: Yeah I get that the _current_ data that’s available is all related to local discovery, but it doesn’t always need to be in the future. Cities can be connected to other parts of our data in many ways.

Generally, viewing a root field as a special node in the graph is limiting, eg consider a new feature on an artwork page that lists shows in the city that the artist of the artwork hails from.

Does that make sense?

l2succes: We touched on this briefly during the data discussion we had but unless there's a good reason not to (backwards compatibility with existing consumers) I think it's safe to augment our existing models (City, Fair, Show, etc...), we're adding this data for the purpose of Local Discovery but it can and may be used for other features as well.

We should definitely double check for existing types but since `City` is new. We might as well.

anandaroop: Yep, I'm good with `City`, and take the point about keeping root fields general. 

As for augmenting vs. new models — I wasn't sure what we would do in the case where we need to resolve shows e.g. to some place other than Gravity (since that data pipeline question is still open). 
Still getting the hang of this kind of orchestration, but all of the above sounds good to me for now 👍

_schema.graphql
anandaroop: Definitely — the actual nature of the BMW data is still unclear to me, but yeah a new field might make sense.

_schema.graphql
anandaroop: Good catch 😄. Looks like that was added 3 years ago(!) and is only used in one place.

Question: does changing the name of this type (even if only for case) qualify as a deprecation, in graphql terms?

src/schema/schema.ts
alloy: The `City` type can specify the `shows` field config, including resolver, so that’s where it would instantiate a list of `Show` entries based on sourced from elsewhere.

_schema.graphql
alloy: Hmm, I think it’s never safe, actually. There’s many places in our client codebases where we have fragments (which are always on types), so if any of our clients has a fragment on `coordinates` then I’m pretty sure this will break (unverified, though!)

orta: 10/10 @alloy  - great point!

I did a [quick look](https://github.com/search?q=org%3Aartsy+%22on+coordinate%22&type=Code) and we don't have it anywhere, so it is good to transform, but realistically we can't do it very often

src/schema/__tests__/city.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    coordinates: { lat: 38.5, lng: -121.8 },
+  },
+}
+
+beforeEach(() => {
+  jest.mock("../city/city_data.json", () => mockCities)
+})
+
+describe("City", () => {
+  it("finds a city by its slug", () => {
```
l2succes: 👍

_schema.graphql
orta: 👍

src/schema/__tests__/city.test.ts
```diff
@@ -0,0 +1,113 @@
...
+        }
+      }
+    `
+
+    const mockShows = [{ id: "first-show" }]
+    const mockShowsLoader = jest.fn(() => Promise.resolve(mockShows))
+    const rootValue = {
+      showsLoader: mockShowsLoader,
+      accessToken: null,
+      userID: null,
```
anandaroop: I found that I had to add L62-63 to make `tsc` happy, even though I don't think they are necessary for these tests to work.

src/schema/local_discovery/city.ts
anandaroop: Yeah, looks like that's where we're ending up anyway, after this morning's platform discussion. Hopefully that smooths the way forward, at least from MP's perspective.

src/schema/__tests__/city.test.ts
alloy: Those properties should be made optional in the `runQuery` interface.

alloy:
Nice job @anandaroop! 👏

(It might also be good to review the GraphQL schema design guide in the README repo man to see if the wording around eg keeping types generic can be improved or other learnings can be added.)

l2succes:
Nice one @anandaroop! I'm gonna go ahead and merge this one, we can follow with the pagination/connections work we discussed after. 

Let's make a ticket in JIRA to keep track of it though

anandaroop:
Awesome, adding that ticket now 👍

PR #1363, opened 2018-10-29
add default to aspect ratio for an image

oxaudo:
Also cleaning up some unused variables to make lint happy.

This PR is a follow up on discussion here: https://github.com/artsy/reaction/pull/1467

### DIFF
--- a/src/schema/image/__tests__/index.test.js
+++ b/src/schema/image/__tests__/index.test.js
@@ -50,6 +50,30 @@ describe("Image type", () => {
     }
   })
 
+  describe("#aspect_ratio", () => {
+    const query = `{
+      artwork(id: "richard-prince-untitled-portrait") {
+        image {
+          aspect_ratio
+        }
+      }
+    }`
+
+    it("returns original aspect_ratio when available", () => {
+      assign(image, { aspect_ratio: 1.5 })
+      return runQuery(query, rootValue).then(data => {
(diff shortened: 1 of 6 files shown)


src/schema/image/__tests__/index.test.js
```diff
@@ -50,6 +50,30 @@ describe("Image type", () => {
...
+    }`
+
+    it("returns original aspect_ratio when available", () => {
+      assign(image, { aspect_ratio: 1.5 })
+      return runQuery(query, rootValue).then(data => {
+        expect(data.artwork.image.aspect_ratio).toBe(1.5)
+      })
+    })
+
+    it("defaults to 1 when origin alsect ratio is not availablwe", () => {
```
alloy:
```suggestion
    it("defaults to 1 when origin aspect ratio is not available", () => {
```

oxaudo: 🙉

alloy: This new GH feature is the best

src/schema/image/index.js
```diff
@@ -25,6 +25,9 @@ const ImageType = new GraphQLObjectType({
   fields: () => ({
     aspect_ratio: {
       type: GraphQLFloat,
```
orta:
```suggestion
      type: new GraphQLNonNull(GraphQLFloat),
```

As it will never return null now - but you'll need to import it at the top though

PR #1365, opened 2018-11-01
[WIP] Add InitialOffer mutation from exchange

ds300:
I marked this as WIP because I haven't figured out certain stuff to do with testing, like what's the relationship between `sampleOrder` and `order.json`? What are we aiming to actually test about mutations and the data they consume/return?


src/schema/__tests__/ecommerce/fulfill_order_at_once_mutation.test.ts
```diff
@@ -53,7 +53,7 @@ describe("Fulfill Order at Once Mutation", () => {
 
     return runQuery(mutation, rootValue).then(data => {
       expect(data!.fulfillOrderAtOnce.orderOrError.order).toEqual(
-        sampleOrder(true, true)
+        sampleOrder({ fulfillments: true })
```
ashkan18: nice!! this is what happens when a JS expert looks at my code 🙌

src/schema/ecommerce/types/initial_offer_input_type.ts
```diff
@@ -0,0 +1,19 @@
...
+  GraphQLNonNull,
+  GraphQLInt,
+  GraphQLID,
+} from "graphql"
+export const InitialOfferInputType = new GraphQLInputObjectType({
+  name: "InitialOfferOrderInput",
+  fields: {
+    orderId: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: "BSON ID of artwork",
```
ashkan18: this should be `ID of the Order`

src/schema/ecommerce/types/offer.ts
```diff
@@ -0,0 +1,46 @@
...
+  name: "Offer",
+  fields: () => ({
+    id: {
+      type: GraphQLID,
+      description: "ID of the offer",
+    },
+    createdAt: date,
+    creatorId: {
+      type: GraphQLString,
+      description: "Id of the user who created the order",
```
ashkan18: i think we can resolve this to a `user`

ds300: Do we have to be careful about auth here? I assume e.g. we don't always want partners to get the buyers' emails.

ashkan18: good question, i don't think we need to be worried in this specific one since this info user data is also in `Order.from`. The part that i think we should be more careful is to not expose partner's contacts to buyers. In the end we may want to make this field a `seller_only` field in Exchange.

ds300: Ok we can iterate on that :+1:

PR #1366, opened 2018-11-01
Add nearby shows field

ashleyjelks:
Adding field needing for Local Discovery Shows view


src/schema/show.js
```diff
@@ -407,6 +432,22 @@ export const ShowType = new GraphQLObjectType({
       type: GraphQLString,
       resolve: ({ fair }) => (isExisty(fair) ? "Fair Booth" : "Show"),
     },
+    nearby_shows: {
```
l2succes: @ashleyjelks We should add some tests for that query

orta:
This is green now, and I'm gonna make a separate PR for the shows -> nearby_shows connectio

l2succes:
Ok awesome, merging then

PR #1370, opened 2018-11-06
Fix missing rename

l2succes:
Not sure why this went green on CI 🤔

- One rename from this PR
- https://github.com/artsy/metaphysics/pull/1368


orta:
Weird, thanks man

PR #1371, opened 2018-11-07
Replace amountCent with offerAmount with amount and currencyCode

ashkan18:
# Problem
Currently Force's UI ask users to input offer amount field in dollars but MP and Exchange expect to receive the amount in cents. Because of this, right now setting offer to $2500 ends up being $25.00 on the Exchange side.

https://artsyproduct.atlassian.net/browse/PURCHASE-639

# Solution
There were some [discussions](https://artsy.slack.com/archives/C1HH3KNJG/p1541532150074000) about where is the best place to do the conversion.
This PR adds it to MP by
- defining `MoneyInput` field which has required `amount` and `currencyCode`
- remove `amountCents` from `initialOffer` and replace it with `offerAmount` with type `MoneyInput`
- added `moneyHelper.ts` which for now only have `moneyFieldToUnit` method which only supports `USD` to cents 

# Followup
Update reaction with this change since its the only current user.

# Lessons Learned
Don't use emoji for Github branch names unless you are really good with emoji keyboard thingy.

### DIFF
--- a/src/lib/moneyHelper.ts
+++ b/src/lib/moneyHelper.ts
@@ -4,10 +4,7 @@ interface MoneyField {
 }
 
 export const moneyFieldToUnit = (moneyField: MoneyField) => {
-  switch (moneyField.currencyCode) {
-    case "USD":
-      return moneyField.amount * 100
-    default:
-      throw new Error("Unknown currency, cannot process.")
-  }
+  // this currently only supports currencies with cents
+  // and multiply the major value to 100 to get cent value
+  return moneyField.amount * 100
 }


_schema.graphql
```diff
@@ -4097,8 +4097,8 @@ input InitialOfferInput {
   # ID of order
   orderId: ID!
 
-  # Offer amount in cents
-  amountCents: Int!
+  # Offer amount
+  offerAmount: MoneyInput
```
joeyAghion: I fear that we're recreating this potential bug in a different form. We call it `cents` to make that data expectation explicit. We should add `dollars` here to avoid the same ambiguity. (I'm open to other ways of referring to the units and subunits, like "major" and "minor," but "dollars" is the natural one compared to "cents.")

joeyAghion: Unless... `MoneyInput` already embeds that information?

src/schema/ecommerce/types/initial_offer_input_type.ts
```diff
@@ -1,19 +1,16 @@
...
     orderId: {
       type: new GraphQLNonNull(GraphQLID),
       description: "ID of order",
     },
-    amountCents: {
-      type: new GraphQLNonNull(GraphQLInt),
-      description: "Offer amount in cents",
+    offerAmount: {
+      type: MoneyInput,
+      description: "Offer amount",
```
sweir27: Maybe to address @joeyAghion 's point above, we talked about making the description here explicit.

_schema.graphql
```diff
@@ -4573,6 +4573,14 @@ type Metaphysics {
   environment: String!
 }
 
+input MoneyInput {
+  # amount
+  amount: Int!
```
joeyAghion: I don't truly understand what's going on here (with the various references to `Int!`, `number`, and `GraphQLInt`), but would expect this to be a float.

ashkan18: oh this is the autogenerated file, but good point it should be float! fixing it in the input definition

src/schema/ecommerce/types/initial_offer_input_type.ts
ashkan18: yes makes sense to add it here! although im really having hard time to find out how to make it generically explicit, `Offer amount in dollars`? `offer amount in major`?

ashkan18: updated description here and in `MoneyField`

src/lib/moneyHelper.ts
```diff
@@ -0,0 +1,13 @@
...
+  amount: number
+  currencyCode: String
+}
+
+export const moneyFieldToUnit = (moneyField: MoneyField) => {
+  switch (moneyField.currencyCode) {
+    case "USD":
+      return moneyField.amount * 100
+    default:
+      return moneyField.amount
```
alloy: This should imo throw an error, the developer should realise ASAP that the thing they’re trying to do is invalid.

_schema.graphql
alloy: I still think this is called an “offer price”. An example can be found [here](https://support.saatchiart.com/hc/en-us/articles/214385258-How-do-I-submit-an-offer-): “Enter your offer price.”

That gets rid of the awkward double use of the word ‘amount’ and I also think that, because this mutation is about making an offer, you can just call the field `price`.

src/schema/fields/money.js
```diff
@@ -74,4 +76,19 @@ const money = ({ name, resolve }) => ({
...
+  name: "MoneyInput",
+  fields: {
+    amount: {
+      type: new GraphQLNonNull(GraphQLFloat),
+      description: "amount in dollars or major unit",
+    },
+    currencyCode: {
+      type: new GraphQLNonNull(GraphQLString),
+      description:
+        "ISO 4217 Currency code of the this amount, https://en.wikipedia.org/wiki/ISO_4217",
```
alloy:
```suggestion
        "The ISO-4217 alphabetic currency code, as per https://en.wikipedia.org/wiki/ISO_4217",
```

src/schema/ecommerce/types/initial_offer_input_type.ts
```diff
@@ -1,19 +1,16 @@
...
     orderId: {
       type: new GraphQLNonNull(GraphQLID),
       description: "ID of order",
     },
-    amountCents: {
-      type: new GraphQLNonNull(GraphQLInt),
-      description: "Offer amount in cents",
+    offerAmount: {
+      type: MoneyInput,
+      description: "Offer amount in dollars or major unit",
```
alloy: I think this can just be:

```suggestion
      description: "Offer price",
```

When the developer reads the docs for the `MoneyInput` type they’ll learn about the other details.

src/schema/ecommerce/types/initial_offer_input_type.ts
```diff
@@ -1,19 +1,16 @@
...
   name: "InitialOfferOrderInput",
   fields: {
     orderId: {
       type: new GraphQLNonNull(GraphQLID),
       description: "ID of order",
     },
-    amountCents: {
-      type: new GraphQLNonNull(GraphQLInt),
-      description: "Offer amount in cents",
+    offerAmount: {
```
alloy: I still think this is called an “offer price”. An example can be found [here](https://support.saatchiart.com/hc/en-us/articles/214385258-How-do-I-submit-an-offer-): “Enter your offer price.”

That gets rid of the awkward double use of the word ‘amount’ and I also think that, because this mutation is about making an offer, you can just call the field `price`.

src/lib/moneyHelper.ts
ashkan18: 👍 whats the best way to through error?

alloy: In this situation you should just throw one, there’s no need to handle this gracefully. `throw new Error(“...”)`

src/lib/moneyHelper.ts
```diff
@@ -0,0 +1,13 @@
+interface MoneyField {
+  amount: number
+  currencyCode: String
+}
+
+export const moneyFieldToUnit = (moneyField: MoneyField) => {
+  switch (moneyField.currencyCode) {
```
sweir27: Doesn't have to be a part of this PR, but I worry that this is too restrictive and the source of the constraint is unclear.

You may just want to allow all currencies put a comment describing that we don't currently support currencies without cents conversions.

ashkan18: alright, going to make this change, going to add comment saying we only support currencies with cent and once we start support currencies without cents we need to add that exception here.

PR #1373, opened 2018-11-08
@xtina-starr => Update Kaws schema

eessex:
Pair session with @l2succes 

- Updates schema to include `categories` from Kaws as `MarketingCollectionCategories`
- Updates name remapping when using Kaws queries in Metaphysics 
- Fixes tests to look for Kaws variables instead of Metaphysics

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -2589,8 +2589,6 @@ type CurrentEvent {
   href: String
 }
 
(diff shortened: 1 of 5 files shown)


l2succes:
I'll go ahead and merge that one as it's blocking some other PRs

PR #1374, opened 2018-11-09
Line item changes

ashkan18:
# Depends On
https://github.com/artsy/exchange/pull/273

# Change
- Deprecate `offerTotal` fields from `Order`
- Add `totalListPrice` fields to `Order`
- Deprecated `priceCents` from `LineItem`
- Added `listPriceCents` to `LineItem`

This should make all these deploys safe as long as Exchange gets deployed first because of accessing `listPriceCents` in MP

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -4872,6 +4872,20 @@ type Order {
     thousand: String = ","
   ): String
 
+  # Total list price in cents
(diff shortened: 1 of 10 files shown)


ashkan18:
I verified Force working with updated Exchange with existing MP, now we can merge this and try Force with updated MP

PR #1375, opened 2018-11-09
SELL-1150: Update exchange.graphql to reflect artsy/exchange#271

dleve123:
Jira: https://artsyproduct.atlassian.net/browse/SELL-1150

Adds `sellerAcceptOfferMutation` to MP, so that it can be consumed by force.

WARNING: If this deployed to production before, Exchange https://github.com/artsy/exchange/commit/b03a4a6f51cb88ca5d9637f5012586b8a0306af0 is deployed to production, then MP might break as MP-prod will have a newer version of Exchange's schema than Exchange-prod.

----

N.B: local MP is on my dev branch, integrating with Gravity staging.

screen shot 2018-11-09 at 6 21 47 pm

### DIFF
--- a/src/schema/ecommerce/seller_accept_offer_mutation.ts
+++ b/src/schema/ecommerce/seller_accept_offer_mutation.ts
@@ -33,17 +33,6 @@ export const SellerAcceptOfferMutation = mutationWithClientMutationId({
             ... on EcommerceOrderWithMutationSuccess {
               order {
                 ${SellerOrderFields}
-                lineItems {
-                  edges {
-                    node {
-                      id
-                      priceCents
-                      artworkId
-                      editionSetId
-                      quantity
-                    }
-                  }
-                }
               }
             }
             ... on EcommerceOrderWithMutationFailure {
--- a/src/schema/ecommerce/seller_accept_offer_mutation.ts
+++ b/src/schema/ecommerce/seller_accept_offer_mutation.ts
@@ -1,0 +1,64 @@
+import { graphql } from "graphql"
+import { mutationWithClientMutationId } from "graphql-relay"
+import { OfferMutationInputType } from "schema/ecommerce/types/offer_mutation_input_type"
+import { OrderOrFailureUnionType } from "./types/order_or_error_union"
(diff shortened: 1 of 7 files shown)


dleve123:
@ashkan18 This is ready for a review btw!

src/schema/ecommerce/seller_accept_offer_mutation.ts
```diff
@@ -0,0 +1,64 @@
...
+          offerId: $offerId,
+        }) {
+          orderOrError {
+            __typename
+            ... on EcommerceOrderWithMutationSuccess {
+              order {
+                ${SellerOrderFields}
+                lineItems{
+                  edges{
+                    node{
```
starsirius: #minor. Missing whitespace before `{` above. We could also move the `lineItems` to `SellerOrderField`, like [`BuyerOrderField`](https://github.com/artsy/metaphysics/blob/b2450f74852f0b4cc7369d06e0ca532974630330/src/schema/ecommerce/query_helpers.ts#L108-L118).

sepans:
We were thinking if it makes more sense for the client (volt here) to run the mutation by `order_id` or `offer_id`? `InitialOfferMutation` accepts `order_id` so do we need to make them consistent or keep it as is since they are being called from the different clients with different contexts?

ashkan18:
I think for `sellerAcceptOfferMutation` makes sense to get `offerId` so we know which offer is actually being accepted, thinking of the case of old open tab in the browser or something were someone tries to accept an already out of date offer.
`initialOffer` gets `orderId` since its adding the first offer to the order. we are actually renaming that one to `addOfferToPendingOrder` as part of https://github.com/artsy/exchange/pull/283 or another PR

ashkan18:
merge on green

dleve123:
@ashkan18 I addressed [@starsirius's comment moving `line_items` to `SellerOrderFields`](https://github.com/artsy/metaphysics/pull/1375#discussion_r233319016) in https://github.com/artsy/metaphysics/pull/1375/commits/9b8589b5d57708d68d4328d1b336dc25192a4a00.

Please check it out, and rebase and/or merge in.

PR #1376, opened 2018-11-09
Update schema with relaxed itemsTotalCents

ashkan18:
# Problem
Making offers is broken on Staging now since MP requires `itemsTotalCents` to be not null based on it's local copy of Exchange Schema.

# Solution
This requirement was update in Exchange, update MP to catch up.


orta:
Yep, looks good

ashkan18:
🙌

PR #1377, opened 2018-11-09
Change sale_artwork.position from int to float

anandaroop:
Gravity consider's a SaleArtwork's `position` to be a float.  Metaphysics had it as an integer.

Auction ops relies on it being a float, so we update Metaphysics here. Downstream clients may need to be updated as well.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -6481,7 +6481,7 @@ type SaleArtwork {
   minimum_next_bid_cents: Float @deprecated(reason: "Favor `minimum_next_bid`")
   opening_bid: SaleArtworkOpeningBid
   opening_bid_cents: Float @deprecated(reason: "Favor `opening_bid`")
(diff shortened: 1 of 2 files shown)


alloy:
Is this a backwards compatible change?

orta:
I think so, any int can be represented by a float

alloy:
In JS, yeah, but maybe GraphQL doesn’t like it because for other languages it could break codegen?

orta:
You think the relay compiler might balk? ( I think our iOS app is fine because mantle should handle the coercion at a loss) and I'm also shipping an update to that

alloy:
To elaborate, does a GraphQL schema checking client, like Relay, i.e. our clients, accept this change?

alloy:
Cooooool, thanks for checking 👍

PR #1379, opened 2018-11-12
Add submitOrderWithOffer, depends on Exchange

ashkan18:
# Change
Adding `submitOrderWithOffer` mutation which gets `offerId` as input and submits the order with this offer.

# Depends On
https://github.com/artsy/exchange/pull/281 This PR should not be merged till Exchange PR is deployed to production.


sweir27:
Seems 👍 !

PR #1384, opened 2018-11-20
Fix exchange link user-agent forwarding

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-644

Small bugfix to properly forward user agent strings from client to exchange.

image

image

### DIFF
--- a/src/lib/stitching/exchange/link.ts
+++ b/src/lib/stitching/exchange/link.ts
@@ -35,21 +35,19 @@ export const createExchangeLink = () => {
 
   const analyticsMiddleware = setContext((_request, context) => {
     const locals = context.graphqlContext && context.graphqlContext.res.locals
-    if (!locals) return context.graphqlContext
-    return {
-      ...context.graphqlContext,
-      headers: {
-        ...context.graphqlContext.headers,
-        "User-Agent": locals.userAgent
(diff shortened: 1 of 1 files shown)


src/lib/stitching/exchange/link.ts
```diff
@@ -35,21 +35,19 @@ export const createExchangeLink = () => {
...
-          : "Metaphysics",
-      },
+    if (!locals) return context
+    const headers = {
+      ...context.headers,
+      "User-Agent": locals.userAgent
+        ? locals.userAgent + "; Metaphysics"
+        : "Metaphysics",
     }
+    return { headers }
```
alloy: Should this not include the rest of the context?

```suggestion
    return { ...context, headers }
```

alloy: Ah ok. I find that a little unintuitive, but definitely no blocker 👍

orta: Would this make sure that we also send requestIDs also? I've not seen a metaphysics -> jaws connection in the dd service map and wondered if it was related

ds300: Just double-checked. Yep, the request ids are going through. I added an extra middleware to make sure.

image

PR #1386, opened 2018-11-21
Offer order interface

ashkan18:
# Change?
An `Order` can either be a `BUY` order or an `OFFER` order based on `order.mode`. When an `Order` is an `Offer` order, we will expose many offer-only related fields like `offers`, `myLastOffer` and etc whereas a `Buy` order would only have a subset of those fields.

This seems to be a good use of a GraphQL `Interface` where in the interface we define all the common fields and then we expose more specific fields for different types, this way a `Buy` Order pretty much defines all fields of `OrderInterface` and then the `Offer` order can expand and add its own set of extra fields which can be accessed with
```graphql
order {
  id
  taxTotalCents
  ... on OfferOrder {
    myLastOffer {
      ...
    }
  }
}
```

This PR does not include any of the Offer specific fields yet (and keeps the old offer fields in the Order around) and introduces the interface so when we merged https://github.com/artsy/exchange/pull/286 we are ready to add offer fields.

This way we can test to make sure reaction and emission work fine with this change and then we won't break things when offer order fields are added.

# Do I need to change anything on the clients?
No! but we should verify things.


src/schema/schema.ts
```diff
@@ -225,4 +226,9 @@ export default new GraphQLSchema({
       viewer: Viewer,
     },
   }),
+  // These are for orphaned types which are types which should be in the schema,
+  // but can’t be discovered by traversing the types and fields from query.
+  //
+  // In this case,
```
orta:
```suggestion
  // In this case, the interface "Offer" is exposed everywhere, but the underlaying type BuyOrder needs to exist 
```

src/test/fixtures/exchange/mockxchange.js
```diff
@@ -36,6 +36,11 @@ export const mockxchange = resolvers => {
         }
       },
     },
+    Order: {
+      __resolveType(obj, _context, _info) {
+        return "BuyOrder"
+      },
+    },
```
orta: These are so that tests can know what an underlaying concrete type is  when asking for an "Order" - it will need to change in the future when there are test for offers

src/schema/ecommerce/types/order.ts
```diff
@@ -181,7 +192,10 @@ export const {
   connectionType: OrderConnection,
   edgeType: OrderEdge,
 } = connectionDefinitions({
-  nodeType: OrderType,
+  // TODO: DO NOT MERGE WITHOUT VERIFYING THIS
+  // and maybe updating @types/graphql
+  // @ts-ignore
+  nodeType: OrderInterface,
```
orta: The d.ts says a nodeType can only be a graphql object type, but it can also be an interface, this'll need adding to the types - @ashkan18 did you test whether the connection was fine?

_schema.graphql
```diff
@@ -2006,6 +2006,230 @@ type BuyersPremium {
...
+  stateReason: String
+
+  # Tracking code of the order
+  code: String
+
+  # Order Requested Fulfillment
+  requestedFulfillment: RequestedFulfillment
+
+  # Latest offer
+  lastOffer: Offer
```
starsirius: If I understand correctly, offer-specific fields will be removed eventually from `Order`interface and `BuyOrder` type, right?

ashkan18: yes, `lastOffer` and `offers` will be deprecated

src/schema/ecommerce/types/order.ts
ashkan18: not yet, need to try that and it's harder to setup than what you think :)

ashkan18: alright! confirmed! connections work! ✅ 

image

orta: I've looked into this from the types perspective, this is the corresponding issue - https://github.com/graphql/graphql-relay-js/issues/41

In *2015* relay could not support interface connections - but maybe relay modern can

orta: The relay compiler was fine with it when we migrated the schema in, so I think it's likely that that changed

src/schema/schema.ts
```diff
@@ -225,4 +226,9 @@ export default new GraphQLSchema({
       viewer: Viewer,
     },
   }),
+  // These are for orphaned types which are types which should be in the schema,
+  // but can’t be discovered by traversing the types and fields from query.
+  //
+  // In this case, the interface "Offer" is exposed everywhere, but the underlaying type BuyOrder needs to exist 
+  types: [BuyOrderType],
```
alloy: I don’t understand this quite yet, but that’s no blocker. If you have any pointers to docs/source/tests, that would be much appreciated.

alloy:
If this indeed is an acceptable schema for released app versions to use, then this looks great to me 👍

src/schema/ecommerce/types/order.ts
```diff
@@ -16,145 +17,155 @@ import { OrderPartyUnionType } from "./order_party_union"
...
+    type: GraphQLString,
+    description: "Currency code of this order",
+  },
+  state: {
+    type: GraphQLString,
+    description: "State of the order",
+  },
+  stateReason: {
+    type: GraphQLString,
+    description: "Reason for current state",
```
sweir27: I realize this was here before, but do clients actually use it?

src/schema/ecommerce/types/order.ts
```diff
@@ -16,145 +17,155 @@ import { OrderPartyUnionType } from "./order_party_union"
...
+  },
+  code: {
+    type: GraphQLString,
+    description: "Tracking code of the order",
```
sweir27: `"Tracking code"` is a little confusing, makes me think it's related to the shipping tracking code.

src/schema/ecommerce/types/order.ts
```diff
@@ -16,145 +17,155 @@ import { OrderPartyUnionType } from "./order_party_union"
...
+    type: RequestedFulfillmentUnionType,
+    description: "Order Requested Fulfillment",
+  },
+  lastOffer: {
+    type: OfferType,
+    description: "Latest offer",
+  },
+  offers: {
+    type: OfferConnection,
+    description: "List of offers made on this order so far",
```
sweir27: `List of submitted offers`

src/schema/schema.ts
ashkan18: Sorry I was driving to the garden state... yes it's same issue as what we saw on the Exchange version of the same PR where if a type is not referenced anywhere it doesn't end up in the schema definition. In Ruby Graphql we solved this by setting _orphan types_ but here looks like we need to explicitly add them to schema.

ashkan18:
@sweir27 in this specific case nothing should really be different, we can exercise the actual benefit of interface when we add new offer fields like `myLastOffer` and etc, wanted to have this PR first so we can make sure we don't impact BN and get ready to introduce new offer fields, once those are added they can only be accessed like the example in the description
```graphql
order {
  id
  taxTotalCents
  ... on OfferOrder {
    myLastOffer {
      ...
    }
  }
}
```

src/schema/ecommerce/types/order.ts
ashkan18: we don't use it in force but i think its used maybe in CMS, def used in Pulse.

ashkan18:
@orta I addressed the comments and also removed the TODO, is there anything else we need to do?

src/schema/schema.ts
alloy: Ah gotcha, thanks 👍

orta:
Yep, ok, cool - I'll test this out in staging too 👍

orta:
Confirming that this is looking fine on staging

screen shot 2018-11-22 at 11 11 33 am

ashkan18:
Awesome! Thanks! 🙌 I'll try the offer flow as well

alloy:
Great stuff @ashkan18 👏

PR #1387, opened 2018-11-23
Switched to OfferOrder fields

ashkan18:
# Depends on
https://github.com/artsy/exchange/pull/289

# Changes
- added `OfferOrder` which implements `OrderInterface` and adds offer specific fields
  - `myLastOffer`
  - `offers`
  - `lastOffer`
  - `awaitingResponseFrom` which also uses newly added `OrderParticipantEnum`
- kept existing offer fields on `OrderInterface` till we switch to `OfferOrder` everywhere.


src/schema/ecommerce/types/order.ts
```diff
@@ -153,12 +146,39 @@ const orderFields = () => ({
...
+}
 
 export const OrderInterface = new GraphQLInterfaceType({
   name: "Order",
-  resolveType: () => BuyOrderType,
-  fields: orderFields,
+  resolveType: ({ mode }) => {
+    if (mode === "BUY") return BuyOrderType
+    else if (mode === "OFFER") return OfferOrderType
+  },
```
orta: 👍

jonallured:
Looking this over, I'm trying to decide if anything here would cause my code in Volt to change, but I think it's fine? I have an `Order` fragment that I use to grab a specific order:

https://github.com/artsy/volt/blob/0e422a1d1debdf1ddb653ce438256e48a194673d/app/operations/metaphysics/_order_fragment.graphql#L1-L4

https://github.com/artsy/volt/blob/0e422a1d1debdf1ddb653ce438256e48a194673d/app/operations/metaphysics/order.graphql#L1-L5

And I use that same fragment when sending mutations so that I can update the state of the order client side. So just to confirm: because both `BuyOrder` and `OfferOrder` implement the `Order` interface, I don't have to change anything, is that correct?

ashkan18:
Thanks @jonallured for verifying, that looks 👌 to me. The only change we need to make is to switch the `offerTotal` to use `itemsTotal`. We have _deprecated_` that field and `itemsTotal` should reflect the last submitted `offer` total.
This PR only impacts how you'd get offer related fields which is now through `... on OfferOrder` fragment.
before we could do 
```graphql
order {
  id 
  lastOffer
}
```
in new style we need to do 
```graphql
order {
  id 
  ... on OfferOrder {
     lastOffer
  }
}
```
Old style would still be working for `lastOffer` and `offers` till we completely remove it

PR #1392, opened 2018-11-26
Exposes AddInitialOfferToOrder mutation

sweir27:
Since https://github.com/artsy/exchange/pull/286, metaphysics is able to send the `AddInitialOfferToOrder` mutation to exchange, which is just a rename of `InitialOffer` to be more explicit that this mutation adds an offer and returns an _order_.


src/schema/ecommerce/initial_offer_mutation.ts
```diff
@@ -8,7 +8,7 @@ import { InitialOfferInputType } from "./types/initial_offer_input_type"
 import { moneyFieldToUnit } from "lib/moneyHelper"
 export const InitialOfferMutation = mutationWithClientMutationId({
   name: "InitialOffer",
-  description: "Creates an order with an artwork",
+  description: "Deprecated. Use AddInitialOfferToOrder instead.",
```
ashkan18: I _think_ for this to be really deprecated (documentation-wise) we need to remove description and add `deprecationReason`, something like
```
deprecationReason: "Use AddInitialOfferToOrder instead.",
```
haven't used this for mutations but I think its the same.

PR #1394, opened 2018-11-27
remove goreplay

izakp:
As you all were in this thread: https://artsy.slack.com/archives/C2TQ4PT8R/p1543350165754200

Remove goreplay as we're not actively using it.  Will pull this boilerplate out into its own Dockerfile in https://github.com/artsy/docker-images

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,17 +9,6 @@ ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_a
 RUN chown deploy:deploy /usr/local/bin/dumb-init
 RUN chmod +x /usr/local/bin/dumb-init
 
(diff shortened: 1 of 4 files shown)


izakp:
Actually it's already set up in https://github.com/izakp/docker-images/blob/master/goreplay/Dockerfile  so to add it back as a sidecar container we just need to add the sidecar container back, so for future reference, in staging.yml...

```
- name: metaphysics-goreplay
  image: artsy/goreplay:latest
  imagePullPolicy: Always
  args: ["/usr/local/bin/goreplay", "--input-raw", ":3000", "--output-tcp goreplay.stg.artsy.systems"]
  securityContext:
    capabilities:
      add:
      - NET_RAW
      - NET_ADMIN
```

PR #1395, opened 2018-11-27
Remove the run-goreplay reference in the dev hokusai yml

orta:
Looks like the file hadn't ran through prettier, so check it with whitespace changes removed


l2succes:
merge on green

PR #1397, opened 2018-11-28
Port daySchedules formatting logic from force

javamonn:
We want to format `Location.day_schedules` for display as part of ongoing local discovery work. Force contains some logic that deduplicates and squashes this data [here](https://github.com/javamonn/force/blob/55e47edd24106806705bef82cdd74314148097af/src/desktop/models/partner_show.coffee#L201-L255). This PR runs that coffeescript through [decaffinate](https://github.com/decaffeinate/decaffeinate), cleans it up a bit, and adds the formatted result as a `displayDaySchedules` on `Location`.

### DIFF
--- a/src/schema/__tests__/formattedDaySchedules.test.ts
+++ b/src/schema/__tests__/formattedDaySchedules.test.ts
@@ -1,0 +1,132 @@
+import {
+  FormattedDaySchedules,
+  DaySchedule,
+} from "../types/formattedDaySchedules"
+
+const formatDaySchedules = FormattedDaySchedules.resolve
+
+describe("FormattedDaySchedules.resolve", () => {
+  it("collapses times duplicated across days", () => {
+    const daySchedules = [
+      {
+        day_of_week: "Monday",
+        start_time: 37800,
+        end_time: 68400,
+      },
+      {
+        day_of_week: "Tuesday",
+        start_time: 37800,
+        end_time: 68400,
+      },
+      {
+        day_of_week: "Wednesday",
+        start_time: 37800,
+        end_time: 68400,
+      },
+      {
+        day_of_week: "Thursday",
+        start_time: 37800,
+        end_time: 68400,
+      },
+      {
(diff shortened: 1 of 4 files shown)


src/schema/__tests__/formattedDaySchedules.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        days: "Monday, Wednesday–Saturday",
+        hours: "11am–2pm, 6pm–8pm",
+      },
+      {
+        days: "Sunday",
+        hours: "11am–2pm, 5pm–8pm",
+      },
+    ])
+  })
+})
```
anandaroop: I was comparing these to [these specs in Force](https://github.com/artsy/force/blob/33ece666e9df1cf32f9c6cbe7914811815a860cc/src/desktop/test/models/partner_show.coffee#L278) — one thing that's not clear to me is if/when we should expect to see "Closed" in the formatted output.

Seems from [this line](https://github.com/artsy/metaphysics/pull/1397/files#diff-4a5993f0c9dcea4bf4fec291505453d7R49) like there is such a condition, and I thought maybe it would apply to Tuesday in the above example?

A spec that clarifies the "Closed" behavior might generally be helpful here.

javamonn: Per our discussion at standup, I've updated this so that "Closed" days are no longer filtered from the result. Thanks for calling this out!

src/schema/__tests__/formattedDaySchedules.test.ts
```diff
@@ -0,0 +1,132 @@
...
+      },
+    ]
+    expect(formatDaySchedules(daySchedules as Array<DaySchedule>)).toEqual([
+      {
+        days: "Monday–Saturday",
+        hours: "10:30am–7pm",
+      },
+      {
+        days: "Sunday",
+        hours: "Closed",
```
anandaroop: 👍 thanks

PR #1400, opened 2018-11-29
[Do not merge] Repo of code-frame suppression issue

damassi:
Repo of code-frame suppression issue due to `throw` in our test utils. See inline for screenshots.

### DIFF
--- a/src/test/utils.ts
+++ b/src/test/utils.ts
@@ -30,14 +30,16 @@ export const runQuery = (
     xForwardedFor: "123.456.789",
   }
 
-  return graphql(schema, query, rootValue, context).then(result => {
-    if (result.errors) {
-      const error = result.errors[0]
-      throw error.originalError || error
-    } else {
-      return result.data
-    }
-  })
+  return graphql(schema, query, rootValue, context)
+    .then(result => {
+      if (result.errors) {
+        const error = result.errors[0]
+        throw error.originalError || error
+      } else {
+        return result.data
+      }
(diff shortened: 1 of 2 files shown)


src/test/utils.ts
```diff
@@ -30,14 +30,16 @@ export const runQuery = (
...
-      throw error.originalError || error
-    } else {
-      return result.data
-    }
-  })
+  return graphql(schema, query, rootValue, context)
+    .then(result => {
+      if (result.errors) {
+        const error = result.errors[0]
+        throw error.originalError || error
```
damassi: without the `catch` down below this is the complete output of the error introduced above: 

screen shot 2018-11-28 at 4 41 07 pm

(The stack trace provides no originating file location.)

src/test/utils.ts
```diff
@@ -30,14 +30,16 @@ export const runQuery = (
...
+      } else {
+        return result.data
+      }
+    })
+    .catch(error => console.error(error))
```
damassi: With `catch` it points to the location, as well as provides a code frame directly to the test that triggered the error: 

screen shot 2018-11-28 at 4 32 17 pm

damassi: That said, the above isn't truly that useful since if the error originates outside of the test in, say, a schema file, the code-frame will point to the test location that triggered the error (failing test) but not the source-code where it originates. Pre-stitching I think the locations were a bit more clear.

PR #1402, opened 2018-11-29
[test] Improve error reporting

alloy:
Fixes #1400 

I’m a little nervous that the `CombinedError` that gets thrown in `graphql-tools` doesn’t automatically report stacks of the individual errors, as we might be getting unhelpful errors reported to Sentry. If that turns out the be an issue at some point, then we should probably submit a PR to `grahql-tools` to do something like the combined stack traces I’ve done in this PR. /cc @freiksenet

### DIFF
--- a/src/test/utils.ts
+++ b/src/test/utils.ts
@@ -1,6 +1,6 @@
 // Please do not add schema imports here while stitching is an ENV flag
 //
-import { graphql } from "graphql"
+import { graphql, GraphQLError } from "graphql"
 
 /**
  * Performs a GraphQL query against our schema.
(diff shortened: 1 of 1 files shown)


joeyAghion:
Any chance the problem described [here](https://github.com/artsy/metaphysics/pull/1074#discussion_r233228464) is related to this?

damassi:
(One thing to note is that we have another function ([runQueryMerged](https://github.com/artsy/metaphysics/pull/1402/files#diff-5ffbfce5647f737c5bc8671db24df41cR112)) down below that might need the same treatment.)

alloy:
@joeyAghion _Could_ well be. Let’s pair tomorrow on that task you had?

@damassi Ah yeah, good point. Though, I expect that variant to be removed if stitching is here to stay , no? /cc @orta

PR #1403, opened 2018-11-29
[depth limit] Make it possible to disable this validation rule.

alloy:
Closes https://artsyproduct.atlassian.net/browse/PLATFORM-1060


alloy:
(Will fix that type error later)

joeyAghion:
I'm still unclear about whether the failing query is _actually_ exceeding the depth limit, or whether there's a bug in the limit's implementation. If it's actually exceeding the limit, do we have a lead on a fix?

(Sorry for messaging about this in multiple forums.)

joeyAghion:
I poked at this and confirmed that this query reaches a depth of 12. I think I understand how depth isn't a great proxy for complexity, but I'm still uncomfortable removing the limit altogether. Could we increase it instead?

My worry is that disabling it _by default_ effectively makes it dangerous to ever enable. Development will tend to happen in environments with no limit imposed, and we risk only surfacing violations after deploys to staging or production.

alloy:
I think we should always have the ability to disable a feature like this. I have a hard time thinking of what the default should be other than disabled, because the whole concept is too arbitrary to me. 20?

joeyAghion:
I agree that we should be able to disable it, so I'll merge. This status is OK with me as long as staging and production remain disabled.

PR #1406, opened 2018-11-30
@damassi => [ArtworkLayer] Add artworkConnection

damassi:
**On hold:** Currently the sale layer [doesn't support](https://github.com/damassi/gravity/blob/f9a7563afa49650f779e7fcbf2f6299bd083b688/app/models/search/layers/for_sale_layer.rb#L21) pagination.

```gql
{
  artwork(id:"lucio-fontana-concetto-spaziale-attese-139") {
    layers {
      artworksConnection(first: 3) {
        edges {
          node {
            id
          }
        }
      }
    }
  }
}
```

```json
{
  todo: add response
}
```


src/schema/artwork/layer.ts
```diff
@@ -0,0 +1,82 @@
...
+        const gravityArgs: GravityArgs = {
+          artwork: [artwork_id],
+          size,
+        }
+
+        return relatedLayerArtworksLoader({ id, type }, gravityArgs).then(
+          body => {
+            return connectionFromArraySlice(body, options, {
+              arrayLength: body && body.length,
+              sliceStart: 0,
```
damassi: Nice -- if we ever need pagination we can return, but for the Artwork page this is sufficient

damassi:
thanks @mzikherman - mergin'

PR #1407, opened 2018-11-30
[Typescript Refactor] Convert remaining files to TS

damassi:
Been working in MP the past week and have been converting files to `.ts` as I went and realized the issues were light; figured I would just convert all the rest. 

There are a few type issues here and there. I added comments where things could be fixed and where `// @ts-ignore` was used, but that can be done in a second pass.


orta:
AMAZING

src/schema/artist/index.ts
```diff
@@ -72,7 +72,7 @@ const artistArtworkArrayLength = (artist, filter) => {
 export const ArtistType = new GraphQLObjectType({
   name: "Artist",
   interfaces: [NodeInterface],
-  fields: () => {
+  fields: (): any => {
```
orta: hah, yeah, I've tried a few times with this one - also, I've never thought of doing the any assignment using this syntax, good work. it feels more obvious (than a `as any` at the bottom)

orta:
Sorry other PRs which may need to rebase.

src/schema/artist/index.ts
damassi: Yah felt it was more important to get everything converted rather than fuss, we can always iterate 😄

damassi:
@orta - q. I just realized that git doesn't track history across file renames automatically. Suggestions about how to do this without blowing up `git blame`? Can revert and take care of this tomorrow / open to suggestions.

orta:
There isn't an option from the GitHub UI etc to follow across files (git tracks content, not files) so folks who are diving into the history will need to use ` git log --follow` to look at diffs for changed files. Afraid it's just an issue we'll have to deal with.

PR #1417, opened 2018-12-05
Howto add exchange operations to MP guide

sepans:
To be linked from [here](https://github.com/artsy/exchange/pull/292)


ashkan18:
Nice! thanks!

PR #1419, opened 2018-12-05
Expose stub shows for local discovery

anandaroop:
Implements https://artsyproduct.atlassian.net/browse/LD-40

Allows clients to request a list of shows that includes local discovery stubs, as well as regular displayable shows.

We do this by taking the new `discoverable` scope that was [added](https://github.com/artsy/gravity/pull/11995) to Gravity's `/shows` endpoint, and bring it into MP in two places:

- as an optional arg provided to a city's `shows` resolver

city

- as an optional arg provided to a show's `nearbyShows` resolver

show

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -2325,6 +2325,12 @@ type City {
   shows(
     sort: PartnerShowSorts
     status: EventStatus
+
+    # Whether to include only displayable shows
+    displayable: Boolean = true
+
+    # Whether to include local discovery stubs as well as displayable shows
+    discoverable: Boolean
     after: String
     first: Int
     before: String
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -2326,9 +2326,6 @@ type City {
     sort: PartnerShowSorts
     status: EventStatus
 
-    # Whether to include only displayable shows
-    displayable: Boolean = true
-
     # Whether to include local discovery stubs as well as displayable shows
     discoverable: Boolean
     after: String
@@ -7735,9 +7732,6 @@ type Show implements Node {
   nearbyShows(
     sort: PartnerShowSorts
     status: EventStatus
-
-    # Whether to include only displayable shows
-    displayable: Boolean = true
 
     # Whether to include local discovery stubs as well as displayable shows
     discoverable: Boolean
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -7729,6 +7729,12 @@ type Show implements Node {
   nearbyShows(
     sort: PartnerShowSorts
     status: EventStatus
+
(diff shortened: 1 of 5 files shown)


src/schema/show.ts
```diff
@@ -495,6 +495,16 @@ export const ShowType = new GraphQLObjectType({
       args: pageable({
         sort: PartnerShowSorts,
         status: EventStatus,
+        displayable: {
+          type: GraphQLBoolean,
+          defaultValue: true,
```
anandaroop: We are still defaulting to `displayable` when asking for a list of `nearbyShows` — but should we default to `discoverable` instead?

@orta would you know if there are any consumers besides the new LD screens that know or care about this field yet?

orta: Not to my knowledge

anandaroop: Cool, I think I'll leave it as is. Seems more conservative, if the idea is not to leak out local discovery stub shows when they're not explicitly requested.

@mbilokonsky I agree about the weird subtleties among displayable/discoverable. I hoped to address that a bit by documenting [here](https://github.com/artsy/metaphysics/pull/1419/files#diff-7b1c236a12a473dffad8b241346f7f28R506) for consumers.

anandaroop:
Thought on it and pushed up a couple of commits:

- Make the inclusion of all `discoverable` nearby shows an opt-in for `city` just as it is for `show`. Seemed confusing for these to have different defaults e80bd0a

- Tried to address @mbilokonsky's comment further by taking `displayable` out of the picture completely for consumers. It's an implementation detail that consumers wouldn't set explicitly anyway. afe49d6

_schema.graphql
```diff
@@ -7729,6 +7732,9 @@ type Show implements Node {
   nearbyShows(
     sort: PartnerShowSorts
     status: EventStatus
+
+    # Whether to include local discovery stubs as well as displayable shows
+    discoverable: Boolean
```
orta: I'm glad we have this comment, because otherwise this wouldn't make sense - did you consider something like `includeStubs`?

orta:
You're welcome to merge if you're happy with `discoverable` vs `includeStubs` 👍

_schema.graphql
anandaroop: Yeah good point. It matches the name of the query param that this maps to over in Gravity.  "Discoverable" is now [one more scope](https://github.com/artsy/gravity/blob/master/app/api/v1/shows_endpoint.rb#L137) you can request, and it includes "displayable" as well as stub shows.

So it's consistent in a _sense_, but I do not dispute that the naming could be better 😬.

PR #1421, opened 2018-12-05
Revert "DISCO-494: Return null shipping info if artwork is not acquirable"

dblandin:
Reverts artsy/metaphysics#1385

ticket: https://artsyproduct.atlassian.net/browse/DISCO-544 :lock: 

This caused a regression in the order flow preventing shipping information from displaying on the receipt once an order is submitted.

We've updated Force (desktop and mobile) to hide shipping information for artworks not enrolled in an e-commerce program ([PR](https://github.com/artsy/force/pull/3161)). We can now remove this logic from Metaphysics.

There's an ongoing discussion on how client-specific responses from Metaphysics should be. There's more context available in the original PR [here](https://github.com/artsy/metaphysics/pull/1385#issuecomment-441789606) and in the Force PR [here](https://github.com/artsy/force/pull/3161#issue-235886148).

### DIFF
--- a/src/schema/artwork/__tests__/artwork.test.js
+++ b/src/schema/artwork/__tests__/artwork.test.js
@@ -1320,31 +1320,13 @@ describe("Artwork type", () => {
       }
     `
 
-    beforeEach(() => {
-      artwork.acquireable = true
-    })
-
-    it("is null if artwork is not enrolled in in an e-commerce program", () => {
-      artwork.acquireable = false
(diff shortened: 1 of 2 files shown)


sweir27:
Thank you! This seems straightforward enough.

Has force been deployed with the new changes yet? Once that happens I think this is good to merge 👍 .

dblandin:
> Has force been deployed with the new changes yet?

Yes! The relevant change in Force should be out in production now. :+1:

PR #1422, opened 2018-12-06
[env] Disable QUERY_DEPTH_LIMIT by default

alloy:
Going to unset them in staging and production


joeyAghion:
👍 as discussed, this addresses my concern about consistency of development and deployed environments.

PR #1423, opened 2018-12-10
Add artworkVersionId to the query for line items

jonallured:
While working on #1326 we missed adding the version id to query to Exchange so this PR is adding it.


ashkan18:
awesome! you may need to update tests to return/include this new field, but lets see :) merge on green

jonallured:
@ashkan18 ok, this is green and I think I found all the places to update fixture data but lmk if you see something I missed!!

ashkan18:
cool! lets do this! 🎉

PR #1424, opened 2018-12-10
LD-85 Expose opening_reception_text on Show

anandaroop:
Implements https://artsyproduct.atlassian.net/browse/LD-85

This supports the opening reception free text that was [added to shows](https://github.com/artsy/gravity/pull/12047) for local discovery stubs

open

### DIFF
--- a/src/schema/show.ts
+++ b/src/schema/show.ts
@@ -550,6 +550,7 @@ export const ShowType = new GraphQLObjectType({
       type: GraphQLString,
       description:
         "Alternate Markdown-supporting free text representation of the opening reception event’s date/time",
+      resolve: ({ opening_reception_text }) => opening_reception_text,
     },
     partner: {
       description:
--- a/src/schema/show.ts
+++ b/src/schema/show.ts
@@ -546,6 +546,11 @@ export const ShowType = new GraphQLObjectType({
         return results
       },
     },
+    opening_reception_text: {
+      type: GraphQLString,
+      description:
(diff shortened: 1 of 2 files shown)


src/schema/show.ts
```diff
@@ -546,6 +546,11 @@ export const ShowType = new GraphQLObjectType({
         return results
       },
     },
+    openingReceptionText: {
+      type: GraphQLString,
+      description:
+        "Alternate Markdown-supporting free text representation of the opening reception event’s date/time",
+    },
```
alloy: I assume that the previous snake_case field worked because the data coming from the upstream service uses that format, no? If so, I'd expect there to need to be a resolver that pulls out the snake_case field.

anandaroop: Yeesh, of course. Thanks!

anandaroop: Thanks for catching that silly error. My feeling is that trivial field definitions (i.e. the ones you don’t write resolvers for) don’t need tests, but maybe this one does now, since it transforms data from an upstream response?

Is that a reasonable rule of thumb?

orta: I think this is also just a bit too simple to warrant writing a test for, you're just testing JS syntax not business logic

anandaroop: Cool, works for me. I think this is done then, if one of y'all will do the honors. Else I'll merge in a bit.

alloy: I agree 👍

PR #1425, opened 2018-12-11
PLATFORM-959: Fix MP Excess Logging

dleve123:
https://artsyproduct.atlassian.net/browse/PLATFORM-959

In the event that the GraphqlError being raised wraps a CombinedError,
then unpack out all the underlying errors in the CombinedError and
instrument those separately to Sentry. We also handle two other types of
errors:

1. When the GraphQL error doesn't have an `originalError`, and as such,
is an error raised independent of stitched services (for example, a
malformed query made to MP), and

2. A GraphqlError that does have an original error, but this original
error isn't a CombinedError. This could maybe be an error representing a
bug from GraphQL Tools.


orta:
Cool - yes, this looks good. Nice abstractions too 👍

PR #1427, opened 2018-12-11
Refactor to fix bug in exposing Fair exhibitors 

ashleyjelks:
This refactor fixes a bug that caused an incomplete list of fair exhibitors to be returned from MP.

Links to the following Jira issues/bugs:

[LD-90](https://artsyproduct.atlassian.net/browse/LD-90)
[LD-31](https://artsyproduct.atlassian.net/browse/LD-31)

screen shot 2018-12-11 at 11 56 06 am


l2succes:
LG! to re-iterate our slack conversation. We will evaluate performance of this field using `allViaLoader` to get all pages from gravity first before grouping. If we run into issues, we should use an exhibitors connection instead

joeyAghion:
Mind linking to the discussion?

On the surface, I'm concerned about designing this to degrade by blasting the API with an unbounded number of concurrent requests. A safer design would be to max out at a reasonable number of paginated requests (say, 5).

PR #1428, opened 2018-12-11
Upgrade node to v10.14.1 for security updates

zephraph:
Bumping to ensure we have the latest security patches.


zephraph:
merge on green

PR #1430, opened 2018-12-11
Add artistsConnection to Fair

javamonn:
~Depends on: https://github.com/artsy/gravity/pull/12048~ (Merged!)

Implements: [LD-83](https://artsyproduct.atlassian.net/browse/LD-83).

Uses Gravity's [FairArtistsEndpoint](https://github.com/artsy/gravity/blob/master/app/api/v1/fairs_artists_endpoint.rb) to expose an `artistsConnection` field on Fair. This field backs the mock seen [here](https://app.zeplin.io/project/5bef4c7dae2b6a3eb6a75f4a/screen/5bf32e8aa6dc5049913296b9). Emission will paginate over this field, sorted by name, and build alphabetical groupings as more data is fetched.

### DIFF
--- a/src/schema/fair.ts
+++ b/src/schema/fair.ts
@@ -1,7 +1,8 @@
-import { omit } from "lodash"
+import { omit, map } from "lodash"
 import { pageable } from "relay-cursor-paging"
 import { connectionFromArraySlice } from "graphql-relay"
 import { convertConnectionArgsToGravityArgs } from "lib/helpers"
+import { artistConnection } from "./artist"
 import moment from "moment"
 import cached from "./fields/cached"
 import date from "./fields/date"
@@ -22,6 +23,7 @@ import {
 import { totalViaLoader } from "lib/total"
 import ShowSort from "./sorts/show_sort"
 import { allViaLoader } from "lib/all"
+import { FairArtistSortsType } from "./sorts/fairArtistSorts"
 
 const FairOrganizerType = new GraphQLObjectType({
   name: "organizer",
@@ -47,6 +49,41 @@ const FairType = new GraphQLObjectType({
   name: "Fair",
   fields: () => ({
     ...GravityIDFields,
+    artists: {
+      type: artistConnection,
+      args: pageable({
+        sort: {
+          type: FairArtistSortsType,
+          description: "Sorts for artists in a fair",
+        },
+      }),
+      resolve: (
+        { id },
+        options,
+        _request,
+        { rootValue: { fairArtistsLoader, artistsLoader } }
+      ) => {
+        const gravityOptions = omit(
+          convertConnectionArgsToGravityArgs(options),
+          ["page"]
+        )
+
+        gravityOptions.total_count = true
+
+        return fairArtistsLoader(id, gravityOptions).then(
+          ({ body, headers }) => {
+            return artistsLoader({ ids: map(body, "artist_id") }).then(
+              artists => {
+                return connectionFromArraySlice(artists, options, {
+                  arrayLength: headers["x-total-count"],
+                  sliceStart: gravityOptions.offset,
+                })
+              }
+            )
+          }
+        )
+      },
+    },
     cached,
     banner_size: {
       type: GraphQLString,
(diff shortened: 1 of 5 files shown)


src/schema/fair.ts
```diff
@@ -47,6 +49,44 @@ const FairType = new GraphQLObjectType({
...
+        const gravityOptions = omit(
+          convertConnectionArgsToGravityArgs(options),
+          ["page"]
+        )
+
+        return Promise.all([
+          totalViaLoader(fairArtistsLoader, id, gravityOptions),
+          fairArtistsLoader(id, gravityOptions),
+        ]).then(([count, { body }]) => {
+          return Promise.all(body.map(({ id }) => artistLoader(id))).then(
```
javamonn: This is less than optimal as it requests each artist individually, so you'll have as many concurrent requests in flight against gravity as the requested page size. I initially tried to get this working with `artistsLoader` but it appears the backing Gravity endpoint only accepts internal Mongo Ids: https://github.com/artsy/gravity/blob/master/app/api/v1/artists_endpoint.rb#L239-L242, as the param is passed directly to an `.in(_id: params[:ids])` filter.

javamonn:
Throwing a WIP label on this to address the discussion here: https://artsy.slack.com/archives/C1HH3KNJG/p1544566057083100

javamonn:
The Gravity work this PR was blocked on has been merged, so this is ready for review.

screen shot 2018-12-13 at 4 31 53 pm

src/schema/fair.ts
```diff
@@ -47,6 +49,44 @@ const FairType = new GraphQLObjectType({
...
+        _request,
+        { rootValue: { fairArtistsLoader, artistsLoader } }
+      ) => {
+        const gravityOptions = omit(
+          convertConnectionArgsToGravityArgs(options),
+          ["page"]
+        )
+
+        return Promise.all([
+          totalViaLoader(fairArtistsLoader, id, gravityOptions),
```
mzikherman: You might be able to drop this call, and add `gravityOptions.total_count = true` somewhere. Then, you can resolve `headers` as well as `body` of `fairArtistsLoader`, and retrieve the count in `headers['x-total=count']`.

src/schema/sorts/fairArtistSorts.ts
```diff
@@ -0,0 +1,18 @@
+import { GraphQLEnumType } from "graphql"
+
+export enum FairArtistSorts {
```
mzikherman: not sure if u need this

src/schema/fair.ts
```diff
@@ -47,6 +49,44 @@ const FairType = new GraphQLObjectType({
   name: "Fair",
   fields: () => ({
     ...GravityIDFields,
+    artistsConnection: {
```
anandaroop: I believe [our preference](https://github.com/artsy/README/blob/f8cc033ec7c71f7f19605a04f0c8b8d5d21b9d7f/playbooks/graphql-schema-design.md#how-to-model-our-graph) now is to just name this field `artists`, so long as a field by that name does not already exist.

src/schema/fair.ts
```diff
@@ -47,6 +49,44 @@ const FairType = new GraphQLObjectType({
...
+      type: artistConnection,
+      args: pageable({
+        section: {
+          type: GraphQLString,
+          description: "Number of artists to return",
```
anandaroop: I'll take your word for it, but just double-checking that this is what "section" means?

javamonn: Nothing apparently! I copied it over from a similar connection and had assumed it was an internal pagination parameter, but things appear to work fine without it - thanks for calling it out.

src/schema/fair.ts
```diff
@@ -47,6 +49,41 @@ const FairType = new GraphQLObjectType({
...
+        options,
+        _request,
+        { rootValue: { fairArtistsLoader, artistsLoader } }
+      ) => {
+        const gravityOptions = omit(
+          convertConnectionArgsToGravityArgs(options),
+          ["page"]
+        )
+
+        gravityOptions.total_count = true
```
l2succes: 👍

javamonn: I don't think so - I think it just exposes an `X-Total-Count` as a response header, though not sure the mechanism of how gravity arrives at that count. Refactored from a separate call to retrieve the count after @mzikherman's feedback here: https://github.com/artsy/metaphysics/pull/1430#discussion_r241569306

l2succes:
Since all feedback has been addressed let's merge

PR #1434, opened 2018-12-14
Add git remote option to MP

dleve123:
This should ensure that Git tags are always pushed when a `hokusai
pipeline promote` or `hokusai staging` are invoked.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   test:
     docker:
-      - image: artsy/hokusai:0.4.6
+      - image: artsy/hokusai:latest
     steps:
       - add_ssh_keys
       - checkout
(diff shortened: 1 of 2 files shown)


joeyAghion:
This requires that the configured CircleCI key has write access to the repo. Just mentioning in case that's not already the case (and it's not the default).

alloy:
The remote looks good to me, but I otherwise have no idea about this configuration, so I’ll defer to @izakp for that.

dleve123:
Considering that we already push tags to GH on the staging deploy, MP should have all the access it needs, but @izakp do please confirm that.

dleve123:
@izakp Addressed your comment in https://github.com/artsy/metaphysics/pull/1434/commits/256c663474408ebb9d089e3724b8e43573af6c47. Please review and let me know what you think.

izakp:
LGTM!

PR #1439, opened 2018-12-18
#donotmerge set ELB connection idle timeout to 5 seconds

izakp:
Update ELB configuration to set the backend connection idle timeout to 5 seconds, matching Node's [server keepalive timeout](https://nodejs.org/api/http.html#http_server_keepalivetimeout) and preventing the ELB using connections Node has closed.

## Migration

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -106,7 +106,7 @@ metadata:
     service.beta.kubernetes.io/aws-load-balancer-ssl-ports: "443"
     service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: "true"
     service.beta.kubernetes.io/aws-load-balancer-connection-draining-enabled: "true"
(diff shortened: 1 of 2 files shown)


izakp:
Just realized that although Node's idle timeout may be set to 5 seconds, MP requests can take significantly longer to complete, so setting this is in danger of prematurely closing the connection from the ELB to Metaphysics.  From https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/config-idle-timeout.html

> By default, Elastic Load Balancing sets the idle timeout to 60 seconds for both connections. Therefore, if the instance doesn't send some data at least every 60 seconds while the request is in flight, the load balancer can close the connection. To ensure that lengthy operations such as file uploads have time to complete, send at least 1 byte of data before each idle timeout period elapses, and increase the length of the idle timeout period as needed.

PR #1441, opened 2018-12-19
add Fairs and Shows to FollowsAndSaves, with mutations.

mbilokonsky:
This PR adds support to view and mutate a user's followed shows and followed fairs.

Shows are followed directly, while fairs are followed indirectly (By following the fair's profile). This PR exposes both through the same flat query, nested under `me.followsAndSaves`, and uses Connection/Paging logic to expose both models at the same level of abstraction.


src/schema/me/followed_shows.ts
```diff
@@ -0,0 +1,44 @@
...
+  fields: {
+    partner_show: {
+      type: ShowType,
+    },
+    ...IDFields,
+  },
+})
+
+export default {
+  type: connectionDefinitions({ nodeType: FollowedShowType }).connectionType,
```
mzikherman: This is ok, but an alternative is to model this similar to 'partner artists'. That is, a partner can have an artist, where the connection should return a node of an `Artist` type, and not a `PartnerArtist` type, however the _edges_ connecting a partner to an artist node, is made up of the `PartnerArtist` data (such as representation status, etc.)

Similarly here, a user/me should be connected to `Show`, since _that's_ the actual interesting object. A `FollowShow` is a relatively uninteresting/meta type, that clients shouldn't have to 'know' about. (What if we change following to some other behaviors, we could still preserve a user -> show graph, but not user -> followshow )

An example definition is https://github.com/artsy/metaphysics/blob/20b27b343ca809a9c2f88b329423f9a86e1d27b0/src/schema/partner_artist.ts#L88-L97 and used https://github.com/artsy/metaphysics/blob/master/src/schema/artist/highlights.ts#L13

This lets you have a connection of partners on an artist (artist -> partners), and the edge between each node contains metadata about that relationship. I would think that for follow shows, a similar setup could work (connection is me -> shows, and the edge contains metadata about that, was it a follow? an auto-follow? some other reason, etc.)

This is totally not a blocker to your setup, feel free to carry on with that! Just wanted to point out an alternative.

mzikherman: I think that's what @alloy was alluding to https://github.com/artsy/gravity/pull/12060#issuecomment-448696507 , where the follows are the edges and not the nodes.

alloy: Yup, that's what I meant! Thanks for the longer explanation @mzikherman 👍

mbilokonsky: I believe that the current configuration satisfies everything outlined in this conversation. I'm opening this up for review now and will tag everyone who commented here, to sanity check me. I'm going to include screenshots of GraphiQL input and output just as the demonstration of functionality, so I'll update this thread with those and then I'll assign.

mbilokonsky:
Initial state: no followed fairs or shows.

image

mbilokonsky:
Mutation 1: Following shows and fairs. Notice that we call `follow_show` for shows but `follow_profile` for fairs - as discussed in slack and elsewhere, the assumption is that a profile ID is always available from a fair object.

image

mbilokonsky:
Query 2: This shows the state of the system after the mutation has been applied. As you can see, both `show` and `fair` point to their respective actual domain model objects - we step through the profile and return the embedded fair, here.

image

mbilokonsky:
Mutation 2: we're going to unfollow the items we've followed.

image

mbilokonsky:
Final state: both fairs and shows unfollowed again.

image

mbilokonsky:
Thanks everyone! I was trying to figure out how to test stateful queries like this and decided that the screenshots of the system working are maybe the best way to demonstrate it - is that what we do for queries?

I did add tests for mutations.

damassi:
@mbilokonsky - one nice to have is to paste the query itself into the description so devs who are testing can copy paste back into graphiql.

mbilokonsky:
BTW whoever merges this, can you do a squash+merge to collapse all of these commits?

src/schema/me/follow_show.ts
```diff
@@ -0,0 +1,43 @@
+import { GraphQLString, GraphQLBoolean } from "graphql"
+import { mutationWithClientMutationId } from "graphql-relay"
+import { ShowType } from "../show"
+
+export default mutationWithClientMutationId({
+  name: "FollowShow",
+  description: "Follow (or unfollow) an show",
```
anandaroop: #minor "a show"

src/schema/me/followed_fairs.ts
```diff
@@ -0,0 +1,62 @@
...
+    ...IDFields,
+  },
+})
+
+// the challenge is that FollowedProfile doesn't directly connect to Fair.
+// we need to go from FollowedProfile to Profile to Owner in order to get the
+// reference to the fair object - how do I tell graphQL how to do that?
+// Is it just manually down in my resolver below?
+export const FollowedFairConnection = connectionDefinitions({
+  name: "FollowedFairConnection",
```
anandaroop: Wondering if this should contain "Connection" in the name? We are getting redundant "...ConnectionConnection" naming [up here](https://github.com/artsy/metaphysics/pull/1441/files#diff-ca178747ca9568ddd30e7dffd824b873R3547), presumably because of [the way the library handles this](https://github.com/graphql/graphql-relay-js/blob/88a2c88f53634cc7cae7ff56e33f678b9f34b7d5/src/connection/connection.js#L111).

[Ditto](https://github.com/graphql/graphql-relay-js/blob/88a2c88f53634cc7cae7ff56e33f678b9f34b7d5/src/connection/connection.js#L93) for "Edge" in [L10](https://github.com/artsy/metaphysics/pull/1441/files#diff-908679437786a6422f3da2c394b44d14R10) above.

src/schema/me/followed_fairs.ts
```diff
@@ -0,0 +1,62 @@
...
+  // FIXME: 'edgeType' does not exist in type 'ConnectionConfig'
+  // @ts-ignore
+  edgeType: FollowedProfileEdge,
+  nodeType: Fair.type,
+})
+
+export default {
+  type: FollowedFairConnection.connectionType,
+  args: pageable({}),
+  description: "A list of the current user’s currently followed profiles",
```
anandaroop: Should this be more specifically "currently followed _fair_ profiles"?

src/schema/me/followed_fairs.ts
```diff
@@ -0,0 +1,62 @@
...
+      type: ProfileType,
+    },
+    ...IDFields,
+  },
+})
+
+// the challenge is that FollowedProfile doesn't directly connect to Fair.
+// we need to go from FollowedProfile to Profile to Owner in order to get the
+// reference to the fair object - how do I tell graphQL how to do that?
+// Is it just manually down in my resolver below?
```
anandaroop: Does the implementation in [L58](https://github.com/artsy/metaphysics/pull/1441/files#diff-908679437786a6422f3da2c394b44d14R58) satisfy your question here? If so can we remove the comment?

src/schema/me/followed_shows.ts
```diff
@@ -0,0 +1,54 @@
...
+      type: ShowType,
+    },
+    ...IDFields,
+  },
+})
+
+export const FollowedShowConnection = connectionDefinitions({
+  name: "FollowedShowConnection",
+  // FIXME: 'edgeType' does not exist in type 'ConnectionConfig'
+  // @ts-ignore
```
anandaroop: I'm not clear on this — here and throughout, what's the significance of this comment? Is `connectionDefinitions` doing anything with this extra arg you are providing?

src/schema/me/followed_fairs.ts
mbilokonsky: Good catch, made the change! I've seen different approaches for Edges and unless someone specifically finds this one problematic I think I'm okay with it but also 100% open to any feedback!

mbilokonsky:
one more commit coming, my formatting fix broke something but I'm still trying to identify what. should be in shortly.

alloy:
This is starting to look good, but it seems like something got lost in translation in [our Slack chat](https://artsy.slack.com/archives/C1HH3KNJG/p1545429833038100) after all, because in reality the `FollowProfile` Gravity model is the edge and so we should probably model it like that. I.e. the graph should look more like the following:

```                                                                           
                         ┌───────────────────────────────┐                    
                         │        FollowsAndSaves        │                    
                         │     (Gravity model: N/A)      │                    
                         │  - profiles(ownerType: FAIR)  │                    
                         └───────────────────────────────┘                    
                                         │                                    
                                         │                                    
                                         ▼                                    
                         ┌───────────────────────────────┐                    
                         │    FollowProfileConnection    │                    
                         │     (Gravity model: N/A)      │                    
                         │            - edges            │                    
                         └───────────────────────────────┘                    
                                         │                                    
                     ┌───────────────────┴───────────────────┐                
                     │                                       │                
                     ▼                                       ▼                
     ┌───────────────────────────────┐       ┌───────────────────────────────┐
     │       FollowProfileEdge       │       │       FollowProfileEdge       │
     │(Gravity model: FollowProfile) │       │(Gravity model: FollowProfile) │
     │            - node             │       │            - node             │
     └───────────────────────────────┘       └───────────────────────────────┘
                     │                                       │                
                     │                                       │                
                     ▼                                       ▼                
     ┌───────────────────────────────┐       ┌───────────────────────────────┐
     │            Profile            │       │            Profile            │
     │   (Gravity model: Profile)    │       │   (Gravity model: Profile)    │
     │            - owner            │       │            - owner            │
     └───────────────────────────────┘       └───────────────────────────────┘
                     │                                       │                
                     │                                       │                
                     ▼                                       ▼                
     ┌───────────────────────────────┐       ┌───────────────────────────────┐
     │             Fair              │       │             Fair              │
     │     (Gravity model: Fair)     │       │     (Gravity model: Fair)     │
     │                               │       │                               │
     └───────────────────────────────┘       └───────────────────────────────┘
```

mbilokonsky:
So, stepping through this, I am definitely using FollowProfile as the edge. I define a schema for it (which just includes `profile` as I don't care about the rest of that edge) here: https://github.com/artsy/metaphysics/pull/1441/files#diff-908679437786a6422f3da2c394b44d14R9

I then use that schema as the `edgeType` in my connection, here, but specify `fair` as the node type: https://github.com/artsy/metaphysics/pull/1441/files#diff-908679437786a6422f3da2c394b44d14R23

To me what this is saying is: in the underlying data model, `FairProfile` -> `Profile` -> `Fair` - but in graphQL we can elid that middle `Profile` and put it into the node resolver that we always return `profile.owner` rather than `profile`, here: https://github.com/artsy/metaphysics/pull/1441/files#diff-908679437786a6422f3da2c394b44d14R52

The purpose of this is to allow a consumer of this API to treat `me.followsAndSaves.fairs` and `me.followsAndSaves.shows` as siblings - I realize their backing implementations have different schemas, but this is my attempt to close over that distinction and provide a unified interface. Does that make sense, or am I still not getting something?

alloy:
That's very fair and I do like a consistent schema! I guess that might be more a question for people with insights into the original modelling to know if there's anything we'd lose by not following the upstream modelling.

alloy:
That's very fair and I do like a consistent schema! I guess that might be more a question for people with insights into the original modelling to know if there's anything we'd lose by not following the upstream modelling.

/cc @joeyAghion

joeyAghion:
In general I would only break from upstream modeling for very good reasons, and I'm not sure this front-end case of treating profile follows like saves or other follows qualifies. The trouble will be when the back-end models evolve or the front-end wants to implement some logic pertaining to profiles (in this case). That will be awkward or impossible with the schema being proposed here. Is it much trouble for the client to make the additional hop from profiles to fairs?

(Goes without saying, but I'm out of my GraphQL depth here so apologies if I misunderstood.)

mbilokonsky:
I guess if GraphQL is literally just exactly modeling upstream relations 1:1 I'm not sure I see what the point of it is. I thought that the idea is that we can decouple the views we serve from the backing structures. Useful indirections that streamline our domain modeling for specific target cases, etc.

Back-end models evolving would always require updates to Metaphysics, so I'm not 100% sold on that as an argument for making the query more complex.

Front-end logic pertaining to profiles can already be managed through the various GQL Profile schemas and mutations - this is not that. This is specifically an attempt to expose followed fairs as simply as possible. I agree that if you try to use this to do profile work you're going to have a bad time - but even if you absolutely had to you can always get the profile from the fair, rather than always having to get the fair from the profile.

It really is an easy enough change, we can just remove that node resolver - but I feel like y'all are putting pressure directly against what I thought I understood about this, so apologies if I'm pushing back and trying to get clarity. 

Who is, like, our internal canonical expert on how GraphQL should be used? @orta? @l2succes? Can ya'll weigh in?

alloy:
The question when modeling this is if `profile` semantically is an attribute of the user->fair association. Had profiles, and the following of fairs through these profiles, not existed then it should definitely have been modeled like you did; but as it does exist I’m inclined to think it serves a purpose that we shouldn’t overlook.

alloy:
A contrived way to view the point of GraphQL is to fetch hierarchical data in a single request and only the attributes the client is interested in.

mbilokonsky:
Ah, sure, batching requests is the other reason for it.

If your concern is at its core a question of "we don't want to lose access to `profile`" then it's fine, because `fair.profile` is always embedded anyway, right?

orta:
It feels like we're trying to figure whether

```graphql
query ($first: Int) {
  me {
    followsAndSaves {
      profiles(first: $first, ownerType: FAIR) {
        edges {
          # This is a FollowProfile
          node {
            # This is a Profile
            owner {
              ... on Fair {
                id
              }
            }
          }
        }
      }
    }
  }
}
```

is more valid than:

```graphql
query ($first: Int) {
  me {
    followsAndSaves {
      fairs(first: $first) {
        edges {
          profile { # This is a Profile
             id
          }
          node { # This is a Fair
              id 
          }
        }
      }

      shows ( ... ) { }
    }
  }
}
```

From my perspective the second queries makes more sense from the perspective of a graph but still offers the same data. The first means you have to understand more about the raw gravity implementation details, but the second tells you what you want upfront. 

I had this problem before with `me { saved_artworks }` - because it's a really weird abstraction in Gravity but no-one would think of when writing a query. It supported both the raw API via root field `collection` , and re-used that abstraction to  provide something that made sense for consumers

mbilokonsky:
@orta that's interesting, so you'd include `profile` as a sibling to `node` even though it's also a child of `node`?

alloy:
Yeah I agree that it feels weird. If there’s no good reason to preserve that relationship then I’m inclined to go with option 1 too, but leave `profile` off of the edge (at least for now). It’s not an attribute of the relationship afaik.

mbilokonsky:
To be clear, I vastly prefer option 2 - I just would leave `profile` off the edge.

mbilokonsky:
And technically it IS an attribute of the relationship, if the edge is the relationship - it has `user` and `profile`. We just pass `profile` in to the `nodeResolver` which returns `profile.owner` instead.

But I think we've started to all say the same thing, and I think the current state of this PR reflects what we're agreeing on?

alloy:
We’re creating a direct relationship from the user to the fair, in that context a profile is unrelated and this shouldn’t be included. At least not until we need it and know how we’d want to use it. So yeah, the current PR is good, except that I think profile should be removed from the edge.

src/schema/me/follow_show.ts
```diff
@@ -0,0 +1,43 @@
+import { GraphQLString, GraphQLBoolean } from "graphql"
+import { mutationWithClientMutationId } from "graphql-relay"
+import { ShowType } from "../show"
+
+export default mutationWithClientMutationId({
+  name: "FollowShow",
+  description: "Follow (or unfollow) a show",
+  inputFields: {
+    partner_show_id: {
```
alloy: Should be cameCase

src/schema/me/followed_fairs.ts
```diff
@@ -0,0 +1,56 @@
+import { ProfileType } from "schema/profile"
+import Fair from "schema/fair"
+import { IDFields } from "schema/object_identification"
+
+import { pageable, getPagingParameters } from "relay-cursor-paging"
+import { connectionDefinitions, connectionFromArraySlice } from "graphql-relay"
+import { GraphQLObjectType } from "graphql"
+
+const FollowedProfileEdge = new GraphQLObjectType({
```
alloy: Rename this to remove the notion of “profile”, seeing as we’re modeling a direct relationship from user to fairs.

src/schema/me/follow_show.ts
mbilokonsky: Other follow fields are not (including FollowProfile), because we're passing this in as a query parameter to gravity which expects snake case like this. 

https://github.com/artsy/metaphysics/blob/master/src/schema/me/follow_gene.ts#L9
https://github.com/artsy/metaphysics/blob/master/src/schema/me/follow_artist.ts#L10
https://github.com/artsy/metaphysics/blob/master/src/schema/me/follow_profile.ts#L9

I'm going to leave it snake case so that there aren't mismatches in the internals unless you feel strongly that it should be different from the others?

mbilokonsky: @orta agrees that it should be camelCase not snake_case - but if we're going to aim for consistency I still think we should have a separate effort for that. If I change just shows now that's easy enough, but to change fairs too I'd have to change everything that consumes that schema etc and this PR feels like it's already done enough - plus then the codebase would be even more inconsistent, where different follow modules do different styles, etc.

Can we do an upkeep ticket for standardizing on variable naming, instead, and get this bad boy merged?

alloy: If that happens soon enough then I'm ok with that. Keep in mind that schema changes _should_ ideally be backwards compatible, we definitely don't want somebody to start using this in the iOS app and deploying it before you'd change it, because then we no longer can.

alloy:
@mbilokonsky Thanks for bearing with us and great job on completing it! 👌 👏 🎉

mbilokonsky:
Likewise, thanks for the help all!

To remove ambiguity I cleared the assignees and made @l2succes the assignee, since he's the one who'll be consuming this so I want to make sure he gets a chance to weigh in.

l2succes:
Thanks to everyone who chimed in! 

Great work @mbilokonsky the screenshots and query examples are very useful 👍 I just caught up with the thorough conversation on modeling nuisances. As we concluded, the important relation is User -> Fair (or User -> Show), the GraphQL model doesn't really need any metadata on the `profile` model, if it turns out we do, we'll have access to it via the node type. Also *option 2* makes the most sense for the UI as we'll be rendering followed Fairs and Shows in separate lists.

src/schema/me/follow_show.ts
alloy: All new fields should be camelCase. Not sure what you’re referring to by other examples in this PR; if they are fields that already existed then they stay as they are.

PR #1442, opened 2018-12-19
Expose and structure artist insights from Gravity

dblandin:
We're beginning to load in new artist insights to show on the artist page.

This will expose insights from Gravity via GraphQL, when available for an artist.

Related to https://github.com/artsy/gravity/pull/12059 :lock: 

ticket: https://artsyproduct.atlassian.net/browse/DISCO-588 :lock: 

2018-12-19-145432_1692x1207_scrot

### DIFF
--- a/src/schema/artist/__tests__/insights.test.ts
+++ b/src/schema/artist/__tests__/insights.test.ts
@@ -1,0 +1,135 @@
+/* eslint-disable promise/always-return */
+import { runQuery } from "test/utils"
+
+describe("ArtistInsights type", () => {
+  let artist = null as any
+  let rootValue = null as any
+
+  beforeEach(() => {
+    artist = {
+      id: "foo-bar",
+      name: "Foo Bar",
+    }
+    rootValue = {
+      artistLoader: () => Promise.resolve(artist),
+    }
+  })
+
+  it("returns an empty list if field values are null", () => {
+    const query = `
+        {
+          artist(id: "foo-bar") {
+            id
+            insights {
+              type
+              label
+              entities
+            }
+          }
+        }
+      `
+
+    return runQuery(query, rootValue).then(data => {
+      expect(data!.artist.insights).toEqual([])
+    })
+  })
+
+  it("does not build an insight if the field contains an empty string", () => {
+    artist.solo_show_institutions = "  "
+
+    const query = `
+        {
+          artist(id: "foo-bar") {
+            id
+            insights {
+              type
+              label
+              entities
+            }
+          }
+        }
+      `
+
+    return runQuery(query, rootValue).then(data => {
+      expect(data!.artist.insights).toEqual([])
+    })
+  })
+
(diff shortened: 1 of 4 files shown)


src/schema/artist/insights.ts
```diff
@@ -0,0 +1,88 @@
...
+})
+
+const ArtistInsights = {
+  type: new GraphQLList(Insight),
+  resolve: artist => {
+    return buildInsights(artist)
+  },
+}
+
+export default ArtistInsights
```
damassi: While this breaks protocol with much of how MP is written we've been trying to avoid default exports since they're not statically analyzable.

dblandin: Gotcha. I've switched to an explicitly named import :+1:

src/schema/artist/insights.ts
```diff
@@ -0,0 +1,86 @@
...
+      type: insightType.name,
+      entities: splitEntities(entitiesString),
+      label: insightType.value,
+    }
+  }
+
+  return compact(
+    Object.keys(ArtistInsightTypeMapping).map(key => {
+      if (artist[key]) {
+        return buildInsight(ArtistInsightTypeMapping[key], artist.major_solo)
```
mzikherman: I think the last param here you meant `artist[key]`

mzikherman:
LGTM! Needs a test, but I'm liking it 😎

dblandin:
Added a test. Looks like I need to convert it to typescript though.

src/schema/artist/__tests__/insights.test.ts
```diff
@@ -0,0 +1,139 @@
...
+  let artist = null as any
+  let rootValue = null as any
+
+  beforeEach(() => {
+    artist = {
+      id: "foo-bar",
+      name: "Foo Bar",
+    }
+    rootValue = {
+      artistLoader: sinon.stub().returns(Promise.resolve(artist)),
```
mzikherman: This doesn't need `sinon.stub()` (prefer Jest mocks if you need it).

Instead, just `artistLoader: () => Promise.resolve(artist)` should work.

mzikherman:
Looks great, one #minor test comment, feel free to merge when ready!

Next step: importing the data 😎

dblandin:
Nice! Here's [an example](https://metaphysics-staging.artsy.net/?query=%7B%0A%20%20artist(id%3A%20%22robert-natkin%22)%20%7B%0A%20%20%20%20insights%20%7B%0A%20%20%20%20%20%20type%0A%20%20%20%20%20%20label%0A%20%20%20%20%20%20entities%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D
) of insights returned on staging:


2018-12-20-113611_1571x905_scrot

mzikherman:
Nice! Yea, didn't realize this would work out of the gate with the existing `collections` data.

PR #1448, opened 2018-12-27
Allow finding a local discovery city nearest to an arbitrary point

anandaroop:
Implements: https://artsyproduct.atlassian.net/browse/LD-64
Follow-up to: https://github.com/artsy/emission/pull/1252#discussion_r243255869

This allows a client to find a local discovery city by latititude/longitude, as obtained e.g. via a device's geolocation API.

To do this we modify the root `city` query to allow lookup by lat/lng (in addition to the already implemented lookup by slug).

The lat/lng lookup is subject to a threshold, such that if you supply a point close enough to an existing city you will get a result:

found

But if you supply a point not close enough, you will get `null`:

null

In the Emission thread we discussed having this query return the city if an appropriate one is found, or else `null` or something representing _no city found_. I've modeled this as a simple `null` for now but am open to suggestion, if that is not idiomatic GraphQL.

### DIFF
--- a/src/lib/geospatial.ts
+++ b/src/lib/geospatial.ts
@@ -1,0 +1,24 @@
+interface LatLng {
+  lat: number
+  lng: number
+}
+
+const EARTH_RADIUS_IN_METERS = 6371e3
+
+const toRadians = (degrees: number): number => degrees * (Math.PI / 180)
+
+const haversineDistance = (point1: LatLng, point2: LatLng): number => {
+  const φ1 = toRadians(point1.lat)
+  const φ2 = toRadians(point2.lat)
+  const Δφ = toRadians(point2.lat - point1.lat)
+  const Δλ = toRadians(point2.lng - point1.lng)
+
+  const a =
+    Math.sin(Δφ / 2) * Math.sin(Δφ / 2) +
+    Math.cos(φ1) * Math.cos(φ2) * Math.sin(Δλ / 2) * Math.sin(Δλ / 2)
+  const c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a))
+
+  return EARTH_RADIUS_IN_METERS * c
+}
+
+export { haversineDistance as distance }
--- a/src/lib/geospatial.ts
+++ b/src/lib/geospatial.ts
@@ -1,12 +1,19 @@
 /**
  * Represents a latitude/longitude coordinate pair
  */
-interface LatLng {
+export interface LatLng {
   /** Latitude in decimal degrees */
   lat: number
 
   /** Longitude in decimal degrees */
   lng: number
+}
+
+/**
+ * Represents a geospatial point feature
+ */
+export interface Point {
+  coordinates: LatLng
 }
 
 const EARTH_RADIUS_IN_METERS = 6371e3 // https://en.wikipedia.org/wiki/Earth_radius#Mean_radius
--- a/src/lib/geospatial.ts
+++ b/src/lib/geospatial.ts
@@ -3,10 +3,29 @@ interface LatLng {
   lng: number
 }
 
-const EARTH_RADIUS_IN_METERS = 6371e3
+const EARTH_RADIUS_IN_METERS = 6371e3 // https://en.wikipedia.org/wiki/Earth_radius#Mean_radius
 
+/**
+ * Convert angular distance from degrees to radians
+ *
+ * @param {number} degrees - decimal degrees
+ * @returns {number} radians
+ */
 const toRadians = (degrees: number): number => degrees * (Math.PI / 180)
 
+/**
+ * Calculates the haversine (spherical) distance between two geographic points.
+ * Arguments are supplied as `LatLng` objects, with coordinates specified
+ * as `lat` and `lng` properties in decimal degrees.
+ *
+ * See:
+ * https://en.wikipedia.org/wiki/Haversine_formula
+ * https://www.movable-type.co.uk/scripts/latlong.html
+ *
+ * @param {LatLng} point1 - an object with `lat` and `lng` properties
+ * @param {LatLng} point2 - an object with `lat` and `lng` properties
+ * @returns {number} Distance between point1 and point2, in meters
+ */
 const haversineDistance = (point1: LatLng, point2: LatLng): number => {
   const φ1 = toRadians(point1.lat)
   const φ2 = toRadians(point2.lat)
--- a/src/lib/geospatial.ts
+++ b/src/lib/geospatial.ts
@@ -1,5 +1,11 @@
+/**
+ * Represents a latitude/longitude coordinate pair
+ */
 interface LatLng {
+  /** Latitude in decimal degrees */
   lat: number
+
+  /** Longitude in decimal degrees */
   lng: number
 }
 
@@ -15,16 +21,12 @@ const toRadians = (degrees: number): number => degrees * (Math.PI / 180)
 
 /**
  * Calculates the haversine (spherical) distance between two geographic points.
- * Arguments are supplied as `LatLng` objects, with coordinates specified
- * as `lat` and `lng` properties in decimal degrees.
  *
  * See:
  * https://en.wikipedia.org/wiki/Haversine_formula
  * https://www.movable-type.co.uk/scripts/latlong.html
  *
- * @param {LatLng} point1 - an object with `lat` and `lng` properties
- * @param {LatLng} point2 - an object with `lat` and `lng` properties
- * @returns {number} Distance between point1 and point2, in meters
+ * @returns Distance between point1 and point2, in meters
  */
 const haversineDistance = (point1: LatLng, point2: LatLng): number => {
   const φ1 = toRadians(point1.lat)
--- a/src/schema/city/constants.ts
+++ b/src/schema/city/constants.ts
@@ -1,1 +1,2 @@
 export const LOCAL_DISCOVERY_RADIUS_KM = 75
+export const NEAREST_CITY_THRESHOLD_KM = 150
--- a/src/schema/city/index.ts
+++ b/src/schema/city/index.ts
@@ -2,7 +2,6 @@ import {
   GraphQLBoolean,
   GraphQLInt,
   GraphQLList,
-  GraphQLNonNull,
   GraphQLObjectType,
   GraphQLString,
 } from "graphql"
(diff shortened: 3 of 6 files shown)


src/lib/geospatial.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @param {LatLng} point1 - an object with `lat` and `lng` properties
+ * @param {LatLng} point2 - an object with `lat` and `lng` properties
+ * @returns {number} Distance between point1 and point2, in meters
+ */
+const haversineDistance = (point1: LatLng, point2: LatLng): number => {
+  const φ1 = toRadians(point1.lat)
+  const φ2 = toRadians(point2.lat)
+  const Δφ = toRadians(point2.lat - point1.lat)
+  const Δλ = toRadians(point2.lng - point1.lng)
```
anandaroop: Yep, [they're legal](https://mothereff.in/js-variables#Δφ) in ES6

src/schema/city/constants.ts
```diff
@@ -1 +1,2 @@
 export const LOCAL_DISCOVERY_RADIUS_KM = 75
+export const NEAREST_CITY_THRESHOLD_KM = 150
```
anandaroop: I picked an arbitrary value here of 2x the city radius we had already defined. Both of these are tunable, of course

src/schema/city/index.ts
```diff
@@ -100,13 +104,29 @@ export const City = {
...
+      throw new Error('The "slug" and "near" arguments are mutually exclusive.')
+    }
+    if (!slug && !near) {
+      throw new Error('One of the arguments "slug" or "near" is required.')
+    }
+    if (slug) {
+      return lookupCity(slug)
+    }
+    if (near) {
+      return nearestCityOrNull(near)
```
anandaroop: Is returning a-value-or-null idiomatic?

src/schema/city/index.ts
```diff
@@ -118,3 +138,27 @@ const lookupCity = slug => {
...
+
+  if (isCloseEnough(latLng, closestCity)) {
+    return closestCity
+  }
+  return null
+}
+
+const citiesOrderedByDistance = latLng => {
+  let cities = Object.values(cityData)
+  cities.sort((a: any, b: any) => {
```
anandaroop: Tried to avoid `any`, but typing a comparator function that reaches into its args was beyond me in this PR.

src/schema/city/index.ts
```diff
@@ -100,13 +104,29 @@ export const City = {
...
   },
-  resolve: (_obj, args) => {
-    return lookupCity(args.slug)
+  resolve: (_obj, { slug, near }) => {
+    if (slug && near) {
+      throw new Error('The "slug" and "near" arguments are mutually exclusive.')
+    }
+    if (!slug && !near) {
+      throw new Error('One of the arguments "slug" or "near" is required.')
+    }
```
anandaroop: Now that we allow `slug` XOR `near` as arguments to the query, I do some manual arg-checking here, as my understanding is that there is no native GraphQL notion of mutually exclusive arguments.

src/schema/city/index.ts
```diff
@@ -17,8 +16,13 @@ import EventStatus from "schema/input_fields/event_status"
 import cityData from "./city_data.json"
 import { pageable } from "relay-cursor-paging"
 import { connectionFromArraySlice } from "graphql-relay"
-import { LOCAL_DISCOVERY_RADIUS_KM } from "./constants"
+import {
+  LOCAL_DISCOVERY_RADIUS_KM,
+  NEAREST_CITY_THRESHOLD_KM,
+} from "./constants"
 import { convertConnectionArgsToGravityArgs } from "lib/helpers"
+import Near from "schema/input_fields/near"
```
anandaroop: TIL… even though `Near` and `LatLngType` have largely the same shape, we can only use the one based on `GraphQLInputObjectType` (i.e. `Near`) as a field argument.

src/lib/geospatial.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const a =
+    Math.sin(Δφ / 2) * Math.sin(Δφ / 2) +
+    Math.cos(φ1) * Math.cos(φ2) * Math.sin(Δλ / 2) * Math.sin(Δλ / 2)
+  const c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a))
+
+  return EARTH_RADIUS_IN_METERS * c
+}
+
+export { haversineDistance as distance }
```
anandaroop: Haversine distance is the most commonly used formula, but it sacrifices some accuracy.

I figure we'll start with that, but leave ourselves room to swap out a different `distance` function under the hood if necessary.

src/lib/geospatial.ts
```diff
@@ -0,0 +1,43 @@
+interface LatLng {
+  lat: number
+  lng: number
+}
+
+const EARTH_RADIUS_IN_METERS = 6371e3 // https://en.wikipedia.org/wiki/Earth_radius#Mean_radius
```
anandaroop: You might be surprised at how [contentious](https://en.wikipedia.org/wiki/Talk:Great-circle_distance/Archive_1) this number is  😅 (_"All you have is a bunch of hand-waving arguments…"_)

Below are some util functions to support distance calculations.

src/schema/city/index.ts
alloy: Alas not.

src/schema/city/index.ts
alloy: Sure. Unless you mean _naming_ it like that? In which case, I don’t believe so.

src/schema/city/index.ts
alloy: Oops. Wrong button, didn’t mean to resolve but also don’t see an undo option (on my phone).

I wanted to say that I would expect these params to get typed based on `cities`. Isn’t that typed as `string[]`?

src/lib/geospatial.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Calculates the haversine (spherical) distance between two geographic points.
+ * Arguments are supplied as `LatLng` objects, with coordinates specified
+ * as `lat` and `lng` properties in decimal degrees.
+ *
+ * See:
+ * https://en.wikipedia.org/wiki/Haversine_formula
+ * https://www.movable-type.co.uk/scripts/latlong.html
+ *
+ * @param {LatLng} point1 - an object with `lat` and `lng` properties
+ * @param {LatLng} point2 - an object with `lat` and `lng` properties
```
alloy: Rather than documenting these objects here, I’d document the properties of the `LatLng` interface.

src/schema/city/index.ts
anandaroop: Cool, I'll just simplify the name then.

src/schema/city/index.ts
anandaroop: `cities` here is an array of objects, shaped [like this](https://github.com/artsy/metaphysics/blob/876d7346a9da828397850094b0fee10f85e5a3fc/src/schema/city/city_data.json#L3-L5).

Without the `any` I got tsc errors like this…

err

Then if I tried to devise a type that has coordinates within it (excuse the goofy naming here) I got something else…

err2

That one left me stumped.

alloy: Ah I see; ofc, I don’t know why I was thinking it would be a string. I don’t recall if `Object.values()` is able to infer the type of the objects. Is the `cityData` typed as you’d expect? If not, then TS isn’t making use of a new compiler setting that is able to infer types from JSON files.

alloy: FYI The type error you get in the second case is, I think, TS complaining about you passing a function that takes `HasCoordinates` parameters to a function that will yield `{}` (empty object).

src/lib/geospatial.ts
damassi: Symbol support is cool :) but its also a little domain specific (for those reading at a glance) vs plain words. (Also hard to figure out how to type from a standard keyboard.)

src/schema/city/index.ts
anandaroop: I think you had the right idea with type inference not working with `Object.values()` — which makes sense since the values could be totally heterogeneous.

Typing that assignment explicitly [here](https://github.com/artsy/metaphysics/pull/1448/files#diff-36aa4512b29ee50fae45b7352a1c9cd6R153) was the key to making this work. Thanks!

src/lib/geospatial.ts
alloy: I'm ok with a bit of an artful touch 🎨, especially in a pace that will likely only be worked on by people that know this domain.

src/schema/city/index.ts
```diff
@@ -100,21 +104,61 @@ export const City = {
   description: "A city-based entry point for local discovery",
   args: {
     slug: {
-      type: new GraphQLNonNull(GraphQLString),
+      type: GraphQLString,
```
alloy: Just to be sure, I'm _assuming_ that no production code was making use of this type yet and so this backwards compatibility breaking change is ok.

src/schema/city/index.ts
```diff
@@ -100,21 +104,61 @@ export const City = {
...
+
+  if (isCloseEnough(latLng, closestCity)) {
+    return closestCity
+  }
+  return null
+}
+
+const citiesOrderedByDistance = (latLng: LatLng): Point[] => {
+  let cities: Point[] = Object.values(cityData)
+  cities.sort((a: Point, b: Point) => {
```
alloy: I don’t think you _need_ to type the function. I would assume `sort` is generic on the contents of the array and thus implicitly type the comparator parameters to be of the content type.

alloy:
No of my new comments are in any way blockers, do as you think is best and merge at will. 👨‍🎤

src/schema/city/index.ts
anandaroop: That is correct. The open PR https://github.com/artsy/emission/pull/1252 is the first potential use case for this.

src/schema/city/index.ts
```diff
@@ -100,21 +104,61 @@ export const City = {
...
+    const distanceA = distance(latLng, a.coordinates)
+    const distanceB = distance(latLng, b.coordinates)
+    return distanceA - distanceB
+  })
+  return cities
+}
+
+const isCloseEnough = (latLng: LatLng, city: Point) => {
+  return distance(latLng, city.coordinates) < NEAREST_CITY_THRESHOLD_KM * 1000
+}
```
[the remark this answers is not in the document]
anandaroop: Heh, 🔍 yeah that is a remnant of inserting a console.log previously.

src/lib/geospatial.ts
damassi: Ooo didn't know about ctrl+command+space!

PR #1451, opened 2018-12-28
[Server] Downgrade Node to 10.13.0

damassi:
**Related:** https://github.com/artsy/force/pull/3235

This is to address the mysterious and periodic 504 issues we've been seeing in staging.


damassi:
Seems trivial. Going to merge and send out to staging to see if it fixes things.

PR #1452, opened 2019-01-03
Add offerable to kaws stitched schema

jonallured:
Worked with @mzikherman to figure out what was missing for https://github.com/artsy/reaction/pull/1744 and found this bit of paperwork to add.


orta:
Hah, nice, the test fail is unrelated but should be fixed from this point onwards

PR #1455, opened 2019-01-04
Add the version id to a couple places

jonallured:
This was missing and was causing the resolver to get an undefined.

### DIFF
--- a/src/schema/ecommerce/order.ts
+++ b/src/schema/ecommerce/order.ts
@@ -21,6 +21,7 @@ export const Order = {
                 id
                 priceCents
                 artworkId
+                artworkVersionId
                 editionSetId
                 quantity
                 fulfillments{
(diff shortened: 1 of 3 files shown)


ashkan18:
nice! Just to be super safe, lets make sure Gravity PR gets deployed before merging this. what do you think?

jonallured:
@ashkan18 yeah I don't feel strongly one way or the other so better to be safe! 🤗 

So the failure on CI seems to be about the ecommerce orders response - maybe there's another place in the fixture data where I need to add this?

ashkan18:
I think we may need to update https://github.com/artsy/metaphysics/blob/master/src/test/fixtures/exchange/orders.json#L65 and possibly https://github.com/artsy/metaphysics/blob/master/src/test/fixtures/results/sample_order.js#L78

jonallured:
@ashkan18 ok the Gravity change is deploying to production right now and the tests are green on this one so I think it's good to merge, what do you think?

ashkan18:
lets do this! 🙌

PR #1457, opened 2019-01-08
exposing partner/exhibitor profile ids in Fairs

ashleyjelks:
This PR adds a new default profile id field in the Fairs exhibitors_grouped_by_name field.

[See LD-115](https://artsyproduct.atlassian.net/browse/LD-115)


damassi:
👍

PR #1459, opened 2019-01-10
feat: expose fair about, links, tickets_link fields

javamonn:
Implements [LD-99](https://artsyproduct.atlassian.net/browse/LD-99)

Exposes fair `about`, `links`, and `tickets_link` fields.

### DIFF
--- a/src/schema/fair.ts
+++ b/src/schema/fair.ts
@@ -50,6 +50,9 @@ const FairType = new GraphQLObjectType({
   name: "Fair",
   fields: () => ({
     ...GravityIDFields,
+    about: {
+      type: GraphQLString,
+    },
     artists: {
       type: artistConnection,
       args: pageable({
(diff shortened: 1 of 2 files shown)


src/schema/fair.ts
```diff
@@ -230,6 +236,9 @@ const FairType = new GraphQLObjectType({
     tagline: {
       type: GraphQLString,
     },
+    tickets_link: {
```
javamonn: On a passthrough field like this, do we want to camelCase?

alloy: Yup, every new field should be camelCase.

damassi: @alloy - there's no way to rename a field with attaching a resolver, right?

orta: Nope, gotta do the `({ snake_case }) => snake_case` dance I'm afraid

PR #1463, opened 2019-01-14
Exchange Stitched Schema

ashkan18:
Deprecates https://github.com/artsy/metaphysics/pull/1320

Stitches the Exchange API under a different namespace: `commerce` instead of `ecommerce`. This means we can turn on the stitching flag in metaphysics staging and test out apps by a quick find & replace.

Differences:

- So far `buyer` and `seller` on an Order are still the exchange representation (basically `{ id: String, __typename: String }`
- _However_, we did successfully add `buyerDetails` and `sellerDetails` with the API you'd expect (a union of the MP User and MP Partner )

We're going to look at re-creating https://github.com/apollographql/graphql-tools/pull/964 with our current version of graphql-tools to remove `buyer` and `seller` on an order, and hopefully replace them with the `buyerParty` and `sellerParty`.

To recap:
- Rename `buyer` to `buyerDetails`
- Rename `seller` to `sellerDetails`
- When using fragments to check for different unions, replace `OrderOffer` with `CommerceOrderOffer` and etc.


orta:
Alright, I've shipped a bunch of tests, it's not 100% comprehensive, but provides a pattern for testing pretty much everything we want to

ashkan18:
Allright! this is ready to go, the plan would be, we merge this, then enable it on staging (or even prod) since its on a different namespace it won't impact merged schema currently used by any clients.
Then switch CMS over to stitched version and eventually move others over.

src/lib/stitching/exchange/__tests__/schema.test.ts
```diff
@@ -1,41 +1,58 @@
...
+  const exchangeTypes = await getTypesFromSchema(exchangeSchema)
+  const mutationFields = await getMutationFieldsFromSchema(exchangeSchema)
+
+  expect(exchangeTypes).toContain("OrderWithMutationSuccess")
+  expect(exchangeTypes).toContain("OrderWithMutationFailure")
+  expect(exchangeTypes).toContain("CreateOrderWithArtworkInput")
+  expect(exchangeTypes).toContain("OrderWithMutationSuccess")
+
+  // What we want
+  // expect(mutationFields).toContain("ecommerceCreateOrderWithArtwork")
```
sweir27: I don't fully understand the messaging here-- should this be uncommented if it's `What we want`?

orta: Look below, when we switch to the default namespace - lines 40/41 will fail and we can replace with 36/37

src/lib/stitching/exchange/__tests__/stitching.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    expect(orderableFields).toContain("buyerDetails")
+    expect(orderableFields).toContain("sellerDetails")
+
+    // Any field inside the CommerceBuyOrder & CommerceOfferOrder which
+    // ends in cents should have a version without cents which is a
+    // string equivalent
+    const fieldsWithCents = orderableFields.filter(f => f.endsWith("Cents"))
+    for (const field of fieldsWithCents) {
+      expect(orderableFields).toContain(field.replace("Cents", ""))
```
joeyAghion: It worries me that we're assuming whole dollar values anywhere. (Am I even understanding that correctly?)

orta: I think that ship has sailed, this is aiming to replicate the setup we have everywhere in MP ( which is using `xxYyCents` and then exposing `xxYy` as a string version where you have to pass in a currency signifier ) 

That said, this test introspects the classes from exchange to verify we do that for everything that has a cents (which happily caught one we missed) which I think is the right kind of test to write (as it means someone can add a new currency field and the stitching tests will fail)

joeyAghion: Are the `xxYy` strings meant to reflect the "formatted" values, or are they expected to be whole numbers only?

I'd love to identify any locations that would actually fail or be wrong when there are non-whole price values, since I'm sure that will be a painful bridge to cross later.

orta: Those strings are the formatted values

src/lib/stitching/exchange/__tests__/stitching.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  artworkResolver({ artworkId: "ARTWORK-ID" }, {}, {}, { mergeInfo })
+
+  expect(mergeInfo.delegateToSchema).toHaveBeenCalledWith({
+    args: { id: "ARTWORK-ID" },
+    operation: "query",
+    fieldName: "artwork",
+    ...restOfResolveArgs,
+  })
+})
```
alloy: I have a bit of a hard time reading these tests and their stated behavioural intent. Is it possible to make these actually work without spying on `delegateToSchema`? E.g. by providing a mocked exchange schema (using grapqhl-tools like we do in Reaction/Emission)? If not, then at least moving some of the boilerplate setup out of the tests might make it easier to focus on the core parts of the tests.

src/lib/stitching/exchange/__tests__/testingUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    // The order should only matter in that extension schemas come after the
+    // objects that they are expected to build upon
+    mergedSchema = mergeSchemas({
+      schemas: [localSchema, cachedSchema, extensionSchema],
+      resolvers: resolvers,
+    })
+
+    const anyMergedSchema = mergedSchema as any
+    anyMergedSchema.__allowedLegacyNames = ["__id"]
```
alloy: Just wondering out-loud if we can instead define it without the cast, like so:

```
Object.defineProperty(mergedSchema, "__allowedLegacyNames", { value: ["__id"] })
```

src/lib/stitching/exchange/schema.ts
```diff
@@ -17,12 +17,35 @@ export const executableExchangeSchema = () => {
...
+  // Rename all the root fields to be camelCased
+  new RenameRootFields(
+    (_operation, name) =>
+      `commerce${name.charAt(0).toUpperCase() + name.slice(1)}`
+  ),
+]
+
+export const legacyTransformsForExchange = [
+  // Apply a prefix to all the typenames
+  // for Legacy merged schema approach
```
alloy: Shouldn't this skip the `eigenWhitelist` ones when stitching exchange is disabled?

src/lib/stitching/exchange/stitching.ts
```diff
@@ -0,0 +1,222 @@
...
+  localSchema: GraphQLSchema,
+  exchangeSchema: GraphQLSchema
+) => {
+  type DetailsFactoryInput = { from: string; to: string }
+
+  /**
+   * This returns a resolver which can take an exchange OrderParty union
+   * and convert it into a metaphysics `OrderParty`.
+   *
+   * You pass in the field to get the details from, and and the new fieldName
```
alloy:
```suggestion
   * You pass in the field to get the details from and the new fieldName.
```

src/lib/stitching/exchange/stitching.ts
```diff
@@ -0,0 +1,222 @@
...
+   */
+  const partyUnionToDetailsFactory = ({ from, to }: DetailsFactoryInput) => {
+    // We abuse the query alias feature to make sure that all
+    // the data we need to generate the full object from
+    // is included.
+
+    // It's possible that this working around a bug.
```
alloy:
```suggestion
    // It's possible that this is working around a bug.
```

alloy: Can you elaborate on _what_ that bug might be?

src/lib/stitching/exchange/stitching.ts
```diff
@@ -0,0 +1,222 @@
...
+            __typename
+            id
+          }
+          ... on CommercePartner {
+            __typename
+            id
+          }
+        }
+      }`
+    }
```
alloy: This closure isn't using any of the in-scope variables, so might as well move its definition outside of the `partyUnionToDetailsFactory` function.

src/lib/stitching/exchange/stitching.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    return {
+      // Bit of a magic in next line, when adding fragment, it seems
+      // all second level fields are ignored, so __typename and id
+      // couldn't be added, so the hack was to alias the fragment field
+      // and that gets the current fields
```
alloy: Can you add an example of what constitutes “second level fields”?

src/lib/stitching/exchange/stitching.ts
```diff
@@ -0,0 +1,222 @@
...
+            .delegateToSchema({
+              schema: localSchema,
+              operation: "query",
+              fieldName: typename === "CommerceUser" ? "user" : "partner",
+              args: {
+                id,
+              },
+              context,
+              info,
+              transforms: (exchangeSchema as any).transforms,
```
alloy: If we're extending these objects, perhaps they should be a subclass instead and thus removing the need to cast?

src/lib/stitching/exchange/stitching.ts
```diff
@@ -0,0 +1,222 @@
...
+            })
+          },
+        },
+        artworkVersion: {
+          // Todo: in the future we may want to use the `artworkVersionId` also
```
alloy: Eh, are we not using it below?

src/lib/stitching/exchange/stitching.ts
```diff
@@ -0,0 +1,222 @@
...
+      CommerceOfferOrder: {
+        ...reduceToResolvers(
+          totalsResolvers("CommerceOfferOrder", orderTotals)
+        ),
+        buyerDetails: buyerDetailsResolver,
+        sellerDetails: sellerDetailsResolver,
+      },
+      CommerceLineItem: {
+        artwork: {
+          // Todo: in the future we may want to use the `artworkVersionId` also
```
alloy: In what way, other than the usage in the next resolver?

src/lib/stitching/exchange/stitching.ts
```diff
@@ -0,0 +1,222 @@
...
+      fromDetails: OrderParty
+      ${offerAmountFieldsSDL.join("\n")}
+    }
+  `,
+
+    // Resolvers for the above
+    resolvers: {
+      CommerceBuyOrder: {
+        // The money helper resolvers
+        ...reduceToResolvers(totalsResolvers("CommerceBuyOrder", orderTotals)),
```
alloy: Is it possible to do the `reduceToResolvers` part as part of `totalsResolvers` to reduce the noise a bit?

src/lib/stitching/lib/getTypesFromSchema.ts
```diff
@@ -52,7 +52,35 @@ export const getRootFieldsFromSchema = async (schema: GraphQLSchema) => {
...
+  const response = await graphql(schema, q, null, null)
+  // Throw an error to make it easy to read in the failing test
+  if (response.errors) {
+    throw new Error(
+      `Got errors from GQL: ${JSON.stringify(response.errors, null, "  ")}`
+    )
+  }
+
+  // Return the real data
+  return response.data!.__type.fields.map(t => t.name) as string[]
```
alloy: Remind me, why do we perform an introspection query like this over using the graphql-js API and operating on the schema object directly?

src/lib/stitching/exchange/__tests__/stitching.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  artworkResolver({ artworkId: "ARTWORK-ID" }, {}, {}, { mergeInfo })
+
+  expect(mergeInfo.delegateToSchema).toHaveBeenCalledWith({
+    args: { id: "ARTWORK-ID" },
+    fieldName: "artwork",
+    ...restOfResolveArgs,
+  })
+})
+
+it("delegates to the right schema when ", async () => {
```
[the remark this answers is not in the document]
ashkan18: @orta this test seems to be as the same one above it which adds even more suspense :) going to remove it, let me know if i missed something.

src/lib/stitching/lib/getTypesFromSchema.ts
orta: This felt simpler (+ very high level) mainly - and provided a pattern for folks wanting to do other test-y queries like this, but I'm open to using a higher level API for this also!

src/lib/stitching/exchange/schema.ts
orta: Nope, I'm reasonably sure - because they are already working right now

src/lib/stitching/exchange/stitching.ts
orta: I get ya, but I don't think it adds much value doing that given that it just moves the function that holds it to be 2 expressions

src/lib/stitching/exchange/stitching.ts
orta: 👍

src/lib/stitching/lib/getTypesFromSchema.ts
orta: That said, I think the implementations can change anytime to use those APIs and we just have a high level set of functions

ashkan18:
We addressed most of comments, this should be 👍 to go

src/integration/__tests__/integration.test.ts
```diff
@@ -0,0 +1,44 @@
+jest.mock("lib/apis/fetch", () => jest.fn())
+import fetch from "lib/apis/fetch"
+const mockFetch = fetch as jest.Mock
+
+const isOnCI = process.env.CI
+
+if (!isOnCI) {
+  // These tests launch new processes, and would change on almost every MP edit
+  // You don't want them in watch mode.
+  it("Skips integration tests in watch mode", () => {})
```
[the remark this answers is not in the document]
alloy: > sub-processes during watch for example have different args

Ah, right, makes sense.

alloy:
Works for me, let's pop that bottle! 🍾

PR #1464, opened 2019-01-15
Paginate ecommerce orders query (GALL-862)

iskounen:
This PR partially addresses [SELL-862](https://artsyproduct.atlassian.net/browse/SELL-862) by including a cursor to be returned with each order, which can be used to paginate the results. It also passes the `first` and `after` parameters to the resolver so that they are passed on to Exchange's GraphQL API.

~This is a WIP because we still need to write a test that confirms the `before`, `after`, `first` and `last` parameters are sent along to Exchange.~


src/schema/ecommerce/orders.ts
```diff
@@ -84,6 +97,10 @@ export const Orders = {
       mode,
       state,
       sort,
+      first,
+      after,
+      last,
+      before,
     }).then(extractEcommerceResponse("ecommerceOrders"))
```
orta: 👍  makes sense

src/schema/ecommerce/orders.ts
```diff
@@ -84,6 +97,10 @@ export const Orders = {
       mode,
       state,
       sort,
+      first,
+      after,
+      last,
+      before,
```
erikdstock: We talked about trying to be clever with something like `Object.keys( pageable({}) )` but that seemed like overkill.

jonallured:
Woo! This is looking great - my hunch is that tests are failing because fixture data needs to be updated, but I'm certainly no expert here. Looking forward to this landing!!

src/schema/ecommerce/orders.ts
```diff
@@ -54,8 +66,10 @@ export const Orders = {
           last: $last
         ) {
           ${PageInfo}
```
ashkan18: we should move this `totalPages` in `PageInfo` in case we wanted to add `PageInfo` later somewhere.

src/schema/ecommerce/orders.ts
```diff
@@ -73,6 +87,28 @@ export const Orders = {
               }
             }
           }
+          pageCursors {
```
ashkan18: i think this can all go into `PageInfo` too so we avoid duplicating it in other places

iskounen: `PageInfo` is also used by the line items, so I decided to put `totalPages` and `pageCursors` in a separate query helper called `PaginationFields`.

PR #1465, opened 2019-01-16
add is_benefit property to auction

oxaudo:
This will be used to accomplish: https://artsyproduct.atlassian.net/browse/DISCO-456 on new artwork page.


orta:
Nice and easy 👍

PR #1466, opened 2019-01-16
Ensures that the images on a carousel are the main image for an artwork

orta:
This helps make AS-285 a bit less painful, but *does not* fix it.

Basically we'd take the first image for an artwork/show, but that could have real bad results 

henri_matisse_reproduction

Probably because first was "last", this moves use to find the default image and use that instead. 

screen shot 2019-01-16 at 5 17 14 pm


damassi:
👍

PR #1469, opened 2019-01-18
First pass resolver data batching

zephraph:
Fixes [PLATFORM-724](https://artsyproduct.atlassian.net/browse/PLATFORM-724)

This is a first pass implementation at data loader batching. I still consider this a WIP, but the implementation is _mostly_ there. Still need to write tests, cover cases which will probably fail.

Thanks to @ds300 for pairing with me on this.  

<details>
  <summary>Test query</summary>

```graphql
query ArtworkFilterArtworkGridQuery(
  $filteredArtworksNodeID: ID!
  $first: Int!
  $after: String
) {
  filtered_artworks: node(__id: $filteredArtworksNodeID) {
    __typename
    ...ArtworkFilterArtworkGrid_filtered_artworks_2HEEH6
    __id
  }
}

fragment ArtworkFilterArtworkGrid_filtered_artworks_2HEEH6 on FilterArtworks {
  __id
  artworks: artworks_connection(first: $first, after: $after) {
    pageInfo {
      hasNextPage
      endCursor
    }
    pageCursors {
      ...Pagination_pageCursors
    }
    ...ArtworkGrid_artworks
    edges {
      node {
        __id
      }
    }
  }
}

fragment Pagination_pageCursors on PageCursors {
  around {
    cursor
    page
    isCurrent
  }
  first {
    cursor
    page
    isCurrent
  }
  last {
    cursor
    page
    isCurrent
  }
  previous {
    cursor
    page
  }
}

fragment ArtworkGrid_artworks on ArtworkConnection {
  edges {
    node {
      __id
      image {
        aspect_ratio
      }
      ...GridItem_artwork
    }
  }
}

fragment GridItem_artwork on Artwork {
  _id
  image {
    placeholder
    url(version: "large")
    aspect_ratio
  }
  is_biddable
  sale {
    is_preview
    __id
  }
  is_acquireable
  is_offerable
  href
  ...Metadata_artwork
  ...Save_artwork
  __id
}

fragment Metadata_artwork on Artwork {
  ...Details_artwork
  ...Contact_artwork
  href
  __id
}

fragment Save_artwork on Artwork {
  __id
  _id
  id
  is_saved
}

fragment Details_artwork on Artwork {
  href
  title
  date
  sale_message
  cultural_maker
  artists(shallow: true) {
    __id
    href
    name
  }
  collecting_institution
  partner(shallow: true) {
    name
    href
    __id
  }
  sale {
    is_auction
    is_live_open
    is_open
    is_closed
    display_timely_at
    __id
  }
  sale_artwork {
    highest_bid {
      display
      __id: id
    }
    opening_bid {
      display
    }
    __id
  }
  __id
}

fragment Contact_artwork on Artwork {
  _id
  href
  is_inquireable
  sale {
    is_auction
    is_live_open
    is_open
    is_closed
    __id
  }
  partner(shallow: true) {
    type
    __id
  }
  sale_artwork {
    highest_bid {
      display
      __id: id
    }
    opening_bid {
      display
    }
    counts {
      bidder_positions
    }
    __id
  }
  __id
}
```
</details>

<details>
  <summary>Test arguments</summary>
```json
{
  "after": "YXJyYXljb25uZWN0aW9uOjE0Mw==",
  "filteredArtworksNodeID": "RmlsdGVyQXJ0d29ya3M6eyJhZ2dyZWdhdGlvbnMiOlsidG90YWwiXSwibWFqb3JfcGVyaW9kcyI6W10sInNpemUiOjAsInNvcnQiOiItZGVjYXllZF9tZXJjaCIsImFydGlzdF9pZCI6Im1heC1lcm5zdCJ9",
  "first": 24
}
```
</details>

--- 

### TODO:
- [x] Write tests
- [x] Ensure a single result to the batch loader works (i.e. call the single id endpoint)
- [x] Determine why three sales from the batched results don't show up
- [x] Allow enabling/disabling batching on an environment variable 
- [x] Set a top limit of 20 items per batched fetch


mzikherman:
Is there any before/after on the queries that were consolidated? Wondering (due to MP caching), if a bunch of requests for individual sales would resolve to a client just as fast as one larger batched call. Of course the latter is much better for backends/cache refreshes. Musing about whether we should expect client-side/front-end wins, better backend performance, mixed, too early to tell, etc.

zephraph:
Thanks to @orta for reminding me to check pagination. Fixing that makes both calls the same.

zephraph:
@mzikherman here's a diff of the results before/after

https://www.diffchecker.com/x5pPEZD3

Left side is w/ batching, right side without. Note that I'm showing these results with a caching miss on the left so you can see what worst case performance would be there.

zephraph:
cached performance looks like this...

```json
{
"requests": {
  "filter/artworks?aggregations%5B%5D=total&artist_id=max-ernst&offset=144&page=7&size=24&sort=-decayed_merch": {
    "time": "0s 7.999ms",
    "cache": true
   },
   "sales?id%5B%5D%5B%5D=592dabef9c18db274a64e96c&id%5B%5D%5B%5D=5942ebbf139b2103c3eeafe2&id%5B%5D%5B%5D=59e770099c18db3c63bc930a&id%5B%5D%5B%5D=5942ebbf139b2103c3eeafe2&id%5B%5D%5B%5D=5ad60d8e8b3b81499b31524c&id%5B%5D%5B%5D=59e50b0e7622dd4d8161d450&id%5B%5D%5B%5D=59ddfc0b9c18db2f5132277b&id%5B%5D%5B%5D=5b759081f25b810026ce5ef2&id%5B%5D%5B%5D=58ee99ff2a893a39aecf634f&id%5B%5D%5B%5D=5a0f4341275b2437589e126e&id%5B%5D%5B%5D=5b9ab8f52835b4633b2fd4d8&id%5B%5D%5B%5D=5ac6e107275b24049a26ad82&id%5B%5D%5B%5D=5b9ab8f52835b4633b2fd4d8&id%5B%5D%5B%5D=59fcc129139b217019abfe60&id%5B%5D%5B%5D=59ddfc0b9c18db2f5132277b&is_auction=true&live=true&size=15": {
     "time": "0s 23.274ms",
     "cache": true
   },
   "sales?id%5B%5D=592dabef9c18db274a64e96c&id%5B%5D=5942ebbf139b2103c3eeafe2&id%5B%5D=5ad60d8e8b3b81499b31524c&id%5B%5D=59e50b0e7622dd4d8161d450&id%5B%5D=5b759081f25b810026ce5ef2&id%5B%5D=58ee99ff2a893a39aecf634f&id%5B%5D=5a0f4341275b2437589e126e&id%5B%5D=5b9ab8f52835b4633b2fd4d8&id%5B%5D=5ac6e107275b24049a26ad82&id%5B%5D=59fcc129139b217019abfe60&id%5B%5D=59ddfc0b9c18db2f5132277b&id%5B%5D=59e770099c18db3c63bc930a&size=12": {
     "time": "0s 26.815ms",
     "cache": true
    }
  }
}
```

joeyAghion:
Good point. If the multipleLoader by default split keys into batches of max 20 items, that would get my 👍.

src/lib/loaders/loaders_without_authentication/gravity.ts
```diff
@@ -2,12 +2,63 @@
...
   const { gravityLoaderWithoutAuthenticationFactory } = factories(opts)
   const gravityLoader = gravityLoaderWithoutAuthenticationFactory
   const gravityUncachedLoader = uncachedLoaderFactory(gravity, "gravity")
 
+  const renderParams = (key) => {
+    if (typeof key === "string") {
+      return ""
+    }
+    const { id, ...params} = key;
+    return Object.entries(params).map(entry => entry.join('=')).sort().join('&')
```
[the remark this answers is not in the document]
zephraph: It only is used to determined the uniqueness of the parameters passed with the object. So for example, sales could have `{ is_live: true, is_auction: true }` with any variation of values for that. We _only_ want to batch the results that have the same parameters, else we'll get weird things. The actual params are built by the `gravityLoader` as they were before.

src/lib/loaders/loaders_without_authentication/gravity.ts
```diff
@@ -2,12 +2,63 @@
...
+        size: keys.length
+      }
+    })
+
+    return Promise.all(
+      groupedKeys
+      .map(keys => {
+        console.log(keys.id)
+      if (keys.id.length === 1) {
+        return singleLoader(keys)
```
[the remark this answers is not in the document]
zephraph: Yeah, for sure. There's a gap here. I'm not sure that this case even works as is. I need to refine this part.

src/lib/loaders/loaders_without_authentication/gravity.ts
```diff
@@ -58,8 +109,10 @@ export default opts => {
     saleArtworksFilterLoader: gravityLoader("filter/sale_artworks"),
     saleArtworksLoader: gravityLoader(id => `sale/${id}/sale_artworks`, {}, { headers: true }),
     saleArtworkLoader: gravityUncachedLoader(({ saleId, saleArtworkId }) => `sale/${saleId}/sale_artwork/${saleArtworkId}`, null),
-    saleLoader: gravityLoader(id => `sale/${id}`),
-    salesLoader: gravityLoader("sales"),
+    // saleLoader: gravityLoader(id => `sale/${id}`),
+    // salesLoader: gravityLoader("sales"),
+    saleLoader: batchSaleLoader,
+    salesLoader: batchSalesLoader,
```
alloy: :ceiling-cat: (landed here due to alert I have for these, which I will disable henceforth) I wonder if the decision to batch shouldn’t be left up to the loader, rather than the call-site? I.e. rather than having a singular and a plural loader, only have one and leave the details up to the loader.

zephraph: 👋 thanks for dropping by. You should definitely not be on here friend 😝.

I'd prefer we avoid opt-in optimizations. I really think batching should be transparent to the resolver, in the same way that caching is. If it's absolutely necessary we can have a method to opt out of batching from the call site, but that's as far as I would want to go with that. Opt-in may seem to be the more incremental way to approach the implementation, but it puts the burden of carrying that implementation forward on everyone who adds or edits a field in the future.

zephraph: 👋 thanks for dropping by. You should definitely not be on here friend 😝.

I'd prefer we avoid opt-in optimizations. I really think batching should be transparent to the resolver, in the same way that caching is. If it's absolutely necessary we can have a method to opt out of batching from the call site, but that's as far as I would want to go with that. Opt-in may seem to be the more incremental way to approach the implementation, but it puts the burden of carrying that implementation forward on everyone who adds or edits a field in the future.

zephraph:
These types are basically all lies. Sweet, sweet lies.

src/lib/loaders/batchLoader.ts
```diff
@@ -0,0 +1,109 @@
...
+  multipleLoader: any
+  defaultResult?: any
+}
+
+export const batchLoader = ({
+  singleLoader,
+  multipleLoader,
+  defaultResult = null,
+}: BatchLoaderArgs) => {
+  if (!ENABLE_RESOLVER_BATCHING) {
```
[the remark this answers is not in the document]
zephraph: That's a fair point. I'll poke into this.

src/lib/loaders/loaders_without_authentication/gravity.ts
alloy: Yeah, I meant that I think we only really need the singular saleLoader. Call-sites don’t need to know if there’s a plural endpoint or not, that seems like a detail for the loader to be concerned about.

Absolutely no blocker on this PR, though!

alloy:
@zephraph What are the type lies?

zephraph:
It's pretty loose all the way down. Loads of anys, forced type coercions, etc. 

On another note, I'm having some issues w/ the data loader tests. The results aren't being batched in the test, but they are being batched when running queries against staging data... not sure if it's due to how jest handles async stuff or if there's potentially a different issue.

src/lib/loaders/loaders_without_authentication/gravity.ts
zephraph: We seem to have distinctions between single resource endpoints and list resource endpoints. For the `/sales` endpoint you can do filtering such as `is_auction` that's not available on the single resource endpoint. Also, I've been told that other endpoints like this might actually only return a subset of data for the list endpoint. It's something we'll have to keep an eye out for.

alloy: I see :/

src/lib/loaders/__tests__/batchLoader.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    await batch(params)
+    expect(multipleLoader).toBeCalledWith(params)
+  })
+
+  it("should group multiple calls together", async () => {
+    const multipleLoader = jest.fn(v => Promise.resolve(v))
+    const batch = batchLoader({ multipleLoader })
+
+    await Promise.all([batch("123"), batch("456"), batch("789")])
```
alloy: DataLoader performs its work on the end of the runloop, so I bet you need to do something like:

```
expect.hasAssertions()
const batches = Promise.all([batch("123"), batch("456"), batch("789")])
setImmediate(() => batches.then(() => {
  expect(multipleLoader).toBeCalledTimes(1)
}))
```

alloy: And otherwise take a look at the test-suite of DataLoader to see how they test them.

alloy: Just saw your updated comment 😸

izakp:
@zephraph so to create a blue/green type deployment setup see https://github.com/artsy/metaphysics/pull/1197

src/lib/loaders/batchLoader.ts
```diff
@@ -56,7 +61,7 @@ export const batchLoader = ({
   multipleLoader,
   defaultResult = null,
 }: BatchLoaderArgs) => {
-  if (!ENABLE_RESOLVER_BATCHING) {
+  if (ENABLE_RESOLVER_BATCHING !== "true") {
```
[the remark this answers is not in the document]
zephraph: I'll do that, but it's not done consistently among other flags.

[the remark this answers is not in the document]
alloy: Once you’ve seen it you gots to patch it! There’s no unseesies ;)

src/lib/loaders/batchLoader.ts
```diff
@@ -0,0 +1,111 @@
...
+        if (keys.id.length === 1 && singleLoader) {
+          return singleLoader(keys.id[0])
+        } else {
+          return multipleLoader(keys)
+        }
+      })
+    ).then(data => {
+      const normalizedResults = data.map((queriedGroup, groupIndex) => {
+        return groupedKeys[groupIndex].id.map(
+          id => queriedGroup.find(r => r._id === id) || defaultResult
```
zephraph: @joeyAghion I don't think it's obvious what's happening here, but I probably should explain it so that you can correct my assumptions if they're invalid. 

Let's say a query makes a request for three sale ids `sale-1, sale-2, sale-3` which results in a call like `/sale?id[]=sale-1&id[]=sale-2&id[]=sale-3`. The results will be an array of sales `[{ _id: sale-1, ... }, ...]`. It's important to note here that the sale id that we pass to the initial dataloader maps over to `_id` in the response object from gravity. At this point in the code, we're actually grouping and reordering results. So I'm taking the initial input ids (that the dataloader batched together) and ordering the results from gravity to match that sequence. 

So `['sale-1', 'sale-2', ...]` -> `{ _id: 'sale-1', ...}, { _id: 'sale-2', ...}, ...`. This is important because sometimes some (or all) of the calls to gravity might result in no data... but the dataloader expects to get a list of results the same length as the input (and in the same order). So the `|| defaultResult` part uses whatever we consider the fallback value for an entry in this list. In most cases that's just something like `null`. 

There could be a corner case here where an input is considered non-nullable or has some other default value... that should likely just be handled in the resolver though. 

Let me know if that makes any sense at all and if you have more questions.

src/lib/loaders/batchLoader.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (!ENABLE_RESOLVER_BATCHING) {
+    return singleLoader ? singleLoader : multipleLoader
+  }
+  const dl = new DataLoader(keys => {
+    let groupedKeys = groupKeys(keys as any)
+
+    return Promise.all(
+      groupedKeys.map(keys => {
+        if (keys.id.length === 1 && singleLoader) {
+          return singleLoader(keys.id[0])
```
zephraph: The thing that gets passed into the single loader is always just the string id. That seems to hold true for the sale endpoint (there aren't other params that I've seen), but that might not hold true for every endpoint.

src/lib/loaders/__tests__/batchLoader.test.ts
```diff
@@ -0,0 +1,179 @@
...
+      await batch("foo")
+      expect(multipleLoader).not.toBeCalled()
+      expect(singleLoader).toBeCalledWith("foo")
+    })
+
+    /**
+     * A list endpoint (like sales) shouldn't provide a singleLoader because the
+     * multiple loader could support filtering options not supported by the single
+     * endpoint.
+     */
```
damassi: ty

src/lib/loaders/batchLoader.ts
```diff
@@ -0,0 +1,111 @@
...
+        return { id: keys, size: keys.length }
+      }
+      return {
+        // @ts-ignore
+        ...keys[0],
+        id: keys.map(k => k.id),
+        size: keys.length,
+      }
+    })
+    .value() as unknown) as GroupKeysResult[]
```
damassi: Tiny nit, this would be a little more introspectable if it has an explicit return vs an illicit return (ignoring `tap`, which folks probably shouldn't be expected to know about)

src/lib/loaders/batchLoader.ts
damassi: Might be worth adding a comment?

joeyAghion: Most Gravity APIs work when passed a true database identifier _or_ a "slug" (e.g., `/sale/1234` or `/sale/christies-sale`) and I believe many clients use a mix of these. In the specific case you're working on, I'd guess that we're dealing with real ids because they were probably sourced from artworks' `sale_ids` field. These end up as the `_id` property in response data, while the "slugs" appear in the `id` property.

Another consideration, if/when we do want to batch slug-based requests, is that many models support a whole "history" of slugs. E.g., you can request artist `frank-smith` and get back an object with `_id: 1234, id: "franklin-smith"` (i.e., the requested id appears nowhere in the response). For that reason, maybe we should just punt on supporting slugs in batching for now. If we care to support it at a later point, we might need to update APIs to respect the order of ids provided in params.

izakp:
@zephraph lmk if you want to pair on some load tests - here's an old ticket re: testing metaphysics with the cache disabled for 10% of traffic

PR #1470, opened 2019-01-22
expose gallery website urls in Shows field

ashleyjelks:
This PR adds a partner url field  in the Shows schema to expose the web address galleries have directly uploaded to CMS.

[Links to LD-148](https://artsyproduct.atlassian.net/browse/LD-148)
[Also links to LD-129](https://artsyproduct.atlassian.net/browse/LD-129)

### DIFF
--- a/src/schema/__tests__/show.test.js
+++ b/src/schema/__tests__/show.test.js
@@ -18,7 +18,6 @@ describe("Show type", () => {
       partner: {
         id: "new-museum",
       },
-      partner_url: "https://www.newmuseum.org/",
       display_on_partner_profile: true,
       eligible_artworks_count: 8,
       is_reference: true,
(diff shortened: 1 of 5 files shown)


anandaroop:
Thanks for jumping on this 👏 .

Couple of thoughts:

1. It would be more in the spirit of GraphQL to add the new `url` field to the `partner` itself, rather than as a `partner_url` on show. That would satisfy _any_ use case where one needs the partner's website, rather than the specific case of going from show &rarr; partner website.

    We have this documented as a best practice [here](https://github.com/artsy/README/blob/master/playbooks/graphql-schema-design.md#how-to-model-our-graph) (see 2nd bullet) — I'd recommend checking that doc out.

2. Let's call this new field `website` instead of `url` to be more consistent with how we generally name external websites.

src/schema/__tests__/show.test.js
```diff
@@ -46,11 +46,14 @@ describe("Show type", () => {
 
     rootValue = {
       showLoader: sinon.stub().returns(Promise.resolve(showData)),
-      showsWithHeadersLoader: sinon
```
mbilokonsky: There seem like there are a couple of formatting-only changes here - this might be fine if it's intentional, but worth checking to see if you're applying prettier rule that are overriding project rules. I recently had a big PR that totally reformatted a file in the same vein, had to rewrite that file after fixing my formatting rules?

src/schema/partner.ts
```diff
@@ -278,6 +278,15 @@ const PartnerType = new GraphQLObjectType({
           return exceptions[type] || type
         },
       },
+      website: {
+        description: "The gallery partner's web address",
+        type: GraphQLString,
+        resolve: root => {
+          if (root.website) {
+            return root.website || ""
```
anandaroop: I don't think the `|| ""`  guard is strictly needed here? (The field is not declared to be non-nullable.)

In that case you can do away with the resolver completely, since the [default field resolver](https://github.com/artsy/metaphysics/pull/1362#discussion_r229079660) will simply expose the same-named field on the object in question.

src/schema/__tests__/show.test.js
ashleyjelks: Oooh good catch, let me check my Prettier settings!

src/schema/partner.ts
ashleyjelks: Yep, just caught that myself...just made the change but looks like CI tests are still running.

PR #1471, opened 2019-01-23
Remove deprecated fields from “Order”

pepopowitz:
*Partially* resolves https://artsyproduct.atlassian.net/browse/PURCHASE-796. 

Removes the deprecated `lastOffer` and `offers` fields from the `Order` type. (They exist on `OrderOffer` instead.)

The following clients have been checked for use of these fields:

* Eigen: does not use them
* Force: does not use them
* Reaction: Was using `lastOffer`. [Reaction PR #1929](https://github.com/artsy/reaction/pull/1929) has been merged ~, but needs to be released to production before this PR can be merged~ [and deployed!!!](https://github.com/artsy/force/pull/3401).
* Volt: Was using `offers`. [Volt PR #3409](https://github.com/artsy/volt/pull/3409) ~needs to be merged & released to production before this PR can be merged~ [has been merged and deployed!!!](https://github.com/artsy/volt/pull/3414).

Once this PR makes it to production, [a followup PR](https://github.com/artsy/exchange/pull/381) that deprecates the fields in Exchange can be merged.

### DIFF
--- a/src/schema/ecommerce/types/order.ts
+++ b/src/schema/ecommerce/types/order.ts
@@ -139,14 +139,6 @@ const orderFields = {
     type: GraphQLString,
     description: "Buyer phone number",
   },
-  lastOffer: {
-    type: OfferType,
-    description: "Latest offer",
-  },
-  offers: {
-    type: OfferConnection,
-    description: "List of submitted offers made on this order so far",
-  },
   offerTotalCents: {
     type: GraphQLInt,
     description: "Total amount of latest offer",
--- a/src/schema/ecommerce/types/order.ts
+++ b/src/schema/ecommerce/types/order.ts
@@ -139,13 +139,6 @@ const orderFields = {
     type: GraphQLString,
     description: "Buyer phone number",
   },
-  offerTotalCents: {
-    type: GraphQLInt,
-    description: "Total amount of latest offer",
-    deprecationReason: "Switch to ItemTotalCents",
-    resolve: ({ itemsTotalCents }) => itemsTotalCents,
-  },
-  offerTotal: amount(({ itemsTotalCents }) => itemsTotalCents),
 }
 
 export const OrderInterface = new GraphQLInterfaceType({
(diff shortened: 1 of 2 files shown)


src/schema/ecommerce/types/order.ts
```diff
@@ -139,14 +139,6 @@ const orderFields = {
...
   },
-  lastOffer: {
-    type: OfferType,
-    description: "Latest offer",
-  },
-  offers: {
-    type: OfferConnection,
-    description: "List of submitted offers made on this order so far",
-  },
   offerTotalCents: {
```
ashkan18: Not related to this PR but this is another field we can deprecate, not sure about the usage in other systems but i think it's not used anymore

pepopowitz: Good eye! 

I'll do a check of client usage again - if it's not used anywhere, I'll remove that field in this PR.

If it is used, and we need to clean those up first...we'll talk about it in slack.

ashkan18: awesome! thanks! sorry i should have founded that one earlier

pepopowitz: I don't see it used anywhere. In fact, it's been [removed from Exchange already](https://github.com/artsy/exchange/commit/d9670c7), so it wouldn't be working if anyone was using it 😸 

I'll remove it and include it in this PR, and update JIRA to reflect the addition of that field.

pepopowitz: I didn’t catch this at first, but it looks like there is an `offerTotal` in addition to `offerTotalCents`. I’m not seeing any usage of that one, either - can it be removed?

pepopowitz: I removed `offerTotal` and `offerTotalCents`. Let me know if I got it wrong!

ashkan18: awesome! this makes sense!  `OfferTotalCents` is the field coming directly from Exchange, `OfferTotal` is the added field in MP which gets some arguments like `symbol` and etc. and returns the display string of that amount in cents. we should be safe, we've already switched it to use `itemsTotalCents` before https://github.com/artsy/metaphysics/blob/master/src/schema/ecommerce/types/order.ts#L154

ashkan18:
Thanks @pepopowitz for detective work 🙌

PR #1472, opened 2019-01-25
set memory requests equal to limits in production to avoid evictions - set CPU requests to 1 core and deprioritize scheduling on API nodes

izakp:
Since we're scheduling concurrently with Causality and Gravity now, I would like that we don't get into the position of nodes evicting pods - so request as much memory as the limit.

Also set CPU requests to 1 core to avoid overscheduling on any given node.  Seeing high CPU consumption and scheduling of up to 4 metaphysics pods on a given `tier=api` node.  Setting this request higher will avoid over-allocation.
```
$ kubectl top pods --selector app=metaphysics
NAME                              CPU(cores)   MEMORY(bytes)
metaphysics-web-c45f9fb4c-4ffcn   853m         1203Mi
metaphysics-web-c45f9fb4c-7hr5g   889m         1143Mi
metaphysics-web-c45f9fb4c-7j9mr   1173m        844Mi
metaphysics-web-c45f9fb4c-cpft6   802m         1219Mi
metaphysics-web-c45f9fb4c-dnh78   759m         690Mi
metaphysics-web-c45f9fb4c-lxrpn   464m         1195Mi
metaphysics-web-c45f9fb4c-qksv6   1209m        1251Mi
metaphysics-web-c45f9fb4c-qrcnl   1150m        1349Mi
metaphysics-web-c45f9fb4c-vstnl   1016m        888Mi
metaphysics-web-c45f9fb4c-wvk5z   1164m        1448Mi
metaphysics-web-c45f9fb4c-xlrbd   922m         1373Mi
```

Removed the preference to schedule on `tier=api` nodes to also avoid over-allocation.

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -38,7 +38,7 @@ spec:
         - containerPort: 3000
         resources:
           requests:
-            cpu: 500m
+            cpu: '1'
             memory: 1536Mi
           limits:
             memory: 1536Mi
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -66,14 +66,6 @@ spec:
                 values:
                 - api
                 - foreground
-          preferredDuringSchedulingIgnoredDuringExecution:
-          - weight: 100
-            preference:
-              matchExpressions:
-              - key: tier
-                operator: In
-                values:
-                - api
 
 ---
 apiVersion: autoscaling/v1
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -39,7 +39,7 @@ spec:
         resources:
           requests:
             cpu: 500m
-            memory: 1Gi
+            memory: 1536Mi
           limits:
             memory: 1536Mi
         readinessProbe:


izakp:
@dleve123 

> The YAML code looks fine, but I don't really understand the mechanism here...

Specifically there are three mechanisms at play here:
1) Allocation of CPU and memory during container scheduling (expressed by the `requests` key in the Yaml configuration
2) Behavior imposed by Docker / cgroups when CPU and memory of a given container grow out-of-bounds (expressed by the `limits` key in Yaml configuration)
3) Pod Evictions when nodes detect they are running out of total allocatable memory

1) Requests:
When Kubernetes schedules pods it determines if a given node has enough free CPU / memory to satisfy the requests specified in the Pod's template (inherited from the deployment pod spec in most all of our cases).  Once it can place a Pod, the Pod's [lifecycle state](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) moves from `Pending` to `ContainerCreating` and the Pods network is created, images are pulled, container started.  At this point, the CPU / resource limits are set via Docker (i.e. the container runtime).  Further reading: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/

2) Limits
If requests limits are set for a container's runtime, Docker will throttle CPU usage / restart a container when it exceeds memory thresholds with an `OOMKilled` error message.  Further reading: https://medium.com/@betz.mark/understanding-resource-limits-in-kubernetes-cpu-time-9eff74d3161b

The interaction of requests and limits are best summarized by the idea of a "Quality of Service" class.  In the case that requests are set lower than limits, a Pod is considered to be "burstable" - i.e. it can consume more CPU and Memory than requested.  If Requests and limits are equal, the QoS class is considered to be "Guaranteed" meaning it has pre-allocated all CPU and memory available to it.

Running burstable Pods means we can over-allocate CPU and memory and use available resource more efficiently, but at the cost of over-allocation, which is what we have been doing with Gravity and Metaphysics.  This moves them closer to a Guaranteed QoS class in terms of memory allocation.

3) Eviction
When kubernetes nodes detect they are running out of CPU and memory, the scheduler will not schedule any further Pods if these requests cannot be met.  But nodes are also can evict pods themselves if they find they are running close to being out of total memory, as a failsafe procedure.  See https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/

In this way, using Guaranteed QoS classes for our API nodes and the services running on them - Gravity, Metaphysics and Causality ensure that nodes won't overallocate their available memory.

dleve123:
@izakp Thank you! Makes a lot of sense. Overallocation leads to more node resource usage fluctuation, which leads to scheduling dynamics, namely eviction, which is generally unfavorable. Very interesting stuff!

izakp:
Updated to also increase CPU requests as I am seeing high CPU consumption

PR #1473, opened 2019-01-25
Fix eslint

mbilokonsky:
Paired with @orta on this in an attempt to debug a pesky dependency cycle. We didn't isolate the issue, but we did finally get ESLint fully integrated with typescript (including getting ESLint errors in VSCode) so that's pretty cool! :)


orta:
This is a good step forwards (and we're at 0 ESLint warnings too)

Not gonna add eslint to the Omakase pack until we have a few more projects using it though, but it is the future ™️

.eslintrc.json
```diff
@@ -6,9 +6,9 @@
     "jasmine": true,
     "es6": true
   },
-  "plugins": ["promise"],
+  "plugins": ["promise", "@typescript-eslint/eslint-plugin"],
```
damassi: nice, so this is now a go

orta: Yeah, this is the official, ran my MS, one too

PR #1474, opened 2019-01-28
Update 'exchange.graphql' with latest schema from Exchange

pepopowitz:
This is the final step (as far as I'm aware) in removing some fields from Exchange, addressing https://artsyproduct.atlassian.net/browse/PURCHASE-796. It updates the copy of the Exchange schema in metaphysics.

There are more changes in here than just mine. I *think* this schema is in metaphysics for reference only, and that whatever I got from Exchange is what we want in MP. If I'm wrong, and I need to track down the owners of the changes, let me know! 

The changes other than my own are:

* Addition of `order` and `shippingTotalCents` to type LineItem
* Addition of `pageCursors`, `totalCount`, and `totalPages` to type OrderConnectionWithTotalCount
* Addition of types PageCursor & PageCursors
* Addition of `competingOrders` and `lineItems` to type Query

### DIFF
--- a/src/data/exchange.graphql
+++ b/src/data/exchange.graphql
@@ -60,7 +60,6 @@ type BuyOrder implements Order {
   # Item total in cents, for Offer Orders this field reflects current offer
   itemsTotalCents: Int
   lastApprovedAt: DateTime
-  lastOffer: Offer @deprecated(reason: "Switch to OfferOrder lastOffer")
   lastSubmittedAt: DateTime
   lineItems(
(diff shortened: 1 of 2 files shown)


ashkan18:
@pepopowitz looks like these changes were also part of https://github.com/artsy/metaphysics/pull/1464/files, should we close this?

pepopowitz:
> @pepopowitz looks like these changes were also part of https://github.com/artsy/metaphysics/pull/1464/files, should we close this?

Yes, closing!

PR #1480, opened 2019-02-01
[package] Update dd-trace client to latest.

alloy:
Part of https://artsyproduct.atlassian.net/browse/PLATFORM-1129

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "compression": "^1.7.2",
     "cors": "^2.7.1",
     "dataloader": "^1.3.0",
-    "dd-trace": "artsy/dd-trace-js#artsy",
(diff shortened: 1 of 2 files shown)


joeyAghion:
This was associated with failures upon deploy. Metaphysics pods would appear to start, but then crash or get restarted as a result of memory constraints. It was reverted in https://github.com/artsy/metaphysics/pull/1486. I'm not sure if there are _problems_ or we simply must expand the memory constraints to accommodate the new overhead.

References: [incident](https://artsy.slack.com/archives/C9RK0BLEP/p1549394685097400),  [follow-up discussion](https://artsy.slack.com/conversation/CA8SANW3W/p1549396045087400)

PR #1482, opened 2019-02-01
Removes GraphQL syntax errors from sentry

orta:
I saw [this in](https://sentry.io/artsynet/metaphysics-production/issues/868039286/?environment=production&referrer=alert_email) my email inbox and got annoyed

screen shot 2019-02-01 at 3 00 04 pm

Now Metaphysics won't send GraphQL syntax errors to Sentry.

### DIFF
--- a/src/lib/__tests__/graphqlErrorHandler.test.ts
+++ b/src/lib/__tests__/graphqlErrorHandler.test.ts
@@ -1,12 +1,39 @@
 import {
   formattedGraphQLError,
   shouldReportParentError,
+  shouldReportError,
 } from "../graphqlErrorHandler"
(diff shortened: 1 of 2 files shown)


joeyAghion:
What HTTP response code does this return? (I know graphql responses don't align with those generally, but if the query has invalid syntax that sounds like an appropriate place for a 4xx response.)

Also, does this also address the `Syntax Error: Expected {, found Name "query"` errors I occasionally notice?

alloy:
> if the query has invalid syntax that sounds like an appropriate place for a 4xx response

Agreed

orta:
Yep, it's a 400 - I think this is handled by our underlying library (this change only will stop that going off to sentry)

screen shot 2019-02-04 at 2 46 17 pm

PR #1483, opened 2019-02-01
[WIP] Add types for all resolvers by default

orta:
This might be new TypeScript things for people, so I'll be verbose in the inline comments below.

This moves all the resolvers in a `GraphQLObject` to be typed with the real dataloaders and variables that we pass in.

WIP cause it causes some TS fails which I don't have time to fix yet.

### DIFF
--- a/patches/@types/graphql+0.13.3.patch
+++ b/patches/@types/graphql+0.13.3.patch
@@ -1,0 +1,15 @@
+patch-package
+--- a/node_modules/@types/graphql/type/definition.d.ts
++++ b/node_modules/@types/graphql/type/definition.d.ts
+@@ -371,7 +371,10 @@ export interface GraphQLResolveInfo {
+     readonly path: ResponsePath;
+     readonly schema: GraphQLSchema;
+     readonly fragments: { [key: string]: FragmentDefinitionNode };
+-    readonly rootValue: any;
++    // Check out ./src/types/graphql.d.ts for the replacement
++    // which uses our types instead.
++    //
++    // readonly rootValue: any;
+     readonly operation: OperationDefinitionNode;
+     readonly variableValues: { [variableName: string]: any };
+ }
--- a/src/index.ts
+++ b/src/index.ts
@@ -31,6 +31,7 @@ import { nameOldEigenQueries } from "./lib/nameOldEigenQueries"
 import { rateLimiter } from "./lib/rateLimiter"
 
 import { logQueryDetails } from "./lib/logQueryDetails"
+import { RootValue } from "types/graphql"
 
 const {
   ENABLE_REQUEST_LOGGING,
@@ -109,8 +110,8 @@ async function startApp() {
     fetchPersistedQuery,
     crunchInterceptor,
     graphqlHTTP((req, res, params) => {
-      const accessToken = req.headers["x-access-token"]
-      const userID = req.headers["x-user-id"]
+      const accessToken = req.headers["x-access-token"] as string | undefined
+      const userID = req.headers["x-user-id"] as string | undefined
       const timezone = req.headers["x-timezone"] as string | undefined
       const userAgent = req.headers["user-agent"]
 
@@ -140,16 +141,18 @@ async function startApp() {
       // Supply userAgent for analytics
       res.locals.userAgent = userAgent
 
+      const rootValue: RootValue = {
+        accessToken,
+        userID,
+        defaultTimezone,
+        exchangeSchema,
+        ...loaders,
+      }
+
       return {
         schema,
         graphiql: true,
-        rootValue: {
-          accessToken,
-          userID,
-          defaultTimezone,
-          exchangeSchema,
-          ...loaders,
-        },
+        rootValue,
         formatError: graphqlErrorHandler(enableSentry, {
           req,
           // Why the checking on params? Do we reach this code if params is falsy?
--- a/src/types/graphql.d.ts
+++ b/src/types/graphql.d.ts
@@ -1,0 +1,29 @@
+import createLoaders from "../lib/loaders"
+
+export interface RootValues {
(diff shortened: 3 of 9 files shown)


patches/@types/graphql+0.13.3.patch
```diff
@@ -0,0 +1,15 @@
...
+     readonly schema: GraphQLSchema;
+     readonly fragments: { [key: string]: FragmentDefinitionNode };
+-    readonly rootValue: any;
++    // Check out ./src/types/graphql.d.ts for the replacement
++    // which uses our types instead.
++    //
++    // readonly rootValue: any;
+     readonly operation: OperationDefinitionNode;
+     readonly variableValues: { [variableName: string]: any };
+ }
```
orta: Step 1: Remove the `readonly rootValue: any` from the GraphQL types. This means that we're free to overwrite it in some upcoming code.

This is odd, and conceptually shouldn't be needed because typescript interfaces are re-openable. e..g

```ts
interface MyThing {
  a: string
}

interface MyThing {
  b: number
}
```

is the same as:

```ts
interface MyThing {
  a: string
  b: number
}
```

Which you can see in action when I'm extending Jest's `expect` [object in a library](https://github.com/jest-community/jest-snapshots-svg/blob/b647082cc5b67bb21a187749f0840755a840ca54/src/index.ts#L6-L16)

```ts
declare global {
    namespace jest {
        interface Matchers<R> {
            /** Checks and sets up SVG rendering for React Components. */
            toMatchSVGSnapshot(width: number, height: number, settings?: { wireframe?: boolean }): void;
        }
    }
}
```

Which means if I should be able to do:

```ts
declare module "graphql" {
  interface GraphQLResolveInfo {
    /**
     * Where we store user/data-loader context, yes there's a context
     * object but without a big refactor it's not happening soon.
     */
    readonly rootValue: RootValue
  }
}

```

Which would mean any reference to `rootValue` in a resolver should look like this:

```ts
// theirs
export interface GraphQLResolveInfo {
    readonly rootValue: any
}

// ours
export interface GraphQLResolveInfo {
    readonly rootValue: RootValue
}

```

But, that doesn't work, see [this playground](https://www.typescriptlang.org/play/#src=interface%20GraphQLResolveInfo%20%7B%0D%0A%20%20readonly%20rootValue%3A%20any%0D%0A%7D%0D%0A%0D%0Ainterface%20GraphQLResolveInfo%20%7B%0D%0A%20%20readonly%20rootValue%3A%20%7B%20a%3A%20string%7D%0D%0A%7D%0D%0A%0D%0Aconst%20a%3A%20GraphQLResolveInfo%20%3D%20%7B%7D%20as%20any%0D%0Aa.rootValue)

So, instead I have patch-package remove the original definition and I have it get filled by our `graphql.d.ts`.

src/index.ts
```diff
@@ -140,16 +141,18 @@ async function startApp() {
       // Supply userAgent for analytics
       res.locals.userAgent = userAgent
 
+      const rootValue: RootValue = {
+        accessToken,
+        userID,
+        defaultTimezone,
+        exchangeSchema,
+        ...loaders,
+      }
```
orta: I extracted this so I could ensure that if someone adds a new property then it will fail tsc without it getting added to `types/graphql.d.ts`

src/types/graphql.d.ts
```diff
@@ -0,0 +1,29 @@
...
+
+declare module "graphql/type" {
+  interface GraphQLResolveInfo {
+    /**
+     * Where we store user/data-loader context, yes there's a context
+     * object but without a big refactor it's not happening soon.
+     */
+    readonly rootValue: RootValue
+  }
+}
```
orta: worth noting this is `d.ts` file - which get special rules. They're called ambient declaration files, and can't have any "runtime" JS code inside them. 

This one exists to inject extra code into the `graphql/types` module ( which resolves to `@types/graphql/types.d.ts` )

src/types/graphql.d.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  /** The schema used by the internal exchange graphql engine */
+  exchangeSchema: GraphQLSchema
+}
+
+/**
+ * The root values we pass into the GraphQL query resolver,
+ * contains optional user information, and all of our data-loaders.
+ * */
+export type RootValue = RootValues & ReturnType<typeof createLoaders>
```
orta: This creates a new type which is a mix of the defined types above, and dynamically inferred types from the `createLoaders` function.

patches/@types/graphql+0.13.3.patch
alloy: Afaik you can only _add_ when merging declarations, only when extending can you make property types more specific. Seeing as we still need to explicitly type all our exported fields anyways, it should be fine to define our own extended interface instead of monkey-patching the default one.

alloy:
I'll pull this into my current refactor work and make the change to drop the monkey-patch.

patches/@types/graphql+0.13.3.patch
alloy: Hmm, this ended up being a bit annoying to do when using `GraphQLObjectType` as we're not creating a subclass and overriding `fields` on `GraphQLObjectTypeConfig` isn't possible due to the `Thunk<…>`. I think we can still improve this and also deal with not having to pass in `TContext` everywhere, but we can iterate on that.

alloy: In reality, this might be the reason for us to move to using the proper ‘context’ arg to put the loaders in, as the context is already typeable. /cc @damassi

orta:
👍

patches/@types/graphql+0.13.3.patch
damassi: Seems worth it

orta:
Deprecated by #1520

PR #1484, opened 2019-02-04
Send offer note in initial + counter mutations

sweir27:
cc @ashkan18 is there currently a way to confirm this via tests?

I verified locally that these three mutations work!


ashkan18:
hhmm... not that i know of but im no MP test expert :) i think current existing tests mock so many things that wouldn't catch these errors.

sepans:
🙏 thanks @sweir27. Yes the MP exchange tests only verify if the field(s) exist in MP types and that's it.

PR #1487, opened 2019-02-06
Runtime modify the eigen queries to conform to the new stitched techniques

orta:
The old implementation was white-listing particular types, which was looking to be mighty tricky when taking into account that there are 3 versions of the same mutation union in the type system ( the one hand-rolled, the one in the legacy "ecommerce" namespace, and the one in the "commerce" namespace) 

Balancing that would have been bunch of complicated code that I'm probably the only one who would grok. This drastically simplifies it and just changes the queries before they get to the main resolvers.


ashkan18:
🙌 ! awesome! thanks for also renaming, new name makes more sense!

PR #1488, opened 2019-02-06
Ensures eigen mutations are only re-written when the stitching config flag is on

orta:
I knew something was bothering me with #1487 last night, but couldn't place it. This was it.


ashkan18:
oh this makes sense!

PR #1491, opened 2019-02-06
Hotfix for the eigen query detection

orta:
Ensures that the Eigen query modifiers aren't overreaching.


orta:
Also adds a type I missed that existed in just one query

orta:
Ashkan just went into a meeting but said it was good to go

PR #1495, opened 2019-02-07
Correct resolver batching behavior

zephraph:
After enabling batching we found a few problems. 

1. List endpoints return an array but single endpoints return an object. I was expecting all responses to be arrays and called `Array.find` on the results which caused metaphysics to crash on single responses
2. Dataloader wasn't properly deduping data. Turns out it does strict checking of the inputs to calculate a cache key (meaning objects are duplicated).
3. For any given endpoint "id" could also be a slug. 
4. Not related to this implementation, but also a problem... list endpoints generally don't support slugs. Thankfully @mzikherman helped me out and [updated the sales endpoint](https://github.com/artsy/gravity/pull/12141) to do so. 
5. We support historical slugs, so the slug passed into the dataloader might not be the slug returned to the response. Given that the order of results isn't guaranteed the dataloader needs some way to match the original key to the results, which it couldn't do if the key was a historical slug.

This PR resolves 1-3. 4 was fixed in @mzikherman's PR and 5 is still an open issue. 

TODO
- [x] Test against a few queries to ensure behavior is as expected


src/lib/loaders/batchLoader.ts
```diff
@@ -69,26 +111,40 @@ export const batchLoader = ({
...
+            ) || defaultResult
+          )
+        })
       })
-      return flatten(normalizedResults)
-    })
-  })
+    },
+    {
+      maxBatchSize: 20,
```
zephraph: I was previously naively splitting batch sizes inside of the dataloader. This eliminates the need for doing that (making the overall logic a lot simpler)

zephraph:
I'm fairly certain I need more coverage on error cases. Two things come to mind.

1. What happens when null is requested from the dataloader? 
2. What happens when the gravity response fails?

It feels like for 1 the gravity loader should just return the default value as a response and for 2 we may need to resolve with an error object... I'll have to dig into how the gravity loader actually handles that.

src/lib/loaders/__tests__/batchLoader.test.ts
```diff
@@ -143,37 +160,94 @@ describe("groupKeys", () => {
...
+            boo: "baz",
+          },
+        ],
+      ],
     ])
   })
 })
+
+describe("cacheKeyFn", () => {
+  const { cacheKeyFn } = require("../batchLoader")
```
damassi: Out of curiosity, are you importing this within the block to reset some kind of `batchLoader` state?

src/lib/loaders/batchLoader.ts
```diff
@@ -1,15 +1,57 @@
...
+}
+
+/**
+ *  This function is used by the dataloader to determine the uniqueness of keys.
+ *  Because keys can be objects, it's important to be able to differentiate between
+ *  their contents. If an object is empty except for an id field, we just use id.
+ *  Otherwise we stringify the object so it can be compared via strict comparison.
+ */
+export const cacheKeyFn = (key: Key): string => {
+  if (typeof key === "object" && key !== null) {
```
damassi: http://2ality.com/2013/04/quirk-undefined.html

src/lib/loaders/__tests__/batchLoader.test.ts
```diff
@@ -143,37 +160,94 @@ describe("groupKeys", () => {
...
+  it("should treat strings and ids no params the same", () => {
```
alloy: Not sure what “ids no params” means here

src/lib/loaders/__tests__/batchLoader.test.ts
```diff
@@ -143,37 +160,94 @@ describe("groupKeys", () => {
...
+  })
+})
+
+describe("normalizeKeys", () => {
+  const { normalizeKeys } = require("../batchLoader")
+
+  it("should turn a string into an object with an id", () => {
+    expect(normalizeKeys(["123"])).toEqual([{ id: "123" }])
+  })
+  it("should preserve an object with an id", () => {
```
alloy: Tiny nit: missing newline.

src/lib/loaders/batchLoader.ts
```diff
@@ -20,35 +62,35 @@ const renderParams = key => {
...
-      return {
-        // @ts-ignore
-        ...keys[0],
-        id: keys.map(k => k.id),
-        size: keys.length,
+/**
+ * Takes an array of normalized keys and groups them based on their parameters.
+ * @returns a tuple of an array of group strings and an array of grouped keys
+ */
+export const groupKeys = (requestedKeys): [KeyGroupList, NormalKey[][]] => {
```
alloy:
```suggestion
export const groupKeys = (requestedKeys: NormalKey[]): [KeyGroupList, NormalKey[][]] => {
```

src/lib/loaders/batchLoader.ts
```diff
@@ -20,35 +62,35 @@ const renderParams = key => {
...
+  const [keyGroupList, groupedKeys] = chain(requestedKeys)
+    .groupBy(getKeyGroup)
+    .entries()
+    .thru(entries => {
+      let result: [KeyGroupList, NormalKey[][]] = [[], []]
```
alloy: Pretty sure mutating the arrays doesn't require `let`. i.e. the pointers to the arrays remain the same.

```suggestion
      const result: [KeyGroupList, NormalKey[][]] = [[], []]
```

src/lib/loaders/batchLoader.ts
```diff
@@ -20,35 +62,35 @@ const renderParams = key => {
...
+  return [keyGroupList, groupedKeys]
 }
 
+/**
+ * Collects all the params into one object to be passed to gravity loader
+ */
+export const compactKeyForRequest = (keys: NormalKey[]): CompactedKey => ({
+  ...keys[0],
+  id: keys.map(key => key.id),
+})
```
alloy: I have a hard time understanding the goal of this function. The singular `id` key holds an array of ids, but otherwise  only the first key is used for the rest of the compacted key's traits? Why is that done in that way?

src/lib/loaders/batchLoader.ts
```diff
@@ -69,26 +111,40 @@ export const batchLoader = ({
...
+          if (keys.id.length === 1 && singleLoader) {
+            return singleLoader(keys.id[0])
+          } else {
+            return multipleLoader(keys)
+          }
+        })
+      ).then((data: Array<GravityResult | GravityResult[]>) => {
+        const normalizedData = data.map(
+          datum => (Array.isArray(datum) ? datum : [datum])
         )
```
alloy: Remind me, a `singleLoader` is _not_ a loader that only hits endpoints that return single entities [it may return lists], it's simply an endpoint that does not take a list of IDs to fetch, right? And which is why this normalisation can't happen on line 122 instead.

src/lib/loaders/batchLoader.ts
```diff
@@ -1,15 +1,57 @@
...
+
+/**
+ * This coerces all keys to an object structure for consistent processing
+ */
+export const normalizeKeys = (keys: Key[]) =>
+  keys.map(key => (typeof key === "string" ? { id: key } : key))
+
 /**
  * This is just used to compare options sent into the dataloader.
  * If the options are an object this stingifies the object in a way
```
alloy: 🐝 

```suggestion
 * If the options are an object this stringifies the object in a way
```

src/lib/loaders/batchLoader.ts
```diff
@@ -69,26 +111,40 @@ export const batchLoader = ({
...
+          }
+        })
+      ).then((data: Array<GravityResult | GravityResult[]>) => {
+        const normalizedData = data.map(
+          datum => (Array.isArray(datum) ? datum : [datum])
         )
+
+        return normalKeys.map(key => {
+          const group = getKeyGroup(key)
+          const groupIndex = keyGroups.indexOf(group)
```
alloy: Perhaps I'm not evaluating all of this correctly in my mind, but I wonder if it's possible to add the key to the data coming out of the data loaders [in the code block starting on line 120] and that removing the need to find the right group here again. Nothing earth-shattering in terms of optimisations, but perhaps a bit simpler.

alloy:
> 1. What happens when null is requested from the dataloader? 

If you mean that the key is `null` then that constitutes an argument error, in my mind. I think the types should disallow it and your code should throw an error.

> 2. What happens when the gravity response fails?

Do you mean a non `2xx` HTTP response? The promise [will reject](https://github.com/artsy/metaphysics/blob/61634d2577483dbbd612970bb26fbc7295ffddc6/src/lib/apis/fetch.ts#L30). (I’m going to change the way that check happens btw, no need to generate a string and compile a regex for each request.)

src/lib/loaders/__tests__/batchLoader.test.ts
zephraph: Ahh, so the object actually represents query parameters for a gravity call. So I'd kind of continued with that language.

zephraph:
@alloy if the gravity request rejects I'll likely need to handle that separately or it'll cause the `promise.all` of the batch loader to fail which would cause all of the batched calls to not return.

src/lib/loaders/batchLoader.ts
```diff
@@ -1,15 +1,57 @@
...
+interface CompactedKey {
+  id: string[]
+  [key: string]: any
+}
+
+interface GravityResult {
+  _id: string
+  _slug: string
+  [key: string]: any
+}
```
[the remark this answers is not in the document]
zephraph: It'd make sense not to have it here. If you have a similar type in your PR we can require it in here.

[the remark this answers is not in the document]
alloy: Nope, not yet, and I feel like your PR will probably land sooner, so I'll move it from here in my PR.

src/lib/loaders/__tests__/batchLoader.test.ts
zephraph: Yeah. There are some tests that mock the import to override the environment variable to enable or disable the feature.

src/lib/loaders/batchLoader.ts
zephraph: The "key" is ultimately parameters that are being passed to gravity loader. To pass multiple ids to a list endpoint you turn id into an array of keys. The goal is to turn a list of like parameters (i.e. they all have the same parameters besides id) into a single object with a list of ids that the list endpoint can understand.

src/lib/loaders/batchLoader.ts
zephraph: My expectation is that a `singleLoader` generally returns single entities, but I guess there's no guarantee of that.  It could potentially happen on 122. That might save some loops. Still, if a list endpoint ended up sending a single result (not as an array) then it would fail. I'm assuming that would never happen, but 🤷‍♂️

src/lib/loaders/batchLoader.ts
zephraph: The point of this section is to make sure that we return results in the exact order that they were passed into the data loader. The only data guarantee that I have is that the groups that I process will be returned in the same order. Results inside the group could be out of order. Adding a key to the results wouldn't really make sense because we already have a key (of sorts) in the results in the form of `_id` or `_slug`. 

I'm not sure. I know this is complicated. I'm open for ideas to make it simpler, I'm just not seeing how shifting this logic and mixing it in with the fetching would do that.

src/lib/loaders/batchLoader.ts
zephraph: I'm not sure what this relates to. If the key is undefined, I'm just returning it like any other value. I'm not sure how to do error handling here, because if I throw an exception it's going to fail the entire query... maybe that's what we want though? Hmmm...

src/lib/loaders/batchLoader.ts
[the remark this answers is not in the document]
zephraph: kk

src/lib/loaders/batchLoader.ts
alloy: Ah yes, ofc 🤦‍♂️ Thanks!

src/lib/loaders/batchLoader.ts
alloy: I'm not really seeing anything concrete, so I'd leave it as-is and later on see if there's opportunity to iterate. Perhaps we can pair, when it all works as you want it to.

zephraph:
So! Thanks to @alloy for some enormously helpful pairing, batching is now a _lot_ closer to what it was. Really the last remaining item is that historical slugs will break batching. 

I've got a few TODO comments in here, mostly about clean up, moving things around, adding error handling. Would definitely appreciate people taking a fresh look! 🙏

zephraph:
After a conversation w/ the platform team we're going to try to add a sorting mode of batch for the sales endpoint. 

This will ensure...

1. results are returned in the proper order
2. results are the same length as the number of items requested (with nils for data that didn't return)

This will greatly simplify the batching code at the cost of slightly more complex api response behavior. I'll link the gravity PR here once it's ready.

zephraph:
OKAY! So, this seems to fully work now. 

Thanks for @joeyAghion, @sweir27, and @orta for the pairing to help me get to this point. 

The final issue was that the `gravityLoader` was sorting params... which changed the order of ids going to gravity and thus caused the data loaders to return things in the wrong order. 

There's still significant complexity here and I've got a lot of commenting to do. Please poke holes. 

Oh and @alloy I have a type issue. Suggestions?

alloy:
Can you point me to the type error?

zephraph:
@alloy I figured it out! It was a simple mistake on my part but the error wasn't very clear.

alloy:
@zephraph Can you confirm if this is done and needs a final review + merge? (I’m slightly confused because it’s still self-assigned.)

zephraph:
@alloy it _is_ done and ready for review. Sorry about that!

src/lib/helpers.ts
```diff
@@ -78,10 +78,19 @@ export const truncate = (string, length, append = "…") => {
...
+   */
+  // @ts-ignore
+  options.batched
+    ? stringify(options, {
+        arrayFormat: "brackets",
+      })
+    : stringify(options, {
+        arrayFormat: "brackets",
+        sort: (a, b) => a.localeCompare(b),
+      })
```
alloy: TIL. Fancy ✨

src/lib/loaders/__tests__/batchLoader.test.ts
```diff
@@ -79,55 +79,45 @@ describe("batchLoader", () => {
...
+describe("groupByParams", () => {
+  const { groupByParams } = require("../batchLoader")
   it("should handle a single key", () => {
-    expect(groupKeys(["a"])).toEqual([
-      {
-        id: ["a"],
-        size: 1,
-      },
+    expect(groupByParams([{ id: "a" }])).toEqual([
+      [""],
```
alloy: This value here makes understanding the behaviour a little harder, it could be improved by first having a test that just focusses on this value and thus explains why this is blank.

src/lib/loaders/batchLoader.ts
```diff
@@ -1,55 +1,78 @@
...
-      if (typeof keys[0] === "string") {
-        return { id: keys, size: keys.length }
-      }
-      return {
-        // @ts-ignore
-        ...keys[0],
-        id: keys.map(k => k.id),
-        size: keys.length,
+/**
+ * Takes an array of parameters and groups them by non-unique
```
alloy: I don’t really understand this sentence, is it incomplete?

src/lib/loaders/batchLoader.ts
```diff
@@ -1,55 +1,78 @@
...
+ * @returns a tuple of an array of group strings and an array of grouped keys
+ */
+export const groupByParams = (
+  params: IdWithParams[]
+): [SerializedParams, IdWithParams[][]] => {
+  const [serializedParams, groupedParams] = chain(params)
+    .groupBy(serializeParams)
+    .entries()
+    .thru(entries => {
+      const result: [SerializedParams, IdWithParams[][]] = [[], []]
```
alloy: You could define this tuple type once and use it here and for the function’s return value.

src/lib/loaders/batchLoader.ts
```diff
@@ -1,55 +1,78 @@
...
+      for (let i = 0; i < entries.length; ++i) {
+        result[0].push(entries[i][0])
+        result[1].push(entries[i][1])
       }
+      return result
     })
     .value()
 
-  return (grouped as unknown) as GroupKeysResult[]
+  return [serializedParams, groupedParams]
```
alloy: Why unpack the tuple above only to pack it here again?

src/lib/loaders/batchLoader.ts
```diff
@@ -59,39 +82,89 @@ interface BatchLoaderArgs {
...
-    ).then(data => {
-      const normalizedResults = data.map((queriedGroup, groupIndex) => {
-        return groupedKeys[groupIndex].id.map(
-          id => queriedGroup.find(r => r._id === id) || defaultResult
-        )
+  const dl = new DataLoader(
+    async (idWithParamsList: IdWithParams[]) => {
+      const [paramGroups, groupedParams] = groupByParams(idWithParamsList)
+      const data = await Promise.all(
+        groupedParams.map(batchParams).map(params_1 => {
```
alloy: Why is this called `params_1`?

alloy: Hrmm, I see, just to clarify to the reader that it’s different from `params_2` 🤔 

Ok, I have no better suggestions right now 🤷‍♂️

src/lib/loaders/batchLoader.ts
```diff
@@ -59,39 +82,89 @@ interface BatchLoaderArgs {
...
+  type Key = string | { id: string[] }
+
+  /**
+   * This is an abstraction around the data loader to keep api parity with the
+   * existing gravityLoader api
+   */
+  return (key: Key) => {
+    /**
+     * This section covers the case when an endpoint is being requested that supports
+     * parameters. An example would the be `sales` endpoint which has filters like
```
alloy:
```suggestion
     * parameters. An example would be the `sales` endpoint which has filters like
```

src/lib/loaders/batchLoader.ts
```diff
@@ -59,39 +82,89 @@ interface BatchLoaderArgs {
...
+        )
+        .then(results => results.filter(result => result !== null))
+      /**
+       * This section covers the case when a single resource endpoint is being requested.
+       * Take the `sale` endpoint for example. `key` is expected to be a string of `id` or `slug`.
+       */
+    } else if (typeof key === "string") {
+      return dl.load({ id: key })
+    } else {
+      console.error("Requested an invalid key type for batchLoader")
```
alloy: Why not throw an error instead?

src/lib/loaders/batchLoader.ts
zephraph: Ummm... That... what. 

I'm not sure how that happened exactly... this seems like a vscode refactor gone wrong, but I _really_ don't remember seeing it. I'll update.

src/lib/loaders/batchLoader.ts
zephraph:
```suggestion
      throw new Error("Requested an invalid key type for batchLoader")
```

zephraph: Like that?

src/lib/loaders/batchLoader.ts
zephraph: 😄 Good catch...

src/lib/loaders/batchLoader.ts
zephraph: Good point. It isn't strictly necessary. Exposing the type will be information enough about what's being returned.

alloy: Ah gotcha. Yeah I can see how in vanilla JS this could help understanding.

src/lib/loaders/batchLoader.ts
alloy: Yup! I’d rather fail early when an API isn’t being used right, especially when there’s types that should prevent it.

If we were using FB’s way of doing [assertions](https://en.wikipedia.org/wiki/Assertion_(software_development)), which I’d really like, this would be classed as an `invariant`.

src/lib/loaders/__tests__/batchLoader.test.ts
zephraph: I see how that can be confusing. I added tests for `serializeParams` right about this that determines what that behavior would be.

src/lib/loaders/__tests__/batchLoader.test.ts
```diff
@@ -82,6 +82,19 @@ describe("batchLoader", () => {
   })
 })
 
+describe("serializeParams", () => {
+  const { serializeParams } = require("../batchLoader")
```
alloy: Btw, why are you loading these functions this way? If you would import them as normal, you could do `describe(serializeParams, () => …)` instead and have the name of the tests updated automatically if we’d ever rename the function (using the refactor tools).

alloy:
@zephraph FYI, re “arg names”–did you know ‘arguments’ are the values that are passed in, whereas ’parameters’ are the names of the variables in a function’s signature? Took me a while to realise this, but really helps.

zephraph:
Ohh, hmmm... I haven't thought about that. JavaScript exposes `arguments` as a way to access parameters. That makes it weird. 

In the context of this file `params` actually refer to query parameters, not function parameters. I probably haven't made that very clear.

src/lib/loaders/__tests__/batchLoader.test.ts
zephraph: That's a good tip!

I'm doing the imports this way because earlier in the tests I mock the dependency to override the feature flag. 

https://github.com/artsy/metaphysics/blob/batching-bugfix/src/lib/loaders/__tests__/batchLoader.test.ts#L10

alloy:
> Ohh, hmmm... I haven't thought about that. JavaScript exposes `arguments` as a way to access parameters. That makes it weird.

Well, not really, it’s a list of the _values_, so these are the arguments.

> In the context of this file `params` actually refer to query parameters, not function parameters. I probably haven't made that very clear.

Nah, that was clear enough to me. I feel like `params` is well understood to refer to HTTP parameters 👍

src/lib/loaders/__tests__/batchLoader.test.ts
alloy: Oh, but actually it doesn’t seem like you’d need to do any mocking for that, you can just change the value and reset it from an `afterEach`?

Anyways, not a blocker for this PR.

src/lib/loaders/batchLoader.ts
alloy:
```suggestion
      const result: ParamGrouping = [[], []]
```

alloy:
As per https://github.com/artsy/README/issues/183, may I squash and merge this one?

PR #1513, opened 2019-02-18
Fixes an issue where Gemini-related mutations always return an error

yuki24:
finishes https://artsyproduct.atlassian.net/browse/AUCT-64

It seems that a `()` is missing when a function is returned from a factory `loader` function, causing the `createGeminiEntryForAsset` and `requestCredentialsForAssetUpload` mutations to fail due to a return value mismatch. This PR adds a missing `()` so the function actually does the work.

### The `createGeminiEntryForAsset` mutation:

create-gemini-entry-for-asset-before

### The `requestCredentialsForAssetUpload` mutation:

request-credentials-for-asset-upload-before

### The `createGeminiEntryForAsset` mutation:

create-gemini-entry-for-asset-after

### The `requestCredentialsForAssetUpload` mutation:

request-credentials-for-asset-upload-after

### DIFF
--- a/src/schema/asset_uploads/create_asset_request_mutation.ts
+++ b/src/schema/asset_uploads/create_asset_request_mutation.ts
@@ -95,6 +95,6 @@ export default mutationWithClientMutationId({
     { rootValue: { createNewGeminiAssetLoader } }
   ) => {
     if (!createNewGeminiAssetLoader) return null
-    return createNewGeminiAssetLoader({ name, acl })
+    return createNewGeminiAssetLoader({ name, acl })()
   },
 })
--- a/src/schema/asset_uploads/__tests__/create_asset_request_mutation.test.js
+++ b/src/schema/asset_uploads/__tests__/create_asset_request_mutation.test.js
@@ -18,7 +18,7 @@ describe("addAssetToConsignmentSubmission", () => {
     `
 
     const rootValue = {
-      createNewGeminiAssetLoader: () =>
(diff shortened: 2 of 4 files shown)


src/schema/asset_uploads/create_asset_request_mutation.ts
```diff
@@ -95,6 +95,7 @@ export default mutationWithClientMutationId({
     { rootValue: { createNewGeminiAssetLoader } }
   ) => {
     if (!createNewGeminiAssetLoader) return null
-    return createNewGeminiAssetLoader({ name, acl })
+    console.log(name, acl)
```
ashkan18: we should remove this 🙌

yuki24: Should we really remove it? I like my console logs!

:trollface: :trollface: :trollface:

yuki24:
@ashkan18 I actually had a chat with @williardx and took over his change from https://github.com/williardx/metaphysics/commit/ab0f593af8a1998d81fac3ddb68bee6c9e55757c. I'm not sure if he has sent a PR, but I thought @williardx and you had some context around this and could review this PR (which includes tests).

ashkan18:
thanks! makes sense! 🙌

ashkan18:
feel free to merge after removing the `console.log` 😍

williardx:
Thanks @yuki24!

PR #1514, opened 2019-02-19
Restructures exhibitors_grouped_by_name

kierangillen:
- Restructures `exhibitors_grouped_by_name` to have one array of `exhibitor` objects and adds correct `profile_id`

screen shot 2019-02-19 at 2 31 57 pm

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1154,7 +1154,7 @@ type ArtworkContextFair {
   tagline: String
   ticketsLink: String
 
-  # The exhibitors with booths in this fair.
+  # The exhibitors with booths in this fair with letter.
   exhibitors_grouped_by_name: [FairExhibitorsGroup]
 
   # Artworks Elastic Search results
@@ -3281,7 +3281,7 @@ type Fair {
   tagline: String
   ticketsLink: String
 
-  # The exhibitors with booths in this fair.
+  # The exhibitors with booths in this fair with letter.
   exhibitors_grouped_by_name: [FairExhibitorsGroup]
 
   # Artworks Elastic Search results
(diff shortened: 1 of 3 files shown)


src/schema/fair.ts
```diff
@@ -291,12 +291,26 @@ const FairType = new GraphQLObjectType({
...
-            },
-            profile_ids: {
-              type: new GraphQLList(GraphQLString),
-              description: "Partner/Exhibitor default profile id",
+              description: "The exhibitor data.",
+              type: new GraphQLList(
+                new GraphQLObjectType({
+                  name: "FairExhibitor",
+                  fields: {
+                    exhibitor_name: {
```
l2succes: From a best practice / convention perspective, these fields should be named `id` and `name`

_schema.graphql
```diff
@@ -3354,15 +3354,23 @@ type FairCounts {
...
   # Letter exhibitors group belongs to
   letter: String
 
-  # Exhibitors sorted by name
-  exhibitors: [String]
-
-  # Partner/Exhibitor default profile id
-  profile_ids: [String]
+  # The exhibitor data.
+  exhibitors: [FairExhibitor]
```
ashfurrow: This is going to be a breaking schema change, which _should_ be okay as long as we aren't consuming this part of the schema yet. Make sure to check Volt, Reaction, and Force before merging.

src/schema/fair.ts
ashfurrow: Agreed.

_schema.graphql
kierangillen: It's not being used anywhere else.

joeyAghion:
Because this is a breaking change and Reaction embeds the old [out-of-date] schema, this is causing compatibility failures on Force PRs (like https://github.com/artsy/force/pull/3582 and https://github.com/artsy/force/pull/3578). If it's not already, mind updating Reaction to match this new schema?

PR #1516, opened 2019-02-21
Add fix_failed_payment mutation

ds300:
I tested against local exchange and it's working :+1:

~This should be merged after artsy/exchange#396 is in production~


_schema.graphql
```diff
@@ -3477,6 +3459,20 @@ type FilterSaleArtworksCounts {
   ): FormattedNumber
 }
 
+input fixFailedPaymentInput {
```
ds300: why is this lower case?

sweir27:
Can we add #donotmerge to this until the exchange-side is deployed to production? 🙏

_schema.graphql
orta: interesting, so is type `fixFailedPaymentPayload`

ds300:
Good catch! 🙇

_schema.graphql
ds300: It was lowercase because I made a typo 😞

ds300:
Alright I think this is 🍏 – As a bonus I've been doing end-to-end tests with force 👌

PR #1518, opened 2019-02-22
Fixes fair booths being filtered out

l2succes:
Fix fair booth slug routing (certain fair booths returning 404s) https://artsyproduct.atlassian.net/browse/LD-239

### DIFF
--- a/src/schema/__tests__/show.test.js
+++ b/src/schema/__tests__/show.test.js
@@ -146,6 +146,27 @@ describe("Show type", () => {
     }
   })
 
+  it("returns a fair booth even with displayable set to true", async () => {
+    showData.fair = {
+      id: "the-art-show-2019",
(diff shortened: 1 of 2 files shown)


src/schema/__tests__/show.test.js
```diff
@@ -146,6 +146,27 @@ describe("Show type", () => {
     }
   })
 
+  it("returns a fair booth even with displayable set to true", async () => {
```
anandaroop: Unless I'm missing something I think the test case should be for when `displayable == false`. 

(And fair is existy, as you've got here.)

So the example name and the fixture data need a little tweak here, right?

l2succes: Yes that's totally right. Thanks @anandaroop

PR #1521, opened 2019-02-25
Filter out online only on city shows

ashkan18:
# Changes
1) By default when looking for shows under a city, filter out shows without location by setting `has_location: true` when calling gravity in the shows resolver.

2) We already support passing `state` to shows query to find shows in specific state but here we update the default to be `current` and allow to override when needed by clients.
 Since default values are not yet properly handled by Graphql (or at least on the version we are on) https://github.com/apollographql/graphql-tools/issues/715 we had manually lowercase `state` in the arguments passed to resolver.


src/schema/city/index.ts
```diff
@@ -42,19 +42,29 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
...
           type: GraphQLBoolean,
           description:
             "Whether to include local discovery stubs as well as displayable shows",
         },
       }),
       resolve: async (city, args, { showsWithHeadersLoader }) => {
+        // default Enum value for status is not properly resolved
+        // so we have to manually resolve it by lowercasing the value
+        // https://github.com/apollographql/graphql-tools/issues/715
+        args.status = args.status.toLowerCase()
```
anandaroop: Bummer! Would it be simpler to just set the default value to `"current"` since that is [the actual value](https://github.com/artsy/metaphysics/blob/master/src/schema/input_fields/event_status.ts#L28)? That would make this line unnecessary.

I'm not sure what the best practice is here 🤷‍♂️

src/schema/city/index.ts
```diff
@@ -42,19 +42,29 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
...
+
         const gravityOptions = {
           ...convertConnectionArgsToGravityArgs(args),
           displayable: true,
           near: `${city.coordinates.lat},${city.coordinates.lng}`,
           max_distance: LOCAL_DISCOVERY_RADIUS_KM,
+          has_location: true,
```

src/schema/city/index.ts
orta: Strictly speaking that would be wrong at GraphQL level (because the enum is ALL_CAPS) so you'd end up with a schema that looks like:

```
# By default we filter shows by current
status: EventStatus = current
```

Which isn't actually something that EventStatus could be, even though the value under the hood is actually the lowercase string

ashkan18:
🙌 shall we merge?

ashkan18:
merging now that Gravity side is deployed on staging

PR #1523, opened 2019-02-26
Make sure we filter out online-only shows for nearbyShows

ashkan18:
And also and set default state to `current` so by default we only show current shows.


orta:
👍

PR #1525, opened 2019-02-26
update kaws schema to include new randomize argument

xtina-starr:
Address: https://artsyproduct.atlassian.net/browse/GROW-1145

This is a part of the work to add support for randomizing collections and updates the kaws schema in metaphysics. 

screen shot 2019-02-26 at 4 55 30 pm


l2succes:
Looks good!

PR #1527, opened 2019-02-27
[City] Make fairs field a connection.

alloy:
Closes https://artsyproduct.atlassian.net/browse/LD-19

PR #1528, opened 2019-02-27
Type headers and parse count header

alloy:
Depends on #1527 being merged first. This PR is solely about c8ba2dea6bc80e6fc833cc10d6e6dfbc26810d84.

### DIFF
--- a/src/lib/loaders/index.ts
+++ b/src/lib/loaders/index.ts
@@ -13,7 +13,9 @@ import {
   PathGenerator,
 } from "./api/loader_interface"
 
-export type BodyAndHeaders<B = any, H = any> = {
+type ResponseHeaders = { [header: string]: string }
+
+export type BodyAndHeaders<B = any, H = ResponseHeaders> = {
   body: B
   headers: H
 }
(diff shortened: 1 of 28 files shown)


src/schema/artist/related.ts
```diff
@@ -68,7 +68,7 @@ export const Related = {
               : relatedContemporaryArtistsLoader
 
           return fetch(gravityArgs).then(({ body, headers }) => {
-            const totalCount = headers["x-total-count"]
+            const totalCount = parseInt(headers["x-total-count"] || "0", 10)
```
anandaroop: Any particular reason we are being so explicit as to provide the radix here?

alloy: The TS types make it a required parameter, because of:

> If the input `string` begins with "0", radix is eight (octal) or 10 (decimal).  Exactly which radix is chosen is implementation-dependent.  ECMAScript 5 specifies that 10 (decimal) is used, but not all browsers support this yet.  For this reason **always specify a radix when using parseInt**.

– https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

anandaroop: Oh snap! Good to know.

orta: today I learned, I just accepted it and moved on - glad you asked

src/lib/loaders/index.ts
```diff
@@ -39,12 +41,12 @@ export interface LoaderFactory {
     globalParams: any,
     pathAPIOptions: { headers: false } & APIOptions
   ): DynamicPathLoader<Body, PathGeneratorParams>
-  <Body = any, Headers = any>(
+  <Body = any, Headers = ResponseHeaders>(
```
alloy: Previously the default type for the headers object was `any`, so `headers["x-total-count"]` was also `any` and passing it to the `arrayLength` property wouldn’t fail, even though it requires a `number`.

So I changed the default of headers to be an object where the value is a `string`, which then started failing on all these places where we were passing a `string` instead of a `number`.

PR #1530, opened 2019-02-27
Exposes lastTransactionFailed from exchange

sweir27:
We'll need to reference this in reaction when deciding whether or not to allow you to visit the `/payment/new` page.


ds300:
:tada:

PR #1532, opened 2019-02-27
Fix LD-177

ashkan18:
Looks like we may not need updated Hong Kong lat/lon if we are using `slug` but this should be ok 

https://artsyproduct.atlassian.net/browse/LD-177


anandaroop:
The Emission change is sort of independent of this — HK change here is good for getting us a more representative set of shows from the `city` query.

PR #1533, opened 2019-02-28
Load all city shows and fairs

alloy:
Part of https://artsyproduct.atlassian.net/browse/LD-284

### DIFF
--- a/src/lib/all.ts
+++ b/src/lib/all.ts
@@ -1,21 +1,39 @@
-import { assign, times, flatten } from "lodash"
+import { times, flatten } from "lodash"
 
-export const allViaLoader = (loader, loaderOptions, apiOptions: any = {}) => {
-  const countOptions = assign({}, apiOptions, {
+// TODO: Type this and refactor, because I think things just kinda work by luck,
+//       because a static and a dynamic path loader take different number of
+//       arguments.
+
+export const MAX_GRAPHQL_INT = 2147483647
+
+export const allViaLoader = (
+  loader,
+  options: {
+    // For dynamic path loaders
+    path?: string | { [key: string]: any }
+    params?: { [key: string]: any }
+    api?: { [key: string]: any }
+  } = {}
+) => {
+  const params = options.params ? { size: 25, ...options.params } : { size: 25 }
+  const invokeLoader = invocationParams =>
+    options.path
+      ? loader(options.path, invocationParams, options.api)
+      : loader(invocationParams, options.api)
+  const countParams = {
+    ...params,
     page: 1,
-    size: 0,
+    size: 1, // FIXME: The /shows endpoint does not return a count with size=0
     total_count: true,
-  })
-  return loader(loaderOptions, countOptions)
+  }
+  return invokeLoader(countParams)
     .then(({ headers }) => {
-      const count = parseInt(headers["x-total-count"] || 0, 10)
-      const pages = Math.ceil(count / (apiOptions.size || 25))
+      const count = parseInt(headers["x-total-count"] || "0", 10)
+      const pages = Math.ceil(count / params.size)
       return Promise.all(
         times(pages, i => {
-          return loader(
-            loaderOptions,
-            assign({}, apiOptions, { page: i + 1 })
-          ).then(({ body }) => body)
+          const pageParams = { ...params, page: i + 1 }
+          return invokeLoader(pageParams).then(({ body }) => body)
         })
       )
     })
--- a/src/schema/__tests__/city.test.ts
+++ b/src/schema/__tests__/city.test.ts
@@ -1,6 +1,3 @@
-import { runQuery } from "test/utils"
-import gql from "lib/gql"
-
 const mockCities = {
   "sacramende-ca-usa": {
     slug: "sacramende-ca-usa",
@@ -14,11 +11,20 @@ const mockCities = {
   },
 }
 
-beforeEach(() => {
-  jest.mock("../city/city_data.json", () => mockCities)
-})
+jest.mock("../city/city_data.json", () => mockCities)
+jest.mock("../../lib/all.ts")
+
+import { runQuery } from "test/utils"
+import gql from "lib/gql"
+import { MAX_GRAPHQL_INT, allViaLoader as _allViaLoader } from "lib/all"
+
+const allViaLoader = _allViaLoader as jest.Mock<typeof _allViaLoader>
 
 describe("City", () => {
+  afterEach(() => {
+    allViaLoader.mockReset()
+  })
+
   describe("finding by slug", () => {
     it("finds a city by its slug", () => {
       const query = gql`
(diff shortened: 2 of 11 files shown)


alloy:
@l2succes Should the same be done for fairs?

src/lib/all.ts
```diff
@@ -1,21 +1,39 @@
...
+  const params = options.params ? { size: 25, ...options.params } : { size: 25 }
+  const invokeLoader = invocationParams =>
+    options.path
+      ? loader(options.path, invocationParams, options.api)
+      : loader(invocationParams, options.api)
+  const countParams = {
+    ...params,
     page: 1,
-    size: 0,
+    size: 1, // FIXME: The /shows endpoint does not return a count with size=0
```
alloy: We need to fix this in Gravity, I presume.

src/lib/all.ts
```diff
@@ -1,21 +1,39 @@
-import { assign, times, flatten } from "lodash"
+import { times, flatten } from "lodash"
 
-export const allViaLoader = (loader, loaderOptions, apiOptions: any = {}) => {
-  const countOptions = assign({}, apiOptions, {
+// TODO: Type this and refactor, because I think things just kinda work by luck,
+//       because a static and a dynamic path loader take different number of
+//       arguments.
+
+export const MAX_GRAPHQL_INT = 2147483647
```
ashkan18: 😕

src/lib/all.ts
ashkan18: so this query is just trying to get the counts?

src/schema/city/index.ts
```diff
@@ -53,31 +54,57 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
...
+            : undefined),
         }
-        delete gravityOptions.page
 
-        if (args.discoverable) {
-          delete gravityOptions.displayable
+        let response
+        let offset
+
+        if (args.first === MAX_GRAPHQL_INT) {
```
ashkan18: im not sure if i understand this correctly, but this means we make one big query to get everything in one request, is it possible to cache things with pagination? basically set page size to 100 and make multiple requests asynch till we fetch everything if there are more than 100 items.

alloy: No it uses the `allViaLoader`, which does what you suggest but in batches of 25.

alloy:
See the linked JIRA ticket regarding the potential load and how/where to cache.

src/lib/all.ts
alloy: Wazzup?

src/lib/all.ts
alloy: Yup

src/schema/city/index.ts
```diff
@@ -53,31 +54,57 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
...
+          // TODO: We could throw an error if the `after` arg is passed, but not
+          //       doing so, for now.
+          offset = 0
+          response = await allViaLoader(showsWithHeadersLoader, {
```
mzikherman: I dont have much (any 😒 ) context , but taking a look at the `/shows` endpoint- can potentially return _all_ shows sort of thing (depending on the query params used). So I'm assuming this is being used with params that will generally return a much more reasonable-sized list (all 'discoverable' shows near a location), but just wanting to point it out.

The few places of loading all data via a paginated API are anti-patterns, basically. Not saying to change your plan for this PR (or the project generally). One possible alternative might be to load in an export of the previous night show's data generated via the Fulcrum tasks.

alloy: Yup, all valid points 👍 We’re aware of the issues and are thinking it through.

PS Is also didn’t add a `all` arg to the field to not encourage it for other uses.

src/schema/__tests__/city.test.ts
```diff
@@ -166,6 +172,30 @@ describe("City", () => {
...
       expect(gravityOptions).not.toHaveProperty("displayable")
     })
+
+    it("can request all shows [that match other filter parameters]", async () => {
+      allViaLoader.mockImplementation(() => Promise.resolve(mockShows))
+
+      query = gql`
+        {
+          city(slug: "sacramende-ca-usa") {
+            shows(first: ${MAX_GRAPHQL_INT}) {
```
l2succes: @alloy so the idea is we'll pass in a really large number for `first` here on the client to get all shows?

l2succes:
@alloy It would make sense to the same for fairs. If it's easy to do so now then let's go for it otherwise I'd say let's nail it for shows first.

src/schema/__tests__/city.test.ts
alloy: Yep. Not just _a_ large number, though, the max int that the GraphQL spec allows for. Max ints are often used as sentinels to e.g. mean ‘everything’. I definitely did not want to add a parameter (such as `all`) because removing it in the future would be breaking and it may suggest that this is a strategy we approve of in general.

alloy:
I'll expand this PR with fairs too 👍

alloy:
Ok, merging this.

joeyAghion:
There are already a few cautious voices here, but I just want to reiterate that it's practically _always_ been a mistake to attempt unbounded loads like this. In certain circumstances (locations with high volumes of data, low quality/duplicate data, slow connections, server-side latency) this will be unhealthy for the back-end systems _and_ not work correctly for the end user.

Sometimes there are design tweaks or practical thresholds that can be applied to mitigate the problem. E.g., can we _sort_ the results by proximity to the user's location (as we do elsewhere), so the early pages are also the most relevant? Can we impose a practical limit on results (like 200) above which we wouldn't want to overwhelm the UI or servers anyway?

alloy:
Let’s please have further discussions on the relevant Jira ticket(s) https://artsyproduct.atlassian.net/browse/LD-282

src/schema/city/index.ts
```diff
@@ -52,68 +55,37 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
...
-        return results
-      },
+          sort: args.sort,
+          // default Enum value for status is not properly resolved
+          // so we have to manually resolve it by lowercasing the value
+          // https://github.com/apollographql/graphql-tools/issues/715
+          status: args.status.toLowerCase(),
+          // This is to ensure the key never makes its way into the `baseParams`
+          // object if not needed,
+          ...(typeof args.discoverable === "undefined"
```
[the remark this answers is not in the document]
anandaroop: Nvm I think I read this wrong, and per https://artsy.slack.com/archives/CFNFWD5F0/p1551803155107100 we are seeing stub shows in results.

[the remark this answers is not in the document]
alloy: Are you _sure_ you were wrong or do we still need to go over this?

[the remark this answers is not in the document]
anandaroop: Pretty sure I was wrong based on (a) re-reading carefully 🤓 , (b)  results IRL and (c) [these tests](https://github.com/artsy/metaphysics/blob/1c1145abd24a3e2c572d672a88dab57d50d4d94f/src/schema/__tests__/city.test.ts#L146-L174) still being green. And happy to be wrong too!

PR #1534, opened 2019-02-28
Fix wrong artwork filtering on show page

ashkan18:
fixes https://artsyproduct.atlassian.net/browse/LD-300

We were passing wrong param to Gravity when filtering by show.


orta:
Yep, cool 👍

ashkan18:
alright, fixing test

PR #1536, opened 2019-03-01
Adds exhibition_period to Fairs

kierangillen:
- Provides `exhibition_period` to fairs for formatted date range

### DIFF
--- a/src/schema/__tests__/fair.test.js
+++ b/src/schema/__tests__/fair.test.js
@@ -130,8 +130,11 @@ describe("Fair", () => {
         artists_count: 1,
         artworks_count: 2,
         partners_count: 3,
+        start_at: "2019-02-15T23:00:00+00:00",
+        end_at: "2019-02-17T11:00:00+00:00",
         partner_shows_count: 4,
         name: "Aqua Art Miami 2018",
+        exhibition_period: "Feb 15 – 17",
         exhibitors_grouped_by_name: [
           {
             letter: "A",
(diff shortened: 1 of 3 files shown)


orta:
Yep!

xtina-starr:
I think something in the might have broken the build...?

https://circleci.com/gh/artsy/metaphysics/4198

src/schema/__tests__/fair.test.js
```diff
@@ -228,6 +231,23 @@ describe("Fair", () => {
...
+  it("includes a formatted exhibition period", async () => {
+    const query = gql`
+      {
+        fair(id: "aqua-art-miami-2018") {
+          exhibition_period
+        }
+      }
+    `
+
+    const data = await runQuery(query, rootValue)
```
alloy: `rootValue` no longer exists in `master`, which was breaking the tests. Will fix that in my PR.

src/schema/__tests__/fair.test.js
```diff
@@ -228,6 +231,23 @@ describe("Fair", () => {
...
+    expect(data).toEqual({
+      fair: {
+        exhibition_period: "Feb 15 – 17",
```
alloy: @kierangillen This test is broken, because the function it uses (`exhibitionPeriod`) is incomplete. It doesn’t take a timezone parameter and thus in my timezone this ends up being rendered as `Feb 16 - 18`. The right fix here is to make `exhibitionPeriod` take a timezone and use `defaultTimezone` from the resolver context.

I'm disabling the test in my PR.

src/schema/__tests__/fair.test.js
xtina-starr: Ah just seeing this! I already opened a PR to fix it because it was blocking one of my PRs. Should I go ahead and close it?

src/schema/__tests__/fair.test.js
kierangillen: I'll create a PR today to fix this test.

src/schema/__tests__/fair.test.js
alloy: Nope, that's fine! 👍

PR #1539, opened 2019-03-04
Cache city shows and fairs for a day

alloy:
Closes https://artsyproduct.atlassian.net/browse/LD-284

This PR also:

* Actually passes one-off API options to the loader, such as `requestThrottleMs` (cc @mzikherman I think you added some of these in the past and so that never really worked previously)
* Disables a broken test, see 
https://github.com/artsy/metaphysics/pull/1536/files#r262078512

### DIFF
--- a/src/lib/loaders/api/loader_without_cache_factory.ts
+++ b/src/lib/loaders/api/loader_without_cache_factory.ts
@@ -1,6 +1,7 @@
 import DataLoader from "dataloader"
 import { loaderInterface } from "./loader_interface"
 import { LoaderFactory } from "../index"
+import { DataLoaderKey } from "./index"
 
 /**
  * This factory provides a short-cut system for our data loader system, it provides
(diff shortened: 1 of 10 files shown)


src/lib/loaders/api/loader_without_cache_factory.ts
```diff
@@ -17,8 +18,16 @@ export const uncachedLoaderFactory = (
...
+    const loader = new DataLoader<DataLoaderKey, any>(
+      keys =>
+        Promise.all(
+          keys.map(({ key, apiOptions }) => {
+            if (apiOptions) {
+              throw new Error("A uncachedLoader does not accept API options.")
+            }
+            return Promise.resolve(api(key, options).then(r => r.body))
+          })
+        ),
```
orta: This is a nice touch

orta:
Cool, yep, this looks good to me 👍

src/lib/loaders/api/loader_without_cache_factory.ts
alloy: 🙇

PR #1543, opened 2019-03-05
[danger] Fail on snake_case additions.

alloy:
This is something we've agreed on for a while now, so might as well have a danger rule for it.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -36,5 +36,6 @@
     "urljoin"
   ],
   "debug.node.autoAttach": "on",
-  "eslint.validate": ["typescript"]
+  "eslint.validate": ["typescript"],
+  "typescript.surveys.enabled": false
 }
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -45,17 +45,20 @@ export default async () => {
 
   const diff = await danger.git.structuredDiffForFile("_schema.graphql")
   if (diff) {
-    const declaredCamel = danger.github.pr.body.includes("#skip_camel")
-    const violation = declaredCamel ? warn : fail
-    var failed = false
+    const skipCamelCaseCheck = danger.github.pr.body.includes("#skip_camel")
+    const violation = skipCamelCaseCheck ? warn : fail
+    let failed = false
     diff.chunks.forEach(chunk => {
       chunk.changes.forEach(change => {
         if (change.type === "add") {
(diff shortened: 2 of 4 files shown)


alloy:
Oh, hrmm, this doesn’t work because `_schema.graphql` is automatically checked in 🤔

_schema.graphql
```diff
@@ -4497,6 +4499,7 @@ type HomePageModuleContextFair {
   # A formatted description of the start to end dates
   exhibition_period: String
   has_full_feature: Boolean
+  has_incorrect_snake_case: Boolean
```
DangerCI: - :no_entry_sign: Found addition using snake_case instead of camelCase: has_incorrect_snake_case

alloy:
@orta Interesting, any idea why the first 2 failures are not inline comments on the diff, but the third is?

orta:
Must be a bug with danger's internal "is this file/line inside the diff" engine. We're quite a few versions behind (it's now 7.0.15 or something) and there has been improvements

_schema.graphql
```diff
@@ -3266,6 +3267,7 @@ type Fair {
...
```
DangerCI: - :no_entry_sign: Found addition using snake_case instead of camelCase: has_incorrect_snake_case

_schema.graphql
```diff
@@ -1115,6 +1115,7 @@ type ArtworkContextFair {
...
```
DangerCI: - :no_entry_sign: Found addition using snake_case instead of camelCase: has_incorrect_snake_case

orta:
Yep, the update fixed it

alloy:
Ace, thanks! I'll finish this up.

alloy:
Ready to go 👍

alloy:
Also created https://github.com/artsy/README/pull/167

orta:
I'm good for this, though I think I'd like an escape of some sort so you can switch it to `warn`, maybe there's things out of your control (or you add docs and it gets included in a diff for example)

alloy:
Is there a builtin danger function that does conditional fail/warn based on a keyword in the OP?

orta:
I added it - see the new dangerfile and if you're happy, merge

alloy:
Yup, looks good. I added a tiny fix and improved copy slightly. Going to merge this 👍

PR #1544, opened 2019-03-05
[trackedEntityLoader] Batch in chunks of 200 IDs.

alloy:
This is to avoid running into a `414 – Request-URI Too Long` from Gravity.

screenshot 2019-03-05 at 12 34 41


joeyAghion:
👀 This worries me (again) about attempting to load unbounded data sets. Is there a limit _in practice_ to the number of records that will be needed here? Can we enforce an overall limit?

PR #1546, opened 2019-03-05
Adds exhibition_period to partner_show_events

kierangillen:
- Need to add this to prevent date formatting in emission client

screen shot 2019-03-05 at 2 28 17 pm


alloy:
I’m not sure I fully understand this. Is this related to that other case which was MP formatting the period and here it’s just a pass-through?

Also, can you rename the field to use `camelCase` please?

orta:
Tsk tsk on new tests, but that's pretty fine for the moment 👍

PR #1547, opened 2019-03-05
Add static sponsored content

anandaroop:
https://artsyproduct.atlassian.net/browse/LD-316

This PR introduces a home for the little bits of static sponsored content — prose descriptions, hyperlinks — that need to show up in certain client contexts such as the city view and the fair view in Eigen.

The content is stored in a [json file](https://github.com/artsy/metaphysics/commit/5f3d301323d707b0fa8179456115c92e38ffa97b#diff-545f2f7a00f714e6b0739c868b93f0b5), and made available through a `sponsoredContent` field that is now available on `city` and `fair`:

sponsored

If these fields are requested when there is no content, it's just `null`:

nullsc

### DIFF
--- a/src/lib/sponsoredContent/data.json
+++ b/src/lib/sponsoredContent/data.json
@@ -1,0 +1,52 @@
+{
+  "cities": {
+    "new-york-ny-usa": {
+      "introText": "The BMW Art Guide is your go-to-guide to discover private collections of modern and contemporary art which are accessible to the public. Please see below for collections in New York.",
+      "artGuideUrl": "https://www.bmw-art-guide.com/",
+      "partnerIds": []
+    },
+    "london-united-kingdom": {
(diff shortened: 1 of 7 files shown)


src/lib/sponsoredContent/data.json
```diff
@@ -0,0 +1,44 @@
+{
+  "cities": {
+    "new-york-ny-usa": {
+      "introText": "The BMW Art Guide is your go-to-guide to discover private collections of modern and contemporary art which are accessible to the public. Please see below for collections in New York.",
```
anandaroop: We discussed using a placeholder here and substituting the current city name at runtime. But in the end I opted for repetitive and simple, vs. concise and clever. This way, each city's text can be editied independently as well. Easy enough to revisit if that turns out to be the wrong call.

kierangillen:
Thanks @anandaroop !

PR #1548, opened 2019-03-05
Exclude fair-booths from shows in a city

ashkan18:
# Problem
When getting shows in a city, we were getting fair booths, these would show up as shows with location id but without coordinates.

image

# Solution
Gravity provides a param for filtering out fair booths, we now by default set `at_a_fair` flag to `false` for these queries.

### DIFF
--- a/src/schema/__tests__/city.test.ts
+++ b/src/schema/__tests__/city.test.ts
@@ -150,6 +150,18 @@ describe("City", () => {
       expect(gravityOptions).toMatchObject({ displayable: true })
       expect(gravityOptions).not.toHaveProperty("discoverable")
     })
+    it("requests shows with location, by default", async () => {
+      await runQuery(query, context)
+      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
+
+      expect(gravityOptions).toMatchObject({ has_location: true })
+    })
+    it("excludes fair booths, by default", async () => {
+      await runQuery(query, context)
+      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
+
+      expect(gravityOptions).toMatchObject({ at_a_fair: false })
+    })
 
     it("can filter by discoverable shows", async () => {
       query = gql`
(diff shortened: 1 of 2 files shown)


src/schema/__tests__/city.test.ts
```diff
@@ -150,6 +150,20 @@ describe("City", () => {
       expect(gravityOptions).toMatchObject({ displayable: true })
       expect(gravityOptions).not.toHaveProperty("discoverable")
     })
+    it("requests shows with location, by default", async () => {
+      await runQuery(query, context)
+      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
+
+      expect(gravityOptions).toMatchObject({ has_location: true })
+      expect(gravityOptions).not.toHaveProperty("discoverable")
```
anandaroop: Let's delete the assertions L158 here and L165 below, since they are not particularly relevant to these two tests.

```suggestion
```

src/schema/__tests__/city.test.ts
```diff
@@ -150,6 +150,20 @@ describe("City", () => {
...
+    })
+    it("requests shows not at a fair, by default", async () => {
+      await runQuery(query, context)
+      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
+
+      expect(gravityOptions).toMatchObject({ at_a_fair: false })
+      expect(gravityOptions).not.toHaveProperty("discoverable")
```
anandaroop:
```suggestion
```

src/schema/__tests__/city.test.ts
```diff
@@ -150,6 +150,20 @@ describe("City", () => {
...
       expect(gravityOptions).not.toHaveProperty("discoverable")
     })
+    it("requests shows with location, by default", async () => {
+      await runQuery(query, context)
+      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
+
+      expect(gravityOptions).toMatchObject({ has_location: true })
+      expect(gravityOptions).not.toHaveProperty("discoverable")
+    })
+    it("requests shows not at a fair, by default", async () => {
```
anandaroop: Suggested wording tweak to convey the intent a little more clearly:

```suggestion
    it("excludes fair booths, by default", async () => {
```

ashkan18: nice! haha i wonder if i can commit this suggestions in one commit!

anandaroop: Oh yeah, I guess each one of these suggestions kicks off a new build. Feel free to ignore the GH suggestions and push up your own commit too!

ashkan18: 🙌 updated

PR #1550, opened 2019-03-06
[rateLimiter] Make disabling with 0 actually work.

alloy:
Closes https://artsyproduct.atlassian.net/browse/PLATFORM-1265
Closes https://artsyproduct.atlassian.net/browse/LD-306


src/config.ts
```diff
@@ -105,6 +105,17 @@ Object.keys(mustHave).forEach(key => {
   }
 })
 
+/**
+ * Use this if "0" should be respected as 0 and only use the default value for
+ * other falsy values, such as `undefined`, `null`, or an empty string.
+ *
+ * Because JS treats 0 as falsy, the following returns 42: `Number("0") || 42`
```
alloy: lol, JS

joeyAghion: ❗️ 

I see a bunch of other `Number(...)` uses below, but none that seem to run into this same problem today. Good catch!

alloy: Yeah I figured we'll know the issue when we run into it again and then update appropriately.

PR #1552, opened 2019-03-07
Add includeStubShows and displayable and deprecate discoverable

ashkan18:
Depends on https://github.com/artsy/gravity/pull/12191

# Change

Deprecate current `discoverable` argument and for time being resolve that to `displayable: true` and `include_local_discovery: true`.

Introduce new `includeStubShows` and `displayable` as alternative of `discoverable`.

### DIFF
--- a/src/schema/__tests__/city.test.ts
+++ b/src/schema/__tests__/city.test.ts
@@ -221,29 +221,6 @@ describe("City", () => {
       })
     })
 
-    it("can filter by displayable", async () => {
-      query = gql`
-        {
-          city(slug: "sacramende-ca-usa") {
-            name
-            shows(first: 1, displayable: true) {
-              edges {
-                node {
-                  id
-                }
-              }
-            }
-          }
-        }
-      `
-      await runQuery(query, context)
-      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
-
-      expect(gravityOptions).toMatchObject({
-        displayable: true,
-      })
-    })
-
     describe("filtering by single partner type", () => {
       it("can filter to gallery shows", async () => {
         query = gql`
--- a/src/schema/__tests__/city.test.ts
+++ b/src/schema/__tests__/city.test.ts
@@ -217,6 +217,7 @@ describe("City", () => {
 
       expect(gravityOptions).toMatchObject({
         include_local_discovery: true,
+        displayable: true,
       })
     })
 
--- a/src/schema/__tests__/city.test.ts
+++ b/src/schema/__tests__/city.test.ts
@@ -191,8 +191,56 @@ describe("City", () => {
       await runQuery(query, context)
       const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
 
-      expect(gravityOptions).toMatchObject({ discoverable: true })
-      expect(gravityOptions).not.toHaveProperty("displayable")
+      expect(gravityOptions).toMatchObject({
+        include_local_discovery: true,
+        displayable: true,
+      })
+    })
+
+    it("can ask for inclduing local_discovery shows", async () => {
+      query = gql`
+        {
+          city(slug: "sacramende-ca-usa") {
+            name
+            shows(first: 1, includeLocalDiscovery: true) {
+              edges {
+                node {
+                  id
+                }
+              }
+            }
+          }
+        }
+      `
+      await runQuery(query, context)
+      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
+
+      expect(gravityOptions).toMatchObject({
+        include_local_discovery: true,
+      })
+    })
+
+    it("can filter by displayable", async () => {
+      query = gql`
+        {
+          city(slug: "sacramende-ca-usa") {
+            name
+            shows(first: 1, displayable: true) {
+              edges {
+                node {
+                  id
+                }
+              }
+            }
+          }
+        }
+      `
+      await runQuery(query, context)
+      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
+
+      expect(gravityOptions).toMatchObject({
+        displayable: true,
+      })
     })
 
     describe("filtering by single partner type", () => {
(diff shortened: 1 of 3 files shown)


src/schema/__tests__/city.test.ts
```diff
@@ -191,8 +191,56 @@ describe("City", () => {
...
 
-      expect(gravityOptions).toMatchObject({ discoverable: true })
-      expect(gravityOptions).not.toHaveProperty("displayable")
+      expect(gravityOptions).toMatchObject({
+        include_local_discovery: true,
+        displayable: true,
+      })
+    })
+
+    it("can ask for inclduing local_discovery shows", async () => {
```
orta:
```suggestion
    it("can ask for including local_discovery shows", async () => {
```

src/schema/__tests__/city.test.ts
```diff
@@ -191,8 +191,56 @@ describe("City", () => {
...
+            }
+          }
+        }
+      `
+      await runQuery(query, context)
+      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
+
+      expect(gravityOptions).toMatchObject({
+        displayable: true,
+      })
```
orta: I think this can use:

```
expect(gravityOptions).toHaveBeenCalledWith(expect.objectContaining({ displayable: true }))
```

instead of the two liner

ashkan18:
@alloy @anandaroop should i rename `inlcudeLocalDiscovery` to `includeStubbedShows`?

alloy:
I think @anandaroop suggested `includeStubShows`, which I think is _better_, but still wonder if we can’t spell it out more, e.g. `includeExternallySourcedShows` ?

I'm naive on the exact details so please correct me if I’m wrong, and also I’m open to the idea of me just not understanding that ‘stub’ captures this well enough to native English speakers.

anandaroop:
I proposed `includeStubShows` because that's the domain language the team has settled on colloquially (devs and non-devs). So to me it seems like the least ambiguous term. 

I get that `includeExternallySourcedShows` is super explicit but it's not the language we use. That's why I vote for `includeStubShows`.

(But I can live with either.)

ashkan18:
cool, i went with `includeStubShows` knowing i will have all sort of typos typing `includeExternallySourcedShows` 😄

src/schema/city/index.ts
```diff
@@ -67,14 +77,9 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
...
-          // object if not needed,
-          ...(typeof args.discoverable === "undefined"
-            ? undefined
-            : { discoverable: args.discoverable }),
-          ...(typeof args.discoverable === "undefined"
-            ? { displayable: true }
-            : undefined),
+          include_local_discovery:
+            args.discoverable === true || args.includeStubShows,
+          displayable: args.discoverable === true || args.displayable,
```
alloy: Just to be sure, this means the `displayable` key will _always_ be included in the params included in the Gravity request; meaning that if _omission_ of the param has any significance in the endpoint then that would no longer work as expected.

ashkan18:
based on https://github.com/artsy/metaphysics/pull/1552/files#diff-0aa0321fab087654cb6a867c724508d6R156 it seemed it was always by default sending it

src/schema/__tests__/city.test.ts
```diff
@@ -191,8 +191,56 @@ describe("City", () => {
...
+              }
+            }
+          }
+        }
+      `
+      await runQuery(query, context)
+      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
+
+      expect(gravityOptions).toMatchObject({
+        include_local_discovery: true,
```
anandaroop: We'd also expect         `displayable: true` here yeah?

src/schema/city/index.ts
anandaroop: We don't ever want to show non-displayable shows (where _displayable_ is a longstanding concept having to do with how many artworks/install shots a show has) — unless they are our own Artsy-created stub shows.

So the fact that…

- `displayable=true` is _always_ sent to Gravity, and 
- `include_local_discovery=true` is _sometimes_ sent (in fact always sent for LD use cases)

… sounds 👍  to me.

What I question is why we have `displayable` as an arg to MP at all on L57. I think we can get rid of that, since it would have no effect as @alloy points out, and doesn't support any actual use case.

src/schema/city/index.ts
```diff
@@ -50,10 +50,20 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
...
+          defaultValue: true,
+          description: "Filter by displayable shows",
+        },
         discoverable: {
           type: GraphQLBoolean,
-          description:
-            "Whether to include local discovery stubs as well as displayable shows",
+          description: "Whether to include stub shows or not",
+          deprecationReason:
+            "Use `includeStubShows` and `displayable` combination",
```

anandaroop: Oh wait I guess based on other comments, this line might be misleading. 

Should be sufficient to say "Use `includeStubShows`" I believe?

ashkan18:
@anandaroop good point about displayable, going to remove that arg and we can add it later if needed 🙌

src/schema/city/index.ts
```diff
@@ -50,10 +50,20 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
...
           description: "By default we filter shows by current",
         },
+        includeStubShows: {
+          type: GraphQLBoolean,
+          description: "Whether to include local discovery stubs",
+        },
+        displayable: {
+          type: GraphQLBoolean,
+          defaultValue: true,
+          description: "Filter by displayable shows",
```
anandaroop: The changes in the last commit are good, but I think I was also suggesting getting rid of this arg, since it's now superfluous, right?

If so let's delete L57-61 ✂️

ashkan18: oh i thought i did 🤦‍♂️ let me do it now

ashkan18: done! (or i think "done")

PR #1553, opened 2019-03-07
Display DaySchedule union

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/LD-221
We already have 3 fields describing day schedules in different ways for a `Location`

- `day_schedules`: Array of `DayScheduleType` set mostly by partners themselves
- `day_schedule_text`: A string set by artsy admins
- `displayDaySchedules`: currently a field that formats `day_schedules` data to array of `FormattedDaySchedules`

Currently clients need to decide which field to use, we want to have MP be the source of truth and have the priority logic about which one to use?

# Solution
Added a new `openingHours` field which is a union of `OpeningHoursArray` and `OpeningHoursText`, if we have `day_schedules` we resolve this field to `OpeningHoursArray` otherwise we use `OpeningHoursText`.

```graphql
{
  show(id: "galerie-crone-emmanuel-bornstein-clinamen") {
    id
    location{
      day_schedule_text
      day_schedules {
        start_time
        end_time
        day_of_week
      }
      openingHours{
        ... on OpeningHoursArray {
          schedules{
            days
            hours
          }
        }
        ... on OpeningHoursText {
          text
        }
      }
    }
  }
}
```

# Question
Currently there doesn't seem to be a way to have a union of Scalar Types, right? something like
```
[FormattedDaySchedule] | GraphQLString
``` 
if we could have ☝️ then the change in emission would have been minimal since it already checks the type

# Selfie
image

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -3077,6 +3077,10 @@ type DaySchedule {
   day_of_week: String
 }
 
+type DayScheduleText {
+  text: String
+}
+
 type DeepZoom {
   Image: DeepZoomImage
 }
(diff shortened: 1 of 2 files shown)


ashkan18:
makes sense @alloy, i was hoping to not introduce another possible field for showing the same things but looks like thats not easily doable :)
going to deprecate existing display field and add a new one 🙌

PR #1556, opened 2019-03-07
[allViaLoader] No longer need to fetch any content to get /shows count.

alloy:
Fixed by https://github.com/artsy/gravity/pull/12196
Related to https://artsyproduct.atlassian.net/browse/LD-305

Still want to test this out first when Gravity is deployed to staging.


alloy:
Ok, confirmed that with my Gravity fix this now works as expected 👍

PR #1559, opened 2019-03-08
Add partner-type filtering to city shows field

anandaroop:
Closes: https://artsyproduct.atlassian.net/browse/LD-315
Implements subtask: https://artsyproduct.atlassian.net/browse/LD-333

This closes the loop on partner type bucketing in the city shows query (following https://github.com/artsy/gravity/pull/12190 and https://github.com/artsy/gravity/pull/12198).

museum

PR #1560, opened 2019-03-08
Allows null partner type for show parameters

ashfurrow:
Discussed here: 
https://github.com/artsy/emission/pull/1385#discussion_r263934065 Follow up from #1559.

When making a query with a null `partnerType` (something the schema allows), Gravity was throwing an error:

```
https://stagingapi.artsy.net/api/v1/shows?at_a_fair=false&displayable=true&has_location=true&include_local_discovery=true&max_distance=25&near=51.51%2C-0.13&page=3&partner_types=&size=10&sort=start_at&status=current&total_count=true - {\"type\":\"param_error\",\"message\":\"Invalid parameters.\",\"detail\":{\"partner_types\":[\"does not have a valid value\"]}}
```

This PR changes metaphysics to only send the `partnerType` param to Gravity if it's truthy.

### DIFF
--- a/src/schema/city/index.ts
+++ b/src/schema/city/index.ts
@@ -82,7 +82,7 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
           max_distance: LOCAL_DISCOVERY_RADIUS_KM,
           has_location: true,
           at_a_fair: false,
-          partner_types: args.partnerType,
+          ...(args.partnerType ? { partner_types: args.partnerType } : {}),
           sort: args.sort,
           // default Enum value for status is not properly resolved
           // so we have to manually resolve it by lowercasing the value
--- a/src/schema/city/index.ts
+++ b/src/schema/city/index.ts
@@ -82,7 +82,7 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
           max_distance: LOCAL_DISCOVERY_RADIUS_KM,
           has_location: true,
           at_a_fair: false,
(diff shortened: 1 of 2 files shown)


src/schema/city/index.ts
```diff
@@ -82,7 +82,7 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
           max_distance: LOCAL_DISCOVERY_RADIUS_KM,
           has_location: true,
           at_a_fair: false,
-          partner_types: args.partnerType,
+          ...(args.partnerType ? { partner_types: args.partnerType } : {}),
```
alloy: FYI `{ ...undefined }` works fine too and would eliminate the need for the empty object returned from your ternary.

ashfurrow: Oh I like that better, good idea.

orta: new to me also

ashfurrow: Since `falsy && whatever` returns `undefined`, I used this:

```ts
 ...(args.partnerType && { partner_types: args.partnerType })
```

Which is just 😚👌

alloy: 🔥

alloy: PS @ashfurrow I’m sure this is what you meant, but just in case: `T && whatever` returns `T`, where `T` is a falsy value. I.e. it will only return `undefined` when `T` is `undefined`.

PR #1562, opened 2019-03-09
Add optional dayThreshold and new ClosingSoon status

ashkan18:
https://artsyproduct.atlassian.net/browse/LD-337

# Changes
Added new `dayThreshold` argument to `city { shows (...)}` which can be used to filter shows in `UPCOMING` and `CLOSING_SOON` states.

# Depends on
https://github.com/artsy/gravity/pull/12204

### DIFF
--- a/src/schema/__tests__/city.test.ts
+++ b/src/schema/__tests__/city.test.ts
@@ -221,7 +221,7 @@ describe("City", () => {
       })
     })
 
-    it("can filter to gallery shows", async () => {
+    it("can filter to shows by status and dayThreshold", async () => {
       query = gql`
         {
           city(slug: "sacramende-ca-usa") {
--- a/src/schema/__tests__/city.test.ts
+++ b/src/schema/__tests__/city.test.ts
@@ -221,6 +221,30 @@ describe("City", () => {
       })
     })
 
+    it("can filter to gallery shows", async () => {
+      query = gql`
+        {
+          city(slug: "sacramende-ca-usa") {
+            name
+            shows(first: 1, status: CLOSING_SOON, dayThreshold: 5) {
+              edges {
+                node {
+                  id
+                }
+              }
+            }
+          }
+        }
+      `
+      await runQuery(query, context)
+      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
+
+      expect(gravityOptions).toMatchObject({
+        day_threshold: 5,
+        status: "closing_soon",
+      })
+    })
+
     describe("filtering by partner type", () => {
       it("can filter to gallery shows", async () => {
         query = gql`
--- a/src/schema/city/index.ts
+++ b/src/schema/city/index.ts
@@ -4,6 +4,7 @@ import {
   GraphQLObjectType,
   GraphQLString,
   GraphQLFieldConfig,
+  GraphQLInt,
 } from "graphql"
 
 import { LatLngType } from "../location"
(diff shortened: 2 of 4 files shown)


src/schema/city/index.ts
```diff
@@ -83,6 +89,7 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
           has_location: true,
           at_a_fair: false,
           ...(args.partnerType && { partner_types: args.partnerType }),
+          ...(args.dayThreshold && { day_threshold: args.dayThreshold }),
```
alloy: Gotta watch out with this if you expect `0` to be a valid value, as JS treats it as falsy and thus you’d try to spread `0` into the object, which I’m sure is invalid—although I have no idea off-hand what would happen, so perhaps test it in a node REPL.

src/schema/__tests__/city.test.ts
```diff
@@ -221,6 +221,30 @@ describe("City", () => {
       })
     })
 
+    it("can filter to gallery shows", async () => {
```
anandaroop: Let's change the name of the spec 😉

src/schema/city/index.ts
anandaroop: Good thinking @alloy.

Couple of thoughts: passing a 0 might be allowed, but would pretty much by definition return an empty result set.

So it would be both permitted and useless. However — I don't think it would even get sent over to G in the first place:

REPL test:

```js
// here is a normal threshold value

> args = {dayThreshold: 42}
{ dayThreshold: 42 }
> gravityArgs = { foo: "bar", ...(args.dayThreshold && { day_threshold: args.dayThreshold }) } 
{ foo: 'bar', day_threshold: 42 }
> 
> 
// here is the questionable 0 threshold value

> args = {dayThreshold: 0}
{ dayThreshold: 0 }
> gravityArgs = { foo: "bar", ...(args.dayThreshold && { day_threshold: args.dayThreshold }) } 
{ foo: 'bar' }
```

Seems safe to me as-is?

orta: yah, looks it 👍

ashkan18: thanks all! it should also be safe from Gravity side 🙌 thanks for checking

src/schema/__tests__/city.test.ts
ashkan18: 🤦‍♂️ CMD+C/CMD+V

ashkan18:
thanks all, merged on green'ed it 🍏

src/schema/city/index.ts
alloy: Thanks for verifying @anandaroop! 🙌

alloy: For completeness I tested the other falsy values I know of:

```
> { foo: "bar", ...("" && { baz: "bla" }) }
{ foo: 'bar' }
> { foo: "bar", ...(Number.NaN && { baz: "bla" }) }
{ foo: 'bar' }
> { foo: "bar", ...(false && { baz: "bla" }) }
{ foo: 'bar' }
> { foo: "bar", ...(null && { baz: "bla" }) }
{ foo: 'bar' }
```

So I guess we can conclude all falsy values give us our preferred behaviour.

ashfurrow: Numbers appear to as well, not just zero.

alloy: @ashfurrow Hmm, not sure I follow:

```
> { foo: "bar", ...(42 && { baz: "bla" }) }
{ foo: 'bar', baz: 'bla' }
```

ashfurrow: Right, since 42 isn’t falsy. But spreading over 42 works: `{...(42)} = {}`, which is kinda neat!

alloy: Ohhh, good to know! I guess anything non-iterable gets ignored

PR #1563, opened 2019-03-09
Include manually curated shows in per-city sponsored content

anandaroop:
https://artsyproduct.atlassian.net/browse/LD-387

This adds a `shows` connection within `city.sponsoredContent` so that the sponsor-curated shows can be served up by MP.

Screen Shot 2019-03-10 at 1 59 48 AM

The list of ids is hard-coded for now — see the Jira ticket for more details.

I started a discussion below about trying to optimize some of these queries for compatibility with future client versions, but I'm not sure that my approach makes sense. So I've resolved the discussion and removed the commit in question. Feel free to read through if you have opinions on this though.


_schema.graphql
```diff
@@ -2477,6 +2477,8 @@ type City {
 type CitySponsoredContent {
   introText: String
   artGuideUrl: String
+  featuredShows: [Show]
+  shows(size: Int, sort: PartnerShowSorts, status: EventStatus): [Show]
```
anandaroop: So this `featuredShows` vs `shows` business...

In the designs we have these content requirements…

Screen Shot 2019-03-09 at 6 13 13 PM


For the featured shows, seems like there are a couple of approaches.

1. Just use the `shows` field in both places

x2


2. Have a dedicated `featuredShows` field (which is the third commit in this PR)

featured shows

Seems to me the advantages of approach 2 is that it allows us to fiddle with the definition of what "featured shows" means server-side, without Emission having to worry about those details — and thus not constraining us to Emission release cycles, esp v1 vs v1.1

But that also seems like a slippery slope we could go down for any number of fields, so I'm not sure if it's the right idea. 

Thoughts?

src/schema/city/index.ts
```diff
@@ -115,6 +117,35 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
...
           },
+          featuredShows: {
+            type: new GraphQLList(ShowType),
+            resolve: (citySponsoredContent, _args, { showsLoader }) => {
+              let gravityOptions = {
+                id: citySponsoredContent.showIds,
+                include_local_discovery: true,
+                status: "current",
+                sort: "end_at",
+                size: 2,
```
anandaroop: Here's the canned query that captures our definition of "featured shows" so that Emission never has to bother with it. 

🤷‍♂️ ?

_schema.graphql
ashfurrow: @anandaroop I don't quite understand the first example – would we add a(n option) parameter to the `shows` resolver requesting sponsored content? 

You raise a really good point about forward-compatible schemas. I think if there's a decent chance that we'll want to include supplementary info beyond the basic `Show` type, we could wrap the `Show` inside a `FeaturedShow` object (even if it only has a single field for now). In that case, we'd need to use a different field name in `sponsoredContent` than `shows()`). But it really depends on what we think we'll want to do, design-wise, after v1.

anandaroop: It's not that I anticipate that we'll be adding more info to these featured show bricks. I think rather I'm anticipating that we'll want to change the criteria for _which_ shows we'll feature in these bricks. And that we might want to do so in a way that newer client versions will just work.

But this is all speculative, so maybe in the interest of keeping things moving I'll get rid of the special `city.sponsoredContent.featuredShows` field, and instead let consumers use the more general `city.sponsoredContent.shows`, as in the screencap for approach # 1 above:

shows x22

The worse case is that v1 clients will show different shows in those preview bricks. Probably not the end of the world.

src/schema/city/index.ts
```diff
@@ -115,6 +117,25 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
...
+              sort: PartnerShowSorts,
+              status: EventStatus,
+            }),
+            resolve: async (
+              citySponsoredContent,
+              args,
+              { showsWithHeadersLoader }
+            ) => {
+              return loadData(args, showsWithHeadersLoader, {
+                id: citySponsoredContent.showIds,
```
ashkan18: nice! so this is why we don't need fetch by `partner_ids` anymore right?

src/schema/city/index.ts
```diff
@@ -115,6 +117,25 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
...
+                include_local_discovery: true,
```
ashkan18: do we need `has_location: true,` and `displayable: true` here or we assume those ids are good for being returned even if they are not displayable and don't have location?

src/schema/city/index.ts
anandaroop: Exactly 🎯

src/schema/city/index.ts
anandaroop: I think those flags are not necessary for this use case.

`has_location` isn't necessary because this query's purpose is to populate a list, not put pins on the map. (In fact you'll notice there's no `near` clause here either. That's because we specifically want to allow these to show up for their associated city, whether or not they are within the normal LD radius.)

`displayable: true` — some of these will be stub shows and will not have enough install shots to be officially `displayable`, so I don't think we need that, but I will double check and follow up. Thanks for raising.

ashfurrow: That makes sense, thanks Roop!

src/schema/__tests__/city.test.ts
```diff
@@ -386,23 +387,70 @@ describe("City", () => {
...
+            }
+          }
+        }
+      `
 
-    expect(result!.city.sponsoredContent).toEqual({
-      introText: "Lorem ipsum dolot sit amet",
-      artGuideUrl: "https://www.example.com/",
+      const result = await runQuery(query, context)
+      const gravityOptions = context.showsWithHeadersLoader.mock.calls[0][0]
```
alloy: FYI You can also use Jest’s mocks like so `expect(mock).toHaveBeenCalledWith(…)`

PR #1566, opened 2019-03-11
[City] Sort city data for consumption by Emission.

alloy:
Part of https://artsyproduct.atlassian.net/browse/LD-335

### DIFF
--- a/src/schema/city/index.ts
+++ b/src/schema/city/index.ts
@@ -13,7 +13,7 @@ import PartnerShowSorts from "schema/sorts/partner_show_sorts"
 import { FairType } from "schema/fair"
 import FairSorts from "schema/sorts/fair_sorts"
 import EventStatus from "schema/input_fields/event_status"
-import cityData from "./city_data.json"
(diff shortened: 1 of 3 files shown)


alloy:
Merge on green

src/schema/city/index.ts
```diff
@@ -201,7 +204,7 @@ const nearestCity = (latLng: LatLng) => {
 }
 
 const citiesOrderedByDistance = (latLng: LatLng): Point[] => {
-  let cities: Point[] = Object.values(cityData)
+  let cities: Point[] = Object.values(cityDataSortedByDisplayPreference)
```
anandaroop: Are we deliberately calling  `Object.values` on an array here? Looked odd to me, but the REPL says that actually works 🤷‍♂️

alloy: Hah, wowzers! So we learn.

I guess TS realised that it was fine and that’s why I missed it too. Even though it works, I’ll send a lil cleanup PR in a bit.

PR #1568, opened 2019-03-11
Fixes shows resolver to accept null status/dayThreshold values

ashfurrow:
The schema permits sending `null` values, and Emission needs to do that so we can use a single query renderer for city section lists (ie: upcoming, galleries, museums, closing soon).

### DIFF
--- a/src/schema/__tests__/city.test.ts
+++ b/src/schema/__tests__/city.test.ts
@@ -246,6 +246,31 @@ describe("City", () => {
       })
     })
 
+    it("works with null status and dayThreshold", async () => {
+      query = gql`
+        {
+          city(slug: "sacramende-ca-usa") {
+            name
+            shows(first: 1, status: null, dayThreshold: null) {
+              edges {
+                node {
+                  id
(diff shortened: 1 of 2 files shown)


src/schema/__tests__/city.test.ts
```diff
@@ -246,6 +246,31 @@ describe("City", () => {
       })
     })
 
+    it("can works with null status and dayThreshold", async () => {
```
alloy: “It can works”? 🇬🇧

src/schema/__tests__/city.test.ts
```diff
@@ -246,6 +246,31 @@ describe("City", () => {
...
+          }
+        }
+      `
+      await runQuery(query, context)
+
+      expect(context.showsWithHeadersLoader).not.toHaveBeenCalledWith(
+        expect.objectContaining({
+          day_threshold: expect.anything(),
+          status: expect.anything(),
+        })
```
alloy: Nice 👌

PR #1569, opened 2019-03-11
Add support for PARTNER_ASC sort option for partner_shows

ashkan18:
# Change
Add support for new `PARTNER_ASC` sort for `PartnerShow` type.

At the moment Gravity doesn't support it yet but this is so Emission could start using it while we figure out how to support it on Gravity end. For now it will just do a `name` sort. 

https://artsyproduct.atlassian.net/browse/LD-380

### DIFF
--- a/src/schema/sorts/partner_show_sorts.ts
+++ b/src/schema/sorts/partner_show_sorts.ts
@@ -76,7 +76,7 @@ const PartnerShowSorts = {
       },
       PARTNER_ASC: {
         // TODO: Change to proper order once Gravity supports this type of sort
-        value: "name",
+        value: "end_at",
       },
     },
   }),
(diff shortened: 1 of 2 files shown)


orta:
👍

src/schema/sorts/partner_show_sorts.ts
```diff
@@ -74,6 +74,10 @@ const PartnerShowSorts = {
       START_AT_DESC: {
         value: "-start_at",
       },
+      PARTNER_ASC: {
+        // TODO: Change to proper order once Gravity supports this type of sort
+        value: "name",
```
anandaroop: This is temporary of course, but for now should this just fall back to whatever our 2nd choice is for sorting (or whatever we're using right now)? 

If so, we might want `end_at` instead of `name`. I'm not positive so maybe we can get product to weigh in. 

Otherwise - awesome!

ashkan18: oh didn't know whats the default sort, let me update.

PR #1571, opened 2019-03-11
Update emission querymap JSON

l2succes:
Updates query map #trivial


alloy:
We're not running any tests against this, so merging it now as I need this in.

PR #1573, opened 2019-03-12
Add optional `city` argument to me.followsAndSaves.shows

mbilokonsky:
This allows consumers of the `me.followsAndSaves.shows` query to use an optional `city` argument. This argument must be one of the pre-determined city slugs that are tracked in `metaphysics/src/schema/city/cityDataSortedByDisplayPreference.json`, and Metaphysics will return an error that enumerates valid city slugs if an invalid string is passed in.

## Data Creation
First, we've got to create some mock data. Let's create two shows, one in New York and one in Hong Kong, and then follow both with the User we're using in GraphiQL. As this PR, the following data already exists - feel free to lookup and reuse for your own testing, or feel free to make your own.

```ruby
user = User.find_by_email("mykola.bilokonsky@artsymail.com").first
location_nyc = PartnerLocation.create!(name: 'nyc_location', coordinates:  [-74, 40.5])
location_hk = PartnerLocation.create!(name: 'hk_location', coordinates: [114.1, 22.3])
  
partner = Partner.create!(given_name: 'ld_test')

nyc_show = PartnerShow.create!(name: 'nyc-based-show', partner: partner, start_at: 10.years.ago, end_at: 10.years.from_now, partner_location: location_nyc)
hk_show = PartnerShow.create!(name: 'hk-based-show', partner: partner, start_at: 10.years.ago, end_at: 10.years.from_now, partner_location: location_hk)

nyc_follow = FollowShow.create!(user: user, partner_show: nyc_show)
hk_follow = FollowShow.create!(user: user, partner_show: hk_show)
```

## GraphiQL Example
The following query tests four different cases - we first make sure that our user is indeed following both of the shows we care about. We then filter to NYC and confirm that we get the one correct show there. We then filter to Hong Kong and confirm that we get the correct one there. Finally, we attempt to filter to an invalid city string.
```graphql
query {
  me {
    followsAndSaves {
      all_shows: shows(first: 4) {
        edges {
          node {
            id            
          }
        }
      }
      new_york_shows: shows(first: 4, city: "new-york-ny-usa") {
        edges {
          node {
            id
          }
        }
      }
      hong_kong_shows: shows(first: 4, city: "hong-kong-hong-kong") {
        edges {
          node {
            id
          }
        }
      }
      invalid_query: shows(first: 4, city: "random string") {
        edges {
          node{
            id
          }
        }
      }
    }
  }
}
```

To which Metaphysics responds with:
```json
{
  "errors": [
    {
      "message": "City slug must be one of: new-york-ny-usa, los-angeles-ca-usa, london-united-kingdom, berlin-germany, paris-france, hong-kong-hong-kong",
      "locations": [
        {
          "line": 25,
          "column": 7
        }
      ]
    }
  ],
  "data": {
    "me": {
      "followsAndSaves": {
        "all_shows": {
          "edges": [
            {
              "node": {
                "id": "ld-test-nyc-based-show"
              }
            },
            {
              "node": {
                "id": "ld-test-hk-based-show"
              }
            }
          ]
        },
        "new_york_shows": {
          "edges": [
            {
              "node": {
                "id": "ld-test-nyc-based-show"
              }
            }
          ]
        },
        "hong_kong_shows": {
          "edges": [
            {
              "node": {
                "id": "ld-test-hk-based-show"
              }
            }
          ]
        },
        "invalid_query": null
      }
    }
  },
  "extensions": {
    "requests": {
      "gravity": {
        "requests": {
          "follow_shows?max_distance=25&near=22.3%2C114.2&offset=0&size=4&total_count=true": {
            "time": "0s 182.636ms",
            "cache": false,
            "length": "2 KB"
          },
          "follow_shows?max_distance=25&near=40.71%2C-74.01&offset=0&size=4&total_count=true": {
            "time": "0s 233.382ms",
            "cache": false,
            "length": "2 KB"
          },
          "follow_shows?offset=0&size=4&total_count=true": {
            "time": "0s 241.167ms",
            "cache": false,
            "length": "3 KB"
          }
        }
      }
    },
    "requestID": "aec265d0-44f6-11e9-889f-a9e2c4d6beff"
  }
}
```

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -3773,7 +3773,10 @@ type FollowsAndSaves {
 
   # A list of the current user’s currently followed shows
   shows(
-    near: Near
+    # A string representing one of the supported cities in the City Guide, which
+    # are: new-york-ny-usa, los-angeles-ca-usa, london-united-kingdom,
+    # berlin-germany, paris-france, hong-kong-hong-kong
+    city: String
     after: String
     first: Int
     before: String
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -3773,6 +3773,7 @@ type FollowsAndSaves {
 
   # A list of the current user’s currently followed shows
   shows(
+    near: Near
     after: String
     first: Int
     before: String
--- a/src/schema/me/followed_shows.ts
+++ b/src/schema/me/followed_shows.ts
@@ -3,9 +3,17 @@ import { IDFields } from "schema/object_identification"
 
 import { pageable, getPagingParameters } from "relay-cursor-paging"
 import { connectionDefinitions, connectionFromArraySlice } from "graphql-relay"
-import { GraphQLObjectType, GraphQLFieldConfig } from "graphql"
+import { GraphQLObjectType, GraphQLFieldConfig, GraphQLString } from "graphql"
 import { ResolverContext } from "types/graphql"
-import Near from "../input_fields/near"
+import cityData from "../city/cityDataSortedByDisplayPreference.json"
+import { LOCAL_DISCOVERY_RADIUS_KM } from "../city/constants"
+
+const location_by_city_slug = cityData.reduce((acc, val) => {
+  acc[val.slug] = val.coordinates
+  return acc
+}, {})
+
+const getValidCitySlugs = () => Object.keys(location_by_city_slug).join(", ")
 
 const FollowedShowEdge = new GraphQLObjectType<any, ResolverContext>({
   name: "FollowedShowEdge",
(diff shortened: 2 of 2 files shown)


mbilokonsky:
Note here to say that @alloy has pointed out in slack that if we query by lat/long then it'll screw up our caching, and that querying by city slug feels safer. I think that sounds right to me, though I'll let the folks do the querying make the final decision - but even if we do decide to add a slug, let's make that an additive change, right? In a separate PR? It'll require gravity and MP updates as well.

joeyAghion:
If the challenge is that Gravity doesn't "understand" city slugs, another option might be to query by stable coordinates corresponding to city centers.

alloy:
MP already has [the data](https://github.com/artsy/metaphysics/blob/master/src/schema/city/cityDataSortedByDisplayPreference.json) to translate a city slug to an epicenter, so the MP->Gravity code doesn’t need to change right now.

I do have some hesitation about merging this PR as-is, as I worry that exposing this schema right now would be setting up a future Emission developer to doing the “wrong” thing–and in general I’d rather only expose API surface when we really need it–so I’d be in favour of changing the arg now.

mbilokonsky:
Oh damn, @alloy I didn't realize that. I can do city-slug in MP and convert to `near` for the gravity query, that's clearly the best of all worlds. Let me make that change. Thank you for clarifying!

mbilokonsky:
Okay, @alloy and @joeyAghion thank you so much for catching that and pointing me in the right direction. This query now takes a city slug rather than a `near` coordinate set, and it uses internal constants to determine what valid slugs are and how they map to latitude and longitude. The queries to Gravity continue to be lat/long based rather than slug, but because there are only six valid locations with hard-coded lat/long pairs we should never be getting non-cacheable queries coming out of MP for this.

mbilokonsky:
...weird, all tests are green locally. Can someone rerun CI on this PR? It's telling me I don't have access to rerun it?

image

alloy:
@mbilokonsky Circle sometimes has some funky authentication state, which requires me to sign-out and sign back in.

anandaroop:
Thanks @mbilokonsky — this looks good. Definitely the right call to map slugs to lat/lngs before tossing over to Gravity.

mbilokonsky:
@anandaroop go ahead and merge when ready, after talking with @alloy we decided that tests will be in a separate pass later.

alloy:
I’ll take this 👍

alloy:
For posterity, ticket for tests https://artsyproduct.atlassian.net/browse/LD-440

PR #1580, opened 2019-03-13
Adds max bounds per city to constrain LD map panning

anandaroop:
Now that (a) we've consolidated LD geo truth in this file, and (b) we might be ready to tackle map viewport constraints, I'm adding these max bounds to our city file.

These results come from my Observable notebook. Here's the NYC example:

- Purple represents our LD search radius of 25km
- Red represents a buffer of 50km beyond that
- Green represents the bounding box that encompasses the buffer area

In my experience we'd want a generous buffer zone around the area of interest so as not to make the map feel overly confining. But this is all tune-able, of course.

Screen Shot 2019-03-12 at 11 21 53 PM

As for how to specify the actual bounding box rect 🙄 of course there are like seventeen ways to do this.

But according to this ["Restrict map panning to an area"](https://docs.mapbox.com/ios/maps/examples/constraining-gestures/) API doc, the native iOS lib wants the bounds specified as a SW and NE point, with each point specified as a lat,lng object, so that's why this data is shaped the way it is in this version.

### DIFF
--- a/src/schema/city/cityDataSortedByDisplayPreference.json
+++ b/src/schema/city/cityDataSortedByDisplayPreference.json
@@ -2,31 +2,55 @@
   {
     "slug": "new-york-ny-usa",
     "name": "New York",
-    "coordinates": { "lat": 40.71, "lng": -74.01 }
+    "coordinates": { "lat": 40.71, "lng": -74.01 },
+    "maxBounds": {
+      "sw": { "lat": 40.144, "lng": -74.754 },
+      "ne": { "lat": 41.272, "lng": -73.266 }
+    }
   },
   {
     "slug": "los-angeles-ca-usa",
     "name": "Los Angeles",
-    "coordinates": { "lat": 34.05, "lng": -118.24 }
+    "coordinates": { "lat": 34.05, "lng": -118.24 },
+    "maxBounds": {
+      "sw": { "lat": 33.453, "lng": -118.959 },
+      "ne": { "lat": 34.644, "lng": -117.521 }
+    }
   },
   {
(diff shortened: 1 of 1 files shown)


src/schema/city/cityDataSortedByDisplayPreference.json
```diff
@@ -2,31 +2,55 @@
   {
     "slug": "new-york-ny-usa",
     "name": "New York",
-    "coordinates": { "lat": 40.71, "lng": -74.01 }
+    "coordinates": { "lat": 40.71, "lng": -74.01 },
+    "maxBounds": {
+      "sw": { "lat": "40.144", "lng": "-74.754" },
+      "ne": { "lat": "41.272", "lng": "-73.266" }
+    }
```
alloy: Why are the new coordinates strings vs numbers? Or rather, I get that often you want to represent floats and large ints as strings in JSON, but then why are we leaving the epicenter coordinates to use numbers?

anandaroop: Lol, bleary-eyed oversight bc I forgot `Number.toFixed` returns a string. Corrected now.

orta:
Nice, is this easy to recreate when we add new cities?

orta:
looks good to me though

anandaroop:
Yep, adding this for new cities is easy. Effortless even, using [the notebook](https://observablehq.com/d/f051b24b671edf38).

l2succes:
Awesome thanks @anandaroop. we'll be able to do this using the Objective-C API

PR #1581, opened 2019-03-13
add Nginx to metaphysics staging and internal cluster service

izakp:
Set up Nginx as a sidecar container in Metaphysics staging, serving both http and https with artsy.net wildcard certificates.  Also create an internal cluster service for Metaphysics mapping ports 80 and 443 to the Nginx container.  This will allow us to resolve, for example, `metaphysics-staging.artsy.net` to `metaphysics-web-internal.default.svc.cluster.local` by means of Kubernetes' internal DNS server, and transparently route all traffic originating from within the cluster through the cluster-internal service (fixing the "hairpin" traffic pattern we now have, where `metaphysics-staging.artsy.net` resolves to the CNAME of an external ELB, and traffic is routed out of Kubernetes' internal network, to the external ELB then back into the Kubernetes' internal network.

Also running a sidecar container requires Hokusai version >=0.5.2 so pin this in hokusai config.  (Running a Hokusai deployment at a version < 0.5.2 will overwrite the Nginx container with a metaphysics container and break internal routing, though if this happens the Nginx readinessProbe will fail and block the deployment).

DNS config changes in https://github.com/artsy/substance/pull/115 - for now just redirecting `metaphysics-staging-test.artsy.net` to `metaphysics-web-internal.default.svc.cluster.local`

### DIFF
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -51,6 +51,29 @@ spec:
               value: https
           initialDelaySeconds: 5
           periodSeconds: 5
+      - name: metaphysics-nginx
+        image: artsy/docker-nginx:1.14.2
+        ports:
+          - containerPort: 80
+          - containerPort: 443
+        lifecycle:
+          preStop:
+            exec:
+              command: ["/usr/sbin/nginx", "-s", "quit"]
+        env:
+          - name: 'NGINX_DEFAULT_CONF'
+            valueFrom:
+              configMapKeyRef:
+                name: nginx-config
+                key: metaphysics
+        volumeMounts:
+          - name: nginx-secrets
+            mountPath: /etc/nginx/ssl
+      volumes:
+        - name: nginx-secrets
+          secret:
+            secretName: nginx-secrets
+            defaultMode: 420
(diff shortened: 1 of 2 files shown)


hokusai/staging.yml
```diff
@@ -51,6 +51,35 @@ spec:
...
+              configMapKeyRef:
+                name: nginx-config
+                key: metaphysics
+        volumeMounts:
+          - name: nginx-secrets
+            mountPath: /etc/nginx/ssl
+      volumes:
+        - name: nginx-secrets
+          secret:
+            secretName: nginx-secrets
```
joeyAghion: This seems like a perfect use case for shared secrets.

ashkan18:
adding another image to pods on staging seems to have broken few hokusai commands, where we do this check https://github.com/artsy/hokusai/blob/830bb7499b8d66a92a7bedb5a95be44748ee466b/hokusai/services/deployment.py#L127

izakp:
@ashkan18 thanks for the heads up!  Now bumped to require Hokusai v 0.5.3

PR #1583, opened 2019-03-13
Adds support for day threshold to the schema

orta:
Adds a new arg for https://github.com/artsy/gravity/pull/12218

The status is already there.

### DIFF
--- a/src/schema/me/followed_shows.ts
+++ b/src/schema/me/followed_shows.ts
@@ -40,6 +40,11 @@ const FollowedShows: GraphQLFieldConfig<void, ResolverContext> = {
   args: pageable({
     status: EventStatus,
     sort: PartnerShowSorts,
+    dayThreshold: {
+      type: GraphQLString,
+      description:
+        "Number of days which will be used to filter upcoming and closing soon shows",
+    },
     city: {
       type: GraphQLString,
       description: `A string representing one of the supported cities in the City Guide, which are: ${getValidCitySlugs()}`,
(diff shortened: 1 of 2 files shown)


ashkan18:
we need `status` too, right? https://github.com/artsy/gravity/pull/12218/files#diff-e7e7c00aa36f3a5f5580878cb1728ed8R14

orta:
It's already in, we added that in the first run (see the schema diff )

src/schema/me/followed_shows.ts
```diff
@@ -40,6 +40,11 @@ const FollowedShows: GraphQLFieldConfig<void, ResolverContext> = {
   args: pageable({
     status: EventStatus,
     sort: PartnerShowSorts,
+    dayThreshold: {
+      type: GraphQLString,
```
ashfurrow: This should be a `GraphQLInt`?

ashkan18: yes to ☝️

ashkan18: I can update this @orta

ashfurrow: He already did here: https://github.com/artsy/metaphysics/pull/1584

ashkan18: nvm looks to be already `int` on `master` :)

alloy:
I think we’ve informally adopted a system where additions to the schema that simply pass-through data already resolved by a resolver doesn't warrant a test _as much_ as actual new logic. But perhaps a topic for the FE practice. /cc @zephraph

PR #1584, opened 2019-03-13
Uses an Int for dayThreshold and removes sort from followed shows

orta:
We don't think this is needed, and isn't supported in gravity - let's not set people up to fail


orta:
Relay caught it: https://travis-ci.org/artsy/emission/jobs/506026969#L521

PR #1586, opened 2019-03-14
Support querying featured sponsored shows explicitly

anandaroop:
https://artsyproduct.atlassian.net/browse/LD-454

This fixes an issue where sponsored content shows are returned from Grav to MP to Emission in unspecified order, which prevents us from relying on the order of results to pull out the first two shows for featuring in illustrated bricks.

Instead we can add a separate field guaranteed to have the correct shows.

```graphql
{
  city(slug: "new-york-ny-usa") {
    slug
    name
    sponsoredContent {
      featuredShows {
        id
      }
    }
  }
}
```

### DIFF
--- a/src/schema/city/index.ts
+++ b/src/schema/city/index.ts
@@ -1,14 +1,15 @@
 import {
   GraphQLBoolean,
   GraphQLEnumType,
-  GraphQLObjectType,
-  GraphQLString,
   GraphQLFieldConfig,
   GraphQLInt,
+  GraphQLList,
+  GraphQLObjectType,
+  GraphQLString,
 } from "graphql"
 
 import { LatLngType } from "../location"
-import { showConnection } from "schema/show"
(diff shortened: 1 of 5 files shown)


ashkan18:
LGTM let me know if you want me to merge

anandaroop:
Thanks! Yeah let's go ahead and merge so we can start getting this out to prod (which is the only place where the the hard-coded IDs will be valid).

src/schema/city/index.ts
```diff
@@ -124,6 +125,16 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
           artGuideUrl: {
             type: GraphQLString,
           },
+          featuredShows: {
+            type: new GraphQLList(ShowType),
```
anandaroop: List not connection, since this is by definition a handful of items.

PR #1587, opened 2019-03-14
Add new running_and_upcoming status

ashkan18:
This is to unblock Emission, for now mapping it to `current` till we figure out how to support this in Gravity.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -3256,6 +3256,9 @@ enum EventStatus {
 
   # End date is in near future
   CLOSING_SOON
+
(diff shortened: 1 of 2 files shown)


orta:
yah, this makes sense to me

mbilokonsky:
Ok, so Emission will ask Metaphysics for FollowShows where `status= RUNNING_AND_UPCOMING`. For now MP will treat that as indistinguishable from a request where `status=CURRENT`, but moving forward we can change how MP communicates this request to Gravity once we know how Gravity will satisfy it. LGTM!

ashkan18:
Yes 🙌

PR #1590, opened 2019-03-14
LD-304 - Updates date exhibitionPeriod format

kierangillen:
- https://artsyproduct.atlassian.net/browse/LD-304
- Previously the date range didn't include the end year if the end year was the same as the present year. E.g. It would display `Apr 10, 2016 - Feb 16` instead of `Apr 10, 2016 - Feb 16, 2019`

### DIFF
--- a/src/lib/date.ts
+++ b/src/lib/date.ts
@@ -3,7 +3,7 @@ import moment from "moment"
 export function exhibitionPeriod(startAt, endAt) {
   const startMoment = moment(startAt)
   const endMoment = moment(endAt)
-  const thisMoment = moment()
+  const thisMoment = moment.utc(Date.now())
   let startFormat = "MMM D"
   let endFormat = "D"
   let singleDateFormat = "MMM D"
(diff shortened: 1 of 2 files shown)


src/lib/date.ts
```diff
@@ -3,25 +3,22 @@ import moment from "moment"
 export function exhibitionPeriod(startAt, endAt) {
   const startMoment = moment(startAt)
   const endMoment = moment(endAt)
-  const thisMoment = moment.utc(Date.now())
+  const thisMoment = moment()
```
ashfurrow: Is there a reason we don't want to use the UTC date?

kierangillen: it doesn't do anything. Moment is smart enough to compare `moment()` to any date provided because `moment()` represents "now" whichever timezone is provided.

ashfurrow: Neat!

kierangillen: Although looking at the failing test, maybe I'm wrong!

PR #1596, opened 2019-03-15
Do not filter out stub shows from show query

anandaroop:
https://artsyproduct.atlassian.net/browse/LD-468

We've observed that it's not possible to tap into _some_ shows in the app, usually (but not always) stub shows.

Gravity returns them but MP doesn't. That is likely because they are being 404'd due to not being [_either a displayable show or a reference show or a fair booth_](https://github.com/artsy/metaphysics/blob/0e1658cb735102ff53640a9e7c8ab4e54f61ce92/src/schema/show.ts#L653-L655)

This PR prevents stub shows by being filtered out in this way, by adding stub shows to the allowlist above.

A good question is why we don't see this more often. I'm not certain, but it's possibly because we tend to tap into stubs that _also_ happen to be displayable by virtue of having install shots (such as the sponsored content shows).

In any case, this change seems warranted.

### Before 

b4

### After
a4

### DIFF
--- a/src/schema/show.ts
+++ b/src/schema/show.ts
@@ -650,7 +650,14 @@ const Show: GraphQLFieldConfig<void, ResolverContext> = {
   resolve: (_root, { id }, { showLoader }) => {
     return showLoader(id)
       .then(show => {
-        if (!show.displayable && !show.is_reference && !isExisty(show.fair)) {
+        if (
+          !(
+            show.displayable ||
+            show.is_local_discovery ||
+            show.is_reference ||
+            isExisty(show.fair)
+          )
+        ) {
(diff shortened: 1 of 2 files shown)


src/schema/show.ts
```diff
@@ -650,7 +650,14 @@ const Show: GraphQLFieldConfig<void, ResolverContext> = {
   resolve: (_root, { id }, { showLoader }) => {
     return showLoader(id)
       .then(show => {
-        if (!show.displayable && !show.is_reference && !isExisty(show.fair)) {
+        if (
+          !(
```
ashkan18: 🙌

PR #1601, opened 2019-03-18
Expose an isActive flag on Fair

anandaroop:
https://artsyproduct.atlassian.net/browse/LD-503

This adds a new `isActive` field (and deprecates the old `is_active` field).

The new `isActive` field's definition is:

- true when the fair's `active_start_at` is in the past and `end_at` is in the future
- false otherwise

Screen Shot 2019-03-18 at 5 31 06 PM

#skip_camel

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1121,9 +1121,10 @@ type ArtworkContextFair {
   hours: String
   href: String
   image: Image
+  is_active: Boolean @deprecated(reason: "Prefer isActive instead")
 
   # Are we currently in the fair's active period?
-  is_active: Boolean
+  isActive: Boolean
   links: String
   mobile_image: Image
   is_published: Boolean
@@ -1150,6 +1151,14 @@ type ArtworkContextFair {
     timezone: String
   ): String
   end_at(
+    # This arg is deprecated, use timezone instead
+    convert_to_utc: Boolean
+    format: String
+
+    # A tz database time zone, otherwise falls back to `X-TIMEZONE` header
+    timezone: String
+  ): String
+  active_start_at(
     # This arg is deprecated, use timezone instead
     convert_to_utc: Boolean
     format: String
@@ -3303,9 +3312,10 @@ type Fair {
   hours: String
   href: String
   image: Image
+  is_active: Boolean @deprecated(reason: "Prefer isActive instead")
 
   # Are we currently in the fair's active period?
-  is_active: Boolean
+  isActive: Boolean
   links: String
   mobile_image: Image
   is_published: Boolean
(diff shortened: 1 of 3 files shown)


src/schema/fair.ts
```diff
@@ -254,6 +264,7 @@ export const FairType = new GraphQLObjectType<any, ResolverContext>({
     },
     start_at: date,
     end_at: date,
+    active_start_at: date,
```
anandaroop: Cargo-culting [this date field definition](https://github.com/artsy/metaphysics/blob/master/src/schema/fields/date.ts) which is already used for `start_at` and `end_at`

_schema.graphql
```diff
@@ -4596,6 +4615,14 @@ type HomePageModuleContextFair {
     # A tz database time zone, otherwise falls back to `X-TIMEZONE` header
     timezone: String
   ): String
+  active_start_at(
+    # This arg is deprecated, use timezone instead
+    convert_to_utc: Boolean
```
DangerCI: - :no_entry_sign: Addition uses `snake_case` instead of `camelCase`: `convert_to_utc`

_schema.graphql
```diff
@@ -3339,6 +3349,14 @@ type Fair {
...
```
DangerCI: - :no_entry_sign: Addition uses `snake_case` instead of `camelCase`: `convert_to_utc`

_schema.graphql
```diff
@@ -3303,9 +3312,10 @@ type Fair {
   hours: String
   href: String
   image: Image
+  is_active: Boolean @deprecated(reason: "Prefer isActive instead")
```
DangerCI: - :no_entry_sign: Addition uses `snake_case` instead of `camelCase`: `is_active`

_schema.graphql
```diff
@@ -4560,9 +4578,10 @@ type HomePageModuleContextFair {
...
```
DangerCI: - :no_entry_sign: Addition uses `snake_case` instead of `camelCase`: `is_active`

_schema.graphql
```diff
@@ -1157,6 +1158,14 @@ type ArtworkContextFair {
     # A tz database time zone, otherwise falls back to `X-TIMEZONE` header
     timezone: String
   ): String
+  active_start_at(
+    # This arg is deprecated, use timezone instead
+    convert_to_utc: Boolean
```
DangerCI: - :no_entry_sign: Addition uses `snake_case` instead of `camelCase`: `convert_to_utc`

_schema.graphql
anandaroop: cc @alloy — There might be some more diff finessing opportunities for this check. Here a deprecation is triggering the warning. No good deed goes unpunished 😛

_schema.graphql
```diff
@@ -1121,9 +1121,10 @@ type ArtworkContextFair {
   hours: String
   href: String
   image: Image
+  is_active: Boolean @deprecated(reason: "Prefer isActive instead")
```
DangerCI: - :no_entry_sign: Addition uses `snake_case` instead of `camelCase`: `is_active`

kierangillen:
Thanks @anandaroop !

PR #1604, opened 2019-03-19
Exclude shows that were blocked from discovery

ashkan18:
# Problem
We want to by default exclude shows that were blocked from discovery in `city` query used in City Guide.

https://artsyproduct.atlassian.net/browse/LD-370

# Solution
Use [newly added Gravity param](https://github.com/artsy/gravity/pull/12214) and send `false` by default to exclude these shows.


src/schema/city/index.ts
```diff
@@ -99,6 +99,7 @@ const CityType = new GraphQLObjectType<any, ResolverContext>({
           displayable: true,
           include_local_discovery:
             args.includeStubShows || args.discoverable === true,
+          include_discovery_blocked: false,
```

PR #1608, opened 2019-03-19
added a test for follow_show queries specifying city by name

mbilokonsky:
Stubs out a test suite for follow_show tests moving forward. Writes assertions against the `city-slug` logic we added a few weeks ago. More tests should probably happen around status, once we decide how we want to play that.


src/schema/me/__tests__/followed_shows.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  it("generates a predictable URL with no parameters", done => {
+    const query = generate_query()
+    runAuthenticatedQuery(query, {
+      followedShowsLoader,
+    }).then(_data => {
+      expectArgs(BASE_GRAVITY_ARGS)
+      done()
+    })
+  })
```
ashfurrow: This test could be rewritten to be a little easier to read with `async/await` and a better Jest matcher:

```ts
it("generates a predictable URL with no parameters", async () => {
  const query = generate_query()
  const data = await runAuthenticatedQuery(query, {followedShowsLoader})
  expect(followedShowsLoader).toHaveBeenCalledWith(BASE_GRAVITY_ARGS)
}
```

mbilokonsky:
@ashfurrow that's great feedback and something I was a little nervous about - I don't want it to feel 'clever' but I do like the fact that this makes assertions against the relation between our city data and the behavior of this query, such that updates to the city data moving forward won't break the test in ways we don't expect.

But I also grok that it's doing some weird stuff.

The async/await approach is so much more readable - let me roll over to that model, and see where the rest of this lands.

ashfurrow:
@mbilokonsky sounds good! Let me know how I can help.

mbilokonsky:
I pushed a change that started using async/await, and it's so much cleaner - thanks!

Re: the 'clever' test, I left a long comment contextualizing and explaining it, but I'll recapitulate here: the behavior under test is specifically a relation between queries and specific arrangements of data in the `cityData` object. By writing this test we're guaranteeing that if the `cityData` changes down the line in some way that creates a corner-case error, we'll be notified immediately. I do think it's worth structuring it this way - but I am open to argument, too.

In my opinion this is a place where a certain amount of complexity is necessary to capture the thing we actually care about, and removing that complexity will reduce the safety that the test gives us, you know?

mbilokonsky:
(I am 100% willing to walk this back and restructure it if people don't find the argument above compelling)

src/schema/me/__tests__/followed_shows.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      // I totally grant that this is a bit weird
+      // but the behavior under test is tightly coupled
+      // to the contents of the cityData file. If a change
+      // to that file breaks this test I want to know about it,
+      // and this will tell us.
+
+      // When Jest hits this point in the code it will treat this almost
+      // like a macro - we'll have a new test generated for each city,
+      // you'll see the test results enumerated as if we'd done this manually.
+      it.call(null, `Handles City: ${name}`, async () => {
```
anandaroop: Yeah, I'm torn here.

An existing and probably simpler to grok pattern is to simply [mock the contents of the city data file](https://github.com/anandaroop/metaphysics/blob/ec1f87ef91feb8924082c9ad7fdd51a09f778417/src/schema/__tests__/city.test.ts#L1-L25) for the purposes of this test. I think that makes sense because we don't care so much about the literal contents of this file, just the logic that makes use of those contents. (Having six specs as opposed to one here, doesn't actually _increase_ our confidence I don't think.)

The `it.call` totally threw me  — I went looking for this in the Jest api before I realized you're just calling `call` on a function, which is what `it` is.

I think I'm coming down slightly on the "too-clever" side (though now that I know how it works, the `it.call` thing is neat trick to know about).

mbilokonsky: So, the mock approach exactly doesn't solve the thing I'm caring about. If we mock the city data for the purposes of this test then future edits to the shape and structure of that won't be reflected in the actual test here, right? That's why this is confusing me so much, because literally the behavior of the function depends on the specific data that currently lives inside of that city data file, you know? If we test a mock, and the mock diverges, then our tests don't give us any confidence as to whether or not we're actually accurately representing the city's parameters in the query.

Am I just overthinking this?

anandaroop: It's true, the shape of the data file could change under your feet and then these tests that rely on mocked data wouldn't catch that. In that sense this test is as brittle as one that mocked out data from a remote service — basically every other test in this repo 😅.

I guess I'm counting on anyone who redesigns the shape of that file to simply check where else the file is used and make the concomitant changes? Exactly that has [already happened once](https://github.com/anandaroop/metaphysics/commit/0ac1bf8f05ee32a305a8b619263780ce1af6aab7).

But your version is admittedly more paranoid, and usually _I'm_ the paranoid one, so I don't know what to believe anymore 😛.

ashfurrow: > If we mock the city data for the purposes of this test then future edits to the shape and structure of that won't be reflected in the actual test here, right?

I definitely hear you here. To address this concern, we could also have a dedicated test file for `cityList.ts`. That way, the complexity of `it.call` can be isolated while still making sure that the data adheres to our expectations. The unit tests remain single-purpose, and we can implement Roop's suggestion of mocking out the file for this test. Thoughts?

mbilokonsky: Just pushed an update which:

- makes an assert against the first city in cityData
- asserts that cityData matches a snapshot
- includes a comment on that snapshot test encouraging reflection on cross-dependencies between these files.

Does this work?

src/schema/me/__tests__/followed_shows.test.ts
```diff
@@ -0,0 +1,85 @@
...
+        lng,
+      }))
+      .forEach(city => {
+        expect(city).toMatchSnapshot({
+          name: expect.any(String),
+          slug: expect.any(String),
+          lat: expect.any(Number),
+          lng: expect.any(Number),
+        })
+      })
```
ashfurrow: This is good, but we're kind of conflating a snapshot test with an object expectation. The object expectation is really good, and probably more valuable a test than the snapshot. We could do something like:

```ts
.forEach(city => {
  expect(city).toMatchObject({
    // Everything you already have
  })
```

Then we can get rid of the snapshot altogether.

src/schema/me/__tests__/followed_shows.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  it("throws an error if presented with an invalid city slug", async () => {
+    const query = generate_query(`(first: 10, city-slug: "this-is-not-a-city")`)
+    try {
+      await runAuthenticatedQuery(query, { followedShowsLoader })
+      // shouldn't get here - if we did then the error wasn't thrown.
+      expect(true).toBeFalsy()
+    } catch (e) {
+      expect(e).toBeDefined()
+    }
+  })
```
ashfurrow: It's great that you're testing this! Just also has a matcher for failure, too:

```ts
await expect(runAuthenticatedQuery(query, { followedShowsLoader })).rejects.toEqual(new WhateverErrorIsThrown("OK"))
```

You can also just expect `.rejects.toMatchInlineSnapshot()` or something to have Jest update the snapshot in the test file.

I learned this recently: https://github.com/artsy/emission/pull/1529/files#diff-18770d90e60e62acb19fdc863c698fcaR75 Jest is by far my favourite testing framework – if I geek out over tests, this is why 😁 Some more of the cool matchers are listed here: https://jestjs.io/docs/en/expect

mbilokonsky: Thank you, your insights into structuring jest tests throughout this process has been really helpful!

mbilokonsky:
Ok, pushed latest updates as per @ashfurrow's feedback

ashfurrow:
Great job, @mbilokonsky! Thanks for sticking with it, these look awesome ✨

alloy:
Nice one, thanks @mbilokonsky 👌

PR #1615, opened 2019-03-20
Shows month if different years but same month

kierangillen:
- Shows ending month if different years but same month


l2succes:
merge on green

PR #1618, opened 2019-03-21
Vortex histogram hookup

ds300:
~DO NOT MERGE: blocked by artsy/vortex#59~

https://artsyproduct.atlassian.net/browse/PURCHASE-925

In this PR I put the `pricingContext` field from vortex in the `Artwork` type and resolve it. 

The main thing I'm unsure about here is that I couldn't find any prior art for adding data loaders that fetch from graphql endpoints. Is that a thing we have an established pattern for?

This will also require some follow up work before going public:

2. For now we are hiding the pricingContext field behind a lab feature, so I'm not concerned about caching. Will add a follow-up ticket for making sure the caching is in place before making the feature public.
3. In a similar vein, requests from metaphysics to vortex are user-authenticated. So for each request a user-specific token for vortex is retrieved from gravity before being used to make requests to vortex. I'm guessing the galleries team will need this feature for adding partner-facing analytics (@ansor4 can you confirm?), but we technically don't need it and don't even want it when making public-facing `pricingContext` queries. So will add another follow-up ticket.

### DIFF
--- a/src/lib/loaders/loaders_without_authentication/vortex.ts
+++ b/src/lib/loaders/loaders_without_authentication/vortex.ts
@@ -9,14 +9,14 @@ const query = gql`
   query artworkPricingContextQuery(
     $artistId: String!
     $category: AnalyticsPricingContextCategoryEnum!
-    $dimensions: AnalyticsPricingContextDimensionsEnum!
-    $listPriceCents: Int!
+    $widthCm: Int!
+    $heightCm: Int!
   ) {
     analyticsPricingContext(
       artistId: $artistId
       category: $category
-      dimensions: $dimensions
-      listPriceCents: $listPriceCents
+      widthCm: $widthCm
+      heightCm: $heightCm
     ) {
       bins {
         maxPriceCents
@@ -34,8 +34,8 @@ export default () => {
       vars: {
         artistId: string
         category: string
-        dimensions: string
-        listPriceCents: number
+        widthCm: number
+        heightCm: number
       },
       context: ResolverContext
     ) {
--- a/src/lib/loaders/loaders_without_authentication/vortex.ts
+++ b/src/lib/loaders/loaders_without_authentication/vortex.ts
@@ -1,0 +1,50 @@
+import { executableVortexSchema } from "lib/stitching/vortex/schema"
+import gql from "lib/gql"
+import { graphql } from "graphql"
+import { ResolverContext } from "types/graphql"
+import { error } from "util"
+
+const schema = executableVortexSchema({ removePricingContext: false })
+const query = gql`
+  query artworkPricingContextQuery(
+    $artistId: String!
+    $category: AnalyticsPricingContextCategoryEnum!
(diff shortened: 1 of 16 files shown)


src/schema/analytics/PricingContext.ts
```diff
@@ -0,0 +1,48 @@
...
+    // this feature is only enbaled for lab users right now
+    if (!context.meLoader) {
+      return null
+    }
+    const me = await context.meLoader()
+    if (!me.lab_features || !me.lab_features.includes("Pricing Context")) {
+      return null
+    }
+
+    // copying vortex to calculate the 'dimensions' field which is actually an enum of "small" | "medium" | "large"
```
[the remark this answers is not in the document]
ds300: I think that makes sense :+1: I'll make a PR to vortex.

src/schema/analytics/PricingContext.ts
```diff
@@ -0,0 +1,48 @@
...
+      price_hidden ||
+      !artist ||
+      !width_cm ||
+      !height_cm ||
+      !category ||
+      !listPriceCents
+    ) {
+      return null
+    }
+    // this feature is only enbaled for lab users right now
```
damassi: typo

ds300:
This is 🍏 again

src/lib/stitching/vortex/link.ts
```diff
@@ -14,5 +18,35 @@ export const createVortexLink = () => {
     uri: urljoin(VORTEX_API_BASE, "graphql"),
   })
 
-  return middlewareLink.concat(responseLoggerLink("Vortex")).concat(httpLink)
+  const authMiddleware = setContext(
```
sepans: 🙏

ansor4:
Thank you @ds300! Galleries team definitely needs authentication added for partner analytics. It looks like this PR already accounts for `pricingContext` to be not gated by authentication. Is there any work/input you need from the partner analytics side of things?

src/lib/loaders/loaders_without_authentication/vortex.ts
```diff
@@ -0,0 +1,50 @@
...
+    async pricingContextLoader(
+      vars: {
+        artistId: string
+        category: string
+        widthCm: number
+        heightCm: number
+      },
+      context: ResolverContext
+    ) {
+      const result = await graphql(schema, query, null, context, vars)
```
alloy: To all those on this PR; when in the future you see the usage of the `graphql` function, it is perhaps incorrectly trying to implement stitching. What is being done here is more a gateway to an upstream service and doesn’t lead to stitching advantages, i.e. we’re selecting _all_ of the fields, rather than only the ones being requested by the client. We don’t want to do this going forward 🙏 

Probably the _only_ case where we have to use this function as a gateway is when we have a _field_ in MP whose value is derived from one or more upstream values (and even there we haven’t spend a lot of time investigating the options within the stitching framework).

PR #1619, opened 2019-03-22
Set up a canary with conventional cache expiration instead of throttling

joeyAghion:
This adds a new `metaphysics-web-canary` deployment with identical set-up except for `replicas: 1`, `DD_TRACER_SERVICE_NAME: metaphysics-canary` and `image: .../metaphysics:ceb17aec2655475edeffd93a124b5c42f5663d5b` with the updated caching scheme.

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -115,3 +115,74 @@ spec:
     component: web
   sessionAffinity: None
   type: LoadBalancer
+
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: metaphysics-web-canary
+  namespace: default
+spec:
+  replicas: 1
+  strategy:
+    rollingUpdate:
+      maxSurge: 1
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: metaphysics
+        layer: application
+        component: web
+      name: metaphysics-web-canary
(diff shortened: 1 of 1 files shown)


hokusai/production.yml
```diff
@@ -115,3 +115,74 @@ spec:
...
+    rollingUpdate:
+      maxSurge: 1
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: metaphysics
+        layer: application
+        component: web
+      name: metaphysics-web
```
izakp: `metaphysics-web-canary` here as well

izakp: So pods that get created are also named `-canary` :)

joeyAghion: Updated

joeyAghion:
Closing as this seemed to operate successfully in production. (I've also deleted the deployment from the k8s dash.) 50th percentile latency was notably worse, but overall performance was comparable and cache hit rate seemed to steadily improve (up to ~30%), presumably as the longer-TTL data like filter results and shows got requested. Would expect even better cache hit rate once fully deployed and shared across the live system.

overall performance: https://cl.ly/2e9bddf52b4d
cache hit rate: https://cl.ly/00aa91bc4965

PR #1622, opened 2019-03-23
Emission queries for 'running_and_upcoming' now redirect correctly.

mbilokonsky:
This is a tiny change in the source file, but I took the opportunity to add a few more tests to our followed_show spec. Now, there's a case to be made that this test doesn't really help us - that the GraphQLEnumType constrains values enough. But this change made me think about the fact that for a while we had:

```
RUNNING_AND_UPCOMING : {
  value: "current"
}
```

I dunno about y'all but once I have enum keys to read I rarely go look at the values. In this case, it seemed reasonable to assert that our enum keys correspond directly to our enum values - if that ever changes again, this test will have to change, and anyone reading the tests to figure out why they keep requesting `status=FOO` but the system keeps returning data as if it had requested `status=BAR` will get a useful piece of information.

### DIFF
--- a/src/schema/input_fields/event_status.ts
+++ b/src/schema/input_fields/event_status.ts
@@ -41,7 +41,7 @@ const EventStatus = {
         description: "End date is in near future",
       },
       RUNNING_AND_UPCOMING: {
-        value: "current",
+        value: "running_and_upcoming",
         description:
           "Special filtering option which is used to show running and upcoming shows",
       },
--- a/src/schema/me/__tests__/followed_shows.test.ts
+++ b/src/schema/me/__tests__/followed_shows.test.ts
@@ -40,44 +40,84 @@ describe("returns followed shows for a user", () => {
     })
   })
 
-  it("generates a predictable URL with a city slug input", async () => {
-    const nyc = cityData[0]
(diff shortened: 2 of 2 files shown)


src/schema/me/__tests__/followed_shows.test.ts
```diff
@@ -40,44 +40,84 @@ describe("returns followed shows for a user", () => {
...
+        }))
+        .forEach(city => {
+          expect(city).toMatchObject({
+            name: expect.any(String),
+            slug: expect.any(String),
+            lat: expect.any(Number),
+            lng: expect.any(Number),
+          })
+        })
+    })
```
orta: good tests all of these 👍

src/schema/input_fields/event_status.ts
```diff
@@ -41,7 +41,7 @@ const EventStatus = {
         description: "End date is in near future",
       },
       RUNNING_AND_UPCOMING: {
-        value: "current",
+        value: "running_and_upcoming",
```
anandaroop: 🎉

kierangillen:
Nice ! @mbilokonsky

PR #1624, opened 2019-03-25
Fixes a bug where CausalityJWT resolver fails due to Cannot read property '_id' of null

yuki24:
There is a [bug][1] where Metaphysics raises an exception due to `sale.partner` being `null`. This commit fixes it so the client can run a more reliable check rather than just checking a raw Javascript error message.

Also, the test needs to call `done()` otherwise the test will call no `expect`s and it will just pass.

[1]: https://sentry.io/organizations/artsynet/issues/841789531/?project=299973&query=is%3Aunresolved&statsPeriod=14d&utc=false

### DIFF
--- a/src/schema/__tests__/causality_jwt.test.js
+++ b/src/schema/__tests__/causality_jwt.test.js
@@ -9,18 +9,18 @@ const { HMAC_SECRET } = config
 describe("CausalityJWT", () => {
   let context
 
+  const sale = {
+    _id: "foo",
+    name: "Foo sale",
+    id: "slug",
+    partner: { _id: "fooPartner" },
+  }
+
   beforeEach(() => {
     const me = {
       _id: "craig",
       paddle_number: "123",
       type: "User",
-    }
-
-    const sale = {
-      _id: "foo",
-      name: "Foo sale",
-      id: "slug",
-      partner: { _id: "fooPartner" },
     }
 
     const mePartners = [{ _id: "fooPartner" }]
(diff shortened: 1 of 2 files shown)


src/schema/__tests__/causality_jwt.test.js
```diff
@@ -127,13 +127,21 @@ describe("CausalityJWT", () => {
     })
   })
 
-  it("does not allow a non-admin user or user not associated with sale partner to be operator", () => {
+  it("does not allow a non-admin user or user not associated with sale partner to be operator", done => {
```
dleve123: @yuki24 There are a fair number of tests here that depend upon `runAuthenticatedQuery`? Should all of there `it` expressions be updated to the `it("...", done) => {})` form? If so, while we're here, I'd recommend updating those tests to improve the robustness of MP.

yuki24: Jest resolves the promise if the test returns a Promise object (so `return` is required) so we don't have to add `done()` to every spec. In this case, the catch block doesn't get invoked entirely unless an error is thrown. The extra `done` was added to prevent the test from passing with no expectations. I've replaced the `done` with `expect.assertions(1)` so the test will fail when the operation being tested does not throw an error and I think this is more explicit than `done`.

yuki24:
I rebased and tested with Prediction master, and this is good for review:

Screen Shot 2019-03-26 at 11 04 20 AM

And Prediction screenshot:

Kapture 2019-03-26 at 11 32 26

yuki24:
I'll go ahead and merge this to fix a bug sooner than later.

PR #1626, opened 2019-03-26
Add docs for graphql microservice authentication

sepans:
Documentation for graphql micro-service authentication was missing.

### DIFF
--- a/docs/adding_a_new_graphql_microservice.md
+++ b/docs/adding_a_new_graphql_microservice.md
@@ -48,6 +48,8 @@ Let's pretend we're mapping a GraphQL API called `Three Body` but provides its o
    }
    ```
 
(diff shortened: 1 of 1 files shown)


docs/adding_a_new_graphql_microservice.md
```diff
@@ -48,6 +48,8 @@ Let's pretend we're mapping a GraphQL API called `Three Body` but provides its o
    }
    ```
 
+   *Note:* If the new microservice requires gravity JWT authentication, anoother middleware is required to pass the token coming from client to gravity and get an application token from gravity and pass it to the microservice. Follow exchange example [here](https://github.com/artsy/metaphysics/blob/master/src/lib/stitching/exchange/link.ts#L20-L40) to add a middleware and [here](https://github.com/artsy/metaphysics/blob/master/src/lib/loaders/loaders_with_authentication/exchange.ts) to add token loader.
+
```
orta: Maybe tie these from `master` to a specific commit sha - then this is good to go. Nice call.

PR #1629, opened 2019-03-27
Real vortex stitching + Add money helper to histogram bin price fields

ds300:
_Edit: The scope of this PR grew. We added in proper stitching for vortex after realising that what I did in #1618 was not real stitching._

https://artsyproduct.atlassian.net/browse/PURCHASE-950


src/lib/stitching/vortex/schema.ts
```diff
@@ -24,9 +25,14 @@ export const executableVortexSchema = ({
   return transformSchema(schema, [
     // we don't want pricingContext to be a root query field, it is
     // accessible through artwork
```
alloy: I’m a little confused about where this stitching occurs, could you link me to the file?

orta: It doesn't yet, it should be in `src/lib/stitching/vortex/stitching.ts` when it does

alloy:
Alas, offhand I can’t think of another way to do this within just MP. The one thing that comes to mind is to port the logic to Ruby that can be shared among Ruby services (and similarly extract our JS version for when an upstream JS service ever needs it).

src/lib/stitching/vortex/schema.ts
alloy: Ah kk, thanks 👍

ds300: oooh that looks like what I should have done in #1618 rather than directly adding to the `Artwork` type 🤔 I wonder if it's worth going back and changing it?

orta: Yep, it is worth it, all the vortex logic should live in those three files (link/schema/stitch)

alloy: Yeah absolutely, please–that is not stitching, it’s a gateway and we shouldn’t be doing that any more.

alloy:
Seeing as I now understand where the data is coming from and that this isn’t done through stitching, I actually think that the _only_ way to do this is by having the upstream service perform the formatting. (Although I’d love to be proven wrong.)

sweir27:
@alloy I think this is a potential beauty/use of metaphysics (to be able to have a consistent formatting). Otherwise this is just logic we'll have to repeat in many upstream services.

alloy:
@sweir27 I absolutely agree. I should probably have rephrased it as: we can make anything work, but I don’t know of a way offhand to do this with existing plumbing.

(Note, though, that I suggest we re-use the code, although in this case we’d first have to port it to Ruby, but also I don’t think there are many such helpers that would need sharing. It would have the added benefit of the upstream service being usable by other internal apps and provide formatting without needing to go through MP. However, that’s an area we haven’t spend much of any time exploring yet.)

ashkan18:
suggested in slack to have an item about this issue (and few other money related issues) in tomorrow's platform practice https://artsy.slack.com/archives/CA8SANW3W/p1553697797061900 we can discuss there too

src/lib/stitching/vortex/schema.ts
ds300: This is done now. Thanks to @orta for pairing :bow:

src/schema/me/__tests__/recently_viewed_artworks.test.ts
```diff
@@ -105,7 +105,7 @@ describe("RecentlyViewedArtworks", () => {
       data: { recordArtworkView: { artwork_id: "percy" } },
     }
 
-    const mockFetch = fetch as jest.Mock<any>
+    const mockFetch = (fetch as unknown) as jest.Mock<any>
```
ds300: typescript started complaining about this cast because I added the types for `node-fetch`

src/schema/artwork/index.ts
```diff
@@ -804,7 +804,18 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+          "If you need to render artwork dimensions as a string, prefer the `Artwork#dimensions` field",
+        type: GraphQLFloat,
+        resolve: ({ width_cm }) => width_cm,
+      },
+      heightCm: {
+        description:
+          "If you need to render artwork dimensions as a string, prefer the `Artwork#dimensions` field",
+        type: GraphQLFloat,
+        resolve: ({ height_cm }) => height_cm,
+      },
```
ds300: When stitching the `pricingContext` into the artwork type, we only have access to the data resolved by graphql on the Artwork type. We weren't exposing the numerical dimensions of an artwork before, only the `dimensions` string values.

It feels unfortunate to expose these extra fields just because we need them for stitching. We couldn't think of an alternative though.

ds300:
OK this is 🍏again

sweir27:
bmo

src/lib/stitching/vortex/__tests__/pricingContext.test.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+`)
+  })
+  it("is null when dimensions not present", async () => {
+    const { width_cm, height_cm, ...others } = artwork
+    artworkLoader.mockResolvedValueOnce(others)
+    const result = (await runQuery(query, context)) as any
+    expect(result.artwork.pricingContext).toBeNull()
+  })
+  it("is null when artist details are not present", async () => {
```
alloy: @ds300 Nit: can you add newlines between these tests?

src/lib/stitching/vortex/stitching.ts
```diff
@@ -0,0 +1,120 @@
...
+  resolvers: {
+    AnalyticsHistogramBin: {
+      minPrice: {
+        fragment: `
+          ... on AnalyticsHistogramBin {
+            minPriceCents
+          }
+        `,
+        resolve: (parent, args, _context, _info) =>
+          amount(_ => parent.minPriceCents).resolve({}, args),
```
alloy: Nice 👌

src/lib/stitching/vortex/stitching.ts
```diff
@@ -0,0 +1,120 @@
...
+              context,
+              info,
+            })
+          } catch (e) {
+            console.error(e)
+            throw e
+          }
+        },
+      },
+    },
```
alloy: Epic ✨

src/schema/artwork/index.ts
alloy: Ah, hmm, I would have expected the `source` parameter for the stitched resolver would get access to that data 🤔 What _does_ it contain then? Is there the opportunity to improve this in the library?

ds300: It contains whatever you query for in the `fragment` field. Feels like there should be a way to allow these stitched fields to optionally access the resolver args of their parent type.

alloy: I agree, it makes sense for augmented fields to receive the same data as sibling fields, rather than having to ‘pollute’ the schema. Sounds like something to contribute.

src/lib/stitching/vortex/__tests__/pricingContext.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    expect(result.artwork.pricingContext).toBeNull()
+  })
+  it("is null when user has no lab features", async () => {
+    meLoader.mockResolvedValueOnce({
+      lab_features: [],
+    })
+    const result = (await runQuery(query, context)) as any
+    expect(result.artwork.pricingContext).toBeNull()
+  })
+})
```
orta: good tests 👍

PR #1637, opened 2019-03-28
Make sure list price is present for pricing context

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-964


sweir27:
Thanks!!

PR #1639, opened 2019-03-29
Do proper category mapping for pricingContext

ds300:
Fixing https://sentry.io/organizations/artsynet/issues/956770079/?project=299973&query=is%3Aunresolved&statsPeriod=14d&utc=false

### DIFF
--- a/src/lib/stitching/vortex/stitching.ts
+++ b/src/lib/stitching/vortex/stitching.ts
@@ -100,7 +100,7 @@ export const vortexStitchingEnvironment = () => ({
 
           const args = {
             artistId: artist._id,
-            category: category.toUpperCase(),
+            category: categoryMap[category] || "OTHER",
             widthCm: Math.round(widthCm),
             heightCm: Math.round(heightCm),
           }
@@ -123,3 +123,26 @@ export const vortexStitchingEnvironment = () => ({
     },
   },
 })
+
+const categoryMap = {
+  Architecture: "ARCHITECTURE",
+  "Books and Portfolios": "BOOKS_AND_PORTFOLIOS",
+  "Design/Decorative Art": "DESIGN_DECORATIVE_ART",
+  "Drawing, Collage or other Work on Paper":
+    "DRAWING_COLAGE_OTHER_WORK_ON_PAPER",
+  "Fashion Design and Wearable Art": "FASHON",
+  Installation: "INSTALLATION",
+  Jewelry: "JEWELRY",
+  "Mixed Media": "MIXED_MEDIA",
+  Other: "OTHER",
+  Painting: "PAINTING",
+  "Performance Art": "PERFORMANCE",
+  Photography: "PHOTOGRAPHY",
+  Posters: "POSTERS",
+  Print: "PRINT",
+  Sculpture: "SCULPTURE",
+  Sound: "SOUND",
+  "Textile Arts": "TEXTILE",
+  "Video/Film/Animation": "VIDEO_FILM_ANIMATION",
+  "Work on Paper": "WORK_ON_PAPER",
+}
--- a/src/lib/stitching/vortex/__tests__/pricingContext.test.ts
+++ b/src/lib/stitching/vortex/__tests__/pricingContext.test.ts
@@ -3,6 +3,7 @@ import gql from "lib/gql"
 import { ResolverContext } from "types/graphql"
 
 jest.mock("../link")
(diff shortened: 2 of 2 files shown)


src/lib/stitching/vortex/stitching.ts
```diff
@@ -123,3 +123,26 @@ export const vortexStitchingEnvironment = () => ({
     },
   },
 })
+
+const categoryMap = {
+  Architecture: "ARCHITECTURE",
+  "Books and Portfolios": "BOOKS_AND_PORTFOLIOS",
+  "Design/Decorative Art": "DESIGN_DECORATIVE_ART",
+  "Drawing, Collage or other Work on Paper":
+    "DRAWING_COLAGE_OTHER_WORK_ON_PAPER",
```
sweir27: Shoot, was this a typo in vortex too? Should be `DRAWING_COLLAGE_OTHER_WORK_ON_PAPER`

ds300: Yeah that's what it is in vortex. Do you think it's worth changing?

sweir27: 😭 if it's quick I would say... yes given there won't be a better time. But not if you feel strongly otherwise!

ds300: Well... it's four coordinated PRs. My preference would be to live with the typo 😅

ds300: We could also just wait until the next breaking change in vortex.

joeyAghion: Meta-observation: this is such a real pain point. How can we make this better? I don't have a strong recommendation but maybe Metaphysics can aim to support more backwards-and forwards-compatible schemas in parallel, or maybe Reaction and Metaphysics releases can be linked more formally or automatically. I'm thinking of tooling improvements (like via renovate-bot and schema compatibility checks). Other ideas?

:ceiling-cat-out:

sweir27: Wait......

we should never have breaking changes 😉

mzikherman: What are the 4 PR's? I'm assuming two PR's (one in Vortex and one in MP), which will always be the minimum required, so that's the baseline.

What are the other 2? I guess, Force would now be blocked by schema validation, so there would need to be a Reaction build with the new schema (1 PR), and then Force needs an update to point to that. (1 PR- although `renovate` should take care of it).

It does seem..onerous. Maybe as @sweir27 says, rather than be a breaking change on Vortex's side, the correctly spelled option should just be _added_ (and not replacing the old one). Then it won't be breaking and so only requires the usual Vortex/MP updates.

Then the cleanup can be deferred to some breaking-change schema cleanup as the feature gets finalized and ready to launch, or something.

ds300: @joeyAghion 
> Other ideas?

image

... but something along the lines you suggested would be lower-hanging fruit for sure 😬

We could add a MP CI job that runs on merges to `master` to create a PR to reaction every time the MP schema changes. Then renovate would handle bumping reaction in force.

Wish we could do the same for vortex, exchange, kaws, et al, (i.e. changes to their schemas creates an automatic PR in MP) but the stitching might not make the process robust :thinking:

ds300: We could also vendor the MP schema via npm and then let renovate deal with updating it wherever it gets used. /cc @orta @alloy  was that considered at some point in the past?

ds300:
@mzikherman yep, the PR chain would be vortex => mp => reaction => force

And yep it would be possible to do it like you suggest. That's twice as many PRs but it's the only way to  do it without breaking changes. (unless of course, the iOS app is using the feature you want to change, then you just have to deprecate and wait a year or two)

src/lib/stitching/vortex/stitching.ts
ds300: @mzikherman yep, the PR chain would be vortex => mp => reaction => force

And yep it would be possible to do it like you suggest. That's twice as many PRs but it's the only way to  do it without breaking changes. (unless of course, the iOS app is using the feature you want to change, then you just have to deprecate and wait a year or two)

joeyAghion: Good point that _more_ PRs can sometimes be better than fewer PRs **if** those PRs are each easy and unblocked, whereas the fewer PRs are burdened with coordination and build warnings.

PR #1642, opened 2019-04-03
Use custom date resolver for partner show events

anandaroop:
https://artsyproduct.atlassian.net/browse/LD-561

@kierangillen and I have been pairing on this fix for incorrect times displaying in City Guide for partner show events.

Currently datetimes for `PartnerShowEvents` — e.g. opening receptions, artist talks, and other one-off events that are associated with a show — are sent from CMS to Gravity without timezone information, and these get stored in the db as having a UTC offset of +00:00. They are returned this way over the wire as well by MP:

```graphql
{
  "data": {
    "show": {
      "id": "helwaser-gallery-anton-ginzburg-views",
      "events": [
        {
          "start_at": "2019-03-28T18:00:00+00:00",
          "end_at": "2019-03-28T21:00:00+00:00"
        },
        {
          "start_at": "2019-04-13T14:00:00+00:00",
          "end_at": "2019-04-13T16:00:00+00:00"
        }
      ]
    }
  }
}
```

Previously other clients [have treated](https://github.com/artsy/force/blob/33ece666e9df1cf32f9c6cbe7914811815a860cc/src/desktop/apps/show/helpers/view_helpers.coffee#L11-L12) these ambiguous datetimes by parsing them with `moment.utc(someDate)` so that e.g. **6pm** in means **6pm** out, regardless of the user's or event's timezone.

The first versions of Eigen that contain ~City Guide~ the new Show screens parse these fields with `moment`, which causes them to be parsed and displayed as local times in the user's timezone, so for the example in the ticket **6pm** becomes **2pm**.

In order to fix this display issue for current versions of Eigen that are _already_ out in the wild, we introduced a custom date field resolver here  — based on the shared one currently in use across MP — that does some user-agent sniffing and returns a modified version of these partner show event datetimes to selected versions of Eigen.

So `2019-03-28T18:00:00+00:00` has its offset stripped out and becomes `2019-03-28T18:00:00`, which will force it to display without timezone conversion, just as if it had been parsed with `moment.utc`.

This fixes the pressing issue with incorrect event times being displayed to users, but note that we have deferred a longer-term fix at the level of the Gravity/CMS or other admin apps; as well as updates to clients that rely on `moment`.

### DIFF
--- a/src/schema/__tests__/partner_show_events.test.ts
+++ b/src/schema/__tests__/partner_show_events.test.ts
@@ -75,10 +75,10 @@ describe("date resolving", () => {
     })
   })
 
-  describe("with Eigen 4.x", () => {
+  describe("with Eigen 4.4.x", () => {
     it("truncates the UTC offset", async () => {
       context.userAgent =
-        "Mozilla/5.0 Artsy-Mobile/4.3.3 Eigen/2018.11.09.15/4.3.3 (iPhone; iOS 12.1.4; Scale/3.00) AppleWebKit/601.1.46 (KHTML, like Gecko); Metaphysics"
+        "iPhone9,1 Mozilla/5.0 Artsy-Mobile/4.4.1 Eigen/2019.03.15.11/4.4.1 (iPhone; iOS 12.1.4; Scale/2.00) AppleWebKit/601.1.46 (KHTML, like Gecko)"
 
       const data = await runQuery(query, context)
       expect(data).toEqual({
--- a/src/schema/__tests__/partner_show_events.test.ts
+++ b/src/schema/__tests__/partner_show_events.test.ts
@@ -132,7 +132,7 @@ describe("date resolving", () => {
     })
   })
 
-  describe("if user agent is returned as an array", () => {
+  describe("When an Eigen user agent is returned as part of an array", () => {
     it("truncates the UTC offset", async () => {
       context.userAgent = [
         "x86_64 Mozilla/5.0 Artsy-Mobile/5.0.1 Eigen/2019.02.28.17/5.0.1 (iPhone; iOS 12.2; Scale/2.00) AppleWebKit/60 1.1.46 (KHTML, like Gecko)",
--- a/src/schema/__tests__/partner_show_events.test.ts
+++ b/src/schema/__tests__/partner_show_events.test.ts
@@ -1,0 +1,154 @@
+import { runQuery } from "test/utils"
+import gql from "lib/gql"
+import moment from "moment"
+import { ResolverContext } from "types/graphql"
+
+// This time-bomb spec will fail starting at some defined point in the future,
+// indicating that we should consider reverting this behavior of sending
+// different raw date representations to different versions of Eigen
+// (in the hope that usage of older versions has dropped below some minimum threshold).
+
+it("is not yet time to rethink this UA-sniffing behavior for resolving dates", () => {
+  const today = moment()
+  const deadline = moment("2020-10-01") // about 18 months after adding this behavior
+
+  let itIsTimeToRethinkThis = today.isAfter(deadline)
+
+  expect(itIsTimeToRethinkThis).toBe(false)
+})
+
+// In the meantime, here is the custom date resolving behavior
+
+describe("date resolving", () => {
+  const showData = {
+    id: "helwaser-gallery-anton-ginzburg-views",
+    displayable: true,
+    partner: {
+      id: "helwaser-gallery",
+    },
+    events: [
+      {
+        event_type: "Opening Reception",
+        start_at: "2019-03-28T18:00:00+00:00",
+        end_at: "2019-03-28T21:00:00+00:00",
+      },
+    ],
+  }
+
+  const mockLoader = jest.fn(() => Promise.resolve(showData))
+
+  const query = gql`
+    {
+      show(id: "helwaser-gallery-anton-ginzburg-views") {
+        events {
+          start_at
+          end_at
+        }
+      }
+    }
+  `
+
+  let context: Partial<ResolverContext>
+
+  beforeEach(() => {
+    context = {
+      showLoader: mockLoader,
+      partnerShowLoader: mockLoader,
+    }
+  })
+
+  describe("default behavior", () => {
+    it("returns dates with UTC offset", async () => {
+      context.userAgent = "some browser"
+
+      const data = await runQuery(query, context)
+      expect(data).toEqual({
+        show: {
+          events: [
+            {
+              end_at: "2019-03-28T21:00:00+00:00",
+              start_at: "2019-03-28T18:00:00+00:00",
+            },
+          ],
+        },
+      })
+    })
+  })
+
+  describe("with Eigen 4.x", () => {
+    it("truncates the UTC offset", async () => {
+      context.userAgent =
+        "Mozilla/5.0 Artsy-Mobile/4.3.3 Eigen/2018.11.09.15/4.3.3 (iPhone; iOS 12.1.4; Scale/3.00) AppleWebKit/601.1.46 (KHTML, like Gecko); Metaphysics"
+
+      const data = await runQuery(query, context)
+      expect(data).toEqual({
+        show: {
+          events: [
+            {
+              end_at: "2019-03-28T21:00:00",
+              start_at: "2019-03-28T18:00:00",
+            },
+          ],
+        },
+      })
+    })
+  })
+
+  describe("with Eigen 5.0.0", () => {
+    it("truncates the UTC offset", async () => {
+      context.userAgent =
+        "iPhone11,6 Mozilla/5.0 Artsy-Mobile/5.0.0 Eigen/2019.03.20.14/5.0.0 (iPhone; iOS 12.1.4; Scale/3.00) AppleWebKit/601.1.46 (KHTML, like Gecko)"
+
+      const data = await runQuery(query, context)
+      expect(data).toEqual({
+        show: {
+          events: [
+            {
+              end_at: "2019-03-28T21:00:00",
+              start_at: "2019-03-28T18:00:00",
+            },
+          ],
+        },
+      })
+    })
+  })
+
+  describe("with Eigen 5.0.1", () => {
+    it("truncates the UTC offset", async () => {
+      context.userAgent =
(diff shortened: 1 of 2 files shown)


src/schema/__tests__/partner_show_events.test.ts
```diff
@@ -0,0 +1,154 @@
...
+import { ResolverContext } from "types/graphql"
+
+// This time-bomb spec will fail starting at some defined point in the future,
+// indicating that we should consider reverting this behavior of sending
+// different raw date representations to different versions of Eigen
+// (in the hope that usage of older versions has dropped below some minimum threshold).
+
+it("is not yet time to rethink this UA-sniffing behavior for resolving dates", () => {
+  const today = moment()
+  const deadline = moment("2020-10-01") // about 18 months after adding this behavior
```
anandaroop: I picked this number based on the number that came up in the Zoom chat, but open to changes here.

src/schema/partner_show_event.ts
```diff
@@ -1,8 +1,46 @@
...
+    result = userAgent.some(ua => checkUserAgent(ua))
+  }
+  return result
+}
+
+const dateFieldForPartnerShowEvent: GraphQLFieldConfig<
+  DateSource,
+  ResolverContext
+> = {
+  ...dateField,
```
anandaroop: This reuses the existing field definition, and only provides this custom resolver below.

src/schema/partner_show_event.ts
```diff
@@ -1,8 +1,46 @@
...
+    { format, timezone },
+    { defaultTimezone, userAgent },
+    { fieldName }
+  ) => {
+    const rawDate = obj[fieldName]
+
+    if (userAgent && isOlderEmissionVersion(userAgent)) {
+      const dateWithoutOffset = rawDate.replace(/[-+]\d\d:\d\d$/g, "")
+      return dateWithoutOffset
+    }
```
anandaroop: These 4 lines are the meaningful change in this PR

src/schema/__tests__/partner_show_events.test.ts
dleve123: I like the idea of an automated test here, but I think the most actionable thing to do is start to work with PM/TLs to prioritize this work and get it on Jira. I encourage keeping this test around and starting the prioritization discussion.

src/schema/__tests__/partner_show_events.test.ts
```diff
@@ -0,0 +1,154 @@
...
+              end_at: "2019-03-28T21:00:00",
+              start_at: "2019-03-28T18:00:00",
+            },
+          ],
+        },
+      })
+    })
+  })
+
+  describe("if user agent is returned as an array", () => {
```
dleve123: Can we include something about Eigen in this docstring as well?

src/schema/partner_show_event.ts
```diff
@@ -1,8 +1,46 @@
-import date from "./fields/date"
-import { GraphQLString, GraphQLObjectType } from "graphql"
+import dateField, { date, DateSource } from "./fields/date"
+import { GraphQLString, GraphQLObjectType, GraphQLFieldConfig } from "graphql"
 import { ResolverContext } from "types/graphql"
 import { exhibitionPeriod } from "lib/date"
 
+const checkUserAgent = (userAgent): boolean =>
```
dleve123: Bit of a nit-pick, but I recommend a less generic name here - maybe hasOldEmissionUserAgent

src/schema/__tests__/partner_show_events.test.ts
anandaroop: Agreed 👍

src/schema/__tests__/partner_show_events.test.ts
```diff
@@ -0,0 +1,154 @@
...
+              start_at: "2019-03-28T18:00:00+00:00",
+            },
+          ],
+        },
+      })
+    })
+  })
+
+  describe("with Eigen 4.x", () => {
+    it("truncates the UTC offset", async () => {
```
alloy: Why are we doing that, instead of only those with the new show view?

src/schema/partner_show_event.ts
```diff
@@ -1,8 +1,46 @@
...
+  userAgent!.indexOf("Artsy-Mobile/4") > 0 ||
+  userAgent!.indexOf("Artsy-Mobile/5.0.0") > 0 ||
+  userAgent!.indexOf("Artsy-Mobile/5.0.1") > 0
+
+const isOlderEmissionVersion = (userAgent): boolean => {
+  let result = false
+  if (typeof userAgent === "string") {
+    result = checkUserAgent(userAgent)
+  } else if (Array.isArray(userAgent)) {
```
alloy: Not: no need for an extra function, just pass `checkUserAgent` as an argument.

src/schema/partner_show_event.ts
```diff
@@ -1,8 +1,46 @@
...
+  resolve: (
+    obj,
+    { format, timezone },
+    { defaultTimezone, userAgent },
+    { fieldName }
+  ) => {
+    const rawDate = obj[fieldName]
+
+    if (userAgent && isOlderEmissionVersion(userAgent)) {
+      const dateWithoutOffset = rawDate.replace(/[-+]\d\d:\d\d$/g, "")
```
alloy: Out of curiosity, why is this regexp made greedy?

src/schema/__tests__/partner_show_events.test.ts
anandaroop: @kierangillen can confirm but our understanding was that the new Show and Fair pages were released to users before the major version bump that actually introduced the City Guide functionality.

@orta is this something you can confirm? (Other folks who would know are OOO at the moment)

src/schema/__tests__/partner_show_events.test.ts
joeyAghion: This might be the first real instance of the kind of time-bomb test I imagined, so thanks. 😍 

I would expect to see a more specific assertion here, like `expect(data).toEqual("... +00:00")` which would fail after the deadline. It looks like that would require piggy-backing on an existing test or lots of repeat stubbing, so this is fine too.

src/schema/partner_show_event.ts
anandaroop: Oh do you mean the "global" `/g` flag? If so, that was an oversight, left in from when I was regex faffing. Will remove.

src/schema/__tests__/partner_show_events.test.ts
alloy: Right, I do recall these rolling out before v5, but not in eg v4.0.0

src/schema/partner_show_event.ts
```diff
@@ -1,8 +1,46 @@
-import date from "./fields/date"
-import { GraphQLString, GraphQLObjectType } from "graphql"
+import dateField, { date, DateSource } from "./fields/date"
+import { GraphQLString, GraphQLObjectType, GraphQLFieldConfig } from "graphql"
 import { ResolverContext } from "types/graphql"
 import { exhibitionPeriod } from "lib/date"
 
+const hasOldEmissionUserAgentString = (userAgent): boolean =>
```
alloy: Btw, you should always type the input parameters, otherwise TS defaults to `any`.

src/schema/__tests__/partner_show_events.test.ts
anandaroop: Good point! I'm happy to tighten up the version spec here, if we can figure out which versions are affected.

src/schema/partner_show_event.ts
anandaroop: Thanks, fixed (think I missed this bc the feedback about it in VSC is so understated).

src/schema/__tests__/partner_show_events.test.ts
alloy: Looks like it was 4.4.0 https://github.com/artsy/eigen/blob/27482286c7b29222228192fad732769bd7e61cf7/CHANGELOG.yml#L20

src/schema/partner_show_event.ts
alloy: This is one of those more strict options that we haven’t been using yet. There’s a compiler option that disallows any implicit `any` and there’s a TSLint rule to require parameter types.

src/schema/partner_show_event.ts
```diff
@@ -1,8 +1,46 @@
-import date from "./fields/date"
-import { GraphQLString, GraphQLObjectType } from "graphql"
+import dateField, { date, DateSource } from "./fields/date"
+import { GraphQLString, GraphQLObjectType, GraphQLFieldConfig } from "graphql"
 import { ResolverContext } from "types/graphql"
 import { exhibitionPeriod } from "lib/date"
 
+const hasOldEmissionUserAgentString = (userAgent: string | string[]): boolean =>
```
alloy: @anandaroop This typing isn’t correct. You want _just_ `string`, because you’re already filtering out the array below when you invoke this function for each entry in the array. The reason this isn’t failing is because TS thinks you’re possibly looking for an exact match of these strings in the array.

anandaroop: 😿  with multiple insertion points comes great responsibility

alloy: It’s still better than implicit `any`!

src/schema/__tests__/partner_show_events.test.ts
anandaroop: 💣 💥 https://artsy.slack.com/archives/C02BC3HEJ/p1601567857100300

alloy: :ceiling-cat: Did something explode?

alloy: Ah, I see it now 😄

anandaroop: Yes, just noting that the bomb went off as planned 😄 

(And hello, friend!)

dleve123: 👋  -so cool to see this worked!

alloy: 👋 Best date. When do we meet next time?

anandaroop: Hey @dleve123 👋 . Yeah this spec was just was disruptive as we... wanted?.. it to be? Lol

I'll go write another time bomb so we can meet again.

PR #1644, opened 2019-04-03
Make sure that we're comparing timestamps in the same zone

dblandin:
When building up the description for shows and fair booth search result items, make sure that we're comparing moment timestamps within the same UTC zone.

Should address this flickering spec: https://circleci.com/gh/artsy/metaphysics/4712


dblandin:
Going to merge this in.

PR #1646, opened 2019-04-04
Fake an upstream GraphQL change to valdiate schema changing

orta:
This PR adds a type that doesn't exist inside the Kaws schema, so it would be classed as missing from the staging schema. This should trigger a peril run which will validate the local data against that schema, which should present a warning.


joeyAghion:
Cleaning this up...

Might be useful again in the future though.

PR #1648, opened 2019-04-04
Pin comparison date to start of day

dblandin:
Compare now, at the beginning of the day, with show / fair booth start/end timestamps.


anandaroop:
Thanks!

PR #1649, opened 2019-04-04
Adds formattedOpeningHours to Fairs

kierangillen:
- This is the first step in https://artsyproduct.atlassian.net/browse/LD-568 which will pull the Fair display dates out of Emission and into Metaphysics

### DIFF
--- a/src/lib/date.ts
+++ b/src/lib/date.ts
@@ -10,7 +10,7 @@ const formattedOpeningHoursDate = date => {
   } else if (momentHour && momentDate) {
     return `${momentDate} at ${momentHour}`
   } else if (momentDate) {
-    return `${momentDate}`
+    return momentDate
   }
 }
 
--- a/src/lib/date.ts
+++ b/src/lib/date.ts
@@ -1,8 +1,32 @@
 import moment from "moment"
 
+const formatDateStatusDisplayDate = date => {
+  const momentToUse = moment.utc(date, "YYYY-MM-DD[T]HH:mm:ss")
+  const momentDate = momentToUse.format("MMM D")
+  const momentHour = momentToUse.format("ha")
+  if (!!momentHour && !!momentDate) {
+    return ` ${momentDate} at ${momentHour}`
+  } else if (!!momentDate) {
+    return ` ${momentDate}`
+  }
+}
+
+export function dateStatusDisplay(startAt, endAt) {
+  const thisMoment = moment()
+  const startMoment = moment(startAt, "YYYY-MM-DD[T]HH:mm:ss")
+  const endMoment = moment(endAt, "YYYY-MM-DD[T]HH:mm:ss")
+  if (thisMoment.isBefore(startMoment)) {
+    return `Opens${formatDateStatusDisplayDate(startAt)}`
+  } else if (thisMoment.isBefore(endMoment)) {
+    return `Closes${formatDateStatusDisplayDate(endAt)}`
+  } else {
+    return "Closed"
+  }
+}
+
 export function exhibitionPeriod(startAt, endAt) {
-  const startMoment = moment(startAt)
-  const endMoment = moment(endAt)
+  const startMoment = moment.utc(startAt, "YYYY-MM-DD[T]HH:mm:ss")
+  const endMoment = moment.utc(endAt, "YYYY-MM-DD[T]HH:mm:ss")
   const thisMoment = moment()
   let startFormat = "MMM D"
   let endFormat = "D"
--- a/src/lib/date.ts
+++ b/src/lib/date.ts
@@ -4,7 +4,11 @@ const formattedDateStatusDate = date => {
   const momentToUse = moment.utc(date, "YYYY-MM-DD[T]HH:mm:ss")
   const momentDate = momentToUse.format("MMM D")
   const momentHour = momentToUse.format("ha")
-  if (!!momentHour && !!momentDate) {
+  const momentMinutes = momentToUse.format("mm")
+  const momentHourWithMinutes = momentToUse.format("h:mma")
+  if (!!momentHour && !!momentDate && momentMinutes !== "00") {
+    return `${momentDate} at ${momentHourWithMinutes}`
+  } else if (!!momentHour && !!momentDate) {
     return `${momentDate} at ${momentHour}`
   } else if (!!momentDate) {
     return `${momentDate}`
--- a/src/lib/date.ts
+++ b/src/lib/date.ts
@@ -1,7 +1,7 @@
 import moment from "moment"
 
 const formattedOpeningHoursDate = date => {
-  const momentToUse = moment.utc(date, "YYYY-MM-DD[T]HH:mm:ss")
+  const momentToUse = moment.utc(date)
   const momentDate = momentToUse.format("MMM D")
   const momentHour = momentToUse.format("ha")
   if (momentHour && momentDate && momentToUse.minutes() !== 0) {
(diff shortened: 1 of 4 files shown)


src/schema/fair.ts
```diff
@@ -144,6 +144,12 @@ export const FairType = new GraphQLObjectType<any, ResolverContext>({
       description: "A formatted description of the start to end dates",
       resolve: ({ start_at, end_at }) => exhibitionPeriod(start_at, end_at),
     },
+    dateStatusDisplay: {
```
alloy: I think this field could have a more self-documenting name, something like ‘formatted opening status’?

src/lib/date.ts
```diff
@@ -1,8 +1,32 @@
 import moment from "moment"
 
+const formatDateStatusDisplayDate = date => {
+  const momentToUse = moment.utc(date, "YYYY-MM-DD[T]HH:mm:ss")
+  const momentDate = momentToUse.format("MMM D")
+  const momentHour = momentToUse.format("ha")
+  if (!!momentHour && !!momentDate) {
+    return ` ${momentDate} at ${momentHour}`
+  } else if (!!momentDate) {
+    return ` ${momentDate}`
```
alloy: Is there a reason we should have the preceding whitespace here instead of below? I think it would make the code slightly more readable.

src/lib/date.ts
```diff
@@ -1,8 +1,32 @@
...
+  } else {
+    return "Closed"
+  }
+}
+
 export function exhibitionPeriod(startAt, endAt) {
-  const startMoment = moment(startAt)
-  const endMoment = moment(endAt)
+  const startMoment = moment.utc(startAt, "YYYY-MM-DD[T]HH:mm:ss")
+  const endMoment = moment.utc(endAt, "YYYY-MM-DD[T]HH:mm:ss")
```
alloy: Is `exhibitionPeriod` not used elsewhere that _shouldn’t_ change?

kierangillen: My test was failing because I didn't include the moment format, so I thought I'd add it here too because I was the original author of this function. That being said I can remove but I believe moving forward moment.js wants the format indicated.

src/lib/date.ts
kierangillen: If for some reason this returns as `null` then the rendered text will be centered without any whitespace, which still workings with the design. I.e. it'll say "Opens" under the countdown clock. Thought this was marginally better.

src/schema/fair.ts
kierangillen: Yes I wasn't super happy with the name either, I'll fix.

kierangillen: That being said it's not just opening but can be used to indicate it's closed so I may go with `formattedDateStatus`

src/lib/date.ts
kierangillen: Fixed this to your recommendation.

src/lib/date.ts
```diff
@@ -1,8 +1,32 @@
 import moment from "moment"
 
+const formattedDateStatusDate = date => {
+  const momentToUse = moment.utc(date, "YYYY-MM-DD[T]HH:mm:ss")
+  const momentDate = momentToUse.format("MMM D")
+  const momentHour = momentToUse.format("ha")
```
anandaroop: This is another case of dropping minute precision right? I'm not sure we should ever use `ha` as a display format, since that is by definition lossy with respect to minutes.

src/schema/__tests__/fair.test.js
```diff
@@ -461,6 +461,90 @@ describe("Fair", () => {
...
+        const mockFairLoader = jest.fn(() => Promise.resolve(fairData))
+        context = {
+          fairLoader: mockFairLoader,
+        }
+
+        const data = await runQuery(query, context)
+
+        expect(data).toEqual({
+          fair: {
+            formattedDateStatus: "Closes Feb 4 at 12pm",
```
anandaroop: So if we agree to my other comment this should assert that we show minutes as well. 

(Unless it is at the top of the hour, which I suspect is the _only_ case in which we may reasonably drop the `:00` — could be good to run this by design to confirm the various scenarios)

kierangillen: Yes good point I'll add the minutes when applicable.

kierangillen: Just updated the function and added a test to check minutes

src/lib/date.ts
```diff
@@ -4,7 +4,11 @@ const formattedDateStatusDate = date => {
   const momentToUse = moment.utc(date, "YYYY-MM-DD[T]HH:mm:ss")
   const momentDate = momentToUse.format("MMM D")
   const momentHour = momentToUse.format("ha")
-  if (!!momentHour && !!momentDate) {
+  const momentMinutes = momentToUse.format("mm")
+  const momentHourWithMinutes = momentToUse.format("h:mma")
+  if (!!momentHour && !!momentDate && momentMinutes !== "00") {
```
alloy: There’s no need to format a string for the minutes check, could just do `momentMinutes.minutes() !== 0`.

I’d also like to move the `momentHourWithMinutes` formatting _inside_ the matching `if` branch so it isn’t done unneeded.

alloy: Btw, is there a good reason for the double bang? Afaik we only need that when we need to pass a boolean value, not when performing a conditional check.

src/lib/date.ts
alloy: Ok, that’s fine then, no need to change if you think that’s good 👍

src/schema/fair.ts
alloy: Perhaps it’s my non-native English (?), but I do think that eg “opening hours” could be “closed on Sunday”. If y’all think that the current iteration is a good self-documenting name, then I’m fine with it.

kierangillen: Ok you've convinced me, I present thee `formattedOpeningHours`

src/lib/date.ts
kierangillen: Cleaned this section up.

src/lib/date.ts
anandaroop: Hmm, this didn't sound right to me, so I did some fiddling with the code here. 

There is an _extremely_ devilish gotcha in this code — many of the fixture dates in the spec file are using en dashes to delimit the date parts instead of hyphens.

So on [L484](https://github.com/artsy/metaphysics/pull/1649/files?file-filters%5B%5D=.js&file-filters%5B%5D=.ts#diff-5ee543dceb7f3086afc41b9ce10c1be2R484) we've got:

```
          start_at: "2019–02–06T12:00:56+00:00",
```

instead of 

```
          start_at: "2019-02-06T12:00:56+00:00",
```
(At the risk of stating the obvious, en dashes are appropriate for _display_ but would never be returned as part of an ISO8601 date string.)

I think if you fix that, you will take care of the unexpected spec failure as well.

kierangillen: Oh wow this was my issue.

kierangillen: Fixed

src/lib/date.ts
```diff
@@ -1,8 +1,35 @@
...
+  const momentToUse = moment.utc(date)
+  const momentDate = momentToUse.format("MMM D")
+  const momentHour = momentToUse.format("ha")
+  if (momentHour && momentDate && momentToUse.minutes() !== 0) {
+    const momentHourWithMinutes = momentToUse.format("h:mma")
+    return `${momentDate} at ${momentHourWithMinutes}`
+  } else if (momentHour && momentDate) {
+    return `${momentDate} at ${momentHour}`
+  } else if (momentDate) {
+    return `${momentDate}`
```
alloy: Last nit, no need to interpolate this string

PR #1651, opened 2019-04-09
Partner operator shouldn't be able to access another partner's auctions

yuki24:
fixes https://artsyproduct.atlassian.net/browse/AUCT-313

The check that got removed in this commit: https://github.com/artsy/metaphysics/commit/9e39a429e85052f54348986ddb7fb8a6356b7b13?w=1#diff-873fe1c7874f0eb5e4c8c9caed5efd90L105 should not have been removed. Because of the lack of the check any partner could open the Operator2 for any auction. This PR brings the check back to fix it.

### DIFF
--- a/src/schema/causality_jwt.ts
+++ b/src/schema/causality_jwt.ts
@@ -102,7 +102,12 @@ const CausalityJWT: GraphQLFieldConfig<void, ResolverContext> = {
 
           if (sale.partner) {
             return mePartnersLoader({ "partner_ids[]": sale.partner._id }).then(
-              () => {
+              mePartners => {
+                // Check if current user has access to partner running the sale
+                if (mePartners.length === 0) {
+                  throw new Error("Unauthorized to be operator")
+                }
+
                 return jwt.encode(
                   {
                     aud: "auctions",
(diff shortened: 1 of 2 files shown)


src/schema/causality_jwt.ts
```diff
@@ -102,7 +102,12 @@ const CausalityJWT: GraphQLFieldConfig<void, ResolverContext> = {
 
           if (sale.partner) {
             return mePartnersLoader({ "partner_ids[]": sale.partner._id }).then(
-              () => {
+              mePartners => {
+                // Check if current user has access to partner running the sale
+                if (mePartners.length === 0) {
+                  throw new Error("Unauthorized to be operator")
```
williardx: Given this, do we need the `throw new Error` statement below? Or maybe we should change the message since it seems like it'd get triggered if we tried to access partner operator for a sale that has no partner?

williardx:
Only one comment but otherwise looks good! Thanks @yuki24!

src/schema/causality_jwt.ts
yuki24: Prediction directly depends on [the error message](https://github.com/artsy/prediction/blob/ca0c0e4da4233078ba9d8c58e425a26bfd9feeee/src/server/helpers.ts#L67) so it would require a few more changes if we want to be more explicit (any other error messages than `"Unauthorized to be operator"` would result in 500 right now), but it shouldn't be too difficult to distinguish two cases.

PR #1653, opened 2019-04-10
update circleci job contexts #PLATFORM-1357

izakp:
AWS IAM credentials set via the "hokusai" circleci context


ashfurrow:
This is why CI is failing for metaphysics right now, right? This PR is green and has 👍 from both requested reviewers, so I'm going to merge.

PR #1654, opened 2019-04-10
Remove placeholder content from sponsored content fair activations

anandaroop:
This simply removes some dummy content to prevent it showing up in the iOS app.

#trivial


ashleyjelks:
Looks good!

PR #1655, opened 2019-04-10
Adds new formatted date fields

kierangillen:
- Creates date formatting fields to match formatting documented here: https://www.notion.so/artsy/Formatting-Time-f6f4aaeeba414583a207013fec5a9794

- `Current.CurrentEvent`, `Show.exhibition_period`, `PartnerShowEvent.dateTimeRange`, `PartnerShowEvent.exhibitionPeriod`, `PartnerShow.exhibition_period`, `Fair.exhibition_period`, `Fair.formattedOpeningHours` now take a hardcoded `"UTC"` timezone which we can replace once corresponding inputs handle timezone entry.

### DIFF
--- a/src/lib/__tests__/date.test.ts
+++ b/src/lib/__tests__/date.test.ts
@@ -226,7 +226,7 @@ describe("date formatting", () => {
         expect(period).toBe("Dec 5, 2019 from 8:00 – 10:00pm UTC")
       })
 
-      it("does not include year both dates are same day and the current year", () => {
+      it("does not include year when both dates are same day and the current year", () => {
         const period = dateTimeRange(
           "2018-12-05T20:00:00+00:00",
           "2018-12-05T22:00:00+00:00",
--- a/src/lib/__tests__/date.test.ts
+++ b/src/lib/__tests__/date.test.ts
@@ -17,7 +17,8 @@ describe("date formatting", () => {
     it("If within same am/pm only display one am or pm", () => {
       const period = timeRange(
         "2022-12-30T08:00:00+00:00",
-        "2022-12-30T10:00:00+00:00"
+        "2022-12-30T10:00:00+00:00",
+        "UTC"
       )
       expect(period).toBe("8:00 – 10:00am UTC")
     })
@@ -25,7 +26,8 @@ describe("date formatting", () => {
     it("If not within same am/pm show both time periods", () => {
       const period = timeRange(
         "2021-12-05T08:00:00+00:00",
-        "2022-12-30T17:00:00+00:00"
+        "2022-12-30T17:00:00+00:00",
+        "UTC"
       )
       expect(period).toBe("8:00am – 5:00pm UTC")
     })
@@ -44,7 +46,8 @@ describe("date formatting", () => {
     it("Dates are the same day returns true", () => {
       const period = datesAreSameDay(
         "2022-12-30T20:00:00+00:00",
-        "2022-12-30T17:00:00+00:00"
+        "2022-12-30T17:00:00+00:00",
+        "UTC"
       )
       expect(period).toBe(true)
     })
@@ -52,7 +55,8 @@ describe("date formatting", () => {
     it("Dates are not the same day returns false", () => {
       const period = datesAreSameDay(
         "2021-12-05T20:00:00+00:00",
-        "2022-12-30T17:00:00+00:00"
+        "2022-12-30T17:00:00+00:00",
+        "UTC"
       )
       expect(period).toBe(false)
     })
@@ -88,7 +92,8 @@ describe("date formatting", () => {
     it("Event opens in the future", () => {
       const period = formattedOpeningHours(
         "2021-12-05T20:00:00+00:00",
-        "2022-12-30T17:00:00+00:00"
+        "2022-12-30T17:00:00+00:00",
+        "UTC"
       )
       expect(period).toBe("Opens Dec 5 at 8:00pm UTC")
     })
@@ -96,7 +101,8 @@ describe("date formatting", () => {
     it("Event is running and closes in the future", () => {
       const period = formattedOpeningHours(
         "2017-12-05T20:00:00+00:00",
-        "2019-12-30T17:00:00+00:00"
+        "2019-12-30T17:00:00+00:00",
+        "UTC"
       )
       expect(period).toBe("Closes Dec 30 at 5:00pm UTC")
     })
@@ -104,7 +110,8 @@ describe("date formatting", () => {
     it("Event ended in the past and is closed", () => {
       const period = formattedOpeningHours(
         "2016-12-05T20:00:00+00:00",
-        "2016-12-30T17:00:00+00:00"
+        "2016-12-30T17:00:00+00:00",
+        "UTC"
       )
       expect(period).toBe("Closed")
     })
@@ -112,7 +119,7 @@ describe("date formatting", () => {
 
   describe(singleTime, () => {
     it("Return singleTime moment by the hour using default UTC timezone", () => {
-      const period = singleTime("2018-12-05T20:00:00+00:00")
+      const period = singleTime("2018-12-05T20:00:00+00:00", "UTC")
       expect(period).toBe("8:00pm UTC")
     })
 
@@ -132,12 +139,12 @@ describe("date formatting", () => {
     })
 
     it("Returns single date and time in this year", () => {
-      const period = singleDateTime("2018-12-05T20:00:00+00:00")
+      const period = singleDateTime("2018-12-05T20:00:00+00:00", "UTC")
       expect(period).toBe("Dec 5 at 8:00pm UTC")
     })
 
     it("Returns single date in different year from now", () => {
-      const period = singleDateTime("2021-12-05T20:00:00+00:00")
+      const period = singleDateTime("2021-12-05T20:00:00+00:00", "UTC")
       expect(period).toBe("Dec 5, 2021 at 8:00pm UTC")
     })
 
@@ -160,12 +167,12 @@ describe("date formatting", () => {
     })
 
     it("Returns single date in this year with day of week", () => {
-      const period = singleDate("2018-12-05T20:00:00+00:00")
+      const period = singleDate("2018-12-05T20:00:00+00:00", "UTC")
       expect(period).toBe("Wed, Dec 5")
     })
 
     it("Returns single date with day of week in different year from now", () => {
-      const period = singleDate("2021-12-05T20:00:00+00:00")
+      const period = singleDate("2021-12-05T20:00:00+00:00", "UTC")
       expect(period).toBe("Sun, Dec 5, 2021")
     })
   })
@@ -182,7 +189,8 @@ describe("date formatting", () => {
     it("Returns start and end datetimes with different years", () => {
       const period = dateTimeRange(
         "2021-12-05T20:00:00+00:00",
-        "2022-12-30T17:00:00+00:00"
+        "2022-12-30T17:00:00+00:00",
+        "UTC"
       )
       expect(period).toBe("Dec 5, 2021 at 8:00pm – Dec 30, 2022 at 5:00pm UTC")
     })
@@ -190,7 +198,8 @@ describe("date formatting", () => {
     it("Returns the start and end date with different year from now", () => {
       const period = dateTimeRange(
         "2040-12-05T20:00:00+00:00",
-        "2040-12-30T17:00:00+00:00"
+        "2040-12-30T17:00:00+00:00",
+        "UTC"
       )
       expect(period).toBe("Dec 5, 2040 at 8:00pm – Dec 30, 2040 at 5:00pm UTC")
     })
@@ -198,7 +207,8 @@ describe("date formatting", () => {
     it("Returns the start and end datetimes in current year", () => {
       const period = dateTimeRange(
         "2018-12-05T20:00:00+00:00",
-        "2018-12-30T17:00:00+00:00"
+        "2018-12-30T17:00:00+00:00",
+        "UTC"
       )
       expect(period).toBe("Dec 5 at 8:00pm – Dec 30 at 5:00pm UTC")
     })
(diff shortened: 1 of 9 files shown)


src/lib/date.ts
```diff
@@ -14,7 +14,60 @@ const formattedOpeningHoursDate = date => {
...
+  } else {
+    // Duration is a range
+    return false
+  }
+}
+
 export function formattedOpeningHours(startAt, endAt) {
+  // Opens Mar 29 at 4pm
+  // Closes Apr 3 at 12:30pm
+  // Closed
```
damassi: Could this comment be expanded slightly, perhaps in a docblock form? Something that makes it a bit clearer rather than floating strings.

src/lib/date.ts
```diff
@@ -27,7 +80,81 @@ export function formattedOpeningHours(startAt, endAt) {
...
+  if (singleHour.minutes() === 0) {
+    hour = singleHour.format("ha")
+  } else {
+    hour = singleHour.format("h:mma")
+  }
+  return hour
+}
+
+export function singleDate(date) {
+  // Aug 5, 5:00pm
```
damassi: Same here -- its idomatic to use DocBlocks annotated at the top of the function, e.g., 
```
/**
 * 
 *
 */
```

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
 
-describe("date", () => {
-  describe("exhibitionPeriod", () => {
-    it("includes the start and end date", () => {
-      const period = exhibitionPeriod(
-        moment("2011-01-01T00:00-0400"),
-        moment("2014-04-19T00:00-0400")
+describe("date formatting", () => {
+  describe("formattedHours", () => {
+    it("If within same am/pm", () => {
```
alloy: This is not a proper sentence and doesn’t describe the behaviour, which is important for people to judge if the code is working as it should. E.g. something like:

```js
it("drops the period component if start and end time are within the same period", () => {
  // …
})
```

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+      const period = formattedHours(
+        "2022-12-30T08:00:00+00:00",
+        "2022-12-30T10:00:00+00:00"
+      )
+      expect(period).toBe("8:00 – 10:00am")
```
alloy: This is missing the TZ name, as per [the spec](https://www.notion.so/artsy/Formatting-Time-f6f4aaeeba414583a207013fec5a9794#16c45f5d8e514e6db248990d97cd7d84) and this should also take a destination TZ for what TZ to consider these times in.

Same applies to test below.

src/lib/__tests__/date.test.ts
alloy: (Same applies to all other tests.)

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+    it("If within across both am/pm", () => {
+      const period = formattedHours(
+        "2021-12-05T08:00:00+00:00",
+        "2022-12-30T17:00:00+00:00"
+      )
+      expect(period).toBe("8:00am – 5:00pm")
+    })
+  })
+
+  describe("datesAreSameDay", () => {
```
alloy:
```suggestion
  describe(datesAreSameDay, () => {
```

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+      expect(period).toBe(true)
+    })
+
+    it("Dates are not the same day", () => {
+      const period = datesAreSameDay(
+        "2021-12-05T20:00:00+00:00",
+        "2022-12-30T17:00:00+00:00"
+      )
+      expect(period).toBe(false)
+    })
```
alloy: What is `datesAreSameDay` behaviour when times are each in different timezones? I think you can really only think of a ‘day’ when times are in the same TZ, as a time in one TZ could be a different day in another TZ in _real-time_.

Something that I could imagine is: 

```js
function datesAreSameDay(time1, time2, consideredInTZ) {
  // …
}
```

src/schema/partner_show_event.ts
```diff
@@ -58,10 +63,20 @@ const PartnerShowEventType = new GraphQLObjectType<any, ResolverContext>({
     },
     start_at: dateFieldForPartnerShowEvent,
     end_at: dateFieldForPartnerShowEvent,
+    datesAreSameDay: {
+      type: GraphQLBoolean,
+      description: "Returns true if dates are on same day",
+      resolve: ({ start_at, end_at }) => datesAreSameDay(start_at, end_at),
+    },
```
alloy: Seeing as I can’t find `datesAreSameDay` in the Notion doc _and_ I have some concerns about this implementation across timezones, do we really need to expose this now? (i.e. is there a client that needs this currently?)

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+  })
+
+  describe("formattedOpeningHours", () => {
```
alloy:
```suggestion
  describe(formattedOpeningHours, () => {
```

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+    it("Event ended in the past and is closed", () => {
+      const period = formattedOpeningHours(
+        "2016-12-05T20:00:00+00:00",
+        "2016-12-30T17:00:00+00:00"
       )
+      expect(period).toBe("Closed")
+    })
+  })
+
+  describe("singleHour", () => {
```
alloy:
```suggestion
  describe(singleHour, () => {
```

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+    it("includes single date in this year", () => {
+      const period = singleHour("2018-12-05T20:00:00+00:00")
+      expect(period).toBe("8:00pm")
+    })
+  })
+
+  describe("singleDate", () => {
```
alloy:
```suggestion
  describe(singleDate, () => {
```

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+      expect(period).toBe("Dec 5 at 8:00pm")
+    })
+
+    it("includes single date in different year from now", () => {
+      const period = singleDate("2021-12-05T20:00:00+00:00")
+      expect(period).toBe("Dec 5, 2021 at 8:00pm")
+    })
+  })
+
+  describe("singleDateWithoutHour", () => {
```
alloy:
```suggestion
  describe(singleDateWithoutHour, () => {
```

src/lib/__tests__/date.test.ts
alloy: How about renaming this to `formattedTime`?

src/lib/__tests__/date.test.ts
alloy: How about renaming this to `formattedDateTime`?

src/lib/__tests__/date.test.ts
alloy: How about renaming this to `formattedDate`?

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+      expect(period).toBe("Wed, Dec 5")
+    })
+
+    it("includes single date in different year from now", () => {
+      const period = singleDateWithoutHour("2021-12-05T20:00:00+00:00")
+      expect(period).toBe("Sun, Dec 5, 2021")
+    })
+  })
+
+  describe("datesWithHours", () => {
```
alloy:
```suggestion
  describe(datesWithHours, () => {
```

alloy: How about renaming this to `formattedDateTimeRange`?

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+    it("date is same day in same year as now", () => {
+      const period = datesWithHours(
+        "2018-12-05T20:00:00+00:00",
+        "2018-12-05T22:00:00+00:00"
+      )
+      expect(period).toBe("Dec 5 from 8:00 – 10:00pm")
+    })
+  })
+
+  describe("datesWithDaysAndHours", () => {
```
alloy:
```suggestion
  describe(datesWithDaysAndHours, () => {
```

alloy: This could perhaps also just be the same function as above with a `includeDayName` parameter?

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+    beforeEach(() => {
+      Date.now = () => new Date("2018-01-30T03:24:00") as any
+    })
+    afterEach(() => {
+      Date.now = realNow
+    })
+
+    it("includes single date in this year", () => {
+      const period = singleDate("2018-12-05T20:00:00+00:00")
+      expect(period).toBe("Dec 5 at 8:00pm")
```
alloy: Missing the TZ component. https://www.notion.so/artsy/Formatting-Time-f6f4aaeeba414583a207013fec5a9794#a9f31923b35d41579c400bd76ce3ba13

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+        "2016-12-30T17:00:00+00:00"
       )
+      expect(period).toBe("Closed")
+    })
+  })
+
+  describe("singleHour", () => {
+    it("includes single date in this year", () => {
+      const period = singleHour("2018-12-05T20:00:00+00:00")
+      expect(period).toBe("8:00pm")
```
alloy: Missing the TZ component. https://www.notion.so/artsy/Formatting-Time-f6f4aaeeba414583a207013fec5a9794#4372bad667d541f991d9ec5d6791c51e

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+      Date.now = realNow
+    })
+
+    it("includes the start and end date with different years", () => {
+      const period = datesWithDaysAndHours(
+        "2021-12-05T20:00:00+00:00",
+        "2022-12-30T17:00:00+00:00"
+      )
+      expect(period).toBe(
+        "Sun, Dec 5, 2021 at 8:00pm – Fri, Dec 30, 2022 at 5:00pm"
```
alloy: Missing the TZ component https://www.notion.so/artsy/Formatting-Time-f6f4aaeeba414583a207013fec5a9794#4e5dea1d5229420793919615a3e4b4be

src/lib/__tests__/date.test.ts
```diff
@@ -1,51 +1,271 @@
...
+    afterEach(() => {
+      Date.now = realNow
+    })
+
+    it("includes the start and end date with different years", () => {
+      const period = datesWithHours(
+        "2021-12-05T20:00:00+00:00",
+        "2022-12-30T17:00:00+00:00"
+      )
+      expect(period).toBe("Dec 5, 2021 at 8:00pm – Dec 30, 2022 at 5:00pm")
```
alloy: Missing the TZ component https://www.notion.so/artsy/Formatting-Time-f6f4aaeeba414583a207013fec5a9794#b0cf40f292634c7b8cc2fe026089919f

alloy:
@kierangillen bump

src/schema/partner_show_event.ts
kierangillen: In Emission there's a component that requires two different designs depending on whether the time is the same day, which would need this logic.

https://app.zeplin.io/project/5bef4c7dae2b6a3eb6a75f4a/screen/5c1ae7415ff90824363b8f42
vs. 
https://app.zeplin.io/project/5bef4c7dae2b6a3eb6a75f4a/screen/5c2e4dd581415cb00682abb9

The design could be pushed back on if we don't want to expose this logic. Currently the logic lives in the client and exposing this field would be in preparation for removing Moment.js from Eigen/Emission.

kierangillen:
@alloy Ready for review when you're available.

src/schema/partner_show_event.ts
alloy: Is there a way to move all of that logic out of the client into MP in a single field, rather than exposing this detail?

src/lib/__tests__/date.test.ts
```diff
@@ -1,59 +1,323 @@
...
+  describe(singleDateTime, () => {
+    const realNow = Date.now
+    beforeEach(() => {
+      Date.now = () => new Date("2018-01-30T03:24:00") as any
+    })
+    afterEach(() => {
+      Date.now = realNow
+    })
+
+    it("Returns single date and time in this year", () => {
```
alloy:
```suggestion
    it("includes only the month day and time for dates in the current year", () => {
```

src/lib/__tests__/date.test.ts
```diff
@@ -1,59 +1,323 @@
...
+      const period = singleDateTime("2018-12-05T20:00:00+00:00")
+      expect(period).toBe("Dec 5 at 8:00pm UTC")
+    })
+
+    it("Returns single date in different year from now", () => {
```
alloy:
```suggestion
    it("also includes the year for dates in a different year from now", () => {
```

src/schema/partner_show_event.ts
kierangillen: Yeah, after looking at it again we can just use a single field here, I wont expose this field.

kierangillen:
@alloy Okay! Ready for one more (hopefully last) pass. :-)

src/lib/__tests__/date.test.ts
```diff
@@ -1,59 +1,365 @@
...
 
-describe("date", () => {
-  describe("exhibitionPeriod", () => {
-    it("includes the start and end date", () => {
-      const period = exhibitionPeriod(
-        moment("2011-01-01T00:00-0400"),
-        moment("2014-04-19T00:00-0400")
+describe("date formatting", () => {
+  describe(timeRange, () => {
+    it("if within same am/pm only display one am or pm", () => {
```
alloy: I think you may not be including the “it” when you’re thinking of these descriptions. Here the description should probably read:

```suggestion
    it("only displays one am or pm if within same am/pm", () => {
```

…as in, “it only displays one am or pm if within same am/pm”.

kierangillen: Ah! Yes I was never considering `it` part of the sentence.

alloy: Also, the ‘it’ refers to your subject specified with the `describe` function.

PR #1656, opened 2019-04-10
Returns null for missing dates

ashfurrow:
I came across some weird behaviour from Metaphysics when working on [this PR](https://github.com/artsy/reaction/pull/2282). It appeared as though various dates on MP's `Sale` type were being resolve to that literal string `"Invalid date"`:

Screen Shot 2019-04-10 at 2 19 14 PM

I confirmed with Reaction's MP logging middleware that these strings were _actually_ being returned from MP, but visiting the in-browser GrapiQL interface shows all the results as `null`:

Screen Shot 2019-04-10 at 2 20 53 PM

So this is weird, eh? We even have code in Reaction that works around this bug:

https://github.com/artsy/reaction/blob/2d9f867189e716f6881d8d8c6792bcd24a47c23b/src/Components/v2/AuctionTimer.tsx#L25-L28

`"Invalid date"` is [moment's default value for unparsable dates](https://github.com/moment/moment/blob/d0f0dd8f40be987593767a3dc5e27c45c6a3fa49/moment.js#L451); I took a closer look and the reason we see different values in the browser GraphiQL and Reaction Storybooks is because Metaphysics behaves differently if it gets an `X-TIMEZONE` HTTP header. Adding this header reproduces the bug:

Screen Shot 2019-04-10 at 2 25 08 PM

This PR fixes the bug by having a check for `null` dates (which avoids moment parsing altogether). This is a pretty wide-reaching change, and while I don't expect any problems after merging, I want to kick the tires around staging before we promote.

### DIFF
--- a/src/schema/fields/__tests__/date.test.js
+++ b/src/schema/fields/__tests__/date.test.js
@@ -1,4 +1,4 @@
-import { date } from "schema/fields/date"
+import dateField, { date } from "schema/fields/date"
 
 describe("date", () => {
   const rawDate = "2020-12-31T12:00:00+00:00"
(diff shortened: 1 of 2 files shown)


mzikherman:
Interesting! Great sleuthing.

Tracing the code, it looks like passing in a `format` or a `timezone` param to the field: `live_start_at(format: "MM/dd/yy")` would also triggered the bug, since having a format or timezone (either on the request or as a param to the resolver) would take the code thru `moment`-parsing, whereas without any of those the raw field would just be returned.

Screen Shot 2019-04-11 at 9 15 16 AM

PR #1666, opened 2019-04-16
When a list endpoint is requested without an id, skip batching

zephraph:
When the sales endpoint is being called _without_ an id, ensure we short circuit the batching process.


alloy:
Coooool 👍

PR #1668, opened 2019-04-17
Add stitched exchange schema back to _schema.graphql

sepans:
I didn't have `ENABLE_COMMERCE_STITCHING` flag locally set so a previous PR removed the stitched exchange schema from `_schema.graphql`. This adds that back.


sepans:
Since only volt consumes stitched version of exchange and it uses the `.json` format it seems the stitched exchange schema has never been added to `_schema.graphql`

PR #1669, opened 2019-04-18
Use edition sets for histogram generation when appropriate

ds300:
The only thing left here (hopefully) is to be tested against staging.


src/schema/edition_set.ts
```diff
@@ -90,6 +93,18 @@ const EditionSetType = new GraphQLObjectType<any, ResolverContext>({
         return "No longer available"
       },
     },
+    widthCm: {
+      description:
+        "If you need to render artwork dimensions as a string, prefer the `Artwork#dimensions` field",
```
ashkan18: thanks for description 🙌

src/schema/fields/priceCents.ts
```diff
@@ -0,0 +1,35 @@
...
+  GraphQLFieldConfig,
+  GraphQLObjectType,
+  GraphQLInt,
+  GraphQLBoolean,
+} from "graphql"
+import { ResolverContext } from "types/graphql"
+
+export const priceCentsField: GraphQLFieldConfig<any, any> = {
+  type: new GraphQLObjectType<any, ResolverContext>({
+    name: "PriceCents",
```
ashkan18: is this more like `PriceRange`? could we use a GraphQL Union here? something like:
```graphql
priceCents {
  ... on ExactPrice {
    value
  }
  ... on Range {
    min
    max
  }
}
```

ds300: I tried this out, have a look and let me know what you think.

src/schema/fields/listPrice.ts
```diff
@@ -0,0 +1,52 @@
...
+    priceCents: {
+      type: new GraphQLNonNull(GraphQLInt),
+    },
+  },
+})
+
+export const listPrice: GraphQLFieldConfig<any, any> = {
+  type: new GraphQLUnionType({
+    name: "ListPrice",
+    types: [PriceRange, ExactPrice],
```
ashkan18: do we need a `resolve_type` definition here? something like 
https://github.com/artsy/metaphysics/blob/7d38d29c8e245b0e78b2f5f16790ab3967981d5a/src/schema/ecommerce/types/order_party_union.ts#L8
i feel like with that you may not need to manually set the `__typename` in 
https://github.com/artsy/metaphysics/pull/1669/files#diff-c8c5574211ce46fb63e10901c34dfee3R43
and then each type could have a `resolve` getting `price_cents`. what do you think?

ds300: `listPrice` still needs a resolver because it's deciding which of the union types to resolve for. So then we'd be moving the resolution logic to four different places (one for each field, listPrice, minPriceCents, maxPriceCents, and priceCents) instead of just one (listPrice), when the concrete types aren't being used independently. Is there any benefit in doing that?

PR #1671, opened 2019-04-18
If an artworks category is OTHER, dont show a histogram for that artwork

sweir27:
Fixes: https://artsyproduct.atlassian.net/browse/PURCHASE-1010

Artworks with this category are too diverse to be grouped together.


saolsen:
👍

PR #1674, opened 2019-04-19
Add appliedFiltersDisplay

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-1032
Follow up on artsy/vortex#89

# Change
add `appliedFilterDisplay` which will use newly returned `appliedFilters` field from Vortex to generate our filter display string.

# Selfie

image

### DIFF
--- a/src/lib/stitching/vortex/__tests__/pricingContext.test.ts
+++ b/src/lib/stitching/vortex/__tests__/pricingContext.test.ts
@@ -14,6 +14,7 @@ describe("PricingContext type", () => {
     artists: [
       {
         _id: "artist-1",
+        name: "Good Artist",
       },
     ],
     category: "Painting",
@@ -38,6 +39,7 @@ describe("PricingContext type", () => {
   const artistLoader = jest.fn(() =>
     Promise.resolve({
       _id: "artist-id",
+      name: "Good Artist",
     })
   )
   const salesLoader = jest.fn(() =>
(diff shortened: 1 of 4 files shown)


src/lib/stitching/vortex/stitching.ts
```diff
@@ -128,14 +150,16 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
...
-            return await info.mergeInfo.delegateToSchema({
+            const vortexContext = await info.mergeInfo.delegateToSchema({
               schema: vortexSchema,
               operation: "query",
               fieldName: "analyticsPricingContext",
               args,
               context,
               info,
             })
+            vortexContext.artistNames = artist_names
```
orta: IMO: worth a comment about this (passing it down from Artwork so it can be used in `appliedFilterDisplay` as a way to work around the resolver only having access to the data in `AnalyticsPricingContext` and not `Artwork`.

src/lib/stitching/vortex/__tests__/pricingContext.test.ts
```diff
@@ -226,3 +239,32 @@ Object {
     expect(result.artwork.pricingContext).toBeNull()
   })
 })
+
+describe("filtersDescription", () => {
```
sweir27: What if both are null?

PR #1676, opened 2019-04-22
Fix case when there is no vortex context

ashkan18:
Fixes https://sentry.io/organizations/artsynet/issues/992134226/?project=299973&query=is%3Aunresolved

cc: @sweir27


sweir27:
merge on green

PR #1678, opened 2019-04-22
[V2] ID transformations and new endpoint

alloy:
* Renames ID fields https://artsyproduct.atlassian.net/browse/PLATFORM-1447
* Expose the v2 schema from a new `/v2` endpoint (`/` needs to remain v1 for backwards compatibility). https://artsyproduct.atlassian.net/browse/PLATFORM-1384

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -10565,6 +10565,9 @@ type UpdateMyProfilePayload {
 }
 
 input UpdateSubmissionMutationInput {
+  # The GUID for the submission
+  id: String!
+
   # The gravity ID for an Artist
   artist_id: String!
 
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -298,7 +298,7 @@ type ApproveOrderPayload {
   clientMutationId: String
 }
 
-type Article implements Node {
+type Article {
   # A globally unique ID.
   __id: ID!
 
@@ -358,7 +358,7 @@ enum ArticleSorts {
   PUBLISHED_AT_DESC
 }
 
-type Artist implements Node & Searchable {
+type Artist {
   # A globally unique ID.
   __id: ID!
 
@@ -700,7 +700,7 @@ type ArtistInsight {
   entities: [String]
 }
 
-type ArtistItem implements Node & Searchable {
+type ArtistItem {
   # A globally unique ID.
   __id: ID!
 
@@ -1006,7 +1006,7 @@ type ArtistStatuses {
   shows: Boolean
 }
 
-type Artwork implements Node & Searchable & Sellable {
+type Artwork {
   # A globally unique ID.
   __id: ID!
 
@@ -1216,7 +1216,7 @@ union ArtworkContext =
   | ArtworkContextPartnerShow
   | ArtworkContextSale
 
-type ArtworkContextAuction implements Node {
+type ArtworkContextAuction {
   # A globally unique ID.
   __id: ID!
 
@@ -1483,7 +1483,7 @@ type ArtworkContextFair {
   sponsoredContent: FairSponsoredContent
 }
 
-type ArtworkContextPartnerShow implements Node {
+type ArtworkContextPartnerShow {
   # A globally unique ID.
   __id: ID!
 
@@ -1588,7 +1588,7 @@ type ArtworkContextPartnerShow implements Node {
   type: String
 }
 
-type ArtworkContextSale implements Node {
+type ArtworkContextSale {
   # A globally unique ID.
   __id: ID!
 
@@ -1729,7 +1729,7 @@ type ArtworkEdge {
 
 union ArtworkFilterFacet = ArtworkFilterTag | ArtworkFilterGene
 
-type ArtworkFilterGene implements Node {
+type ArtworkFilterGene {
   # A globally unique ID.
   __id: ID!
 
@@ -1790,7 +1790,7 @@ type ArtworkFilterGene implements Node {
   ): FilterArtworks
 }
 
-type ArtworkFilterTag implements Node {
+type ArtworkFilterTag {
   # A globally unique ID.
   __id: ID!
 
@@ -1884,7 +1884,7 @@ type ArtworkInquiryEdge {
   cursor: String!
 }
 
-type ArtworkItem implements Node & Searchable & Sellable {
+type ArtworkItem {
   # A globally unique ID.
   __id: ID!
 
@@ -2205,7 +2205,7 @@ type AuctionLotImages {
   thumbnail: Image
 }
 
-type AuctionResult implements Node {
+type AuctionResult {
   # A globally unique ID.
   __id: ID!
 
@@ -2493,7 +2493,7 @@ type BuyersPremium {
   percent: Float
 }
 
-type BuyOrder implements Order {
+type BuyOrder {
   # ID of the order
   id: ID
 
@@ -2731,7 +2731,7 @@ type City {
     sort: PartnerShowSorts
 
     # Filter shows by chronological event status
-    status: EventStatus = CURRENT
+    status: EventStatus = null
 
     # Filter shows by partner type
     partnerType: PartnerShowPartnerType
@@ -2774,7 +2774,7 @@ type CitySponsoredContent {
   ): ShowConnection
 }
 
-type Collection implements Node {
+type Collection {
   # A globally unique ID.
   __id: ID!
 
@@ -2948,7 +2948,7 @@ type CommerceBuyerRejectOfferPayload {
   orderOrError: CommerceOrderOrFailureUnion!
 }
 
-type CommerceBuyOrder implements CommerceOrder {
+type CommerceBuyOrder {
   buyer: CommerceOrderPartyUnion!
   buyerPhoneNumber: String
   buyerTotalCents: Int
@@ -3396,7 +3396,7 @@ type CommerceOfferEdge {
   node: CommerceOffer
 }
 
-type CommerceOfferOrder implements CommerceOrder {
+type CommerceOfferOrder {
   awaitingResponseFrom: CommerceOrderParticipantEnum
   buyer: CommerceOrderPartyUnion!
   buyerPhoneNumber: String
@@ -4015,7 +4015,7 @@ type ConfirmPickupPayload {
 }
 
 # A work to be consigned to the user
-type ConsignmentSubmission implements Node {
+type ConsignmentSubmission {
   # A globally unique ID.
   __id: ID!
 
@@ -4110,7 +4110,7 @@ type Convection {
 }
 
 # A conversation.
-type Conversation implements Node {
+type Conversation {
   # A globally unique ID.
   __id: ID!
 
@@ -4643,7 +4643,7 @@ input EditableLocation {
   state_code: String
 }
 
-type EditionSet implements Sellable {
+type EditionSet {
   # A globally unique ID.
   __id: ID!
 
@@ -4964,7 +4964,7 @@ type Feedback {
   message: String!
 }
 
-type FilterArtworks implements Node {
+type FilterArtworks {
   # The ID of the object.
   __id: ID!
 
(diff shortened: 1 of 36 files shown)


zephraph:
This makes me happy.

damassi:
Me too!

_schema.graphql
```diff
@@ -2699,7 +2726,7 @@ type City {
     sort: PartnerShowSorts
 
     # Filter shows by chronological event status
-    status: EventStatus = CURRENT
+    status: EventStatus = null
```
zephraph: I don't suspect this was intentional? 🤔

_schema.graphql
```diff
@@ -10077,7 +10125,7 @@ type Show implements Node {
     sort: PartnerShowSorts
 
     # By default show only current shows
-    status: EventStatus = CURRENT
+    status: EventStatus = null
```
zephraph: Or this?

patches/graphql-tools+4.0.4.patch
```diff
@@ -0,0 +1,38 @@
...
+--- a/node_modules/graphql-tools/dist/stitching/schemaRecreation.js
++++ b/node_modules/graphql-tools/dist/stitching/schemaRecreation.js
+@@ -62,7 +62,9 @@ function recreateType(type, resolveType, keepResolvers) {
+         var newValues_1 = {};
+         values.forEach(function (value) {
+             newValues_1[value.name] = {
+-                value: value.value,
++                // Simply pass-through the name of the enum entry such that the enum type in the original schema will
++                // receive that and will be able to do its work as if it were never proxied.
++                value: value.name,
```
alloy: @freiksenet Reverting this to [how you fixed it](https://github.com/apollographql/graphql-tools/commit/c500fdf82773b117b59a27aeffd3a90f22cd3a6d#diff-d5dcbf958f513509b72e378b5a9c0178R91) at the time did indeed fix the issue entirely 👍 

Once this current work is out we'll figure out a failing test upstream and perhaps the fix.

zephraph:
So as an update, I've been testing this and (at least) `/artist` and `/collections` are failing due to `__id` being null in `SaleArtworkHighestBid`. I haven't chased down the exact cause yet though.

zephraph:
Here's the error message that's happening when trying to navigate to an Artist page...

image

zephraph:
Okay, so the problem is what it says... `SaleArtworkHighestBid` is getting a `null` for `__id`... because it wasn't defined in the original schema. 

Here's the diff of the schema between this branch and master

```diff
diff --git a/_schema.graphql b/_schema.graphql
index deb3461b..440d45b3 100644
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1861,8 +1861,15 @@ type ArtworkInfoRow {
 
 # An inquiry on an Artwork
 type ArtworkInquiry {
+  # A globally unique ID.
+  __id: ID!
+
+  # A type-specific ID.
+  id: String!
+
+  # A type-specific Gravity Mongo Document ID.
+  _id: String!
   artwork: Artwork!
-  id: ID
   impulse_conversation_id: String
 }
 
@@ -2123,8 +2130,14 @@ enum ArtworkSorts {
 }
 
 type ArtworkVersion {
-  # ID of the order line item
-  id: ID!
+  # A globally unique ID.
+  __id: ID!
+
+  # A type-specific ID.
+  id: String!
+
+  # A type-specific Gravity Mongo Document ID.
+  _id: String!
 
   # Artwork title
   title: String
@@ -2144,8 +2157,8 @@ type ArtworkVersion {
 
 # An asset which is assigned to a consignment submission
 type Asset {
-  # Convection asset ID.
-  id: String!
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # The convection submission ID
   submission_id: String
@@ -2159,8 +2172,8 @@ type Asset {
 
 # Fields of an attachment (currently from Radiation)
 type Attachment {
-  # Attachment id.
-  id: String!
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # Content type of file.
   content_type: String!
@@ -2172,8 +2185,15 @@ type Attachment {
 
 # Collection of fields that describe attribution class
 type AttributionClass {
+  # A globally unique ID.
+  __id: ID!
+
+  # A type-specific ID.
   id: String!
 
+  # A type-specific Gravity Mongo Document ID.
+  _id: String!
+
   # Shortest form of attribution class display
   name: String
 
@@ -2494,8 +2514,8 @@ type BuyersPremium {
 }
 
 type BuyOrder implements Order {
-  # ID of the order
-  id: ID
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # Order Mode
   mode: OrderModeEnum
@@ -2712,7 +2732,14 @@ enum CancelReasonType {
 
 # Fields of partner category (currently from Gravity).
 type Category {
+  # A globally unique ID.
+  __id: ID!
+
+  # A type-specific ID.
   id: String!
+
+  # A type-specific Gravity Mongo Document ID.
+  _id: String!
   category_type: String
   internal: Boolean
   name: String
@@ -4114,8 +4141,8 @@ type Conversation implements Node {
   # A globally unique ID.
   __id: ID!
 
-  # Impulse id.
-  id: String
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # Gravity inquiry id.
   inquiry_id: String
@@ -4212,8 +4239,8 @@ type ConversationEdge {
 
 # The participant who started the conversation, currently always a User
 type ConversationInitiator {
-  # Impulse id.
-  id: String!
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # The type of participant, e.g. Partner or User
   type: String!
@@ -4232,8 +4259,8 @@ union ConversationItemType = Artwork | Show
 
 # The participant responding to the conversation, currently always a Partner
 type ConversationResponder {
-  # Impulse id.
-  id: String!
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # The type of participant, e.g. Partner or User
   type: String!
@@ -4545,8 +4572,8 @@ type DeleteCreditCardPayload {
 
 # Fields of a delivery (currently from Radiation)
 type Delivery {
-  # Delivery id.
-  id: String!
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # Masked email w/ display name.
   full_transformed_email: String!
@@ -4908,12 +4935,18 @@ type FairEdge {
 }
 
 type FairExhibitor {
+  # A globally unique ID.
+  __id: ID!
+
+  # A type-specific ID.
+  id: String!
+
+  # A type-specific Gravity Mongo Document ID.
+  _id: String!
+
   # Exhibitor name
   name: String
 
-  # Exhibitors id
-  id: String
-
   # Exhibitors _id
   partner_id: String
 
@@ -4944,7 +4977,7 @@ type FairSponsoredContent {
 }
 
 type FeaturedLinkItem {
-  # Attempt to get the ID of the entity of the FeaturedLink
+  # A type-specific ID.
   id: String
   href: String
   image: Image
@@ -5503,7 +5536,14 @@ type GeneEdge {
 }
 
 type GeneFamily {
+  # A globally unique ID.
+  __id: ID!
+
+  # A type-specific ID.
   id: String!
+
+  # A type-specific Gravity Mongo Document ID.
+  _id: String!
   name: String!
   genes: [Gene]
 }
@@ -6407,9 +6447,10 @@ type HomePageModuleContextTrending {
 }
 
 type HomePageModulesParams {
+  # A type-specific ID likely used as a database ID.
+  id: ID!
   followed_artist_id: ID
   gene_id: String
-  id: ID
   medium: String
   price_range: String
   related_artist_id: ID
@@ -6873,8 +6914,8 @@ type Message implements Node {
   # A globally unique ID.
   __id: ID!
 
-  # Impulse message id.
-  id: String!
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # Impulse message id.
   impulse_id: String!
@@ -7250,8 +7291,8 @@ enum NotificationsFeedItemStatus {
 }
 
 type Offer {
-  # ID of the offer
-  id: ID
+  # A type-specific ID likely used as a database ID.
+  id: ID!
   createdAt(
     # This arg is deprecated, use timezone instead
     convert_to_utc: Boolean
@@ -7366,8 +7407,8 @@ type OfferEdge {
 }
 
 type OfferOrder implements Order {
-  # ID of the order
-  id: ID
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # Order Mode
   mode: OrderModeEnum
@@ -7595,8 +7636,8 @@ type OpeningHoursText {
 union OpeningHoursUnion = OpeningHoursArray | OpeningHoursText
 
 interface Order {
-  # ID of the order
-  id: ID
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # Order Mode
   mode: OrderModeEnum
@@ -7837,8 +7878,8 @@ type OrderedSet {
 }
 
 type OrderFulfillment {
-  # ID of the fulfillment
-  id: ID
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # Fulfillment Courier
   courier: String
@@ -7879,8 +7920,8 @@ enum OrderFulfillmentType {
 }
 
 type OrderLineItem {
-  # ID of the order line item
-  id: ID
+  # A type-specific ID likely used as a database ID.
+  id: ID!
 
   # Artwork that is being ordered
   artwork: Artwork
@@ -9479,7 +9520,14 @@ type SaleArtworkEdge {
 }
 
 type SaleArtworkHighestBid {
-  id: ID
+  # A globally unique ID.
+  __id: ID!
+
+  # A type-specific ID.
+  id: String!
+
+  # A type-specific Gravity Mongo Document ID.
+  _id: String!
   created_at(
     # This arg is deprecated, use timezone instead
     convert_to_utc: Boolean
@@ -10517,7 +10565,7 @@ type UpdateMyProfilePayload {
 }
 
 input UpdateSubmissionMutationInput {
-  # A type-specific ID.
+  # The GUID for the submission
   id: String!
 
   # The gravity ID for an Artist
@@ -10591,7 +10639,7 @@ type User {
   # A globally unique ID.
   __id: ID!
 
-  # User id.
+  # A type-specific ID.
   id: String!
 
   # A type-specific Gravity Mongo Document ID.
```

I still haven't narrowed down the exact issue, but I'm closer.

alloy:
I wonder if the issue is triggered by aliasing the field selection? 🤔

alloy:
@zephraph As discussed during standup, that error must be because of this addition https://github.com/artsy/metaphysics/pull/1678/files#diff-477176756e1e4cb78dcbc378d27ad0a6R167

alloy:
@zephraph If you look at the previous v1 schema, [the `id` field](https://github.com/artsy/metaphysics/blob/6c494aa57ff4560ecb537cc3016c86b5f15ae6a4/src/schema/sale_artwork.ts#L168-L170) was optional. In fact, it would _always_ return `null`, because [the resolve function](https://github.com/artsy/metaphysics/blob/6c494aa57ff4560ecb537cc3016c86b5f15ae6a4/src/schema/sale_artwork.ts#L191) does not include any `id` value from the `source` argument. Our change replaced the `id` field with a _required_ value, which is why this is now failing.

I think we should just remove the `id` field altogether, as long as the iOS app did not rely on it, and then do a synchronised deploy of MP and Force.

----

If you’re interested in debugging this yourself, here’s a query to repro:

```graphql
{
  artwork(id: "banksy-get-out-while-you-can-28") {
    sale_artwork {
      highest_bid {
        id
      }
    }
  }
}
```

zephraph:
@alloy I just added a temporary object to represent a nullable ID. That way it passes without having to update the queries in other clients and we have a way to trace it back down for later. 

I went down a whole rabbit whole of trying to get the v1 schema to be the same as it was before the PR, but there really aren't any good answers for how to do that (that I was able to figure out). I wasted too much time on that. 

I've done some testing, but we should just get this to staging and try to get people to poke at it for a while.

src/schema/object_identification.ts
```diff
@@ -161,6 +161,13 @@ export const GlobalIDField = {
   },
 }
 
+export const SingleIDField = {
+  id: {
+    description: "A type-specific ID.",
+    type: GraphQLID,
+  },
+}
+
```
alloy: We should exclude this field entirely from the v2 schema, as it doesn’t do anything.

alloy: Done

zephraph:
After testing, I feel like this is good 👍

alloy:
:shipit:

PR #1680, opened 2019-04-23
WIP: Metaphysics exposes size_score in its artwork schema

lilyfromseattle:
Added sizeScore to both Artwork and EditionSet and added tests for both.
This is necessary for the Price in context feature.

WIP because it depends on gravity


ashkan18:
Nice! 🙌

PR #1681, opened 2019-04-24
Enable commerce stitching by default

ds300:
We need the stitched commerce bits (prefixed with `commerce`) in reaction now, as we're trying to migrate away from our 'fake stitching' things (prefixed with `ecommerce`)


src/config.ts
```diff
@@ -150,7 +150,8 @@ export default {
   DISABLE_SCHEMA_STITCHING,
   ENABLE_RESOLVER_BATCHING: ENABLE_RESOLVER_BATCHING === "true",
   ENABLE_CONSIGNMENTS_STITCHING,
-  ENABLE_COMMERCE_STITCHING,
+  ENABLE_COMMERCE_STITCHING:
+    ENABLE_COMMERCE_STITCHING === undefined ? true : ENABLE_COMMERCE_STITCHING,
```
starsirius: Note that an env var is a string, so setting
```
ENABLE_COMMERCE_STITCHING=false
```
the [condition](https://github.com/artsy/metaphysics/blob/a11bc65ea8fbb7be285b0b3ebd0b38a558a7991e/src/lib/stitching/mergeSchemas.ts#L47) would still be true,
```
if ("false") {
  // is actually truthy.
}
```
Maybe we can do

```javascript
ENABLE_COMMERCE_STITCHING === undefined || ENABLE_COMMERCE_STITCHING === "true"
```

ds300: Good point. This PR preserved the previous semantics in that regard, but yeah those semantics are wonky :/ I'll make another PR

PR #1682, opened 2019-04-25
Smarter env var truthiness

ds300:
Context: https://github.com/artsy/metaphysics/pull/1681#discussion_r278333721


starsirius:
Thanks for following up! 💯

PR #1684, opened 2019-04-29
Use sizeScore when fetching histogram

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-1021

- [x] Tested manually against staging


saolsen:
👍

PR #1695, opened 2019-05-02
Switch to Apollo Server

zephraph:
https://artsyproduct.atlassian.net/browse/PLATFORM-1377

This is a relatively substantial change. I'm migrating our GraphQL server implementation over to use apollo server. Doing this will allow us to hook into apollo platform to get more insights into performance and usage of our graphql services. It'll also provide us with mechanisms to do schema syncing. 

To enable apollo server locally add `ENABLE_APOLLO=true` to your `.env` file.


src/lib/graphqlErrorHandler.ts
```diff
@@ -81,7 +81,11 @@ const reportErrorToSentry = (
   })
 }
 
-type WriteablePartial<T> = { -readonly [P in keyof T]+?: T[P] }
+export type WriteablePartial<T> = { -readonly [P in keyof T]+?: T[P] }
+
+export type GraphQLErrorHandler = (
+  topLevelError: GraphQLError
+) => WriteablePartial<GraphQLFormattedError>
```
alloy: The writeable part of this should probably remain private to the module scope, as it’s only needed for the newly code that assigns to the object.

zephraph: Yep, that's a left over from when I was exporiting it (before I exported the `GraphQLErrorHandler` explicitly). Good call.

zephraph:
This is mostly good to go. I should still probably write tests for the extensions, but everything else is reviewable. 

Note that the shape of the error response object has changed a little. See this convo in slack: https://artsy.slack.com/archives/C1HH3KNJG/p1556912399065100

PR #1696, opened 2019-05-03
'amount' helper should handle zero correctly [PURCHASE-995]

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-995

Need to check that we're not breaking usages of this before merging (I'm checking in reaction, emission, and volt):

 ### BuyOrder/OfferOrder
  - [x] itemsTotal
  - [x] sellerTotal
  - [x] commissionFee
  - [x] totalListPrice
  - [x] buyerTotal
  - [x] taxTotal
  - [x] shippingTotal
  - [x] transactionFee
 ### LineItem
  - [x] shippingTotal
  - [x] listPrice
  - [x] commissionFee
 ### Offer
  - [x] amount
  - [x] taxTotal
  - [x] shippingTotal
  - [x] buyerTotal
 ### AnalyticsHistogramBin
  - [x] minPrice
  - [x] maxPrice
 ### HighestBid
  - [x] amount
 ### Invoice
  - [x] total
 ### BuyersPremium
  - [x] amount
 ### SaleArtworkHighestBid
  - [x] amount

### DIFF
--- a/src/schema/artwork/index.ts
+++ b/src/schema/artwork/index.ts
@@ -560,11 +560,12 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           )
             return "Free shipping worldwide"
           var domesticShipping = amount(
-            ({ domestic_shipping_fee_cents }) => domestic_shipping_fee_cents
+            ({ domestic_shipping_fee_cents }) =>
+              domestic_shipping_fee_cents || null
           ).resolve(artwork, { precision: 0 })
           var internationalShipping = amount(
             ({ international_shipping_fee_cents }) =>
-              international_shipping_fee_cents
+              international_shipping_fee_cents || null
           ).resolve(artwork, { precision: 0 })
 
           if (
--- a/src/schema/fields/__tests__/money.test.ts
+++ b/src/schema/fields/__tests__/money.test.ts
@@ -1,0 +1,38 @@
+import { amount } from "../money"
+
+describe(amount, () => {
+  let amountCents: number
+  let args: any
+  beforeEach(() => {
+    args = { symbol: "$" }
+    amountCents = 1234
+  })
+
+  const getResult = () => amount(() => amountCents).resolve({}, args)
+
+  it("formats dollars", () => {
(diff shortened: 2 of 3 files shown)


sweir27:
_this is so exciting_

src/schema/artwork/index.ts
```diff
@@ -560,11 +560,12 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           )
             return "Free shipping worldwide"
           var domesticShipping = amount(
-            ({ domestic_shipping_fee_cents }) => domestic_shipping_fee_cents
+            ({ domestic_shipping_fee_cents }) =>
+              domestic_shipping_fee_cents || null
```
sweir27: Why do we need the added `|| null` here? Don't we already return `null` from the money helper if this is undefined?

sweir27:
Just verifying-- did you take a look at deployed clients (i.e. emission) to ensure they're not depending on `null` values here?

(a la https://github.com/artsy/metaphysics/pull/1325#issuecomment-428853308)

src/schema/artwork/index.ts
ds300: In case domestic_shipping_free_cents is 0, to preserve the old behaviour.

ds300:
😬 no I didn't check the deployed version, only master.

ds300:
Alright @sweir27 checked some older versions of emission and the usage is all safe. So that's emission, reaction, metaphysics, and volt checked. I don't think positron uses any of these fields. Can you think of anywhere else to check? If not I think we're good to go :+1: :tada:

sweir27:
Thank you! I think Eigen, Force, and Prediction might be the last ones? (I don't think Eigen consumes metaphysics _much_ but it does some).

Also thank you! This is tedious but so so important 🙌 🙌

ds300:
Ah yes forgot that I also checked prediction and it's 👍 but not force or eigen.

Just checked them both now and we're good :+1: :+1:

sweir27:
Woooo! This looks great.

I'll assign it to you since once this is out in staging we should be sure to spot-check a few places!

PR #1703, opened 2019-05-07
Removes lab feature requirement for pricing context

sweir27:
Addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-978

cc @ds300 I wasn't sure what you were imagining for caching here (or if we should wait until after release)-- feel free to push up to this branch!

Leaving this as "do not merge" until we get the green light!

This will also require setting `VORTEX_TOKEN` in staging and production.


.env.example
```diff
@@ -28,6 +28,7 @@ EMBEDLY_KEY=**
 GALAXY_TOKEN=**
 GOOGLE_CSE_CX=**
 GOOGLE_CSE_KEY=**
+VORTEX_TOKEN=**
```
alloy: I just now realise that this `**` placeholder will make it pass the `mustHave` check incorrectly. Perhaps we should just not have placeholder values?

ds300:
@sweir27 I'm not sure about caching now that we're using real schema stitching. I can't find prior art in MP, and a quick googling suggests I don't know enough about apollo and our usage of it to make a good estimate of complexity. It's definitely a 3 for me :sweat_smile:

@alloy do you know anything about stiching + caching?

If we think vortex can take the load, maybe we can get away with no caching for now? I have no idea how many requests/sec we'd be sending to vortex and how many it can handle.

sweir27:
I think we're probably okay to wait on this for now, although I'm also curious how we want to do this in general.

We should triple-check that there isn't anything we're missing, but vortex prod should be equipped to handle this load.

alloy:
Thus far @zephraph and I have experimented with batching of stitched parts of a schema, not with caching specifically yet. I’d love to learn more about the details of what you’d want to cache, though, so perhaps write up a backlog ticket for it and link me to it?

PR #1708, opened 2019-05-14
update to use most recent version of vortex schema

mdole:
This PR brings MP up to date with current Vortex schema, which includes info on Pageviews that will be consumed by the Activity section of the new Partner Analytics dashboard ([spec](https://app.zeplin.io/project/5aabc6f28d835c7b2585fdf4/screen/5c9a75f087814d603e229d7a)).

Note that the change to `timeSeries` is theoretically breaking (but only for the cms.artsy.net/analytics route, which is only accessible to users with the CMS lab flag) - that said, as far as @ansor4 and I can tell it should already be breaking things and isn't. Might be because the format of the data hasn't changed at all, only the type expected.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -48,15 +48,8 @@ type AggregationCount {
 type AnalyticsArtworksPublishedStats {
   percentageChanged: Int!
   period: AnalyticsQueryPeriodEnum!
-  timeSeries: [AnalyticsArtworksPublishedTimeSeriesStats!]
+  timeSeries: [AnalyticsPartnerTimeSeriesStats!]!
   totalCount: Int!
-}
-
-# Publish artwork Series Stats
-type AnalyticsArtworksPublishedTimeSeriesStats implements AnalyticsPartnerTimeSeriesStats {
-  count: Int
-  endTime: AnalyticsDateTime
-  startTime: AnalyticsDateTime
 }
 
 # An ISO 8601 datetime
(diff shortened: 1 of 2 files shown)


mdole:
@ansor4 now that the previous deploy is complete, I think this should be ready (just rebased on top of master as well)

_schema.graphql
```diff
@@ -48,17 +48,10 @@ type AggregationCount {
 type AnalyticsArtworksPublishedStats {
   percentageChanged: Int!
   period: AnalyticsQueryPeriodEnum!
-  timeSeries: [AnalyticsArtworksPublishedTimeSeriesStats!]
+  timeSeries: [AnalyticsPartnerTimeSeriesStats!]!
```
starsirius: Once metaphysics is deployed to production with this change, the following [Force](https://github.com/artsy/force/blob/13bb4efd42e0c872beefed7c4406264fda682521/.circleci/config.yml#L16-L23) and [Volt](https://github.com/artsy/volt/blob/41ab4719e339930911e8a23d032e7525d1d4186a/.circleci/config.yml#L35-L42) production deploy might fail due to breaking schema changes. We might want to update the MP schema in reaction, volt, (and emission?).

Screen Shot 2019-05-16 at 2 14 33 PM

Screen Shot 2019-05-16 at 2 14 02 PM

mdole:
yeah definitely - @ansor4 maybe we take a stab at that this afternoon?

PR #1709, opened 2019-05-14
Revert apollo server changes

zephraph:
Reverting #1695 failed due to changes being made on top of the files since that point, so I did a reset of those files. 

After the [last metaphysics deploy](https://github.com/artsy/metaphysics/compare/dc7c2875...6ddaa9fb) we started seeing high latency spikes and timeouts in metaphysics which was causing some cascading failures. This PR was the largest change and the most likely culprit. We're reverting to unblock deploys and just to verify this is what was causing the issue.


damassi:
@zephraph - isn't this work behind an ENV feature flag?

zephraph:
It is, yes. Still, of the things that changed in the release, this is the largest. If it starts failing again we'll know it's something else.

PR #1712, opened 2019-05-16
Use most recent version of vortex schema

mdole:
Vortex schema has been updated with new Sales data. After this PR is merged, I'll follow up by updating Force, Reaction, and Volt's schemas. Possibly Emission as well.


mdole:
closed in favor of https://github.com/artsy/metaphysics/pull/1711

Issue #134, opened 2016-03-04
Preparations for use in Relay


broskoski:
I haven't researched at all what is needed here, but whatever work is needed to make metaphysics work nicely with Relay will be needed for both web and mobile.

----

[Updated by @alloy]

- [x] Add [Global Object Identification](https://facebook.github.io/relay/graphql/objectidentification.htm).
  - [x] Need to change the `id` fields to be a composite key, so that the type and specific instance can be inflected from it. E.g. `Artist#banksy`. The common implementation deployed in the GraphQL community also Base64 encodes that key.
  - [x] Ensure existing clients use the appropriate ID field. Which, for those not using Relay atm, means they probably need to use `_id`, which maps to the Gravity ID. However from short discussion with @joeyAghion, it seems like searching by `_id` could be a problem, because only _some_ models would support searching by slug on that field. Alternatively we’ll have to add e.g. `gravityID`, which would be the equivalent of `id` right now.

- [ ] Expose associations as [connections](https://facebook.github.io/relay/graphql/connections.htm), which allow for slicing and paginating.

- [ ] See if we need to do anything else to support [deferring query fragments](https://github.com/facebook/relay/issues/288).

dzucconi:
https://github.com/graphql/graphql-relay-js

alloy:
Aye aye. So basically, since our current sprint where the new artist view is scheduled is slated for 21st of March, this means I’d need to be able to use Relay with metaphysics before then, e.g. 16th.

orta:
@alloy where did this end up ( I know we use relay in Eigen now )

alloy:
We do use Relay, but there are certain features we can’t use yet. I will update the original comment with a TODO list.

alloy:
@mzikherman Can you shed some light on that issue with `_id` on Gravity supporting search by slug, but only on some models?

alloy:
Moved connections part to a new ticket https://github.com/artsy/metaphysics/issues/367

PR #1718, opened 2019-05-21
GALL-1753: Add order response time to Vortex schema

mdole:
This PR updates MP's copy of Vortex schema with a new field, `OrderResponseTime`. That field is implemented in [Vortex #134](https://github.com/artsy/vortex/pull/134).

Although this is a non-breaking change and is not currently being consumed by anything, I marked as `Do not merge` in case we decide to do something like change the field name. Once the associated Vortex PR is merged, I'll update the title and we can go for it.


mdole:
@sepans I think this is now ready for review + merge!

PR #1720, opened 2019-05-23
Update global ids

zephraph:
A small update to switch the types of some of our global ids from `String` to `ID`


alloy:
Did you verify that this works with a client? If so, go ahead and merge

PR #1723, opened 2019-05-24
Make image id nullable again

zephraph:
Sentry is reporting the following error:

```
Cannot return null for non-nullable field Image.__id.
       __id
```

Previously, `__id` wasn't present on the `Image` and its `id` field is nullable. I'm restoring it back to the old nullable field and we can follow up later on how to resolve it. 

It _does_ seem like there are places in reaction that's actually querying `__id` on the image.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -6460,11 +6460,8 @@ type HomePageModulesParams {
 }
 
 type Image {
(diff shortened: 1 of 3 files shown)


joeyAghion:
For posterity: this was associated with widespread 500 errors in staging.

PR #1724, opened 2019-05-24
[Related Collections] sync kaws schema

xtina-starr:
Related to: https://github.com/artsy/kaws/pull/97

This updates the kaws schema in metaphysics.


xtina-starr:
Now that https://github.com/artsy/kaws/pull/97 is merged, this is ready to go.

PR #1727, opened 2019-05-28
Going for a deploy...

zephraph:
This is a huge update. Has MP v2 stuff in it.


src/index.ts
```diff
@@ -27,7 +26,7 @@ import {
 import { middleware as requestIDsAdder } from "./lib/requestIDs"
 import { nameOldEigenQueries } from "./lib/modifyOldEigenQueries"
 import { rateLimiter } from "./lib/rateLimiter"
-import { graphqlErrorHandler } from "./lib/graphqlErrorHandler"
+// import { graphqlErrorHandler } from "./lib/graphqlErrorHandler"
```
alloy: @zephraph Hmm, are we going to miss runtime exceptions being recorded in Sentry?

src/index.ts
```diff
@@ -160,11 +154,12 @@ async function startApp() {
...
     })
 
-    server.applyMiddleware({ app, path: "/" })
+    server.applyMiddleware({ app, path })
   } else {
     const graphqlHTTP = require("express-graphql")
     app.use(
-      graphqlHTTP((req, res, params) => {
+      graphqlHTTP((req, res /*, params */) => {
+        console.log("Request from", path)
```
alloy: Stray loggy?

zephraph: Somewhat, yeah. It does tell if the request is coming from v1 or v2 which is useful for the logs.

alloy:
Based on skimming, this looks good to me. I’m only afraid that we’re losing sentry reporting, which may be something we want especially with this big release?

zephraph:
Why would we be losing sentry reporting?

alloy:
Because that’s done from the error handler https://github.com/artsy/metaphysics/blob/4447c0f2d2d25f67f62e7a7a5758370d460bc019/src/lib/graphqlErrorHandler.ts#L71-L81

PR #1728, opened 2019-05-28
Shore up nullable ids

zephraph:
Upon performing an audit of the `id` fields that have changed since #1678 these types had `id`s that went from nullable to non-nullable:

- BuyOrder
- Offer
- OfferOrder
- Order
- OrderFulfillment
- OrderLineItem
- Conversation
- FairExhibitor

According to @ashkan18 the order/offer types should add have ids present and the reason they were left nullable was a misunderstanding of the `ID` type. (It was assumed that a field typed as `ID` was always non-nullable). 

The last two on the list we're not confident about so I've reinstated their nullability just to be on the safe side. 

Getting this in before #1727.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -4158,8 +4158,8 @@ type Conversation implements Node {
   # A globally unique ID.
   __id: ID!
 
-  # A type-specific ID likely used as a database ID.
-  id: ID!
+  # A type-specific ID.
+  id: ID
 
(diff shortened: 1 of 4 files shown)


alloy:
LGTM 👍

ashkan18:
🕺 we can try orders on staging to make sure things are working

PR #1731, opened 2019-05-29
 [v2] Properly rename nullable id field on non-Gravity types

alloy:
Noticed that `Conversation`, which comes from impulse, had a `gravityID` instead of `internalID`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,6 +32,9 @@ const enableQueryTracing = ENABLE_QUERY_TRACING === "true"
 
 let server, isShuttingDown
 
+// Always load the source on startup so any file level issues surface sooner.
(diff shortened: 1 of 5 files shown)


alloy:
Going for it

zephraph:
After you ship it I'll take care of the reaction update

index.js
```diff
@@ -32,6 +32,9 @@ const enableQueryTracing = ENABLE_QUERY_TRACING === "true"
 
 let server, isShuttingDown
 
+// Always load the source on startup so any file level issues surface sooner.
+require("./src")
```
damassi: 👍

PR #1732, opened 2019-05-30
 [ConsignmentSubmission] Remove ID fields other than DB ID

alloy:
`ConsignmentSubmission` has all the id fields right now, but no clients should be using anything other than `id` on v1, so I’m just going to remove the others as it keeps transforming slightly simpler.


alloy:
Going to self-merge

PR #1734, opened 2019-05-31
Update schema for Vortex order count (GALL-1503)

starsirius:
https://artsyproduct.atlassian.net/browse/GALL-1503

This also updated v2 schema and want to make sure it's ok. cc @zephraph @alloy.

PR #1735, opened 2019-05-31
Update schema for order count from Vortex (GALL-1503)

starsirius:
https://artsyproduct.atlassian.net/browse/GALL-1503

This also updated v2 schema and want to make sure it's ok. cc @zephraph @alloy.

Closed the [previous one](https://github.com/artsy/metaphysics/pull/1734) and PRed from upstream branch.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -105,6 +105,7 @@ type AnalyticsPageviewStats {
 
 # Sales stats of a partner
 type AnalyticsPartnerSalesStats {
+  orderCount: Int!
   orderResponseTime: Int!
   partnerId: String!
   period: AnalyticsQueryPeriodEnum!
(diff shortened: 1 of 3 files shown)


_schemaV2.graphql
```diff
@@ -9200,6 +9201,7 @@ type Query {
   # Find PartnerStats
   analyticsPartnerStats(partnerId: String!): AnalyticsPartnerStats
   marketingCollections(
+    category: String
```
iskounen: This change seems unrelated to the order count field. Is it supposed to be included in this PR?

alloy:
Yep, that’s expected, v2 is simply a transformed version of v1, also so that it stays up to date with in progress v1 work.

iskounen:
Thanks for the clarification. For reference, this change was introduced to v1 in #1724.

PR #1737, opened 2019-06-03
Update schema for inquiry stats from Vortex (GALL-1503)

iskounen:
### What?
This PR adds the `PartnerInquiryStats` type from the Vortex GraphQL schema into Metaphysics and exposes it as `AnalyticsPartnerInquiryStats`.

### Why?
This change allows the CMS to display information about a partner's inquiries (specifically, how many they received and how soon they were able to respond to them in a given time period) to the Sales section of the analytics dashboard.

### Show me!
query

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -103,6 +103,14 @@ type AnalyticsPageviewStats {
   uniqueVisitors: Int
 }
 
+# Inquiry stats of a partner
+type AnalyticsPartnerInquiryStats {
+  inquiryCount: Int!
+  inquiryResponseTime: Int
+  partnerId: String!
+  period: AnalyticsQueryPeriodEnum!
+}
+
 # Sales stats of a partner
 type AnalyticsPartnerSalesStats {
   orderCount: Int!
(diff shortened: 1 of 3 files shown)


_schema.graphql
```diff
@@ -103,6 +103,14 @@ type AnalyticsPageviewStats {
   uniqueVisitors: Int
 }
 
+# Inquiry stats of a partner
+type AnalyticsPartnerInquiryStats {
+  inquiryCount: Int!
+  inquiryResponseTime: Int
```
mdole: Hmm so inquiry count is non-nullable but inquiry response time is - should we make those consistent? I'm not sure why we'd prefer one over the other off the top of my head, or why they might be different

mdole:
Looks like the tests failed due to a missing API key, I think? 
Screen Shot 2019-06-03 at 1 48 35 PM

_schema.graphql
iskounen: Good catch. `AnalyticsPartnerSalesStats` has both of these fields typed as non-nullable so I think we should also be consistent across both types. I am leaning towards making `inquiryResponseTime` non-nullable since it will involve the least amount of changes.

iskounen: I spoke with @sepans and it confirmed that this field was intentionally left nullable. Doing so allows clients the option of taking an action if the response time is `nil` - in the case of CMS analytics dashbaord that means we have the option of not displaying a breakdown item for inquiry response times.

We have a ticket in our backlog to further explore this issue: https://artsyproduct.atlassian.net/browse/GALL-1783

mdole: ah that's right, thanks for investigating @iskounen. Let's leave as-is!

iskounen:
I'm closing this in favor of #1738 because PRs can't be merged from forked repos to Metaphysics until we find a way for the `danger` step of the build process to run from forked repos.

PR #1738, opened 2019-06-03
Update schema for inquiry stats from Vortex (GALL-1503)

iskounen:
### What?
This PR adds the `PartnerInquiryStats` type from the Vortex GraphQL schema into Metaphysics and exposes it as `AnalyticsPartnerInquiryStats`.

### Why?
This change allows the CMS to display information about a partner's inquiries (specifically, how many they received and how soon they were able to respond to them in a given time period) to the Sales section of the analytics dashboard.

### Show me!
query

### DIFF
--- a/src/data/vortex.graphql
+++ b/src/data/vortex.graphql
@@ -62,6 +62,25 @@ type PageviewStats {
 }
 
 """
+Audience stats of a partner
+"""
+type PartnerAudienceStats {
+  partnerId: String!
+  period: QueryPeriodEnum!
+  uniqueVisitors: Int!
+}
+
+"""
+Inquiry stats of a partner
+"""
+type PartnerInquiryStats {
+  inquiryCount: Int!
+  inquiryResponseTime: Int
+  partnerId: String!
+  period: QueryPeriodEnum!
+}
+
+"""
 Sales stats of a partner
 """
 type PartnerSalesStats {
@@ -95,6 +114,16 @@ type PartnerStats {
   Time series data on number of artworks published
   """
   artworksPublished(period: QueryPeriodEnum!): ArtworksPublishedStats
+
+  """
+  Audience stats
+  """
+  audience(period: QueryPeriodEnum!): PartnerAudienceStats
+
+  """
+  Inquiry stats
+  """
+  inquiry(period: QueryPeriodEnum!): PartnerInquiryStats
 
   """
   Different types of partner pageviews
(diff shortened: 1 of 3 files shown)


src/data/vortex.graphql
```diff
@@ -137,6 +166,7 @@ type PartnerStats {
   Number of unique visitors
   """
   uniqueVisitors(period: QueryPeriodEnum!): Int
+    @deprecated(reason: "Use audience() { uniqueVisitors } instead")
```
mdole: just to double check - does the use of `@deprecated` stop schema changes from breaking the services that use it, or will we need to update Reaction/Force/Volt schemas with this change?

starsirius: I think this is to avoid breaking changes, but we can double check locally.

iskounen: My guess is that this directive offers a warning to clients about this field, and leaves it up to users of the schema to heed or ignore it. In my experience with Java, any instantiation/call/access to a class/method/field that is annotated with `@deprecated` throws a warning in the development IDE without preventing those actions.

I've reached out to the #graphql channel because I'm not finding search results that explain what happens as a result of applying this directive to a GraphQL schema.

ansor4: I think @iskounen is right, all `@deprecated` does is warn the developer that the field is deprecated. It does not restrict usage, but serves as a warning for clients using that field. For example, there's a small tooltip that pops up when querying vortex from insomnia.
Screen Shot 2019-06-03 at 8 01 27 AM
This may be confusing because I had made another change to vortex that removed the resolver for the top level `uniqueVisitors` field, so now that field returns `null`. If I did not remove this resolver, this field would continue to work as normal, but that deprecation message would continue to pop up.

Below, I stubbed a resolver function for `uniqueVisitors` to always return `12345`.
Screen Shot 2019-06-04 at 8 15 10 AM
And the `uniqueVisitors` field would return that value along with the deprecation message.
Screen Shot 2019-06-04 at 8 14 16 AM

iskounen: Thanks for clarifying that, @ansor4.

@mzikherman made a good point about whether we should be deprecating a new field, which made me think that it may be better to rename this field in Vortex and propagate the changes through Volt at this stage of development. If we deprecate it now we'll need to eventually _actually_ deprecate it, so we may as well do that work now. What do you all think?

cc: @sepans @mdole

ansor4: My only concern would be if volt is still expecting a `uniqueVisitors` field. I _think_ volt is equipped to handle `null` as a value for `uniqueVisitors`, I'm not sure if it's equipped to not expect the field at all. @sepans I believe you're more familiar with the error handling on volt's side, would you know if this field is ok to deprecate now?

sepans: I believe so. Going to merge this and will PR a fix if volt breaks (I think that is faster) 😎

PR #1739, opened 2019-06-03
Exclude danger for now

joeyAghion:
Metaphysics PR builds have been failing ever since we stopped sharing secrets with forked PRs.

I honestly don't know how critical some of [Danger's current checks](https://github.com/artsy/metaphysics/blob/master/dangerfile.ts) are to our workflow. If they are, seems like our best option is migrating them to something like Peril. I don't think we should tolerate build failures until then, so I suggest we exclude Danger for now.


alloy:
Not sure I understand this change. I assume builds are failing for PRs _from forks_? If so, I would prefer people not to use forks instead of disabling extra checks.

PR #1744, opened 2019-06-05
does not pass NaN as a page value

mbilokonsky:
Metaphysics right now seems to be passing `NaN` as the value of the `page` parameter when it queries for collections through `filter_artworks`, leading to e.g. no artworks showing up on pages like this: https://www.artsy.net/collection/andy-warhol-jean-michel-basquiat

This is an emergency patch to get these things rendering correctly again. Additional debugging is ongoing.

### DIFF
--- a/src/schema/filter_artworks.ts
+++ b/src/schema/filter_artworks.ts
@@ -113,7 +113,9 @@ export const FilterArtworksType = new GraphQLObjectType<any, ResolverContext>({
         }
       ) => {
         const relayOptions = convertConnectionArgsToGravityArgs(args)
-        if (!!gravityOptions.page) relayOptions.page = gravityOptions.page
+        if (!!gravityOptions.page && !isNaN(gravityOptions.page)) {
+          relayOptions.page = gravityOptions.page
+        }
 
         const { page, size } = relayOptions
         const {


src/schema/filter_artworks.ts
```diff
@@ -113,7 +113,9 @@ export const FilterArtworksType = new GraphQLObjectType<any, ResolverContext>({
         }
       ) => {
         const relayOptions = convertConnectionArgsToGravityArgs(args)
-        if (!!gravityOptions.page) relayOptions.page = gravityOptions.page
+        if (!!gravityOptions.page && !isNaN(gravityOptions.page)) {
+          relayOptions.page = gravityOptions.page
+        }
```
alloy: Hmm, I don’t quite understand this 🤔 

```
$ node
> !!NaN
false
> !isNaN(NaN)
false
```

Can you elaborate? Perhaps in the form of a regression test.

(Unrelated, but that double bang doesn’t really seem necessary anyways.)

mbilokonsky: Yeah I'm trying to figure out how to write this test but struggling with the typescript/graphql ecosystem, wanted to get eyes on this first and ideally ship since there are broken collections in prod right now.

In the meantime I'm not sure you read that right? Here's some node tinkering to demonstrate that this does, I think, do what we want?

```node
> const gravityOptions1 = { page: 'hello' }
> const gravityOptions2 = { page: 7 }
> if (!!gravityOptions1.page && !isNaN(gravityOptions1.page)) { console.log('this does not work') } else { console.log('this does work') }
this does work
> if (!!gravityOptions2.page && !isNaN(gravityOptions2.page)) { console.log('this does work') } else { console.log('this does not work') }
this does work
```

mbilokonsky: Oh @alloy are you reading `!!isNaN` instead of `!isNaN`?

mbilokonsky: oooh, wait, I think I know what's up. Don't merge this PR, hang on.

alloy: Ahhh, no, I was thinking of `page` being the `NaN` sentinel, not a different value such as a string–that makes sense 👍 

```
> const gravityOptions3 = { page: NaN }
undefined
>  if (!!gravityOptions3.page) { console.log('this does not work') } else { console.log('this does work') }
this does work
```

mzikherman: Inspecting the request for the staging collection (which returns no works) seems to reveal normal looking filter requests to Gravity in the logs:
Screen Shot 2019-06-05 at 2 00 21 PM

Scrolling reveals `page=1` being passed in, so there might be something else going on?

mbilokonsky: Hmm, @l2succes flagged a log where `page` was `NaN`. @mzikherman are you looking at requests for the specific collection linked above? Because some collections are working just fine.

mzikherman: yup, the screenshot is from the collection linked above (but on staging, since that's what gets you the logs).

Making that request to gravity reveals [no data](http://api.artsy.net/api/v1/filter/artworks?acquireable=false&aggregations%5B%5D=total&artist_ids%5B%5D=4d8b92b34eb68a1b2c0003f4&height=%2A-%2A&keyword=Basquiat%2C%20References%20to%20Jean-Michel%20Basquiat&keyword_match_exact=true&offset=0&page=1&price_range=%2A-%2A&size=30&sort=-decayed_merch&width=%2A-%2A)

mbilokonsky: Ok so I think this PR is actually something that should be merged. We recognized that the `NaN` is not what was causing the problem in this case, it was actually a symptom of not getting any hits back.

Still, we should be aggressive around stuff like this - `page=NaN` shouldn't be possible, and I believe this PR fixes it. (But no I have not written any tests around it yet, I tried at first but then when I realized the real problem was in gravity I started working on that. This one feels like a no-brainer to me, though?)

alloy: If `page` is `NaN` then the existing code should already work, as per my code examples.

mbilokonsky: oh I see what you mean, you're right. gosh I'm trying to keep too many things straight, thanks, closing!

PR #1748, opened 2019-06-06
[ci] Re-enable danger, but only print failures to stdout

alloy:
Reverts #1739

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "build:lib": "babel src --copy-files --extensions '.ts,.js' --ignore src/test,src/integration,src/**/__tests__ -s inline -d build/src",
     "build:remote-schemas": "cp src/data/*.graphql build/src/data/",
     "build": "yarn build:lib && yarn build:index && yarn build:fixtures && yarn build:remote-schemas",
-    "ci": "yarn test",
+    "ci": "yarn test && yarn danger ci --text-only",
     "dev": "DEBUG=info,warn,error nf start --procfile Procfile.dev -w",
     "dump-remote-schema": "babel-node --extensions '.ts,.js' ./scripts/dump-remote-schema.ts",
     "dump-schema": "babel-node --extensions '.ts,.js' ./scripts/dump-schema.ts",


alloy:
```
Skipping Danger due to this run not executing on a PR.
```

Hrmm. Unsure why this is 🤔

alloy:
Going to create a new PR to see if enabling the “only build pull-requests” CircleCI setting fixes this.

alloy:
It turns out that danger insists on having an API key, as it needs to be able to grab PR metadata and doing so without API key would lead to rate-limiting quickly. So this doesn’t appear to be an option 😞 

```
metaphysics_1            | $ /app/node_modules/.bin/danger ci --text-only
metaphysics_1            | The DANGER_GITHUB_API_TOKEN/DANGER_BITBUCKETSERVER_HOST environmental variable is missing
metaphysics_1            | Without an api token, danger will be unable to comment on a PR
metaphysics_1            | Error:  Error: Cannot use authenticated API requests.
metaphysics_1            |     at Object.getPlatformForEnv (/app/node_modules/danger/distribution/platforms/platform.js:51:11)
metaphysics_1            |     at Object.<anonymous> (/app/node_modules/danger/distribution/commands/ci/runner.js:81:57)
metaphysics_1            |     at step (/app/node_modules/danger/distribution/commands/ci/runner.js:32:23)
metaphysics_1            |     at Object.next (/app/node_modules/danger/distribution/commands/ci/runner.js:13:53)
metaphysics_1            |     at fulfilled (/app/node_modules/danger/distribution/commands/ci/runner.js:4:58)
metaphysics_1            |     at process._tickCallback (internal/process/next_tick.js:68:7)
metaphysics_1            |     at Function.Module.runMain (internal/modules/cjs/loader.js:744:11)
metaphysics_1            |     at startup (internal/bootstrap/node.js:285:19)
metaphysics_1            |     at bootstrapNodeJSCore (internal/bootstrap/node.js:739:3)
```

joeyAghion:
Re. needing a key to avoid rate-limiting when pulling PR metadata, is that an intrinsic requirement, or just a choice in Danger? On the surface, it seems like requesting a single open source PR's metadata shouldn't encounter limits.

alloy:
A choice in Danger, as I understand it.

alloy:
I guess we could try to see how fast we run into it by applying a patch-package path to danger that disables this check?

PR #1749, opened 2019-06-06
[ci] Re-enable danger, but only print failures to stdout

alloy:


alloy:
https://github.com/artsy/metaphysics/pull/1748#issuecomment-499507156

PR #1753, opened 2019-06-07
Removes moment without TZ from date tests

kierangillen:
- The tests for `dateRange` used `moment()` without a timezone which were failing the tests when tested outside of EST.


alloy:
Short and sweet, thanks! 🙏 

Merge on green

PR #1756, opened 2019-06-07
[v2] Rename node __id arg to id

alloy:
Closes https://artsyproduct.atlassian.net/browse/PLATFORM-1441

### DIFF
--- a/patches/graphql-tools+4.0.4.patch
+++ b/patches/graphql-tools+4.0.4.patch
@@ -1,4 +1,17 @@
 patch-package
+--- a/node_modules/graphql-tools/dist/stitching/resolvers.js
++++ b/node_modules/graphql-tools/dist/stitching/resolvers.js
+@@ -51,7 +51,9 @@ function createProxyingResolver(schema, operation, fieldName, transforms) {
+         schema: schema,
+         operation: operation,
+         fieldName: fieldName,
+-        args: {},
++        // It is unclear if passing on `args` here is without downsides or this
++        // _has_ to be `{}`.
++        args: args,
+         context: context,
+         info: info,
+         transforms: transforms,
 --- a/node_modules/graphql-tools/dist/stitching/schemaRecreation.js
 +++ b/node_modules/graphql-tools/dist/stitching/schemaRecreation.js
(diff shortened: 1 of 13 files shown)


patches/graphql-tools+4.0.4.patch
```diff
@@ -1,4 +1,17 @@
...
+--- a/node_modules/graphql-tools/dist/stitching/resolvers.js
++++ b/node_modules/graphql-tools/dist/stitching/resolvers.js
+@@ -51,7 +51,9 @@ function createProxyingResolver(schema, operation, fieldName, transforms) {
+         schema: schema,
+         operation: operation,
+         fieldName: fieldName,
+-        args: {},
++        // It is unclear if passing on `args` here is without downsides or this
++        // _has_ to be `{}`.
++        args: args,
```
alloy: Need to test if this change breaks tests in graphql-tools.

alloy: Does not break any tests 👍

alloy:
Also made the test output clean again, which involved replacing deprecated `Buffer` usage, so do look at that closely.

src/lib/cache.ts
```diff
@@ -111,7 +111,7 @@ function _get<T>(key) {
         if (CACHE_COMPRESSION_DISABLED) {
           resolve(JSON.parse(data.toString()))
         } else {
-          zlib.inflate(new Buffer(data, "base64"), (er, inflatedData) => {
+          zlib.inflate(Buffer.from(data, "base64"), (er, inflatedData) => {
```
zephraph: 👍

src/schema/sale/__tests__/index.test.js
```diff
@@ -175,7 +175,7 @@ describe("Sale type", () => {
     })
   })
 
-  describe("sale_artworks_connection", async () => {
+  describe("sale_artworks_connection", () => {
```
zephraph: I fixed some of these yesterday in reaction. I think we should find/write a lint rule to catch these.

alloy: Perhaps, yeah

zephraph:
I did some local testing on this and it looks good! Gonna merge. 

Thanks man!

PR #1761, opened 2019-06-10
Dynamically fetch artwork details for SaleArtwork 

ashkan18:
# Problem
This is a follow up on https://github.com/artsy/metaphysics/pull/1745. The problem is, we currently rely on the artwork json coming back from Gravity when resolving artwork in a `SaleArtwork`. But response from Gravity is the `short` version of artwork which wouldn't include for example edition set data.

# Solution
Inspect incoming query and based on requested fields, decide if we need to fetch Artwork from Gravity or just rely on current embedded one.


src/schema/__tests__/sale_artwork.test.js
```diff
@@ -90,6 +90,35 @@ describe("SaleArtwork type", () => {
...
+            artwork{
+              id
+              edition_sets{
+                id
+              }
+            }
+          }
+        }
+      `
+      expect(await execute(query)).toEqual({
```
ashkan18: in this test, can I check to see if the mocked `artworkLoader` above was used? we can kind of assume that based on the results but that's with assumption that original mocked artwork is not returning editionset.

src/schema/sale_artwork.ts
```diff
@@ -364,4 +382,24 @@ const SaleArtwork: GraphQLFieldConfig<void, ResolverContext> = {
   },
 }
 
+const shouldFetchArtwork = (
+  fieldNode: Readonly<FieldNode>,
+  fragments: any
+): boolean => {
+  const fetchArtworkFields = ["edition_of", "edition_sets"]
```
ashkan18: for now only adding these two fields as fields triggering extra call, we can add more fields when needed.

src/schema/__tests__/sale_artwork.test.js
alloy: You can change the loader into a `jest.fn(...)` and store a reference to it and make assertions against it, e.g. `expect(loader).toHaveBeenCalledWith(...)`

mzikherman:
I love how readable the GraphQL AST introspection code is! This is overall an improvement and a bug fix over the existing behavior (which would return incorrect data for edition set information), so I'm 👍 it.

But, worth mentioning that with this behavior, Prediction (which fetches an entire sale's worth of sale artworks up front, so potentially many sale artwork/bidding related fetches, _and_ requests artwork edition set info for each, so now we've added N+1 artwork fetches), might now become quite a bit slower when initially launched, and so should be monitored. It wouldn't be great for folks to be unable to or take an excessively long amount of time to join for large sales.

So, we may want to make sure Prediction will be ok with this fix, prior to or just after merging. I wouldn't be surprised if there are some follow-ups that might come out of that:
  - Prediction able to defer individual lot data fetching
  - Implement batching for artworks (similar to sales) cc @zephraph

_schemaV2.graphql
```diff
@@ -6657,6 +6657,7 @@ type MarketingCollection {
 
   # Collection has prioritized connection to artist
   is_featured_artist_content: Boolean!
+  relatedCollections: [MarketingCollection!]!
```
ashkan18: hmm.. not sure how this got in there, but seems harmless

src/schema/sale_artwork.ts
```diff
@@ -364,4 +382,24 @@ const SaleArtwork: GraphQLFieldConfig<void, ResolverContext> = {
...
+    },
+    FragmentSpread(node) {
+      const fragmentDef = fragments[node.name.value]
+      requestedArtworkFields.push(
+        ...fragmentDef.selectionSet.selections.map(s => s.name.value)
+      )
+    },
+  })
+  return requestedArtworkFields.some(f => fetchArtworkFields.includes(f))
+}
```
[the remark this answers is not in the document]
ashkan18: looks like `visit` kind of visit's the fields recursively, do you know if there is a way to limit it to only one level deep?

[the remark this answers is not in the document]
ashkan18: @alloy updated a logic a little should be slightly better since i don't check if we already know we need to fetch, but i think `visit` still traverses all nodes.

ashkan18:
@mzikherman yes, I already started working on prediction part in parallel, the thing is, this seems to be the first usecase for fetching things on-demand on prediction and seems little more involved, but that would be the ultimate solution :)

src/schema/sale_artwork.ts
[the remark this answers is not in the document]
alloy: Check the source (and tests) for the best docs https://github.com/graphql/graphql-js/blob/61f993604bbc3e0aa2dcfc9191b6f6d4cb060bee/src/language/visitor.js#L135-L165

So you should be able to return `false` to stop the visitor from recursing further down into a sub-selection and you can probably do that based on the `path` parameter to the visitor function.

Additionally, I think you should make your visitor recursive when it encounters a fragment, as a fragment could be spread into a fragment into a fragment and so on… 🐢 

```js
	const visitor = (rootNode: FieldNode) => {
	  visit(rootNode, {
	    Field(node, _key, _parent, path, _ancestors) {
	      if (path.length > 1) {
	        return false
	      }
	      requestedArtworkFields.push(node.name.value)
	    },
	    FragmentSpread(node) {
	      const fragmentDef = fragments[node.name.value]
	      // Recurse into fragment
	      visitor(fragmentDef)
	    },
	  })
	}
```

(This is untested code!)

src/schema/sale_artwork.ts
alloy: (Off-topic If we had gravitype up and running we could reflect on the different json versions and know that way if shallow is enough or not.)

src/schema/sale_artwork.ts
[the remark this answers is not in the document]
ashkan18: @alloy let me know what you think about this approach? had to check `path.length <= 3` since for fragments we have `selectionSets.selections` also in the path.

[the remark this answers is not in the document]
alloy: You’re not actually visiting the fragment yet, though. See the second comment in my example code above or let me know if it still isn’t clear 🙏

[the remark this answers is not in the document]
ashkan18: oh meant to ask you about this, right now im checking fragment's field manually within that block so this kind of takes care of first level fields. But i guess you are talking about 
```graphql
fragment TestF1 on Artwork {
  edition_of
}
fragment TestF2 on Artwork {
 id
}
query {
  artwork(id: "...") {
     ...testF1 {
       ....testF2 
     }
  }
}
```
is this even valid?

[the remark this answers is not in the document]
alloy: Sure is!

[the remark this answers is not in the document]
alloy: Well, actually TestF2 would spread inside TestF1.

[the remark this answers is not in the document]
ashkan18: trying this now, i don't think we can call original `visit` method here, we basically need to right the visit method or use `shouldFetchArtwork`
```javascript
FragmentSpread(node) {
      const fragmentDef = fragments[node.name.value]
      visit(fragmentDef, {
        Field(node) {
          ...
        }
      })
    },
```
or 
```js
 FragmentSpread(node) {
      const fragmentDef = fragments[node.name.value]
      needsArtworkFetch = shouldFetchArtwork(fragmentDef, fragments)
    },
```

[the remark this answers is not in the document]
alloy: Hmm, not sure I follow, if you would define the function inline then the fragments should still be in scope, right?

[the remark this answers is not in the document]
alloy: Btw, we could probably extract this code afterwards and use it for other places where we’ve been accessing things manually (eg in the blacklist fields function).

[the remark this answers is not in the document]
ashkan18: yes, @sweir27 mentioned the same thing about extracting it, let me see if i find a good common place for it.

ashkan18:
ended up making a more generic method for solving this issue and also other cases where we check queried fields. as part of this i removed old helper methods and switched all places i could find to use this new approach. As part of this i _think_ we now get a proper `FragmentSpeard` support which i don't think old code was supporting.

While all tests work, Typescript is complaining about some stuff, for now pushed with no-verify to make sure we want to go this route and then will work on fixing them.

thanks for bearing on this one.

src/lib/fieldSelectionHelpers.ts
```diff
@@ -0,0 +1,49 @@
...
+export const hasFieldSelection = (
+  fieldNode: Readonly<FieldNode | FragmentDefinitionNode>,
+  fragments: { [key: string]: FragmentDefinitionNode },
+  match: (nodeName: string, depth: number) => boolean
+): boolean => {
+  if (!fieldNode) return false
+  let matched: boolean = false
+  visit(fieldNode.selectionSet!.selections, {
+    Field(node, _key, _parent, path, _ancestors) {
+      if (match(node.name.value, path.length)) {
```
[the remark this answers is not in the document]
alloy: And I still think we should not be passing the depth to the call-site, we should never go further than depth 1, anything else doesn’t make sense for the call-site (as it is about an entirely different type and should probably be handled in the resolver for the nested field of interest).

src/lib/fieldSelectionHelpers.ts
```diff
@@ -0,0 +1,49 @@
...
+      }
+    },
+    FragmentSpread(node) {
+      const fragmentDef = fragments[node.name.value]
+      matched = hasFieldSelection(fragmentDef, fragments, match)
+    },
+  })
+  return matched
+}
+
```
alloy: Not a fan of these below names. I wonder if we can draw some inspiration from set theory?

src/lib/fieldSelectionHelpers.ts
```diff
@@ -0,0 +1,49 @@
...
+  fragments: { [key: string]: FragmentDefinitionNode },
+  match: (nodeName: string, depth: number) => boolean
+): boolean => {
+  if (!fieldNode) return false
+  let matched: boolean = false
+  visit(fieldNode.selectionSet!.selections, {
+    Field(node, _key, _parent, path, _ancestors) {
+      if (match(node.name.value, path.length)) {
+        matched = true
+        return false
```
alloy: Per my above comment, I think we should return `false` for any depth > 1.

src/lib/fieldSelectionHelpers.ts
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable require-yield */
+
+import { FieldNode, visit, FragmentDefinitionNode } from "graphql"
+
+export const hasFieldSelection = (
+  fieldNode: Readonly<FieldNode | FragmentDefinitionNode>,
+  fragments: { [key: string]: FragmentDefinitionNode },
```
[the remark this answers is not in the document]
ashkan18: that def helps in fixing the typescript complains :) and makes this look lot less awkward! :raised

src/lib/fieldSelectionHelpers.ts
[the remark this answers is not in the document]
ashkan18: so interesting enough, looks like `parent` in this case is always the same node! I tried to limit based on parent and say `parent === fieldNode` but with some debugging noticed `parent` is somehow the current `node` hm.... so ended up going with `path` checking.

src/lib/hasFieldSelection.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  let matched: boolean = false
+
+  const visitor = (
+    fieldNode: FieldNode | FragmentDefinitionNode,
+    fragments?: { [key: string]: FragmentDefinitionNode }
+  ): void => {
+    visit(fieldNode, {
+      Field(node, _key, _parent, path, _ancestors) {
+        if (path.length === 3 && match(node.name.value)) {
```
alloy: It looks like you’re now still recursing to deeper levels, I would expect something like this instead:

```suggestion
        if (path.length > 3) {
          return false
        }
        if (match(node.name.value)) {
```

Also, did you check if the `parent` parameter provides a saner way to determine depth?

src/lib/hasFieldSelection.ts
```diff
@@ -0,0 +1,53 @@
...
+  FragmentDefinitionNode,
+  BREAK,
+  GraphQLResolveInfo,
+} from "graphql"
+
+export const hasFieldSelection = (
+  resolveInfo: GraphQLResolveInfo,
+  match: (nodeName: string) => boolean
+): boolean => {
+  if (!resolveInfo.fieldNodes) return false
```
alloy: Why would this not exist? Is that just for tests?

src/lib/hasFieldSelection.ts
```diff
@@ -0,0 +1,53 @@
...
+  const firstFieldNode = resolveInfo.fieldNodes[0]
```
alloy: Perhaps leave a TODO that us only reflecting on the first field node may be suspect to breaking in edge cases.

src/lib/hasFieldSelection.ts
```diff
@@ -0,0 +1,53 @@
...
+        const fragmentDef = resolveInfo.fragments[node.name.value]
+        visitor(fragmentDef, fragments)
+      },
+    })
+  }
+  visitor(firstFieldNode)
+  return matched
+}
+
+export const includesFieldsSelection = (
```
alloy: How about?

```suggestion
export const hasIntersectionWithSelectionSet = (
```

src/lib/hasFieldSelection.ts
```diff
@@ -0,0 +1,53 @@
...
+  resolveInfo: GraphQLResolveInfo,
+  fields: string[]
+): boolean => {
+  return hasFieldSelection(resolveInfo, nodeName => fields.includes(nodeName))
+}
+
+export const includesOtherFieldsSelection = (
```
alloy: …and?

```suggestion
export const hasNoIntersectionWithSelectionSet = (
```

src/lib/hasFieldSelection.ts
ashkan18: ended up addressing this as part of last commit 👍

src/lib/hasFieldSelection.ts
ashkan18: no, i actually got this from one of the places where it was checking it and removing it was failing the test, I can remove and see which test breaks :)

src/lib/hasFieldSelection.ts
ashkan18: renamed in last commit

src/lib/hasFieldSelection.ts
ashkan18: ended up renaming but not to this name, `hasNoIntersectionWithSelectionSet` doesn't reflect what it does, this checks to see if there are fields other than passed in list of fields in the query.I thought of super set but then it's not really superset too since its totally fine to have other fields and non of the passed in ones.

src/schema/sale_artwork.ts
```diff
@@ -84,7 +86,18 @@ export const SaleArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
       ...GravityIDFields,
       cached,
-      artwork: { type: Artwork.type, resolve: ({ artwork }) => artwork },
+      artwork: {
+        type: Artwork.type,
+        resolve: ({ artwork }, _options, { artworkLoader }, info) => {
+          if (!info.fieldNodes) return artwork
+          const fetchArtworkFields = ["edition_of", "edition_sets"]
+          if (hasIntersectionWithSelectionSet(info, fetchArtworkFields)) {
+            return artworkLoader(artwork.id).catch(() => artwork)
```
alloy: What’s the catch for? That would only break later on when trying to pluck fields from the artwork that aren’t there, right? By that time the original error is gone, so I think it would be better to error here.

src/lib/hasFieldSelection.ts
```diff
@@ -0,0 +1,56 @@
...
+  FieldNode,
+  visit,
+  FragmentDefinitionNode,
+  BREAK,
+  GraphQLResolveInfo,
+} from "graphql"
+
+export const hasFieldSelection = (
+  resolveInfo: GraphQLResolveInfo,
+  match: (nodeName: string) => boolean
```
alloy:
```suggestion
  match: (fieldName: string) => boolean
```

src/lib/hasFieldSelection.ts
```diff
@@ -0,0 +1,56 @@
...
+      },
+    })
+  }
+  visitor(firstFieldNode)
+  return matched
+}
+
+export const hasIntersectionWithSelectionSet = (
+  resolveInfo: GraphQLResolveInfo,
+  fields: string[]
```
alloy:
```suggestion
  fieldNames: string[]
```

src/lib/hasFieldSelection.ts
```diff
@@ -0,0 +1,56 @@
...
+): boolean => {
+  return hasFieldSelection(resolveInfo, nodeName => fields.includes(nodeName))
+}
+
+export const includesFieldsOtherThanSelectionSet = (
+  resolveInfo: GraphQLResolveInfo,
+  fields: string[]
```
alloy:
```suggestion
  fieldNames: string[]
```

src/schema/sale_artwork.ts
ashkan18: the catch was a safe back up plan in case there was an issue with fetching the artwork we basically fallback to artwork returned from sale artwork.

alloy: Ok, let’s not do that and let it fail with the helpful error when it happens.

src/schema/__tests__/tag.test.js
```diff
@@ -8,7 +8,7 @@ describe("Tag", () => {
     // is not working correctly. This test is to make sure we don't
     // request to gravity.
 
-    it("returns filtered artworks", () => {
+    it.only("returns filtered artworks", () => {
```
alloy:
```suggestion
    it("returns filtered artworks", () => {
```

alloy: Committing this in your stead.

alloy: Oh… it lives on your fork… then I cannot :disappointed:

alloy: Oh… it lives on your fork… then I cannot :disappointed:

alloy:
Yasssss! :shipit:

ashkan18:
So we ended up reverting this PR in #1780 , this PR broke `/collect` page, thanks to @mzikherman we found out the query that was broken is 
```graphql
query {
  viewer {
    ...CollectApp_viewer_3XOIsA
  }
}

fragment CollectApp_viewer_3XOIsA on Viewer {
  filter_artworks(aggregations: [MEDIUM, TOTAL], size: 0) {
    __id
  }
  ...CollectFilterContainer_viewer_3XOIsA
}

fragment CollectFilterContainer_viewer_3XOIsA on Viewer {
  filter_artworks(aggregations: [MEDIUM, TOTAL], size: 0) {
    aggregations {
      slice
      counts {
        name
        id
        __id
      }
    }
    __id
  }
}
```
Did some debugging and found out the issue is the line where we only use first `fieldNode` out of `resolveInfo.fieldNodes` 
```js
const firstFieldNode = resolveInfo.fieldNodes[0]
```
with this logic, we only visit the first node. In this example most of the selected fields are in the second `fieldNode`. Not exactly sure why thats the case for this since we start from `query` but going to update the logic to handle this.

PR #1763, opened 2019-06-11
PURCHASE-1168: Serves the artwork attribution class descriptions from Metaphysics instead of storing them in-app

lilyfromseattle:
Subtask of https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=41&modal=detail&selectedIssue=PURCHASE-1105

__User Story__
As a user, I can tap on the attribution class below the tombstone to learn more. Tapping on the link will open a new page with all of Artsy’s classifications.
I can click the back button or “OK” to go back to the artwork page.

__Spec__
https://zpl.io/aX3O198

### DIFF
--- a/src/schema/artworkAttributionClasses.ts
+++ b/src/schema/artworkAttributionClasses.ts
@@ -1,0 +1,12 @@
+import { GraphQLList, GraphQLFieldConfig } from "graphql"
+import { ResolverContext } from "types/graphql"
+import AttributionClass from "./artwork/attributionClass"
+import attributionClasses from "lib/attributionClasses"
+
+const ArtworkAttributionClasses: GraphQLFieldConfig<void, ResolverContext> = {
+  type: new GraphQLList(AttributionClass),
+  description: "List of all artwork attribution classes",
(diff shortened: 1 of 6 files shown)


src/schema/artworkAttributionClasses.ts
```diff
@@ -0,0 +1,12 @@
+import { GraphQLList, GraphQLFieldConfig } from "graphql"
+import { ResolverContext } from "types/graphql"
+import AttributionClass from "./artwork/attributionClass"
+import attributionClasses from "lib/attributionClasses"
+
+const ArtworkAttributionClasses: GraphQLFieldConfig<void, ResolverContext> = {
+  type: new GraphQLList(AttributionClass),
+  description: "List of all artwork attribution classes",
+  resolve: (_root, _args, _context) => Object.values(attributionClasses),
```
sweir27: If we're doing the extra work to expose this, we should probably cast the snake case as camelcase (given that's what we prefer in our graphQL schemas). So this would expose `shortDescription` and `longDescription`.

PR #1769, opened 2019-06-12
[v2] Formalize deprecation and add ability to drop all in v2

alloy:
Closes https://artsyproduct.atlassian.net/browse/PLATFORM-1468

Not yet enabled, so clients have a longer window of opportunity to clean deprecated usage up while still receiving info about what fields to use instead, but this should be enabled before v2 goes live in an App Store version of the iOS app.

See [this commit](https://github.com/artsy/metaphysics/commit/2ead0cff34126313f32ca71ca26abf297062e271) for all dropped fields/types.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -487,13 +487,13 @@ type Artist implements Node & Searchable {
     exclude_artists_without_artworks: Boolean = true
   ): [Artist]
   consignable: Boolean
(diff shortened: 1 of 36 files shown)


alloy:
Going to self-merge this so I can work on removing deprecated usage in Emission.

PR #1762, opened 2019-06-11
[WIP] Ranked stats stitching

sepans:
**Update:**
~renamed `statType` to `rankedEntity` and `typeId` to `entityId` so some sections of queries below are slightly different.~ (updated it in queries below)

TODO:
- [ ] fix the failing test 😓 
- [ ] do I need to update `v2` schema? It fails when doing `dump-schema`

Sample query:
```graphql
query {
  partner(id: "4d8b92c44eb68a1b2c0004cb") {
   analytics {
     topArtworks: rankedStats(first: 1, objectType: ARTWORK, period: ONE_YEAR) {
      edges {
        node {
          value
          # this is what vortex returns. rankedEntity is a Union of Artwork, Artist and Show
          rankedEntity {
            ... on AnalyticsArtwork {
               __typename
               entityId  # ID of the object
	     }
           }
           # this is being stitched in MP, works fine
           artwork {
             id
             title
             artist_names
           }
           # But I don't want to even be able to query show (the same about subquery below)
           show { 
             name
          }
        }
      }
    }  
    topShows: rankedStats(first: 1, objectType: SHOW, period: ONE_YEAR) {
      edges {
        node {
            value
            rankedEntity {
              ... on AnalyticsArtwork {
                 __typename
                 entityId
               }
            }          
            show {
              id
              name
              artists {
                name
              }
            }
            # artwork shouldn't be available here!
            artwork {
              title
            }
          }
        }
      }     
    }
  }
}
```

It works fine and it returns the correct data
```
"data": {
    "partner": {
      "analytics": {
        "topArtworks": {
          "edges": [
            {
              "node": {
                "value": 2945,
                "rankedEntity": {
                  "__typename": "AnalyticsArtwork",
                  "entityId": "4dbd8b36db8e540221001f96"
                },
                "artwork": {
                  "id": "gregory-crewdson-untitled-75",
                  "title": "Untitled",
                  "artist_names": "Gregory Crewdson"
                },
                "show": null
              }
            }
          ]
        },
        "topShows": {
          "edges": [
            {
              "node": {
                "value": 1775,
                "rankedEntity": {},
                "show": {
                  "id": "gagosian-takashi-murakami-prints",
                  "name": "Takashi Murakami: Prints",
                  "artists": [
                    {
                      "name": "Takashi Murakami"
                    }
                  ]
                },
                "artwork": null
              }
            }
          ]
        }
      }
    }
  }
```

~but it also returns an error:~ (not anymore. see 👇 )
```JSON
[...]

PR #1770, opened 2019-06-12
Stitch AnalyticsRankedStats union

sepans:
~depends on https://github.com/artsy/vortex/pull/149~ (deployed to production)

We tried two other approaches ([1](https://github.com/artsy/metaphysics/pull/1762) , [2](https://github.com/artsy/metaphysics/compare/master...sepans:ranked-stats-2?expand=1)) but with this one we got it right:

cc: @alloy 

query:
```graphql
query {
    analytics {
      topArtworks: rankedStats(first: 1, objectType: ARTWORK, period: ONE_YEAR) {
        edges{
          node{
            entity {
              ... on Artwork{
                title
                artists {
                  name
                }
              }
              ... on PartnerShow{
                name
              }
            }
          }
        }
      }      
        edges{
          node{
            entity {
              ... on Artwork{
                title
                artists {
                  name
                }
              }
              ... on PartnerShow{
                name
              }
            }
          }
        }
      }
    }
  }
}
```

response:
```JSON
{
"data": {
    "partner": {
      "analytics": {
        "topArtworks": {
          "edges": [
            {
              "node": {
                "entity": {
                  "title": "Untitled",
                  "artists": [
                    {
                      "name": "Gregory Crewdson"
                    }
                  ]
                }
              }
            }
          ]
        },
        "topShows": {
          "edges": [
            {
              "node": {
                "entity": {
                  "name": "Takashi Murakami: Prints"
                }
              }
            }
          ]
        }
      }
    }
  }
}
```

### DIFF
--- a/src/lib/stitching/vortex/stitching.ts
+++ b/src/lib/stitching/vortex/stitching.ts
@@ -16,7 +16,7 @@ const getMaxPrice = (thing: { listPrice: any }) => {
 export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
   // The SDL used to declare how to stitch an object
   extensionSchema: gql`
-    union AnalyticsEntityUnion = Artwork | PartnerShow | Artist
+    union AnalyticsEntityUnion = Artwork | Show | Artist
     extend type AnalyticsPricingContext {
       appliedFiltersDisplay: String
     }
@@ -293,24 +293,9 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
           }
         `,
         resolve: (parent, _args, context, info) => {
-          const toSnakeCase = s =>
-            s
-              .replace(/\.?([A-Z])/g, (_x, y) => "_" + y.toLowerCase())
-              .replace(/^_/, "")
-          const getMPNameFromTypeName = (typename: string): string => {
-            switch (typename) {
-              case "AnalyticsArtwork":
-                return "Artwork"
-              case "AnalyticsArtist":
-                return "Artist"
-              case "AnalyticsShow":
-                return "PartnerShow"
-              default:
-                return ""
-            }
-          }
+          const removeVortexPrefix = name => name.replace("Analytics", "")
           const typename = parent.rankedEntity.__typename
-          const fieldName = toSnakeCase(getMPNameFromTypeName(typename))
+          const fieldName = removeVortexPrefix(typename).toLowerCase()
           const id = parent.rankedEntity.entityId
           return info.mergeInfo
             .delegateToSchema({
@@ -325,7 +310,7 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
               transforms: vortexSchema.transforms,
             })
             .then(response => {
-              response.__typename = getMPNameFromTypeName(typename)
+              response.__typename = removeVortexPrefix(typename)
               return response
             })
         },
--- a/src/lib/stitching/vortex/stitching.ts
+++ b/src/lib/stitching/vortex/stitching.ts
@@ -16,7 +16,7 @@ const getMaxPrice = (thing: { listPrice: any }) => {
 export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
   // The SDL used to declare how to stitch an object
   extensionSchema: gql`
-    union AnalyticsEntityUnion = Artwork | Show | Artist
+    union AnalyticsRankedEntityUnion = Artwork | Show | Artist
     extend type AnalyticsPricingContext {
       appliedFiltersDisplay: String
     }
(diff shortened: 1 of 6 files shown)


src/lib/stitching/vortex/stitching.ts
```diff
@@ -292,6 +274,63 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
...
+              info,
+              transforms: vortexSchema.transforms,
+            })
+            .then(response => {
+              response.__typename = getMPNameFromTypeName(typename)
+              return response
+            })
+        },
+      },
+    },
```
alloy: Seeing as the number of types is limited, I wonder if it isn’t much more readable to just hardcode the values.

```js
          const [mpTypeName, mpFieldName] = () => {
	            switch (typename) {
	              case "AnalyticsArtwork":
	                return ["Artwork", "artwork"]
	              case "AnalyticsArtist":
	                return ["Artist", "artist"]
	              case "AnalyticsShow":
	                return ["PartnerShow", "partner_show"]
	              default:
	                return ""
	            }
	          })()
```

sepans: Yes, and now that `PartnerShow` should be `Show` I can just do `const fieldname = typename.replace('Analytics', '').toLowercase()`

sepans: made the change in https://github.com/artsy/metaphysics/pull/1770/commits/0e0bc4f826af7a2b17c3993e096c4fe9bc36c514

src/lib/stitching/vortex/stitching.ts
```diff
@@ -16,6 +16,7 @@ const getMaxPrice = (thing: { listPrice: any }) => {
 export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
   // The SDL used to declare how to stitch an object
   extensionSchema: gql`
+    union AnalyticsEntityUnion = Artwork | Show | Artist
```
starsirius: Maybe we can call it `AnalyticsRankedEntityUnion` to be explicit?

src/lib/stitching/vortex/stitching.ts
```diff
@@ -44,9 +45,6 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
     extend type Partner {
       analytics: AnalyticsPartnerStats
     }
-    extend type AnalyticsTopArtworks {
-      artwork: Artwork
-    }
```
starsirius: `TopArtworks` is now only deprecated in Vortex. Would removing the stitching here make clients querying the stitched `artwork` fail?

src/lib/stitching/vortex/stitching.ts
```diff
@@ -292,6 +274,48 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
           amount(_ => parent.totalCents).resolve({}, args),
       },
     },
+    AnalyticsRankedStats: {
+      entity: {
+        fragment: gql`
+          ... on AnalyticsRankedStats {
+            rankedEntity{
+              __typename
+              ... on AnalyticsArtwork{
```
starsirius: #minor/human linting: space before `{`.

src/lib/stitching/vortex/stitching.ts
alloy: Perfect

src/lib/stitching/vortex/stitching.ts
sepans: Yes it would.

The deprecated `topArtwork` in vortex now returns `null` so leaving this here could also create error in MP.

Since this feature is not on production based on our previous discussion around 'fail fast and fix' we can quickly follow up with Volt PR and fix this.

src/lib/stitching/vortex/stitching.ts
starsirius: And... this is the magical part! 🎩

sepans: 👍 fixed in https://github.com/artsy/metaphysics/pull/1770/commits/ea759d1d59c8d27f800f88e3df7605a4051e374e

src/lib/stitching/vortex/stitching.ts
sepans: Thanks for catching that. The deprecated messages are even less visible when referencing those fields in the code 😞

src/lib/stitching/vortex/stitching.ts
starsirius: Commented https://github.com/artsy/vortex/pull/149/files#r293502277, and I wonder if we should keep it for backward compatibility and exercise removal of deprecated stuff separately later.

sepans: Also cross posting: https://github.com/artsy/vortex/pull/149/files#r293509762

ashkan18:
going to merge since i was last reviewer 🙌 exciting!

PR #1773, opened 2019-06-13
[v2] Refactor and extract generic RenameFields and FilterFields transforms

alloy:
As you can see there are no changes to the dumped schemas, besides formalising some new deprecations that were added today.

Having a generic `RenameFields` transform puts us in the position to rename `snake_case` fields to `camelCase`. I also heard @ashkan18 and @ds300 were interested in this transform.


alloy:
Going to self-merge this and work on aliasing.

PR #1781, opened 2019-06-18
Dynamic field fetch take ii

ashkan18:
# Problem
Once we deployed #1761 , `/collect` page broke.

# Cause
Thanks to @mzikherman we found the breaking query:
```graphql
query {
  viewer {
    ...CollectApp_viewer_3XOIsA
  }
}

fragment CollectApp_viewer_3XOIsA on Viewer {
  filter_artworks(aggregations: [MEDIUM, TOTAL], size: 0) {
    __id
  }
  ...CollectFilterContainer_viewer_3XOIsA
}

fragment CollectFilterContainer_viewer_3XOIsA on Viewer {
  filter_artworks(aggregations: [MEDIUM, TOTAL], size: 0) {
    aggregations {
      slice
      counts {
        name
        id
        __id
      }
    }
    __id
  }
}
```
Trying things locally, we found out that the fact that we limited visiting nodes only to  `fieldNodes[0]` in this specific query we were not traversing `CollectFilterContainer_viewer_3XOIsA`. 

# Solution
The fix here was to traverse all `fieldNodes` instead of only the first one.

## Do I need help?
YES! i wrote a test for this but having hard time getting it to work since I don't think my way of creating `GraphQLResolveInfo` is correct in this approach.

### DIFF
--- a/src/lib/__tests__/hasFieldSelection.test.ts
+++ b/src/lib/__tests__/hasFieldSelection.test.ts
@@ -1,4 +1,9 @@
-import { parse, GraphQLResolveInfo } from "graphql"
+import {
+  parse,
+  GraphQLResolveInfo,
+  FragmentDefinitionNode,
+  FieldNode,
+} from "graphql"
 import gql from "lib/gql"
 import {
   hasFieldSelection,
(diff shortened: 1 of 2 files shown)


src/lib/hasFieldSelection.ts
```diff
@@ -39,7 +35,7 @@ export const hasFieldSelection = (
       },
     })
   }
-  visitor(firstFieldNode)
+  resolveInfo.fieldNodes.map(f => visitor(f))
```
alloy: Perhaps we can optimise this slightly:

```ts
return resolveInfo.fieldNodes.some(firstFieldNode => {
  let matched: boolean = false

  const visitor = (
    fieldNode: FieldNode | FragmentDefinitionNode,
    fragments?: { [key: string]: FragmentDefinitionNode }
  ): void => {
    // ...
  }
  visitor(firstFieldNode)
  
  return matched
})
```

alloy: That way we don’t visit the second node if we already find it in the first one.

src/lib/__tests__/hasFieldSelection.test.ts
```diff
@@ -30,6 +30,53 @@ const info: GraphQLResolveInfo = {
...
+  }
+`)
+
+const [
+  query,
+  firstFragment,
+  secondFragment,
+] = multipleFieldNodes.definitions as any
+
+let fragmentDefs = [firstFragment, secondFragment]
```
alloy:
```suggestion
const fragmentDefs = [firstFragment, secondFragment]
```

src/lib/__tests__/hasFieldSelection.test.ts
```diff
@@ -30,6 +30,53 @@ const info: GraphQLResolveInfo = {
...
+let multipleFieldNodesFragments = {}
```
alloy:
```suggestion
const multipleFieldNodesFragments = {}
```

src/lib/__tests__/hasFieldSelection.test.ts
```diff
@@ -30,6 +30,53 @@ const info: GraphQLResolveInfo = {
...
+
+fragmentDefs.map(node => {
```
alloy:
```suggestion
fragmentDefs.forEach(node => {
```

alloy:
> YES! i wrote a test for this but having hard time getting it to work since I don't think my way of creating `GraphQLResolveInfo` is correct in this approach.

I tried to see what the issue is just from reading the current code, but wasn’t able to. Can you elaborate or should we pair?

ashkan18:
sorry i was OOO today, now im back.

So the issue is, the same query that I have in test, works fine when you actually make query in graphiQL hittin the server but my test is failing and my guess is it has something to do with difference between how a real `GraphQLResolveInfo` is generated for that query and how I create it using `parse` in my test, lets pair tomorrow if you have time.

alloy:
Alrighty, this is looking good to me. Here we go again!

alloy:
Oh wait, there’s merge conflicts... once fixed, merge on green 👍

PR #1784, opened 2019-06-18
[WIP] Let `Country` go through `AnalyticsRankedStats` stitching

sepans:
`Country` doesn't need stitching


sepans:
No longer applicable.

PR #1787, opened 2019-06-21
[v2] Extract RenameFields take 2

alloy:
This re-applies the previously reverted changes from #1773, but fixed.

The previous issue would occur with multiple fields (on different types) with the same old name
but different new names.


alloy:
Seeing as you’re OOO, I’ll self-merge this.

PR #1788, opened 2019-06-21
Adds new fields to Auction and creates new date fields

kierangillen:
-  Adds `formattedStartDateTime` and `initials` to auctions for the `AuctionsEntityItem` (see image below)

Screen Shot 2019-06-21 at 2 50 16 PM

- Removes unnecessary `formattedOpeningHoursDate` from dates.ts
- Creates `formattedStartDateTime`
- Removes hours displaying from `singleDate` and creates new field `singleDateTime` to show date with hours. `singleDate` isn't used yet, so this fix will have no breaking changes.

### DIFF
--- a/src/lib/__tests__/date.test.ts
+++ b/src/lib/__tests__/date.test.ts
@@ -5,9 +5,11 @@ import {
   exhibitionStatus,
   dateTimeRange,
   singleDate,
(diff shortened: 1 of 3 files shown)


alloy:
Great 👌✨

PR #1792, opened 2019-06-24
[WIP] First pass in replacing types on V2

ashkan18:
# What we want to do?
https://artsyproduct.atlassian.net/browse/PURCHASE-1210
Basically, replace `PartnerShow` with `Show` in V2 for all types who currently use `PartnerShow`

# So Far
We've managed to replace the type of single `partner_show` in the schema BUT 👇 

# Issues
- `Show` is a superset of `PartnerShow`, looks like at the moment we are missing something in our  resolver, all fields are that in `Show` but not `PartnerShow` are not being resolved properly and are always `null`.
- We had trouble getting this worked under fragments, meaning `... on Show` on the fields where returning `null` for all included fields under fragment, outside of fragment we can use `PartneShow` fields.
- We still need to replace Array and connection fields.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1407,8 +1407,12 @@ type ArtworkContextAuction implements Node {
     # A tz database time zone, otherwise falls back to `X-TIMEZONE` header
     timezone: String
   ): String
+
+  # A formatted description of when the auction starts or ends or if it has ended
+  formattedStartDateTime: String
   href: String
   name: String
+  initials(length: Int = 3): String
   is_auction: Boolean
   is_benefit: Boolean
     @deprecated(reason: "Prefer to use `isBenefit`. [Will be removed in v2]")
@@ -1876,8 +1880,12 @@ type ArtworkContextSale implements Node {
     # A tz database time zone, otherwise falls back to `X-TIMEZONE` header
     timezone: String
   ): String
+
+  # A formatted description of when the auction starts or ends or if it has ended
+  formattedStartDateTime: String
   href: String
   name: String
+  initials(length: Int = 3): String
   is_auction: Boolean
   is_benefit: Boolean
     @deprecated(reason: "Prefer to use `isBenefit`. [Will be removed in v2]")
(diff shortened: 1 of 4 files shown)


_schema.graphql
```diff
@@ -1407,8 +1407,12 @@ type ArtworkContextAuction implements Node {
     # A tz database time zone, otherwise falls back to `X-TIMEZONE` header
     timezone: String
   ): String
+
+  # A formatted description of when the auction starts or ends or if it has ended
+  formattedStartDateTime: String
   href: String
   name: String
+  initials(length: Int = 3): String
```
zephraph: I didn't expect these changes..

alloy: This is a matter of rebasing. It was already updated in an earlier PR.

src/schema/v2/ReplaceType.ts
```diff
@@ -0,0 +1,140 @@
...
+      if (replacementTypes.includes(name)) {
+        return this.typesToReplace[name].type
+      }
+      return type
+    })
+
+    Object.entries(fields).forEach(([fieldName, fieldDefinition]) => {
+      const fieldConfig = fieldToFieldConfig(fieldDefinition, resolveType, true)
+      // If it's not a type we want to replace, just skip it
+      if (!replacementTypes.includes(fieldDefinition.type.name)) {
```
alloy: You could do this in the `transformSchema` step too where we deal with types instead of fields.

src/schema/v2/ReplaceType.ts
```diff
@@ -0,0 +1,140 @@
...
+        newFields[fieldName] = fieldConfig
+      } else {
+        madeChanges = true
+        const replacementType = this.typesToReplace[fieldDefinition.type.name]
+          .type
+        newFields[fieldName] = {
+          ...fieldConfig,
+          type: replacementType,
```
alloy: Isn’t the type already switched by providing the `resolveType` to `fieldToFieldConfig`?

ashkan18: good point! checking it now.

ashkan18:
@alloy do you have any thoughts on the issues we are facing (in PR description)

ashkan18:
closing in favor of #1793

PR #1793, opened 2019-06-26
Add `shows` as `Show` in Artwork v2

ashkan18:
We did this by adding `v2Shows` to `Artwork` and then rename `shows` to `deprecatedShows` and then `v2Shows` to `shows` on V2 version of `Artwork`.

follow up on #1792 

# Selfie
image

### DIFF
--- a/src/schema/v2/RenameFields.ts
+++ b/src/schema/v2/RenameFields.ts
@@ -86,18 +86,25 @@ export class RenameFields implements Transform {
     const resolveType = createResolveType((_name, type) => type)
 
     Object.keys(fields).forEach(oldName => {
-      const field = fields[oldName]
-      const newField = fieldToFieldConfig(field, resolveType, true)
-      const newName = this.renamer(type, field)
-      if (newName) {
-        madeChanges = true
-        newFields[newName] = {
-          ...newField,
-          resolve: source => source[oldName],
+      /**
+       * If the key already exists, it means another field got renamed to this
+       * `oldName` and we should not override it with an old implementation.
+       * I.e. the old implementation got replaced by the newly renamed one.
+       */
+      if (!this.changedFields[fieldKey(type, oldName)]) {
+        const field = fields[oldName]
+        const newField = fieldToFieldConfig(field, resolveType, true)
+        const newName = this.renamer(type, field)
+        if (newName) {
+          madeChanges = true
+          newFields[newName] = {
+            ...newField,
+            resolve: source => source[oldName],
+          }
+          this.changedFields[fieldKey(type, newName)] = oldName
+        } else {
+          newFields[oldName] = newField
         }
-        this.changedFields[fieldKey(type, newName)] = oldName
-      } else {
-        newFields[oldName] = newField
       }
     })
 
--- a/src/schema/v2/RenameFields.ts
+++ b/src/schema/v2/RenameFields.ts
@@ -9,6 +9,10 @@ import {
   visit,
   visitWithTypeInfo,
   Kind,
+  getNamedType,
+  GraphQLList,
+  isLeafType,
+  GraphQLUnionType,
 } from "graphql"
 import {
   visitSchema,
@@ -91,7 +95,7 @@ export class RenameFields implements Transform {
           ...newField,
           resolve: source => source[oldName],
         }
-        this.changedFields[newName] = oldName
+        this.changedFields[fieldKey(type, newName)] = oldName
       } else {
         newFields[oldName] = newField
       }
@@ -116,7 +120,9 @@ export class RenameFields implements Transform {
               return
             }
 
-            const oldName = this.changedFields[node.name.value]
+            const type = getTypeWithSelectableFields(typeInfo)
+            const oldName =
+              type && this.changedFields[fieldKey(type, node.name.value)]
             if (oldName) {
               return {
                 ...node,
@@ -126,7 +132,7 @@ export class RenameFields implements Transform {
                 },
               }
             }
-            return undefined
+            return
           },
         },
       })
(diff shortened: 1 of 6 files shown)


src/schema/v2/RenameFields.ts
```diff
@@ -94,8 +91,7 @@ export class RenameFields implements Transform {
           ...newField,
           resolve: source => source[oldName],
         }
-
-        this.changedFields[fieldKey(type, newName)] = oldName
```
ashkan18: @alloy this was the cause of the issue we were running into with rename, basically this line would have set `Artwork.shows` to be `v2Shows` in that array while later in https://github.com/artsy/metaphysics/pull/1793/files#diff-e2335796e14ecc9ac96c36ef4b25a34eL124 in `transformRequest`, `fieldKey(type, node.name.value)` was returning `Show.shows` which meant we were not finding `oldName` in the dict and not getting into https://github.com/artsy/metaphysics/pull/1793/files#diff-e2335796e14ecc9ac96c36ef4b25a34eR120

ashkan18: im not sure if ☝️ is the fix but it seems to work for our current case an `ID` fields as far as i can tell.

alloy: This should break something, as we need to scope fields to types. Consider the case where we’d rename `Artwork.id` to `Artwork.slug` and `ArtworkVersion.id` to `ArtworkVersion.internalID`, in your change one of these would override the other. (Clearly we’re missing some test coverage for that case.)

As for the actual issue, are you saying that in `transformSchema` the type would be `Artwork` but in `transformRequest` it would be `Show`?

ashkan18: that makes sense, i knew that code is there for a reason :)

yes, `const type = getTypeWithSelectableFields(typeInfo)` in `transformRequest` was returning `Show` for `Artwork.shows`.

alloy: Let me know if you need assistance with figuring out the cause of that! At a glance, it might unexpectedly be going into https://github.com/artsy/metaphysics/pull/1793/files#diff-e2335796e14ecc9ac96c36ef4b25a34eL157 ?

ashkan18:
Alright.... 2 days and 1210938109283109283 breakpoints later, the fix was to check if `typeInfo.getType()` is `GraphQLList`, then return parent type...  while this worked, this broke all normal `id` fields. 
The fix for this second issue was to use a separate `RenameFields` instead of existing one for ids, added that one at the end of current transformers.

src/schema/v2/RenameFields.ts
```diff
@@ -152,7 +153,12 @@ function fieldKey(type: TypeWithSelectableFields, fieldName: string) {
 function getTypeWithSelectableFields(
   typeInfo: TypeInfo
 ): TypeWithSelectableFields {
-  const type = getNamedType(typeInfo.getType())
+  let type
+  if (typeInfo.getType() instanceof GraphQLList) {
+    return getNamedType(typeInfo.getParentType())
+  } else {
+    type = getNamedType(typeInfo.getType())
+  }
```
alloy: I think it might read better if `typeInfo.getType()` or the parent is done using the ternary operator.

ashkan18: the thing is, the true case of `if` ends up early returning, i did some updates hopefully this makes it more readable.

ashkan18:
@alloy interesting point about automatically rename all `v2_`s does that even work? so we want to rename `v2_shows` to `shows` but at the same time we need to rename existing `shows` to `deprecatedShows`. How would we do that?

alloy:
> @alloy interesting point about automatically rename all `v2_`s does that even work? so we want to rename `v2_shows` to `shows` but at the same time we need to rename existing `shows` to `deprecatedShows`. How would we do that?

We don’t. It should just be gone.

Take a look at my change that I had sent you https://github.com/artsy/metaphysics/commit/2c87bb33c9b7302a720ba439a0959c79dfa83ab3#diff-e2335796e14ecc9ac96c36ef4b25a34eR88

alloy:
Also copy the test from my commit and let’s rename to prefix with `v2_` instead of having to change casing of the first letter after `v2`.

ashkan18:
@alloy oh sorry i missed that part of your logic! makes total sense and its working now.

`v2_` makes sense, we should just make sure anything after `_` is camelCase but that could even be a peril rule or something.

src/schema/v2/index.ts
```diff
@@ -142,12 +142,9 @@ export const transformToV2 = (
...
       : []),
-    new RenameFields((type, field) => {
-      if (field.name === "v2Shows" && type.name === "Artwork") {
-        return "shows"
-      }
-      if (field.name === "shows" && type.name === "Artwork") {
-        return "deprecatedShows"
+    new RenameFields((_type, field) => {
+      if (field.name.startsWith("v2_")) {
+        return field.name.substring(3)
```
alloy: I think this transform should be _before_ the one that removes deprecated fields, otherwise those originals will no longer be there and can’t be renamed.

ashkan18: yeah! good catch! moved it further up!

src/schema/v2/RenameFields.ts
```diff
@@ -121,7 +130,8 @@ export class RenameFields implements Transform {
             }
 
             const type = getTypeWithSelectableFields(typeInfo)
-            const oldName = this.changedFields[fieldKey(type, node.name.value)]
+            const oldName =
+              type && this.changedFields[fieldKey(type, node.name.value)]
```
alloy: Why would there ever not be a type?

ashkan18: yes, i was little surprised by that too, but in case of 👇 query, `__id`'s `type` is `undefined`
```graphql
{
  artwork(id: "retna-original-limited-edition-skateboard-skate-deck-blue-with-coa-signed-by-retna") {
    id
    gravityID
    shows{
      id
      gravityID
      name
    }
  }
}
```
```console
// debug console
node.name.value
"__id"
type
undefined
````

ashkan18: actually with little more digging, looks like any field under `shows` has would have `undefined` type:
image

sweir27:
Merging this! 🎉 🎉

src/schema/v2/RenameFields.ts
alloy: Man, this `TypeInfo` API and its state during AST traversal. I should spend some time really understanding its implementation. Anyways, if tests pass and queries work–great job! 🎉

PR #1794, opened 2019-06-26
hokusai updates

eessex:
Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-1254

- Use apk to install docker libs
- Update orb to 0.3.0
- Update hokusai required version
- Remove deprecated aws config

I attempted to add alpine but ran into some issues related to the `appmetrics` lib, which had issues with node versions. It looks like there is some movement recently on this issue, see context here: https://github.com/RuntimeTools/appmetrics/issues/582


Dockerfile
```diff
@@ -1,38 +1,28 @@
...
+RUN yarn install
 
-# Set up yarn
-RUN npm install -g yarn@1.9.4
-RUN chmod +x /usr/local/bin/yarn
+# Copy application code
+COPY . ./
 
-# Set up mc
-ADD https://dl.minio.io/client/mc/release/linux-amd64/mc /usr/local/bin/mc
```
izakp: Maybe leave `mc` set up as it is used in https://github.com/artsy/metaphysics/blob/master/scripts/create-heapsnapshot.sh#L10

Dockerfile
```diff
@@ -1,38 +1,28 @@
...
-# Set up dumb-init
-ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64 /usr/local/bin/dumb-init
-RUN chown deploy:deploy /usr/local/bin/dumb-init
-RUN chmod +x /usr/local/bin/dumb-init
+# Set up deploy user and working directory
+RUN apt-get update && apt-get install \
+  dumb-init \
+  libgpm2 \
+  libslang2 \
+  mc \
```
eessex: `mc` and its dependencies are added here, is there a specific preference to import in the way below?

izakp: Doh!  Okay then, no that's fine (better even) that it's installed via a package :)

izakp:
Not sure why tests are failing here - these changes look good to go but seeing a ton of these errors in CI logs...

```
metaphysics_1            | src/schema/artwork/index.ts:56:50 - error TS2558: Expected 0 type arguments, but got 2.
metaphysics_1            | 
metaphysics_1            | 56 export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
```

eessex:
My best guess is that the patches are not being applied correctly, will continue to debug.

PR #1799, opened 2019-06-28
PURCHASE-1231: Update resolvers for framed and certificate_of_authenticity

lilyfromseattle:
# Problem
Previously, when framed and `certificate_of_authenticity` were `true` they resolved to an object with a label field with the value "Framed"/"Certificate of authenticity" and a details field that was `null`.

When they were `false` or `null` they resolved to `null`. 

Because of this we had no way to distinguish between a case where the partner checked the "no" box and a case where they left both boxes unchecked. We want a different behavior for these cases in emission, so this needed to change. 

# Solution

## `framed`

When `true` it returns
```
{ label: "Framed", details: "Included" }
``` 
when `false` it returns 
```
{ label: "Framed", details: "Not included" }
```
and when `null` it returns `null`.

## `certificate_of_authenticity`

When `certificate_of_authenticity` is `true` it returns
```
{ label: "Certificate of authenticity", details: "Included" }
``` 

when `false` it returns 
```
{ label: "Certificate of authenticity", details: "Not included" }
```
and when `null` it returns `null`.

This will allow us to display "Included" or "Not included" or hide the field in Emission and Reaction.

### DIFF
--- a/src/schema/artwork/__tests__/artwork.test.js
+++ b/src/schema/artwork/__tests__/artwork.test.js
@@ -1666,18 +1666,22 @@ describe("Artwork type", () => {
         expect(data).toEqual({ artwork: { framed: null } })
       })
     })
-    it("is null when framed is false", () => {
-      artwork.framed = false
-      return runQuery(query, context).then(data => {
-        expect(data).toEqual({ artwork: { framed: null } })
-      })
-    })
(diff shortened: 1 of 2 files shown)


ashkan18:
🙌 since this change is in MP, the changes are reflected in current places using these fields, right? do we need to verify artsy.net and iOS app?

PR #1803, opened 2019-07-03
Fix Exchange CommerceDateTime issue

ashkan18:
# Problem
Exchange date fields are exposed as `CommerceDateTime`, when switched to Exchange stitched version, this is different than our old merged version of MP's `dateField`. This makes our switch to stitched Exchange schema painful and almost not possible in a clean way.

# Solution
Before stitchign, transform Exchange's schema by adding a new `Transformer` that replaces `CommerceDateTime` fields with MP's `dateField`.

# Selfie
image

### DIFF
--- a/src/lib/stitching/exchange/__tests__/replaceComerceDateTime.test.ts
+++ b/src/lib/stitching/exchange/__tests__/replaceComerceDateTime.test.ts
@@ -30,7 +30,7 @@ const schema = transformSchema(originalSchema, [
 ])
 
 describe(ReplaceCommerceDateTimeType, () => {
-  it("renames fields on object types", async () => {
+  it("replaces CommerceDateTime fields", async () => {
     const data = await runQueryOrThrow({
       schema,
       source: gql`
--- a/src/lib/stitching/exchange/__tests__/replaceComerceDateTime.test.ts
+++ b/src/lib/stitching/exchange/__tests__/replaceComerceDateTime.test.ts
@@ -1,0 +1,56 @@
+import { GraphQLSchema, GraphQLObjectType, GraphQLScalarType } from "graphql"
+import { runQueryOrThrow } from "test/utils"
+import gql from "lib/gql"
+import { transformSchema } from "graphql-tools"
+import { ReplaceCommerceDateTimeType } from "../transformers/replaceCommerceDateTimeType"
+
+const originalSchema = new GraphQLSchema({
+  query: new GraphQLObjectType({
+    name: "Query",
+    fields: {
+      someRootField: {
+        type: new GraphQLObjectType({
+          name: "SomeType",
+          fields: {
+            someDateField: {
+              type: new GraphQLScalarType({
+                name: "CommerceDateTime",
+                serialize: x => x,
+              }),
+            },
+          },
+        }),
+      },
+    },
+  }),
+})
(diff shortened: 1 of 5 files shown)


src/lib/stitching/exchange/transformers/replaceCommerceDateTimeType.ts
```diff
@@ -0,0 +1,90 @@
...
+  createResolveType,
+  fieldToFieldConfig,
+} from "graphql-tools/dist/stitching/schemaRecreation"
+import dateField from "schema/fields/date"
+
+type TypeWithSelectableFields =
+  | GraphQLObjectType<any, any>
+  | GraphQLInterfaceType
+
+export class ReplaceCommerceTimeDateType implements Transform {
```
ds300: `ReplaceCommerceTimeDateType` => `ReplaceCommerceDateTimeType` ?

src/lib/stitching/exchange/transformers/replaceCommerceDateTimeType.ts
```diff
@@ -0,0 +1,90 @@
...
+            description: type.description,
+            astNode: type.astNode,
+            extensionASTNodes: type.extensionASTNodes,
+            isTypeOf: type.isTypeOf,
+            interfaces: type.getInterfaces(),
+          })
+        )
+      }) as TypeVisitor,
+
+      [VisitSchemaKind.INTERFACE_TYPE]: ((type: GraphQLInterfaceType) => {
```
ds300: Might be we can just visit `FIELD_TYPE` and return `{...fieldConfig, ...dateField}` if it is a `CommerceDateTime`? That would dramatically simplify this code if possible 😅

ashkan18: ooh how would that be? i did `...dateField` but in `transformFields` take a look but im all for any possibility of simplifying this.

ds300: looks like it's not possible to visit field definitions directly with the visitSchema helper, so my idea wouldn't work. Oh well. What you've got already is easy enough to follow :+1:

ashkan18:
@ds300 ran into issues in creating test for this, tried to add test for our `ReplaceCommerceDateTime` transformer but having hard time defining a sample schema where a type is `CommerceDateTime` for one of the fields.

src/lib/stitching/exchange/__tests__/replaceComerceDateTime.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    },
+  }),
+})
+
+const schema = transformSchema(originalSchema, [
+  new ReplaceCommerceDateTimeType(),
+])
+
+describe(ReplaceCommerceDateTimeType, () => {
+  it("renames fields on object types", async () => {
```
ds300:
```suggestion
  it("replaces CommerceDateTime fields", async () => {
```

zephraph:
Does Emission/Eigen use any of these fields?

ds300:
@zephraph nope, the affected date fields are exchange-only and so far the only exchange things eigen is aware of are the mutations to create an order, which don't reference dates.

PR #1804, opened 2019-07-04
[HOLD] [v2] Remove duplicated types

alloy:
Closes https://artsyproduct.atlassian.net/browse/PLATFORM-1582

- [x] `ArtworkContext*` (replace w/ actual types i.e. sale)
- [x] `ArtworkFilter*` (review for possible replacement w/ Tag)
- [x] `Highlighted*`
- [x] `HomePageModuleContext*`
- [x] `Category` (Remove in V2 in favor of `PartnerCategory`)
- [ ] ~~`CategoryType` (Remove in V2 in favor of `PartnerCategoryType`)~~ Deferring this till we get rid of transforms, see JIRA ticket for details.
- [x] `ItemType`
- [ ] ~~Make things work again when enabling removal of deprecated types.~~ Deferring this till we get rid of transforms, see JIRA ticket for details.


_schemaV2.graphql
```diff
@@ -1678,28 +1679,19 @@ type ArtworkContextFair {
 type ArtworkContextPartnerShow implements Node {
   # A globally unique ID.
   id: ID!
-    @deprecated(
```
zephraph: Why were these removed?

src/schema/home/context.ts
```diff
@@ -180,9 +181,78 @@ const Context: GraphQLFieldConfig<
...
     return moduleContext[key](context, params)
   },
 }
 
-export default Context
+const HomePageArtworkOfRelatedToFollowedArtistModuleType = new GraphQLObjectType<
+  any,
+  ResolverContext
+>({
+  name: "HomePageArtworkOfRelatedToFollowedArtistModule",
```
alloy: @zephraph Perhaps this should just be `Artist` and we can add to the description of the `context` field that `Artist` represents a followed artist?

src/schema/home/context.ts
```diff
@@ -180,9 +181,78 @@ const Context: GraphQLFieldConfig<
...
+      type: Artist.type,
+    },
+    basedOn: {
+      type: Artist.type,
+      resolve: ({ based_on }) => based_on,
+    },
+  }),
+})
+
+const HomePageArtworkOfFollowedArtistModuleType = new GraphQLObjectType<
```
alloy: @zephraph Perhaps this should just be `Artist` and we can add to the description of the `context` field that `Artist` represents a followed artist?

zephraph: Yeah, I'm not sure. The name is both very descriptive and a bit confusing so I'm definitely up for changing it. We could name the field `followedArtist` and just return the artist type?

src/schema/home/context.ts
zephraph: I don't have any suggestions. I'm not in love with it, but it also seems good enough.

src/schema/home/context.ts
zephraph: I don't know if my suggested made sense in the context of your question. I'd be wary of using a description on `context` as being the sole differentiator just because the description might be overlooked.

alloy: Fair enough, as well. I’ll leave this as-is then 👍

src/schema/home/context.ts
alloy: Fair enough 👍

alloy:
@zephraph Ok, this is ready for final review, but no merging yet please! I will do so myself after updating Emission to consume these changes.

alloy:
Going to close this as we’re no longer going to be using transformers.

PR #1785, opened 2019-06-20
Add support for slug fields in metaphysics v2

zephraph:
Will resolve PLATFORM-1552.

The purpose of this task is to remove `gravityID` from metaphysics v2 in favor of `slug` and `internalID` where appropriate. 

It requires manually combing through the list below to ensure the IDs are correct in MP v1 and match the below mapping for v2. 

- [x] "AggregationCount": { "id": "internalID" } 
- [x] "Artist": { "_id": "internalID", "id": "slug" }
- [x] "Artwork": { "_id": "internalID", "id": "slug" }
- [x] "ArtworkVersion": { "id": "internalID" }
- [x] "AuctionLot": { "id": "internalID" } (mapped to `AuctionResult`)
- [x] "Bidder": { "id": "internalID" }
- [x] "BidderPosition": { "id": "internalID" }
- [x] "BuyersPremium": { "_id": "internalID", "id": "slug" }
- [x] "Collection": { "id": "slug" }
- [x] "CollectorProfile": { "id": "internalID" }
- [x] "CreditCard": { "id": "internalID" }
- [x] "DaySchedule": { "_id": "internalID" }
- [x] "EditionSet": { "id": "internalID" }
- [x] "Fair": { "_id": "internalID", "id": "slug" }
- [x] "FairOrganizer": { "_id": "internalID", "id": "slug" }
- [x] "FeaturedLink": { "_id": "internalID", "id": "??? BOTH SAME" }
- [x] "Feedback": { "id": "internalID" }
- [x] "FollowArtist": { "id": "internalID" }
- [x] "FollowGene": { "id": "internalID" }
- [x] "FollowProfile": { "id": "internalID" }
- [x] "FollowShow": { "id": "internalID" }
- [x] "Gene": { "_id": "internalID", "id": "slug" }
- [x] "GeneFamily": { "_id": "internalID", "id": "slug" }
- [x] "GeneType": { "_id": "internalID", "id": "slug" }
- [x] "Image": { "id": "internalID" }
- [x] "InquiryRequest": { "id": "internalID" }
- [x] "Location": { "id": "internalID" }
- [x] "OrderedSet": { "_id": "internalID", "id": "??? BOTH SAME" }
- [x] "Partner": { "_id": "internalID", "id": "slug" }
- [x] "PartnerArtist": { "id": "internalID" }
- [x] "PartnerCategory": { "_id": "internalID", "id": "slug" }
- [x] "PartnerShow": { "_id": "internalID", "id": "slug" }t
[...]

PR #1818, opened 2019-07-11
Add support for slug fields in metaphysics v2

zephraph:
Reverts artsy/metaphysics#1813

This PR will reinstate https://github.com/artsy/metaphysics/pull/1785 when it's merged.


zephraph:
Okay, this is now updated with a query map that is _additive_. Old versions of the app pointing at MP will still have their queries defined 😅.

zephraph:
Ultimately, yes. Current emission master will use v2 queries (and is configured to target the v2 endpoint). Having both query types in the same map is fine.

PR #1835, opened 2019-07-22
Rename schema to schema/v1

zephraph:
Renames `schema` to `schemaV1`.

This is a step towards the [v2 schema duplication plan :lock:](https://www.notion.so/artsy/MPv2-Schema-Duplication-Success-Plan-a6f3ea7e295344c2973c51cd7e0472cc). Taking this step will provide the distinct difference between files in `v1` and files in `v2`.


alloy:
Perhaps we should've discussed the option to use `schema/v1/…` and `schema/v2/…` etc, but that is not an important change to warrant a lot of rework, so, unless that would be trivial to do, let’s merge this as-is at your own convenience 👍

_schema.graphql
```diff
@@ -204,7 +204,7 @@ type AnalyticsPartnerInquiryStats {
 # Sales stats of a partner
 type AnalyticsPartnerSalesStats {
   orderCount: Int!
-  orderResponseTime: Int!
+  orderResponseTime: Int
```
zephraph: I don't understand why this would have changed... 🤔

docs/adding_a_new_graphql_microservice.md
```diff
@@ -31,7 +31,7 @@ Let's pretend we're mapping a GraphQL API called `Three Body` but provides its o
    import fetch from "node-fetch"
    import urljoin from "url-join"
 
-   import { middlewareLink } from "../lib/middlewareLink"
+   import { middlewareLink } from "lib/middlewareLink"
```
zephraph: Nesting in `/v1` caused several of these relative paths to break. I've update them to use the root import provided via ts.

docs/create_exchange_operations.md
```diff
@@ -6,113 +6,122 @@ This is a temporary solution for now. Graphql has a mechanism called sticking th
 
 Follow this PR as a reference: https://github.com/artsy/metaphysics/pull/1375
 
-1. Copy updated exchange schema which includes the new mutation here: src/data/exchange.graphql (from exchange/_schema.graphql)
```
zephraph: Prettier?

src/schema/v1/auction_result.ts
```diff
@@ -9,12 +9,12 @@ import {
...
 import { indexOf } from "lodash"
-import { connectionWithCursorInfo } from "schema/fields/pagination"
-import Image, { normalizeImageData } from "schema/image"
+import { connectionWithCursorInfo } from "schema/v1/fields/pagination"
+import Image, { normalizeImageData } from "schema/v1/image"
 import { ResolverContext } from "types/graphql"
 
 // Taken from https://github.com/RubyMoney/money/blob/master/config/currency_iso.json
-const currencyCodes = require("../lib/currency_codes.json")
+import currencyCodes from "lib/currency_codes.json"
```
zephraph: Okay, cc @alloy this change probably needs direct eyes. I'm not sure why this was a require statement, but I removed the relative import and made it fall back to the project global import. We do have some tests around this and they _are_ passing.

src/schema/v1/me/__tests__/record_artwork_view.test.js
```diff
@@ -19,7 +19,7 @@ describe("Recording artwork views", () => {
   `
 
   const typeDefs = fs.readFileSync(
-    path.resolve(__dirname, "../../../data/gravity.graphql"),
+    path.resolve(__dirname, "../../../../data/gravity.graphql"),
```
zephraph: I left this one with the relative path imports but just updated it to point to the right place.

src/schema/v1/sale/index.ts
```diff
@@ -1,22 +1,22 @@
...
 import { pageable, getPagingParameters } from "relay-cursor-paging"
 import { connectionFromArraySlice, connectionDefinitions } from "graphql-relay"
-import { amount } from "schema/fields/money"
+import { amount } from "schema/v1/fields/money"
 import { convertConnectionArgsToGravityArgs } from "lib/helpers"
 import { map } from "lodash"
-import { NodeInterface } from "schema/object_identification"
-import { allViaLoader } from "../../lib/all"
+import { NodeInterface } from "schema/v1/object_identification"
+import { allViaLoader } from "lib/all"
```
zephraph: Another relative path update.

src/schema/v1/show.ts
```diff
@@ -38,7 +38,7 @@ import {
   GraphQLUnionType,
   GraphQLFieldConfig,
 } from "graphql"
-import { allViaLoader } from "../lib/all"
+import { allViaLoader } from "lib/all"
```
zephraph: ☝️ one mo

src/schema/v1/suggested_genes.ts
```diff
@@ -1,4 +1,4 @@
-import fetch from "../lib/apis/fetch"
+import fetch from "lib/apis/fetch"
```
zephraph: I should probably stop pointing these out now

damassi: I notice that VSCode auto-import will relatively import items that are one level deep but then absolutely import everything else. I wonder if that's configurable.

zephraph: It might be. That relative import path would've been two levels if corrected. Using the root src directory import notation just makes it a little more... portable?

zephraph:
Okay, @alloy I think this is good to go.

src/schema/v1/auction_result.ts
alloy: Probably predates TS gaining the ability to import JSON directly.

alloy:
I obviously only skimmed, but this seems good to me 👍

zephraph:
Holding on the MP deploy, then will merge.

alloy:
🥳 :tada:

PR #1838, opened 2019-07-23
[v2] Preparations to make codemodable

alloy:
For https://github.com/artsy/codemods/pull/2

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1763,27 +1763,27 @@ type ArtworkContextPartnerShow implements Node {
   # A globally unique ID.
   __id: ID!
     @deprecated(
-      reason: "The `PartnerShow` type is deprecated in favor of the `Show` type. [Will be removed in v2]"
+      reason: "The `PartnerShow` type has been deprecated. Prefer to use the `Show` type instead. [Will be removed in v2]"
     )
 
   # A slug ID.
   id: ID!
     @deprecated(
-      reason: "The `PartnerShow` type is deprecated in favor of the `Show` type. [Will be removed in v2]"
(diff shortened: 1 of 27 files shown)


alloy:
This is reviewable 👍

_schema.graphql
```diff
@@ -10925,7 +10934,7 @@ type Show implements Node {
   artworks(
     # List of artwork IDs to exclude from the response (irrespective of size)
     exclude: [String]
-    for_sale: Boolean
+    for_sale: Boolean = false
```
zephraph: Why the addition of this default?

alloy: The dev had tried to add it, but used `default` as the key rather than `defaultValue`, so it was a bug, but no runtime change (as it would have been `null`).

src/schema/sellable.ts
```diff
@@ -25,6 +27,10 @@ export const Sellable = new GraphQLInterfaceType({
     },
     price: {
       type: GraphQLString,
+      deprecationReason: deprecate({
+        inVersion: 2,
+        preferUsageOf: "sale_message",
```
zephraph: What's the context behind this? I'm assuming this is just a known change that I'm unaware of.

alloy: The `EditionSet` type has this field as being deprecated, which means that when it goes away it no longer conforms to this interface. At the time I checked usage of the others and I didn’t spot any, in Emission at least, so I went ahead and deprecated it in the interface and the other type that conforms to it (Artwork iirc). If in the future it turns out we still need it we can always add it back to v2.

PR #1842, opened 2019-07-23
[WIP] Add duplicated v2 schema

zephraph:
This PR does three things

1. Adds a duplicate schema at `src/schema/v2`
2. Removes v2 transforms from v1
3. Only applies v2 transforms to v2

As per our success plan, the intent is to incrementally disable transforms as we implement functionality directly into the v2 schema. 

Also note that there's currently no mechanism for helping developers reduce drift. I'm going to be taking that on as next steps.


zephraph:
I'm actually going to be throwing this PR away. 😬. Too much in one PR.

PR #1846, opened 2019-07-25
Duplicate V2

zephraph:
This PR consists of the follow changes:

1. duplicating `v1` to `v2`
2. duplicating `src/test/utils.ts` to `src/schema/v1/test/utils.ts` and `src/schema/v2/test/utils.ts`. This was needed because each file specifies the schema to query. 
3. Updating the import path to the correct `test/utils` file in both `v1` and `v2`. 
4. Updating `incrementalMergeSchemas` in `src/lib/mergeSchemas` to take in a schema. This was causing a test failure in `src/schema/v2/__tests__/city.test.ts` because it has a mock file that was being bypassed by stitching actually loading files from v1. 

What this PR does **not** do:

- Wire up `/v2` to anything
- Make any substantial or logical differences in `v1` when compared to `v2`

It is, for all intents and purposes, a straight copy the the basic due diligence to ensure that tests are passing for both `v1` and `v2`. 

Once we merge this the sync battles begin.


zephraph:
Heh, I don't know why I was thinking that this would be remotely easy. Tests were hanging/failing due to memory issues. I thought I could potentially alleviate this situation by pulling the tests out of the hokusai container. Our integration tests don't work outside of that environment. (Makes sense). Separated out the integration tests from the unit tests and we're still running into memory issues. Boo.

Here's the script I've been using to test memory usage in jest

```
node --expose-gc ./node_modules/.bin/jest --runInBand --logHeapUsage
```

We got a leak all right and it ain't pretty. 

Here's a demonstration: 

image

(It keeps growing from there)

I've got meh news and bad news. The meh news is that this actually happens in master too. The bad news is that because the number of tests have effectively doubled, we can't keep truckin' along with our memory sieve spewin' bytes like a bad breakfast.

We've gotta figure out what's up with this before moving on. Le sigh.

zephraph:
The memory leak feels a lot like https://github.com/facebook/jest/issues/6814. There's a PR for it at https://github.com/facebook/jest/pull/8331 and a patch-package update linked from there that we could potentially test.

ashkan18:
ayaya.. this kind of (i hope) explains why running MP tests kind of kills my laptop while its running. Great find!

alloy:
Turned out to be a different type of leak in our own code https://github.com/artsy/metaphysics/pull/1848

alloy:
Tested locally with Emission master, let’s do this 👍

PR #1848, opened 2019-07-25
[test] Fix leak that retains memory across tests.

alloy:
The combination of dd-trace and (probably) statsd seems to be causing
this. I dug a bit into dd-trace upto TracerProxy, but still unsure where
the exact cycle is. For now, and also in general, not loading these
libraries at all during tests is a good enough solution.

### DIFF
--- a/src/lib/cache.ts
+++ b/src/lib/cache.ts
@@ -3,8 +3,8 @@ import { isArray } from "lodash"
 import config from "config"
 import { error, verbose } from "./loggers"
 import Memcached from "memcached"
-import { cacheTracer } from "./tracer"
-import { statsClient } from "./stats"
+import { createCacheTracer } from "./tracer"
+import { createStatsClient } from "./stats"
 
 const {
   NODE_ENV,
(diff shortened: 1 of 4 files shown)


alloy:
cc @izakp

zephraph:
Beautiful, thank you!

alloy:
FWIW This is what put me on the path of the `statsClient` variable and the only place it was being used (`cache.ts`):

Screenshot 2019-07-25 at 18 18 21

src/lib/cache.ts
```diff
@@ -80,6 +80,12 @@ function createMemcachedClient() {
 
 export const client = isTest ? createMockClient() : createMemcachedClient()
 
+const cacheTracer: ReturnType<typeof createCacheTracer> = isTest
+  ? { get: x => x, set: x => x, delete: x => x }
+  : createCacheTracer()
```
damassi: 👍

damassi:
👍

PR #1852, opened 2019-07-26
GALL-1920 Use other show endpoint

ansor4:
- Add a new dataloader: authenticated `partnerShowLoader`
- Change show resolver in vortex stitching to use new authenticated `partnerShowLoader`, passing `partner_id` and `show_id`

Screen Shot 2019-07-26 at 3 38 13 PM

- [ ] We most likely will need to create a new graphql type in MP, it seems like `partner_show` is used a bunch in other services.
- [ ] Blocked by https://github.com/artsy/vortex/pull/185


_schema.graphql
```diff
@@ -414,7 +421,7 @@ enum AnalyticsQueryPeriodEnum {
   SIXTEEN_WEEKS
 }
 
-union AnalyticsRankedEntityUnion = Artwork | Show | Artist
+union AnalyticsRankedEntityUnion = Artwork | PartnerShow | Artist
```
alloy: The `PartnerShow` type is deprecated, `Show` is the preferred one.

src/lib/stitching/vortex/stitching.ts
```diff
@@ -282,29 +282,34 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
...
+                partnerId
               }
             }
           }
         `,
         resolve: (parent, _args, context, info) => {
           const removeVortexPrefix = name => name.replace("Analytics", "")
           const typename = parent.rankedEntity.__typename
-          const fieldName = removeVortexPrefix(typename).toLowerCase()
+          const fieldName = snakeCase(removeVortexPrefix(typename))
```
alloy: If I’m reading this right this is adding a snake case field? All new MP fields should be camelCase.

_schemaV2.graphql
```diff
@@ -11871,8 +11877,11 @@ type Viewer {
 
   # A Partner Show
   partner_show(
+    # The slug or ID of the Partner
+    partner_id: String!
+
     # The slug or ID of the PartnerShow
-    id: String!
+    show_id: String!
```
alloy: I assume these are those new snake_case fields?

src/schema/v1/partner_show.ts
```diff
@@ -363,16 +362,20 @@ const PartnerShow: GraphQLFieldConfig<void, ResolverContext> = {
...
-  resolve: (_root, { id }, { showLoader }) => {
-    return showLoader(id).then(show => {
-      if (!show.displayable) {
-        return new HTTPError("Show Not Found", 404)
-      }
+  resolve: (_root, { partner_id, show_id }, { partnerShowLoader }) => {
+    return partnerShowLoader({
+      partner_id: partner_id,
+      show_id: show_id,
+    }).then(show => {
```
alloy: As mentioned above, the `PartnerShow` type is deprecated so these changes should go to the `Show` type instead.

_schema.graphql
```diff
@@ -237,6 +239,11 @@ type AnalyticsPartnerSalesTimeSeriesStats {
   ): String
 }
 
+type AnalyticsPartnerShow {
+  entityId: String!
+  partnerId: String!
+}
```
alloy: Just out of curiosity—all these types seem simple join models, is there a reason they can’t all be the same type?

_schema.graphql
starsirius: @alloy For partner's use case in CMS (e.g. fetching their own non-displayable shows), we need to hit Gravity [partner shows](https://github.com/artsy/gravity/blob/2e667561217b8da7e22e5beab55385dbc3ac6b05/app/api/v1/partners_shows_endpoint.rb#L49) endpoint, instead of [shows](https://github.com/artsy/gravity/blob/2e667561217b8da7e22e5beab55385dbc3ac6b05/app/api/v1/shows_endpoint.rb#L16) endpoint that's for public users. Therefore, we need an authenticated version of the [partnerShowLoader](https://github.com/artsy/metaphysics/blob/ff305c7f5996d264d9465fabbdbb8efe841ac8ce/src/lib/loaders/loaders_without_authentication/gravity.ts#L50), and a type that can resolve with that loader.

The recommended `Show` type is using the [dataloader without authentication](https://github.com/artsy/metaphysics/blob/ff305c7f5996d264d9465fabbdbb8efe841ac8ce/src/schema/v1/show.ts#L668) for pubic use cases, which makes sense. We are hoping to use the other `PartnerShow` for CMS/partner's use cases, but it's also using the [dataloader _without_ authentication](https://github.com/artsy/metaphysics/blob/ff305c7f5996d264d9465fabbdbb8efe841ac8ce/src/schema/v1/partner_show.ts#L372).

It seems like we have to either add a new type (although `PartnerShow` sounds perfect) or make `Show` (or `PartnerShow`) smart enough to conditionally use the new authenticated partner show dataloader (e.g. we might have to differentiate a logged-in user and a logged-in partner). We could miss something here, and would love to get your recommendation in this case!

starsirius: This may be a new pattern once we have more partner's use cases that could be different than logged-in user's use cases.

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -89,6 +90,7 @@ export default (accessToken, userID, opts) => {
     ),
     savedArtworksLoader: gravityLoader("collection/saved-artwork/artworks", { user_id: userID, private: true }),
     sendFeedbackLoader: gravityLoader("feedback", {}, { method: "POST" }),
+    showLoader: gravityLoader(id => `show/${id}`),
```
starsirius: We probably don't need this loader.

_schema.graphql
alloy: Right, in that case the existing resolver should be made smarter. We shouldn't have 2 types to describe essentially the same modelled data with the only difference being some technical details. The reason is that it makes the graph unintuitive and now associated types need to be added in multiple places.

starsirius: That makes sense! A followup question: Depending on the request (i.e. from unauthenticated user, logged-in user, partner or admin, etc), some fields on the `Show` type may be handled/exposed differently. So we'll likely apply similar technique to resolve them differently?

ansor4:
Closing in favor of https://github.com/artsy/metaphysics/pull/1856

_schema.graphql
alloy: Yup :+1:

You could do it all in the `show` field’s resolver, if there’s not too much differences/overhead, which keeps it close to the different data loaders and thus self-documenting where those differences really are.

PR #1853, opened 2019-07-26
[HOLD] Remove duplicated types

zephraph:
Addresses [PLATFORM-1582](https://artsyproduct.atlassian.net/browse/PLATFORM-1582). 

- [x] `ArtworkContext*` (replace w/ actual types i.e. sale)
- [x] `ArtworkFilter*` (review for possible replacement w/ Tag)
- [x] `Highlighted*`
- [x] `HomePageModuleContext*`
- [x] `Category` (Remove in V2 in favor of `PartnerCategory`)
- [x] `CategoryType` (Remove in V2 in favor of `PartnerCategoryType`)
- [x] `ItemType`


src/schema/v2/artwork/context.ts
```diff
@@ -79,6 +68,7 @@ const Context: GraphQLFieldConfig<any, ResolverContext> = {
       .then(first)
       .then(show => {
         if (!show) return null
+        // FIXME: Update this to `Show`?
```
alloy: Sure, yep 👍

alloy:
The amount of deletions from the schema feels sooo good!

src/schema/v2/artwork/context.ts
alloy:
```ts
console.log()
```

(testing some stuff)

alloy:
So this is done, just waiting for a Emission PR that migrates to use these changes before merging.

_schemaV2.graphql
```diff
@@ -9544,7 +9530,7 @@ type Viewer {
 
   # A list of PartnerCategories
   partner_categories(
-    category_type: CategoryType
+    category_type: PartnerCategoryType
```
anandaroop: 🙌 thanks!

src/schema/v2/input_fields/partner_category_type.ts
```diff
@@ -1,8 +1,8 @@
 import { GraphQLEnumType } from "graphql"
 
-const CategoryType = {
+const PartnerCategoryTypeEnum = {
```
anandaroop: Should enumerated types now always be appended with `...Enum`?

src/schema/v2/partner_category.ts
```diff
@@ -1,36 +1,42 @@
...
-      resolve: ({ id }, options, { partnersLoader }) =>
-        partnersLoader(
-          _.defaults(options, {
-            partner_categories: [id],
-          })
-        ),
-    },
-  }),
+  fields: () => {
+    const Partners = require("./partners").default
```
anandaroop: Why did this change from a regular `import` to this `require` style instead?

zephraph: I think @alloy updated this because cyclical imports don't play well with debugging in node.

src/schema/v2/input_fields/partner_category_type.ts
zephraph: It was either `PartnerCategoryTypeType` or `PartnerCtegoryTypeEnum`. 😄. We actually _do_ have a precedent of appending Enum to the end of an enum type. It doesn't actually effect the name that's in the schema though.

src/schema/v2/partner_category.ts
alloy: Circular require dependencies 😞 Doing it here means it won’t get loaded until the `fields` ‘thunk’ (a lazy function) is evaluated.

src/schema/v2/input_fields/partner_category_type.ts
alloy: Probably a good idea yeah, so this needs to be added to the design doc and have a lint/peril rule.

PR #1854, opened 2019-07-29
[v2] Drop deprecated fields

alloy:
Part of: https://artsyproduct.atlassian.net/browse/PLATFORM-1638
Unblocks: https://github.com/artsy/emission/pull/1756


alloy:
Going to self-merge as @zephraph isn’t going to be in until later.

PR #1857, opened 2019-07-29
[AUCT-528] Fixes ID fields for CreditCards and Bidders (?)

ashfurrow:
Last week, we saw that auction registration wasn't working. We traced it down to an error returned from Metaphysics:

```
[ { message: 'Cannot return null for non-nullable field CreditCard.internalID.',
  locations: [ { line: 9, column: 11 } ],
  path: 
  [ 'createCreditCard',
    'creditCardOrError',
    'creditCard',
    'internalID' ] } ],
```

Looking into MP's v2 schema, we found that Bidders and CreditCards were expecting `_id` fields from Gravity, but were actually getting `id` fields.

Daniel and I _think_ this is the right change to make, but we have two questions:

- Is this the correct change? 
- How can we be rigorous about testing these kinds of mappings, from MP v2 and Gravity IDs? Specifically, how can we make sure there aren't more Gravity models that are breaking this expectation?

We saw that the v2 tests were ignored and weren't sure of the best approach to test this PR.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2513,7 +2513,7 @@ type Bidder {
   # A globally unique ID.
   id: ID!
 
-  # A type-specific ID likely used as a database ID.
+  # A type-specific ID.
   internalID: ID!
   created_at(
     format: String
@@ -4665,7 +4665,7 @@ type CreditCard {
   # A globally unique ID.
   id: ID!
 
-  # A type-specific ID likely used as a database ID.
+  # A type-specific ID.
   internalID: ID!
 
   # Brand of credit card
--- a/src/schema/v2/bidder.ts
+++ b/src/schema/v2/bidder.ts
@@ -6,13 +6,13 @@ import {
 } from "graphql"
 import { ResolverContext } from "types/graphql"
 import date from "./fields/date"
-import { InternalIDFields } from "./object_identification"
+import { IDFields } from "./object_identification"
 import Sale from "./sale/index"
 
 const BidderType = new GraphQLObjectType<any, ResolverContext>({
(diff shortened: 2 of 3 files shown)


dleve123:
wanna cc @joeyAghion on this for his context as well.

_schemaV2.graphql
```diff
@@ -4665,7 +4665,7 @@ type CreditCard {
   # A globally unique ID.
   id: ID!
 
-  # A type-specific ID likely used as a database ID.
+  # A type-specific ID.
   internalID: ID!
```
alloy: Other than comments I’m not seeing any actual changes in the schema, so this doesn’t appear to actually do anything. Perhaps you were initially working on the codebase that was still using transformations for the ID fields (which identified the fields based on their description)?

ashfurrow: The schemes is unchanged, it’s the resolver that’s different.

alloy:
Ah yes, of course. Makes sense 👍

I’m unsure if @zephraph will be around today, so I’ll take this.

zephraph:
These changes were reverted in #1859. The reason why it was failing is because the resolver on `InternalIDFields` was incorrectly referencing `_id` when it should've been referencing `id`. Being as that issue is fixed now, this will no longer be a problem.

dleve123:
@zephraph Can we double check that the iOS bidder registration flow now successfully works after the patch in https://github.com/artsy/metaphysics/pull/1859? I understand that the solution in this PR and the solution in that PR are the same in theory, but would feel more comfortable with explicit QA as @ashfurrow and I did that for this patch.

PR #1858, opened 2019-07-29
PURCHASE-1282: Use currencyCode to determine symbol in amount() helper, if it exists

sweir27:
We have an `amount` wrapper that we use to resolve money-related fields, such as those returned from Exchange.

To determine the relevant currency symbol, the helper uses either a `symbol` arg that's passed in, or the `symbol` on the parent object passed in.

This PR updates the helper to also use the parent object's `currencyCode` to determine which symbol to show.

This will allow us to return the correct amounts from metaphysics (via Exchange) when an order is in a currency other than USD.

Note that `currencyCode` is suggested naming [here](https://www.notion.so/artsy/Currency-Best-Practices-95081d370e664509928509e3dbd74c81), although metaphysics does not yet comply with the best practices set in that doc. That feels out of scope of this!

image

### DIFF
--- a/src/schema/v1/fields/__tests__/money.test.ts
+++ b/src/schema/v1/fields/__tests__/money.test.ts
@@ -55,4 +55,9 @@ describe(amount, () => {
     obj = { symbol: "£", currencyCode: "GBP" }
     expect(getResult()).toMatchInlineSnapshot(`"$12.34"`)
   })
+
+  it("doesn't break when it can't find a currencyCode", () => {
+    obj = { currencyCode: "BLAH" }
+    expect(getResult()).toMatchInlineSnapshot(`"$12.34"`)
+  })
 })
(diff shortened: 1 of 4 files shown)


src/schema/v1/fields/__tests__/money.test.ts
```diff
@@ -3,12 +3,15 @@ import { amount } from "../money"
...
+  let obj: any
+
   beforeEach(() => {
+    obj = {}
     args = { symbol: "$" }
     amountCents = 1234
   })
 
-  const getResult = () => amount(() => amountCents).resolve({}, args)
+  const getResult = () => amount(() => amountCents).resolve(obj, args)
```
alloy: Nit-pick: Instead of defining `obj` and `args` as `let` variables, it seems like it would be simpler to just pass parameters to `getResult`?

alloy: Same goes for `amountCents`.

ashkan18:
Nice! 🙌 #mergeOnGreen

src/schema/v1/fields/__tests__/money.test.ts
alloy: @sweir27 Nice, thanks 🙏

PR #1862, opened 2019-07-30
[v2] Drop id transform for exchange

alloy:
Part of https://artsyproduct.atlassian.net/browse/PLATFORM-1448

Foremost this uses Exchange’s new `internalID` fields instead of us transforming `id` to `internalID`, as a follow-up to https://github.com/artsy/exchange/pull/453

* Unlike us filtering `id` out (although in a bit of a tacky place) this does add a bunch of `internalID` fields to v1
* The `CommercePartner` and `CommerceUser` types no longer have a `internalID` field, but these types should not be used by the end-user anyways

As I didn’t want to make any large changes at this point, especially not to v1, I’m deferring to the near future some expansions to our stitching setup:
* Remove types from the final schema that are only meant for stitching, such as `CommercePartner` and `CommerceUser`
* Filter `internalID` from v1
* Filter `id` from v2


_schema.graphql
```diff
@@ -3614,6 +3615,22 @@ enum CommerceCancelReasonTypeEnum {
   SELLER_REJECTED_SHIPPING_UNAVAILABLE
 }
 
+# Autogenerated input type of ConfirmFulfillment
+input CommerceConfirmFulfillmentInput {
+  # A unique identifier for the client performing the mutation.
+  clientMutationId: String
+  id: ID!
```
alloy: @ashkan18 I’m noticing the mutations also have `id` fields.

_schemaV2.graphql
```diff
@@ -4278,7 +4297,7 @@ type CommerceSubmitPendingOfferPayload {
 }
 
 type CommerceUser {
-  internalID: String!
+  id: String!
```
alloy: @ashkan18  So these I’m not filtering out, as it would leave the type empty and then break things. However, as I understand it, these are only for stitching purposes and actually shouldn’t stay in the final public schema, is that correct?

src/schema/v2/v2/index.ts
```diff
@@ -58,6 +58,17 @@ export const transformToV2 = (
...
+    new FilterFields(
+      (type, field) =>
+        !type.name.startsWith("Commerce") ||
+        // TODO: These two exchange types are only meant for stitching and
+        //       should simply be filtered out after stitching, but I don't want
+        //       to make such a large sweeping change right now.
+        type.name === "CommercePartner" ||
+        type.name === "CommerceUser" ||
+        field.name !== "id"
+    ),
```
alloy: In the long run, this shouldn’t remain here, in fact this entire file should be retired, but as stated I will do that after the v2 work is ready for Emission as it requires a bit more thoughtful work around stitching and v1.

alloy:
Going to self-merge this and continue.

#MergeOnGreen

_schema.graphql
alloy: Scratch that, this is an input type.

_schemaV2.graphql
ashkan18: yes, this makes sense. We use the id, type field in stitching to resolve `buyer`, `seller` to a `user` or `partner`. These don't have to exist in the generated stitched schema.

ashkan18: https://github.com/artsy/metaphysics/blob/e07150646f4230f9fc3786436deaf45838fa4693/src/lib/stitching/exchange/stitching.ts#L41

PR #1863, opened 2019-07-30
Update with latest Exchange schema

ashkan18:
this updates the schema with changes in https://github.com/artsy/exchange/pull/455


ashkan18:
this is now unblocked by https://github.com/artsy/exchange/pull/468

PR #1867, opened 2019-07-30
Add `Artwork.formattedMetadata` field to address `Artwork.to_s` deprecation

damassi:
Addresses a v1 deprecation, removed in v2: 

Screen Shot 2019-07-30 at 12 27 03 PM


src/schema/v1/artwork/index.ts
```diff
@@ -780,19 +795,11 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         resolve: ({ title }) => (_.isEmpty(title) ? "Untitled" : title),
       },
       to_s: {
-        type: GraphQLString,
         deprecationReason: deprecate({
           inVersion: 2,
-          reason: "Unused field named using Ruby idiom.",
+          reason: "formattedMetadata",
```
alloy: Replace `reason` with `preferUsageOf` for a nicer deprecation message

PR #1868, opened 2019-07-30
Update `Artwork.price` deprecation to point to `listPrice`

damassi:
The pass-through field `Artwork.price` is used in a few different ways in Reaction, from analytics to JSONLD, with the suggested field `sale_message` being different in form from the parseable `price`. Since `price` usage is being discouraged, this updates the deprecation pointer to suggest `listPrice`.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1376,7 +1376,7 @@ type Artwork implements Node & Searchable & Sellable {
   ): Partner
   pickup_available: Boolean
   price: String
-    @deprecated(reason: "Prefer to use `sale_message`. [Will be removed in v2]")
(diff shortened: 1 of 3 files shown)


joeyAghion:
We've been trying to discourage clients from parsing the `price` field for a while, since it results in duplicate, inconsistent, and non-currency-aware handling in different places. The API's `price_cents` array is a better source for numeric data and I think there are other display-oriented fields for formatted values.

damassi:
I've updated the PR to point to `listPrice` as the suggested replacement, since `price_cents`, too, has been deprecated.

alloy:
👍 #MergeOnGreen

PR #1869, opened 2019-07-30
GALL-1920: YAPR on authenticated show dataloader

ansor4:
This serves the same function as the last two: to load non-displayable shows if the requestor is a partner on that show or if the requestor is an admin.

However, this is an improvement on the last one in that it grabs the userid of the requestor from the jwt, so doesn't require vortex to pass back a partner id.

Any feedback, especially surrounding how we get the JWT and parse it, would be great appreciated!

See the last two here: https://github.com/artsy/metaphysics/pull/1856, https://github.com/artsy/metaphysics/pull/1852

-------------

- [x] Pull in two separate loaders to get added benefit of caching, as detailed here: https://github.com/artsy/metaphysics/pull/1856#issuecomment-516456221


src/schema/v1/show.ts
```diff
@@ -664,16 +665,20 @@ const Show: GraphQLFieldConfig<void, ResolverContext> = {
       description: "The slug or ID of the Show",
     },
   },
-  resolve: (_root, { id }, { showLoader }) => {
+  resolve: (_root, { id }, { showLoader, accessToken }) => {
+    const decodedJwt = decodeArtsyJWT(accessToken)
+    const partnerIds: Array<String> = decodedJwt ? decodedJwt.partner_ids : []
+    const isAdmin: boolean =
+      decodedJwt && decodedJwt.roles.split(",").includes("admin")
```
alloy: So seeing as we know this is an admin, perhaps then it would make sense to explicitly use the authenticated loader only for these cases (and have all other requests cached)?

src/lib/artsy_jwt.ts
```diff
@@ -0,0 +1,10 @@
+import jwt from "jwt-simple"
+import config from "config"
+const { HMAC_SECRET } = config
+
+export const decodeArtsyJWT = token => {
+  if (!token) {
+    return null
+  }
+  return jwt.decode(token as string, HMAC_SECRET, true, "HS256")
```
ashkan18: `HMAC_SECRET` would be different for different apps, for example if a token coming into this method was a token generated for calling Vortex it would have been signed with Vortex's `ClientApplication`'s `internal_secret` and if the token was what was used to access CMS it would have been signed with `Volt`'s `internal_secret`. My guess is this `jwt.decode(...)` would fail if you pass a Vortex token. Looks like current `HMAC` secret config variable is basically to generate Causality's JWT.

src/lib/artsy_jwt.ts
```diff
@@ -0,0 +1,10 @@
+import jwt from "jwt-simple"
+import config from "config"
+const { HMAC_SECRET } = config
+
+export const decodeArtsyJWT = token => {
```
alloy: Adding the type here also means the call-site is checked.

```suggestion
export const decodeArtsyJWT = (token: string) => {
```

src/lib/artsy_jwt.ts
alloy: …and then:

```suggestion
  return jwt.decode(token, HMAC_SECRET, true, "HS256")
```

ansor4: @ashkan18 ahh ok, so this could break in production as it is. Could we work around this by keeping volt's secret in the config and decode with that? Or is there another way to only validate against user JWTs I think that could work too. `jwt.decode` seems to work with user JWTs (generated from gravity), so maybe we can only enforce this check for those.

I guess another broader question is if we can solve this without involving JWTs at all, I'd be open to any of these options.

joeyAghion: The user's ID (or roles) can be _decoded_ from the JWT without _validating_ its signature. It's only the validation that requires the secret. I would look at the implementation of this `decode` method to understand how, or the `jwt-simple` library probably has a helper to allow this.

joeyAghion: You're actually passing `true` as the 3rd argument (`noVerify`) already, so I think no secret should be required here.

src/lib/artsy_jwt.ts
```diff
@@ -0,0 +1,10 @@
+import jwt from "jwt-simple"
```
damassi: #minor - We've been making all new filenames `camelCase`

ansor4:
after talking it over with @ashkan18, we landed on making an additional call to gravity with `meLoader` to fetch a user and verify if the user can see this show. We could potentially only make this call when the show is non-displayable. While this would result in an additional round trip, it's the proper way to validate if a user can see a show, and it should be a pretty inexpensive call to make. I'm going to give this a shot and see what happens.

ansor4:
Ok, after another chat with @ashkan18 and @joeyAghion, I'm going to go ahead with decoding JWTs (without validating them) as a half step to getting partners and admin status. There were questions on whether or not MP should be making those roundtrips to gravity on every call (although currently it has become commonplace), or if MP should be validating JWTs at all. In addition, I found that I'd also need to call `mePartnersLoader` in addition to `meLoader`, resulting in two additional roundtrips for this one resolver. That and my fear of code rot make me lean towards merging what I have now, but with clearly worded variables at @joeyAghion's behest.

There are basically two stances we could take if we wanted to both decode and validate JWTs on every request in MP:
1) MP should not do any authentication, we make those roundtrips to gravity (which acts as source of truth for auth), and we verify in MP
2) All of Artsy's system should share a single secret, so any system (including MP) can validate JWTs with each request. In this case, we'd be able to easily add validation in this PR.

Both have immense tradeoffs, and neither can really be addressed in any single PR. I personally think it's a great discussion topic for the platform practice meeting, so we may bring it up there.

sepans:
I think we decided to do this authorization in MP because it seemed easier and we didn't want to change the behavior of an endpoint that other services may depend on through kinetic. 

So the other option is to implement all the authorization in Gravity controller and remove it all from MP?

src/schema/v1/show.ts
```diff
@@ -664,16 +665,37 @@ const Show: GraphQLFieldConfig<void, ResolverContext> = {
...
+    const isAdmin: boolean =
+      decodeUnverifiedJwt &&
+      decodeUnverifiedJwt.roles.split(",").includes("admin")
+    const isDisplayable = show =>
+      show.displayable || isAdmin || partnerIds.includes(show.partner._id)
+
+    const loader =
+      decodeUnverifiedJwt && loaderWithoutCache
+        ? loaderWithoutCache
+        : loaderWithCache
```
starsirius: If I understand correctly, we don't need to differentiate `loaderWithCache` and `loaderWithoutCache` here because they are both hitting the same Gravity endpoint (`/show/:id`) the same way. Since we name the loaders the same, depending if there is `accessToken` in the request, we will automatically use authenticated/non-authenticated loader and get desired caching behavior.

I could be totally wrong here. cc @mzikherman

src/lib/decodeUnverifiedJWT.ts
```diff
@@ -0,0 +1,10 @@
+import jwt from "jwt-simple"
+
+export const decodeUnverifiedJWT = (token: string) => {
+  if (!token) {
+    return null
+  }
+
+  // NOTE: We don't verify the token when decoding here
+  return jwt.decode(token, "", true, "HS256")
```
starsirius: This makes sense to me. So MP doesn't verify the signed token, and blindly pass that that to upstream service (i.e. the authority) that will verify.

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -89,6 +89,7 @@ export default (accessToken, userID, opts) => {
     ),
     savedArtworksLoader: gravityLoader("collection/saved-artwork/artworks", { user_id: userID, private: true }),
     sendFeedbackLoader: gravityLoader("feedback", {}, { method: "POST" }),
+    showLoader: gravityLoader(id => `show/${id}`),
```
starsirius: This Gravity endpoint only returns [`:public` properties](https://github.com/artsy/gravity/blob/7d49dca9bac65c88cce8b681f899a78cb1a04dcc/app/api/v1/shows_endpoint.rb#L16). It's sufficient for this particular use case, but in the future CMS is likely to need more data that's authorized for partners.

Given MP only wants to have 1 type handle the same domain model, I guess in the future we still need to either hit the [partner shows endpoint](https://github.com/artsy/gravity/blob/7d49dca9bac65c88cce8b681f899a78cb1a04dcc/app/api/v1/partners_shows_endpoint.rb#L37-L51) that currently CMS is hitting (but the extra `:partner_id` param is annoying), or consolidate the two Gravity endpoints.

Definitely not blocking for this PR and we can follow up.

ansor4:
After talking it over with @starsirius, I think we'll keep the implicit loader since the uncached loader depends on the access token anyway. While this makes the caching behavior more opaque, I think that should be a larger discussion on all of MP's loaders. Is the best practice now to fan out both uncached and cached loaders? If so we should be doing it everywhere.

Happy to make changes that result from that discussion, but we'll keep this implementation for now since I think it's cleaner and less confusing.

PR #1873, opened 2019-07-31
Stitch creditCard on Exchange Order

ashkan18:
# Problem
When we stitched Exchange schema we missed resolving `Order.creditCardId` and adding `Order.creditCard`.

# Solution 
Stitch `creditCard` using `Order.creditCardId` on both `buy` and `offer` orders.

# Problem
When we try to `delegateToSchema` the `fieldName` on V1 is `credit_card` while this has change in V2 to `creditCard`. Current stitched code is shared between V1 and V2 and current code would only work for V1.
Knowing this is only used by Reaction (`V1`) at the moment, this may be fine for now but we may want to duplicate exchange stitching to get `V2` version actually working.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2118,6 +2118,7 @@ type CommerceBuyOrder implements CommerceOrder {
   updatedAt: String!
   buyerDetails: OrderParty
   sellerDetails: OrderParty
+  creditCard: CreditCard
   itemsTotal(
     decimal: String = "."
 
@@ -2630,6 +2631,7 @@ type CommerceOfferOrder implements CommerceOrder {
   updatedAt: String!
   buyerDetails: OrderParty
   sellerDetails: OrderParty
+  creditCard: CreditCard
   itemsTotal(
     decimal: String = "."
 
@@ -2780,6 +2782,7 @@ interface CommerceOrder {
   updatedAt: String!
   buyerDetails: OrderParty
   sellerDetails: OrderParty
+  creditCard: CreditCard
   itemsTotal(
     decimal: String = "."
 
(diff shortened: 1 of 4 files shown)


_schemaV2.graphql
```diff
@@ -4474,46 +4477,6 @@ type Gene implements Node {
     last: Int
   ): ArtistConnection
   artworksConnection(
-    acquireable: Boolean
```
ashkan18: not sure why these are here! seem pretty unrelated to my changes!

src/lib/stitching/exchange/stitching.ts
```diff
@@ -105,6 +105,24 @@ export const exchangeStitchingEnvironment = (
...
   })
 
+  const creditCardResolver = {
+    fragment: `fragment CommerceOrderCreditCard on CommerceOrder { creditCardId }`,
+    resolve: (parent, _args, context, info) => {
+      const id = parent.creditCardId
+      return info.mergeInfo.delegateToSchema({
+        schema: localSchema,
+        operation: "query",
+        fieldName: "credit_card",
```
ashkan18: for this to work for `V2` this has to change to `creditCard`

alloy:
> # Problem
> When we try to `delegateToSchema` the `fieldName` on V1 is `credit_card` while this has change in V2 to `creditCard`. Current stitched code is shared between V1 and V2 and current code would only work for V1.
> Knowing this is only used by Reaction (`V1`) at the moment, this may be fine for now but we may want to duplicate exchange stitching to get `V2` version actually working.

https://artsyproduct.atlassian.net/browse/PLATFORM-1672

alloy:
LGTM 👍

PR #1875, opened 2019-08-01
Commit generated changes from v2 schema

zephraph:
This just removes the `__allowedLegacyNames` entry with `__id` in it. 

@alloy the `_schemaV2.graphql` updates might be from it not properly being updated in #1874?

I can revert that part if needed.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -4477,6 +4477,46 @@ type Gene implements Node {
     last: Int
   ): ArtistConnection
   artworksConnection(
+    acquireable: Boolean
+    offerable: Boolean
+    aggregationPartnerCities: [String]
+    aggregations: [ArtworkAggregation]
+    artistID: String
+    artistIDs: [String]
+    atAuction: Boolean
+    attributionClass: [String]
+    color: String
+    dimensionRange: String
+    extraAggregationGeneIDs: [String]
+    includeArtworksByFollowedArtists: Boolean
+    includeMediumFilterInAggregation: Boolean
+    inquireableOnly: Boolean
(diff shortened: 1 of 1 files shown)


_schemaV2.graphql
```diff
@@ -4477,6 +4477,46 @@ type Gene implements Node {
...
+    priceRange: String
+    page: Int
+    saleID: ID
+    size: Int
+    sort: String
+    tagID: String
+    keyword: String
+
+    # When true, will only return exact keyword match
+    keywordMatchExact: Boolean
```
alloy: Ah yep, this is good 👍

src/lib/stitching/mergeSchemas.ts
```diff
@@ -78,11 +78,6 @@ export const incrementalMergeSchemas = (localSchema, testConfig?: any) => {
     resolvers: extensionResolvers,
   })
 
-  // Because __allowedLegacyNames isn't in the public API
-  Object.defineProperty(mergedSchema, "__allowedLegacyNames", {
-    value: ["__id"],
-  })
-
```
alloy: Don’t we still need this for v1, though?

zephraph: 🤦‍♂ 

You're right. I'll change this to just have the thing above.

zephraph:
Okay @alloy this just the generated changes.

src/lib/stitching/mergeSchemas.ts
alloy: Hah. Great reminder that we can remove the graphql-js patch from Emission, though!

PR #1877, opened 2019-08-01
Fix skipping not null CommerceDateTime in stitching replacement

ashkan18:
# Problem
As part of #1803 , we replaced all `CommerceDateTime` fields with MP's `dateTime`. But `createdAt` field under `Order` is still coming as `CommerceDateTime`.

# Cause
When we did replacement, the transformer replaces the field resolver based on `type.name` of the field, but when a field is not nullable, the type is `GraphQLNotNull` and `type.ofType.name` for that field would be `CommerceDateTime`.

# Solution
Check for `type.ofType` when there is available also.

### DIFF
--- a/src/lib/stitching/exchange/transformers/replaceCommerceDateTimeType.ts
+++ b/src/lib/stitching/exchange/transformers/replaceCommerceDateTimeType.ts
@@ -63,7 +63,10 @@ export class ReplaceCommerceDateTimeType implements Transform {
     const fields = type.getFields()
     const newFields: GraphQLFieldConfigMap<any, any> = {}
     const resolveType = createResolveType((_name, type) => {
-      if (type.name === "CommerceDateTime") {
+      if (
+        type.name === "CommerceDateTime" ||
+        (type.ofType && type.ofType.name === "CommerceDateTime")
+      ) {
         return dateField.type
       }
       return type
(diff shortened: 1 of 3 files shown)


src/lib/stitching/exchange/transformers/replaceCommerceDateTimeType.ts
```diff
@@ -72,14 +75,18 @@ export class ReplaceCommerceDateTimeType implements Transform {
     Object.entries(fields).forEach(([fieldName, fieldDefinition]) => {
       const fieldConfig = fieldToFieldConfig(fieldDefinition, resolveType, true)
       // If it's not a type we want to replace, just skip it
-      if (fieldDefinition.type.name !== "CommerceDateTime") {
-        newFields[fieldName] = fieldConfig
-      } else {
+      if (
+        fieldDefinition.type.name === "CommerceDateTime" ||
+        (fieldDefinition.type.ofType &&
+          fieldDefinition.type.ofType.name === "CommerceDateTime")
```
alloy: There’s a function in GraphQL-JS to get an unwrapped type (unwrapping nonnull and lists). I believe it’s called `getNamedType` or something like that.

ashkan18: Nice! switched to use that for checking non null fields.

PR #1879, opened 2019-08-01
Remove legacy exchange stitching

zephraph:
This is an attempt to remove the legacy (manual) exchange stitching.


zephraph:
There's still one `ecommerce` reference. 

https://github.com/artsy/metaphysics/blob/bd2a9b97224f46b6e96c665bca08825c1fd53803/_schemaV2.graphql#L820

I don't think it's directly related though, so we can follow up with that in a separate PR if it needs to be removed/renamed.

alloy:
Screenshot 2019-08-02 at 06 54 52

💥

src/schema/v2/schema.ts
```diff
@@ -184,31 +163,7 @@ if (!ENABLE_CONSIGNMENTS_STITCHING) {
...
-stitchedMutations.ecommerceConfirmPickup = ConfirmPickupMutation
-stitchedMutations.ecommerceFulfillOrderAtOnce = FulfillOrderAtOnceMutation
-stitchedMutations.ecommerceRejectOrder = RejectOrderMutation
-stitchedMutations.ecommerceSubmitOrder = SubmitOrderMutation
-stitchedMutations.ecommerceAddInitialOfferToOrder = AddInitialOfferToOrderMutation
-stitchedMutations.ecommerceSubmitOrderWithOffer = SubmitOrderWithOfferMutation
-stitchedMutations.ecommerceFixFailedPayment = FixFailedPaymentMutation
-
 export default new GraphQLSchema({
-  allowedLegacyNames: ["__id"],
```
alloy: 👌

alloy:
When we separate stitching for v1 and v2 we should also see if there’s an opportunity to remove the `exchangeSchema` inserted into `ResolverContext`.

alloy:
`ecommerceArtworks` seems to be a pass-through field from Gravity, so that’s just a matter of similar naming 👍

PR #1882, opened 2019-08-02
Fix CommerceOrder stitching

sweir27:
@mdole noticed that our resolvers for `CommerceOrder` weren't working-- in particular, the fields under `CommerOrder` wrapped in the `amount` helper weren't resolving correctly.

It appears this issue is due to the fact that we weren't actually including the `CommerceOrder` in our resolvers for the stitched version of the schema.

This should fix!
image

### DIFF
--- a/src/lib/stitching/exchange/__tests__/stitching.test.ts
+++ b/src/lib/stitching/exchange/__tests__/stitching.test.ts
@@ -28,6 +28,18 @@ it("extends the Order objects", async () => {
   }
 })
 
+it("resolves amount fields on CommerceOrder", async () => {
+  const { resolvers } = await getExchangeStitchedSchema()
+  const totalListPriceResolver = resolvers.CommerceOrder.totalListPrice.resolve
+
+  expect(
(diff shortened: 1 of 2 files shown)


ashkan18:
interesting! so if you had it under `... on CommerceBuyOrder` it would have worked?

As for tests, i think we can add something in https://github.com/artsy/metaphysics/blob/master/src/lib/stitching/exchange/__tests__/stitching.test.ts

src/lib/stitching/exchange/__tests__/stitching.test.ts
```diff
@@ -28,6 +28,18 @@ it("extends the Order objects", async () => {
   }
 })
 
+it("resolves amount fields on CommerceOrder", async () => {
```
sweir27: @ashkan18 added a test for the resolver specifically-- this would fail before because `resolvers.CommerceOrder` didn't exist, but I think it's still good 👍 .

ashkan18:
🙌

PR #1886, opened 2019-08-06
[v2] Drop non-connection lists for which we have connections

alloy:
Part of https://artsyproduct.atlassian.net/browse/PLATFORM-1650
Precedes https://github.com/artsy/emission/pull/1774


src/schema/v1/me/followed_fairs.ts
```diff
@@ -1,23 +1,14 @@
...
-  },
-})
-
 export const FollowedFairConnection = connectionDefinitions({
   name: "FollowedFair",
-  // FIXME: 'edgeType' does not exist in type 'ConnectionConfig'
-  // @ts-ignore
-  edgeType: FollowedFairEdge,
+  // This never ended up being used in the underlying lib.
+  // edgeType: FollowedFairEdge,
```
alloy: @zephraph I should’ve called this out, but these v1 changes are no-ops. This never actually worked and enabling it breaks things.

PR #1897, opened 2019-08-20
Disable failing snapshot test

zephraph:
This particular test seems to have a field that flip flops based on whether jest cache is available or not. I'm just gonna disable it for now until we can dig into it further.


alloy:
Can you expand on what you mean by flip-flopping? Are consecutive build succeeding, failing, and then succeeding again? (Got any links?)

zephraph:
isActive can change depending on if jest cache is present

PR #1899, opened 2019-08-21
Adds priceIncludesTax to Artwork

kierangillen:
- Metaphysics side to expose whether tax is included in the price of an artwork
- Gravity side https://github.com/artsy/gravity/pull/12518


kierangillen:
@alloy Yes some clearer definition of price related fields would be helpful in a refactor. Added this to v1.

PR #1901, opened 2019-08-23
Update emission querymap JSON

alloy:
📖


zephraph:
Weird native dep issue caused the test failure. I kicked them back off.

PR #1915, opened 2019-08-29
[WIP] [GALL-1627] Partner blacklist

sepans:
Fixes [GALL-1627](https://artsyproduct.atlassian.net/browse/GALL-1627)

## Problem:
`parseFieldASTsIntoArray` function assumes the top level `fieldASTs`s are of type `Field` so with `map(flatMap(fieldASTs, "selectionSet.selections"), "name.value")` it gets the name of all of them, but when that assumption is not true (e.g. top level `fieldAST` is an inline fragment) it doesn't go deeper into the tree.

This hasn't been previously a problem but when we stitched `analytics` under `partner` the top level AST is [this](https://github.com/artsy/metaphysics/blob/master/src/lib/stitching/vortex/stitching.ts#L239-L241) inline fragment and `parseFieldASTsIntoArray` returns `[undefined]` and the `queriedForFieldsOtherThanBlacklisted` function fails to recognize that there are no other fields in the query and always queries gravity for partner data even when the `id` is already provided in the query and no other data is needed from gravity.

## Solution
The naive solution would have been adding an extra if block and handle the case where AST is an `InlineFragment` but the elegant solution which makes  `parseFieldASTsIntoArray` handle all cases was to recursively iterate over the tree.
I was going to implement the tree traversal code but randomly came across [graphql-list-fields](https://github.com/jakepusateri/graphql-list-fields) library. It has decent amount of tests and simple enough that we can contribute/maintain it. I am open to having our own implementation of it but it would be pretty much the same code.

Finally I updated the specs to test the `partnerLoader` is only being called when the `id` passed to the query is the slug. 

## TODO:
- [ ] V2. (Just copy V1?)

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "graphql": "0.13.2",
     "graphql-depth-limit": "1.1.0",
     "graphql-extensions": "0.5.7",
+    "graphql-list-fields": "^2.0.2",
     "graphql-middleware": "1.2.6",
     "graphql-relay": "0.5.4",
     "graphql-tools": "4.0.4",
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -3,21 +3,21 @@ import {
   camelCase,
   compact,
   difference,
-  flatMap,
   flow,
   includes,
   isEmpty,
   isObject,
   isString,
-  map,
   omit,
   reject,
   trim,
+  uniq,
 } from "lodash"
 import now from "performance-now"
 import { stringify } from "qs"
 import { getPagingParameters, CursorPageable } from "relay-cursor-paging"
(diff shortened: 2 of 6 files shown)


sepans:
Nice! didn't know about that util 👍

alloy:
@sepans Actually I just realised that the backfield function can just be updated to use https://github.com/artsy/metaphysics/blob/7876a86455667e820dde8f4bc0967c54e49ce557/src/lib/hasFieldSelection.ts#L62, no further work needed. (Although I do wonder what the performance looks like if we continuously visit the AST from the start again, rather than finding all selections once, but I guess we can keep an eye on that.)

sepans:
Closing in favor of https://github.com/artsy/metaphysics/pull/1918

PR #1918, opened 2019-08-29
[GALL-1627] Partner loader query fix for partner analytics

sepans:
Fixes [GALL-1627](https://artsyproduct.atlassian.net/browse/GALL-1627)

## Problem:
`parseFieldASTsIntoArray` function assumes the top level `fieldASTs`s are of type `Field` so with `map(flatMap(fieldASTs, "selectionSet.selections"), "name.value")` it gets the name of all of them, but when that assumption is not true (e.g. top level `fieldAST` is an inline fragment) it doesn't go deeper into the tree.

This hasn't been previously a problem but when we stitched `analytics` under `partner` the top level AST is [this](https://github.com/artsy/metaphysics/blob/master/src/lib/stitching/vortex/stitching.ts#L239-L241) inline fragment and `parseFieldASTsIntoArray` returns `[undefined]` and the `queriedForFieldsOtherThanBlacklisted` function fails to recognize that there are no other fields in the query and always queries gravity for partner data even when the `id` is already provided in the query and no other data is needed from gravity.

## Solution
In the [first attempt](https://github.com/artsy/metaphysics/pull/1915) using a library but @alloy pointed out that we already have the exact same helper function available ❤️ 

So now using `includesFieldsOtherThanSelectionSet` instead of `parseFieldASTsIntoArray` in `queriedForFieldsOtherThanBlacklisted`

Finally I updated the specs to test the `partnerLoader` is only being called when the `id` passed to the query is the slug. 

It also required updating a bunch of loader classes to pass `GraphQLResolveInfo` instead of just `FieldNodes` to `queriedForFieldsOtherThanBlacklisted`

Other than tests passing also verified the result of `includesFieldsOtherThanSelectionSet` is exactly the same as the previous implementation when running tests.

## TODO:
- [ ] V2. (Just updating `analytics` to `_id` in `v2/partner.ts` is it still called `_id`?)

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -15,8 +15,6 @@ import now from "performance-now"
 import { stringify } from "qs"
 import { getPagingParameters, CursorPageable } from "relay-cursor-paging"
 import { formatMarkdownValue } from "schema/v1/fields/markdown"
-import { includesFieldsOtherThanSelectionSet } from "./hasFieldSelection"
-import { GraphQLResolveInfo } from "graphql"
 
 const loadNs = now()
 const loadMs = Date.now()
@@ -83,14 +81,6 @@ export const stripTags = (str?: string) => {
 }
 export const markdownToText = str => {
   return stripTags(formatMarkdownValue(str, "html"))
-}
-
-export const queriedForFieldsOtherThanBlacklisted = (
-  info: GraphQLResolveInfo,
-  blacklistedFields
-) => {
-  if (!info) return true
-  return includesFieldsOtherThanSelectionSet(info, blacklistedFields)
 }
 
 export const convertConnectionArgsToGravityArgs = <T extends CursorPageable>(
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -2,14 +2,11 @@ import {
   assign,
   camelCase,
   compact,
-  difference,
-  flatMap,
   flow,
   includes,
   isEmpty,
   isObject,
   isString,
-  map,
   omit,
   reject,
   trim,
@@ -18,6 +15,8 @@ import now from "performance-now"
 import { stringify } from "qs"
 import { getPagingParameters, CursorPageable } from "relay-cursor-paging"
 import { formatMarkdownValue } from "schema/v1/fields/markdown"
+import { includesFieldsOtherThanSelectionSet } from "./hasFieldSelection"
+import { GraphQLResolveInfo } from "graphql"
 
 const loadNs = now()
 const loadMs = Date.now()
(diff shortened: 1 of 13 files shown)


src/schema/v1/partner.ts
```diff
@@ -290,14 +290,15 @@ const Partner: GraphQLFieldConfig<void, ResolverContext> = {
       description: "The slug or ID of the Partner",
     },
   },
-  resolve: (_root, { id }, { partnerLoader }, { fieldNodes }) => {
-    const blacklistedFields = ["analytics"]
+  resolve: (_root, { id }, { partnerLoader }, info) => {
+    // when the query is partner analytics, vortex stitching fragment only has _id
+    const blacklistedFields = ["_id"]
```
sepans: I originally thought we had to blacklist `analytics` for queries like this:
```graphql
query {
    analytics {
         ...
      }
    }
  }
}
```

but realized when the analytics query gets to this point the query is the [stitching fragment](https://github.com/artsy/metaphysics/blob/master/src/lib/stitching/vortex/stitching.ts#L239-L241) so need to blacklist (whitelist) `_id`

src/lib/helpers.ts
```diff
@@ -85,20 +84,13 @@ export const stripTags = (str?: string) => {
...
   blacklistedFields
 ) => {
-  if (!fieldASTs) return true
-  const queriedFields = parseFieldASTsIntoArray(fieldASTs)
-  return difference(queriedFields, blacklistedFields).length > 0
-}
-export const queryContainsField = (fieldASTs, soughtField) => {
-  return parseFieldASTsIntoArray(fieldASTs).includes(soughtField)
+  if (!info) return true
+  return includesFieldsOtherThanSelectionSet(info, blacklistedFields)
```
oxaudo: Did not have time to go into too much details but stealing @joeyAghion usual comment here. I think we had a discussion how `blacklisted` is not a preferred terminology now. Maybe something with `deny` is a better option?

sepans:
@alloy I am a bit confused about `v2`. As I understood it that v1 `_id` comes from [this](https://github.com/artsy/metaphysics/blob/master/src/lib/stitching/vortex/stitching.ts#L239-L241) fragment. Does it go through the schema transformation before getting to the loader in v2, or we need to also update the stitching code for v2? (well not that either of those are required now but wanted to keep that in mind for future)

alloy:
@sepans Hmm, this is an excellent point, thanks for raising it. I think we’ll notice soon enough when we migrate reaction 😬 /cc @zephraph

alloy:
@sepans Sorry, I lost track of this. Is it ready for merging?

sepans:
I believe so. I addressed the comments and the [v2 problem](https://github.com/artsy/metaphysics/pull/1918#issuecomment-526604387) doesn't seem to be something we need to worry about now.

alloy:
Great, thanks for bearing with me! 🙌

PR #1922, opened 2019-09-02
[v2] Move services into System

alloy:
Consignments needs this, but it wasn't using relay and thus was silently failing.


alloy:
#MergeOnGreen

PR #1941, opened 2019-09-09
PURCHASE-1471, PURCHASE-1468, 1476: Updates to inquiry fees description, shippingInfo returns non-country specific strings, new fields added related to shipping

lilyfromseattle:
#### User story: 
As a user on an artwork page with BN or MO where the price listed is in GBP
If there is a shipping fee 
Then I should see the price of shipping in GBP and the domestic price should refer to shipping within the UK

I also added a new field, `onlyShipsDomestically`, and deprecated the field `shipsToContinentalUSOnly` because the logic behind it would return true for artworks shipping from the UK but only within the UK, so the name no longer made sense. 

#### Screenshot of previous state:
https://cl.ly/1c0f8f80ac3d

Current state (I didn't want to link to reaction/force just for a screenshot, but here is proof the field is returning the correct messaging with appropriate currency symbols):
Screen Shot 2019-09-09 at 5 45 19 PM

Changed to "Shipping, tax, and _aditional_ _fees_ quoted by the seller"

As a user, If I'm buying a work that can only be shipped domestically, When I enter my shipping information, the shipping form has the artwork's origin country selected, and I cannot select any other country.

[PURCHASE-1471]: https://artsyproduct.atlassian.net/browse/PURCHASE-1471

[PURCHASE-1468]: https://artsyproduct.atlassian.net/browse/PURCHASE-1468

### DIFF
--- a/src/schema/v1/artwork/index.ts
+++ b/src/schema/v1/artwork/index.ts
@@ -41,7 +41,7 @@ import AttributionClass from "schema/v1/artwork/attributionClass"
 // Mapping of attribution_class ids to AttributionClass values
 import attributionClasses from "lib/attributionClasses"
 import { LotStandingType } from "../me/lot_standing"
-import { amount } from "schema/v1/fields/money"
+import { moneyDisplay, symbolFromCurrencyCode } from "schema/v1/fields/money"
 import { capitalizeFirstCharacter } from "lib/helpers"
 import artworkPageviews from "data/weeklyArtworkPageviews.json"
 import { ResolverContext } from "types/graphql"
@@ -585,6 +585,20 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLBoolean,
         description:
           "Is this work available for shipping only within the Contenental US?",
+        deprecationReason: deprecate({
+          inVersion: 2,
+          preferUsageOf: "onlyShipsDomestically",
+        }),
+        resolve: artwork => {
+          return Boolean(
+            artwork.domestic_shipping_fee_cents &&
+              artwork.international_shipping_fee_cents == null
+          )
+        },
+      },
+      onlyShipsDomestically: {
+        type: GraphQLBoolean,
+        description: "Is this work only available for shipping domestically?",
         resolve: artwork => {
           return Boolean(
             artwork.domestic_shipping_fee_cents &&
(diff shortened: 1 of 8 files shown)


src/schema/v1/artwork/index.ts
```diff
@@ -585,6 +585,20 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+          preferUsageOf: "onlyShipsDomestically",
+        }),
+        resolve: artwork => {
+          return Boolean(
+            artwork.domestic_shipping_fee_cents &&
+              artwork.international_shipping_fee_cents == null
+          )
+        },
+      },
+      onlyShipsDomestically: {
```
sweir27: I took a look at where this is used, and it highlights something else we'll have to do before launching to the UK--

https://github.com/artsy/reaction/blob/8faaae88e740ec99c61ae1990f590cc0726bb2fe/src/Apps/Order/Routes/Shipping/index.tsx#L338

Right now if `shipsToContinentalUSOnly` is true, we disable the country select dropdown on the shipping form and default it to "United States".

In order to make this country-agnostic, we'll have to update that logic to pin the country select to the country that's being shipped from. I made https://artsyproduct.atlassian.net/browse/PURCHASE-1476 to track!

src/schema/v1/artwork/index.ts
```diff
@@ -601,31 +615,36 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+            return "Free domestic shipping"
           if (
             artwork.domestic_shipping_fee_cents === 0 &&
             artwork.international_shipping_fee_cents === 0
           )
             return "Free shipping worldwide"
-          var domesticShipping = amount(
-            ({ domestic_shipping_fee_cents }) =>
-              domestic_shipping_fee_cents || null
-          ).resolve(artwork, { precision: 0 })
```
sweir27: I believe that instead of refactoring the `money` helper here to return a display helper, you can pass `{ precision: 0, symbol: symbolFromCurrencyCode(artwork.price_currency)  }` and it should resolve how you want it to.

src/schema/v1/fields/money.ts
```diff
@@ -53,26 +53,32 @@ export const amount = centsResolver => ({
...
-      : null
+export const symbolFromCurrencyCode = currencyCode => {
+  return currencyCode
+    ? currencyCodes[currencyCode.toLowerCase()] &&
+        currencyCodes[currencyCode.toLowerCase()].symbol
+    : null
+}
 
-    const symbol = options.symbol || obj.symbol || symbolFromCurrencyCode
+export const moneyDisplay = (cents, symbol, options) => {
```
sweir27: As stated above, if you pass in the symbol directly then you can remove this exported helper and rely on the existing resolver.

src/schema/v1/artwork/index.ts
```diff
@@ -649,6 +671,16 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           return artwork.shipping_origin && artwork.shipping_origin.join(", ")
         },
       },
+      shippingCountry: {
+        type: GraphQLString,
+        description: "The country an artwork will be shipped from.",
+        resolve: artwork => {
+          return (
+            artwork.shipping_origin &&
+            artwork.shipping_origin[artwork.shipping_origin.length - 1]
```
sweir27: This worries me because we're making an assumption about the structure of an arbitrary array. Would probably be better to return a `shipping_origin_country` from gravity that we could use here, but given all of the logic is in this resolver I think we can follow up on that.

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -1553,6 +1566,63 @@ describe("Artwork type", () => {
...
 
+  describe("#onlyShipsDomestically", () => {
+    const query = `
+      {
+        artwork(id: "richard-prince-untitled-portrait") {
+          onlyShipsDomestically
+        }
+      }
+    `
+    it("is true when domestic_shipping_fee_cents is present and international_shipping_fee_cents is null", () => {
```
sweir27: Awesome test coverage 💯

PR #1955, opened 2019-09-16
Add back the Bidder query to the top-level root field

yuki24:
This PR adds back the top-level `Bidder` query to the root field.

In https://github.com/artsy/metaphysics/pull/1952 we initially discussed not adding a root filed unless absolutely necessary or it truly makes sense. I then switched the gear to changed it to an implementation of the `Node` type, which seems to be working fine [except the fact that the client needs to know how MP's global id is decoded/encoded](https://github.com/artsy/pulse/pull/415#discussion_r324298623). However, the use cases we have in Pulse are heavily relying on raw MongoDB ids and we can't easily take advantage of the generic  `Node` type as much as we do in other frontend repos.

So, I am proposing we add the `Bidder` query to the root field. The risk here is that this could start the whole series of "adding another type to root" PRs. I believe we have to add another top-level query for `SaleArtwork` which [will be used here](https://github.com/artsy/pulse/blob/1234d4a79a4b79df52b912c1e5db5014420de09b/app/services/lot_placed_on_block_service.rb#L5) and there might be a lot more going forward. let me know what you all think.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -5399,6 +5399,12 @@ type Query {
     id: String!
   ): Artist
 
+  # A Bidder
+  bidder(
+    # The ID of the bidder
+    id: String!
+  ): Bidder
+
   # A city-based entry point for local discovery
   city(
     # A slug for the city, conforming to Gravity's city slug naming conventions
--- a/src/schema/v2/schema.ts
+++ b/src/schema/v2/schema.ts
@@ -9,6 +9,7 @@ import {
 import Artwork from "./artwork"
 import ArtworkAttributionClasses from "./artworkAttributionClasses"
 import Artist from "./artist"
+import Bidder from "./bidder"
 import Fair from "./fair"
 import Gene from "./gene"
 import HomePage from "./home"
@@ -137,6 +138,7 @@ export default new GraphQLSchema({
       // artworks: Artworks,
       artist: Artist,
       // artists: Artists,
+      bidder: Bidder,
       // causalityJWT: CausalityJWT, // TODO: Perhaps this should go into `system` ?
       city: City,
       // collection: Collection,


alloy:
> So, I am proposing we add the Bidder query to the root field. The risk here is that this could start the whole series of "adding another type to root" PRs. I believe we have to add another top-level query for SaleArtwork which will be used here and there might be a lot more going forward. let me know what you all think.

Right, that's what worries me. I don’t think most clients need these as entry points into the schema. Can we consider having Gravity encode the Node IDs and enqueue those? I'd be fine with _some_ duplication of that, as long as it resides in canonical systems.

alloy:
@yuki24 @dleve123 I think we had that short meeting after this thread, but unsure where that leaves this PR. Can you update it in whatever way needed?

dleve123:
@alloy 

> @yuki24 @dleve123 I think we had that short meeting after this thread, but unsure where that leaves this PR. Can you update it in whatever way needed?

I believe the takeaway from our short meeting was that we will circle back to this item after we are doing with our initial email implementation. We are done with that implementation, and here is some relevant code: 

- call site to a MP query: https://github.com/artsy/pulse/blob/e27e804148cc9dc8be24bb2be86bbb8ba55b17ee/app/operations/metaphysics.rb#L18
- the actual MP query https://github.com/artsy/pulse/blob/master/app/operations/metaphysics/user_and_lot_by_bidder_and_lot_id.graphql#L2

My impression is that while it's not terribly complicated in Pulse ATM, if we are concerned about the the possibility of changing the name of the `Bidder` type in MP at some point in time (due to stitching or other reasons), I'd much rather add `bidder` as a top level query than risk a breakage to our transactional emails.

I think the minimalism argument is a compelling one, but not compelling enough here, so would like to move forward with this PR.

@yuki24 Curious to get your thoughts as well!

Thanks so much for circling back here @alloy – are you comfortable with that?

damassi:
@yuki24 - can this be merged / closed?

yuki24:
We do want to bring it back but at the same time this isn't extremely required. Closing.

PR #1966, opened 2019-09-24
Add time_zone/timeZone to Sale

bhoggard:
https://artsyproduct.atlassian.net/browse/AUCT-626

We are only adding time zone, with the expectation that clients will manage formatting to local time zones for now. Times are still stored in UTC.

I had one question: Is it expected that adding time zone to Sale would cause it to appear in multiple places in the generated V1 schema file?

### DIFF
--- a/src/schema/v1/sale/index.ts
+++ b/src/schema/v1/sale/index.ts
@@ -344,7 +344,10 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       symbol: { type: GraphQLString },
-      timeZone: { type: GraphQLString },
+      timeZone: {
+        type: GraphQLString,
+        resolve: ({ time_zone }) => time_zone,
+      },
     }
   },
 })
--- a/src/schema/v1/sale/index.ts
+++ b/src/schema/v1/sale/index.ts
@@ -344,7 +344,7 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       symbol: { type: GraphQLString },
-      time_zone: { type: GraphQLString },
+      timeZone: { type: GraphQLString },
     }
   },
 })
--- a/src/schema/v1/sale/index.ts
+++ b/src/schema/v1/sale/index.ts
@@ -344,6 +344,7 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       symbol: { type: GraphQLString },
+      time_zone: { type: GraphQLString },
     }
   },
 })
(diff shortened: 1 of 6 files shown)


src/schema/v1/sale/index.ts
```diff
@@ -344,6 +344,7 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       symbol: { type: GraphQLString },
+      time_zone: { type: GraphQLString },
```
sweir27: We might as well make this camelCase as well!

src/schema/v2/sale/__tests__/index.test.js
```diff
@@ -22,6 +22,24 @@ describe("Sale type", () => {
...
+  describe("sale timeZone", () => {
+    const query = `
+    {
+      sale(id: "foo-foo") {
+        timeZone
+      }
+    }`
+
+    it("returns the correct value for time_zone", async () => {
+      sale.timeZone = "America/Chicago"
```
dblandin: The other examples I see in this file still show snake_case getters and setters but in the query, the query field is camelCased, which is a bit confusing.

Maybe @zephraph can clarify the way this should work.

bhoggard:
OK, pushed the `timeZone` change for v1.

src/schema/v2/sale/__tests__/index.test.js
dblandin: @bhoggard Is this resolved? I'm still curious if this should be `sale.time_zone` vs `sale.timeZone`. This file has some conflicting examples.

sweir27: We prefer camelcase (which is why _all_ fields are camelcased in metaphysics v2). We had started to migrate some fields already in v1, but are not consistent (as you're seeing here!).

Making it camelcased now means we'll have one less field to not change later when we migrate, but it's not a huge issue either way.

src/schema/v1/sale/index.ts
```diff
@@ -344,6 +344,7 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       symbol: { type: GraphQLString },
+      timeZone: { type: GraphQLString },
```
sweir27: Does this work? I think it should be:
```suggestion
timeZone: {
  type: GraphQLString,
  resolve: ({ time_zone }) => time_zone,
},
```

sweir27: (Spacing might be off)

src/schema/v2/sale/__tests__/index.test.js
dblandin: Gotcha. Thanks! Sounds good :+1:

src/schema/v1/sale/index.ts
bhoggard: Thanks, you are correct! Pushing a fix now.

PR #1906, opened 2019-08-27
[v2] Remove Viewer type. Relay Modern no longer needs this proxy.

alloy:
Emission PR: TODO

PR #1969, opened 2019-09-25
[V2] Readd Viewer type

damassi:
Related: https://github.com/artsy/metaphysics/pull/1906

Re-adding `Viewer` type during our MPv2 migration in Reaction. We can revisit once the migration is complete.


alloy:
Yep 👍

PR #1972, opened 2019-09-26
Adds StaticContent

kierangillen:
- Adds `staticContent` to resolve page content from Gravity's `Page` model

Example:

```
{
  staticContent(id: "how-auctions-work-bidding") {
    content
    name
  }
}
```

Screen Shot 2019-09-26 at 2 36 54 PM

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -5479,7 +5479,7 @@ type Query {
 
   # Content for a specific page or view
   staticContent(
-    # The slug for the view
+    # The slug or id for the view
     id: String!
   ): StaticContent
 
@@ -6534,6 +6534,14 @@ enum sort {
 }
 
 type StaticContent {
+  # A globally unique ID.
+  id: ID!
+
+  # A slug ID.
+  slug: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
   name: String
   content: String
 }
@@ -6904,7 +6912,7 @@ type Viewer {
 
   # Content for a specific page or view
   staticContent(
-    # The slug for the view
+    # The slug or id for the view
     id: String!
   ): StaticContent
 
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -5477,6 +5477,12 @@ type Query {
     id: String!
   ): Show
 
+  # Content for a specific page or view
+  staticContent(
+    # The slug for the view
+    id: String!
+  ): StaticContent
+
   # Fields related to internal systems.
   system: System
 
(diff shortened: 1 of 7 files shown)


src/schema/v2/static_content.ts
```diff
@@ -0,0 +1,35 @@
...
+  },
+})
+
+const StaticContent: GraphQLFieldConfig<void, ResolverContext> = {
+  type: StaticContentType,
+  description: "Content for a specific page or view",
+  args: {
+    id: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: "The slug for the view",
```
zephraph: Is this _just_ a slug, or does it also support an `id` field?

_schemaV2.graphql
```diff
@@ -6527,6 +6533,11 @@ enum sort {
   ASC
 }
 
+type StaticContent {
```
zephraph: `StaticContent` for v2 should also expose `id`, `internalID`, and `slug`. One sec and a I can point to an example of that.

src/schema/v2/static_content.ts
```diff
@@ -0,0 +1,35 @@
...
+  GraphQLObjectType,
+  GraphQLFieldConfig,
+  GraphQLNonNull,
+  GraphQLString,
+} from "graphql"
+import { ResolverContext } from "types/graphql"
+
+const StaticContentType = new GraphQLObjectType<any, ResolverContext>({
+  name: "StaticContent",
+  fields: {
```
zephraph: You'll want to spread this in fields for the id, slug, internalID, etc. 

https://github.com/artsy/metaphysics/blob/da72816d9999383846ead5aa334676048d0ee6d3/src/schema/v2/artist/index.ts#L89

src/schema/v2/static_content.ts
zephraph: I talked to @ashkan18 and it and he said it could be both. Might be good to reflect that: https://artsy.slack.com/archives/CDR806U7J/p1569526745006400

src/schema/v2/static_content.ts
zephraph: You should actually probably add this to v1 too. That'll add the `__id` field which relay uses for caching.

zephraph:
A few minor things to make sure some of the fields we expect to be there on types are present, but otherwise looks good to me. Once the updates are made I'll merge it. 👍

src/schema/v1/static_content.ts
```diff
@@ -0,0 +1,37 @@
...
+const StaticContentType = new GraphQLObjectType<any, ResolverContext>({
+  name: "StaticContent",
+  fields: {
+    ...SlugAndInternalIDFields,
+    name: {
+      type: GraphQLString,
+    },
+    content: {
+      type: GraphQLString,
+    },
```
alloy: Can these fields ever really be null?

PR #1980, opened 2019-10-04
DO NOT MERGE [PURCHASE-1539: "Ended invalid date" in auction entity header

lilyfromseattle:
# [PURCHASE-1539]: "Ended invalid date" in auction entity header

https://www.notion.so/artsy/Ended-invalid-date-in-auction-entity-header-f981deec4bdd43acbee9febb789806f9

`formattedStartDateTime` was returning "ended Invalid date" when endAt was null. I added and `else if` so that `formattedStartDateTime` returns null when endAt is null. For all other cases behavior will remain the same.

[PURCHASE-1539]: https://artsyproduct.atlassian.net/browse/PURCHASE-1539

### DIFF
--- a/src/lib/date.ts
+++ b/src/lib/date.ts
@@ -262,6 +262,8 @@ export function formattedStartDateTime(startAt, endAt, timezone) {
     return `Starts ${singleDateTime(startAt, timezone)}`
   } else if (thisMoment.isBefore(endMoment)) {
     return `Ends ${singleDateTime(endAt, timezone)}`
+  } else if (endAt === null) {
+    return null
   } else {
     return `Ended ${singleDate(endAt, timezone)}`
   }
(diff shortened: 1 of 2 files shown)


src/lib/date.ts
```diff
@@ -262,6 +262,8 @@ export function formattedStartDateTime(startAt, endAt, timezone) {
     return `Starts ${singleDateTime(startAt, timezone)}`
   } else if (thisMoment.isBefore(endMoment)) {
     return `Ends ${singleDateTime(endAt, timezone)}`
+  } else if (endAt === null) {
```
sweir27: It might be clearer if we reordered this so we check for whether the current time is _after_ the `endMoment` and then if none of those cases match, return null. That way, this catch-all will also work if `startAt` is null, for instance.

i.e.
```javascript
} else if (thisMoment.isAfter(endMoment)) {
  return `Ended ${singleDate(endAt, timezone)}`
} else {
  return null
}

```

PR #1983, opened 2019-10-08
Adds isVerifiedSeller to Partner

sepans:
Fixes [PURCHASE-1574](https://artsyproduct.atlassian.net/browse/PURCHASE-1574)

Screen Shot 2019-10-10 at 3 34 49 PM

Is there a test strategy for these fields? I just found specs for `website` [field](https://github.com/artsy/metaphysics/blob/master/src/schema/v1/__tests__/partner.test.js#L33) and not anything else

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -9034,7 +9034,7 @@ type Partner implements Node {
   website: String
 
   # Indicates the partner is a trusted seller on Artsy
-  commerceVerified: Boolean
+  isVerifiedSeller: Boolean
   analytics: AnalyticsPartnerStats
 }
 
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -9032,6 +9032,9 @@ type Partner implements Node {
 
   # The gallery partner's web address
   website: String
+
+  # Indicates the partner is a trusted seller on Artsy
+  commerceVerified: Boolean
   analytics: AnalyticsPartnerStats
 }
 
--- a/src/schema/v2/partner.ts
+++ b/src/schema/v2/partner.ts
@@ -213,6 +213,11 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      commerceVerified: {
+        description: "Indicates the partner is a trusted seller on Artsy",
+        type: GraphQLBoolean,
+        resolve: ({ commerce_verified }) => commerce_verified,
+      },
     }
   },
 })
--- a/src/schema/v2/partner.ts
+++ b/src/schema/v2/partner.ts
@@ -213,10 +213,10 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
-      commerceVerified: {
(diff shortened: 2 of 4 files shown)


_schema.graphql
```diff
@@ -9032,6 +9032,9 @@ type Partner implements Node {
 
   # The gallery partner's web address
   website: String
+
+  # Indicates the partner is a trusted seller on Artsy
+  commerceVerified: Boolean
```
zephraph: Is there any reason _not_ to name this `trustedSeller`?

sepans: We had a little bit of discussion in the [gravity PR](https://github.com/artsy/gravity/pull/12609#discussion_r332211884) about the name of the flag and it just got carried over to here. You mean renaming it to `trustedSeller` only in MP or both here an gravity/kinetic. Happy to do either if it makes it more clear.

zephraph: I think only in MP. The name doesn't have to match underlying systems IMO, but generally I'd like to keep the MP as self evident as possible. Maybe @alloy has opinions?

alloy: I agree. Maybe it’s my naive English, but “commerce verified” doesn’t really mean anything to me whereas `isTrustedSeller` or `isVerifiedSeller` do. While having docs is always good, the greatest of names don’t _need_ docs.

alloy: (Docs could expand on eg what criteria is considered for a partner to have that status.)

sepans: Yes I also agree. Now that we have the opportunity to also rename it in gravity would you suggest doing that to avoid confusion? May be `verified_seller` in gravity and `isVerifiedSeller` in MP to follow the variable naming conventions in both places? (cc: @ashkan18 since you love naming variables)

ashkan18: haha! Sorry missed this! as suggester of `commerce_verified` i have to say i don't like it either but i didn't have any better suggestion. `verified_seller` sounds better to me though! let me know if I can help in updating name elsewhere.

sepans: renamed in https://github.com/artsy/metaphysics/pull/1983/commits/fb4a117cba5c779b275e0ff09209a6adcddc784c. Also added #dontmerge since it now depends on gravity PR: https://github.com/artsy/gravity/pull/12615

src/schema/v2/partner.ts
```diff
@@ -213,6 +213,11 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      isVerifiediSeller: {
```
alloy:
```suggestion
      isVerifiedSeller: {
```

sepans: 🙌

sepans: 🙌 @alloy. updated https://github.com/artsy/metaphysics/pull/1983/commits/3075870c4f55fad53dd0adaad0324eaa11aafa99

ashkan18:
thanks! merging 🙌

PR #1985, opened 2019-10-09
Update formattedStartDateTime to account for live sales and fix bugs

lilyfromseattle:
Semi-related to [PURCHASE-1584] and [PURCHASE-1539]

This will allow the context card to show messaging about auction start/end times that will be consistent with the auction timer (I have another PR in emission to update that). I also fixed a bug where the string was coming back with "Invalid date" in some cases. I also made the dateTimes timezone specific but defaulting to UTC, whereas before they were always in UTC. 

__Before:__
Screen Shot 2019-10-09 at 3 11 57 PM

__After:__
Screen Shot 2019-10-09 at 3 12 17 PM

__After (when it gets timezone from the client):__
Screen Shot 2019-10-09 at 4 29 15 PM

[PURCHASE-1584]: https://artsyproduct.atlassian.net/browse/PURCHASE-1584
[PURCHASE-1539]: https://artsyproduct.atlassian.net/browse/PURCHASE-1539

### DIFF
--- a/src/lib/date.ts
+++ b/src/lib/date.ts
@@ -254,15 +254,27 @@ export function formattedOpeningHours(startAt, endAt, timezone) {
  * Ends Apr 3 at 12:30pm
  * Ended Apr 3 2017
  */
-export function formattedStartDateTime(startAt, endAt, timezone) {
-  const thisMoment = moment()
+export function formattedStartDateTime(startAt, endAt, liveStartAt, timezone) {
+  const thisMoment = moment.tz(moment(), timezone)
   const startMoment = moment.tz(startAt, timezone)
   const endMoment = moment.tz(endAt, timezone)
+  const liveStartMoment = moment.tz(liveStartAt, timezone)
+
   if (thisMoment.isBefore(startMoment)) {
     return `Starts ${singleDateTime(startAt, timezone)}`
+  } else if (liveStartAt && thisMoment.isBefore(liveStartMoment)) {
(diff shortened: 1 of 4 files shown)


src/lib/date.ts
```diff
@@ -254,15 +254,27 @@ export function formattedOpeningHours(startAt, endAt, timezone) {
  * Ends Apr 3 at 12:30pm
  * Ended Apr 3 2017
  */
-export function formattedStartDateTime(startAt, endAt, timezone) {
-  const thisMoment = moment()
+export function formattedStartDateTime(startAt, endAt, liveStartAt, timezone) {
```
ashkan18: this method seems pretty generic, adding `liveStartAt` makes it very "auction" related should we make a separate method that possibly uses this method and extends it with `liveStartAt`?

lilyfromseattle: If liveStartAt is null it should resolve to the same strings as it did before. I also checked and confirmed that this is only used in emission for auction times.

ashkan18: cool, thanks for confirming. I would still rename it to `auctionFormattedDateTime` to reflect the auction specific logic in there since this method is defined in pretty generic area `src/lib/date.ts`. what do you think?

lilyfromseattle: I discussed that with Sarah, but because prod still depends on the other name we thought it was more trouble than its worth. Also it still works for non-auctions.

ashkan18: 👍

PR #1988, opened 2019-10-14
 Add `hasCertificateOfAuthenticity`

sepans:
discussion about adding this field in addition to `certificateOfAuthenticity`:[here](https://github.com/artsy/reaction/pull/2896/files#r334499431)

### DIFF
--- a/src/schema/v1/artwork/index.ts
+++ b/src/schema/v1/artwork/index.ts
@@ -919,8 +919,17 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      hasCertificateOfAuthenticity: {
+        type: GraphQLBoolean,
+        description:
+          "Returns true when artwork has certificate of authenticity",
+        resolve: ({ certificate_of_authenticity }) =>
+          certificate_of_authenticity === true,
+      },
       certificateOfAuthenticity: {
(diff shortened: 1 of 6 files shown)


src/schema/v1/artwork/index.ts
```diff
@@ -919,8 +919,17 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      hasCertificateOfAuthenticity: {
+        type: GraphQLBoolean,
+        description:
+          "Returns true when artwork has certificate of authenticity",
```
zephraph:
```suggestion
          "Returns true when artwork has a certificate of authenticity",
```

src/schema/v2/artwork/index.ts
```diff
@@ -812,8 +812,17 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      hasCertificateOfAuthenticity: {
+        type: GraphQLBoolean,
+        description:
+          "Returns true when artwork has certificate of authenticity",
```
zephraph:
```suggestion
          "Returns true when artwork has a certificate of authenticity",
```

zephraph:
Some minor updates you can accept if you want, but overall looks good 👍

PR #1991, opened 2019-10-14
Add ids argument to SaleArtworkConnections

yuki24:
Basically the same change as https://github.com/artsy/metaphysics/pull/1982 but to v1.


_schema.graphql
```diff
@@ -1617,6 +1617,8 @@ type ArtworkContextAuction implements Node {
     all: Boolean = false
   ): [SaleArtwork]
   sale_artworks_connection(
+    # List of sale artwork IDs to fetch
+    ids: [String]
```
zephraph: Small nit, shouldn't this be `ids: [ID]`?

yuki24: All existing `*_ids` arguments are `[String]`, but perhaps we could start using `[ID]`?

cc/ @alloy

alloy: Yes please 👌

yuki24: Okay, updated the PR and now they are using `[ID]`.

PR #1995, opened 2019-10-17
Fix not honoring ended_at for live auctions

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-1598

# Cause
We were not honoring `ended_at` field of the auction and we were only using `end_at`.

`end_at` can be `null` and `ended_at` is for live auctions where we store the actual time the auction ended.

# Solution
When calling `formattedStartDateTime` use `ended_at` and if it's null, use `end_at`.

### DIFF
--- a/src/lib/date.ts
+++ b/src/lib/date.ts
@@ -262,18 +262,23 @@ export function formattedStartDateTime(startAt, endAt, liveStartAt, timezone) {
 
   if (thisMoment.isBefore(startMoment)) {
     return `Starts ${singleDateTime(startAt, timezone)}`
-  } else if (liveStartAt && thisMoment.isBefore(liveStartMoment)) {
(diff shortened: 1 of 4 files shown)


sweir27:
This looks great!! Is there a test case that we should add/update to capture this case?

ashkan18:
yes looks like we could add some tests similar to here  https://github.com/artsy/metaphysics/blob/678bd7df6c3ea35b87a7d647b0cb1da8313e8448/src/schema/v2/sale/__tests__/index.test.js#L524 for `formattedStartDate`

ashkan18:
alright all sort of tests added in query level, this makes me feel better for sure!

ashkan18:
merging this since @sweir27 is OOO and we added tests for this.

PR #480, opened 2016-11-26
@craigspaeth => use float so we get 64 bits

sweir27:
High estimates for auctions are running above 25M HKD, which is larger than 32 bits and thus borking GraphQL.

PR #2004, opened 2019-10-18
Add CalculatedCost type to SaleArtwork

yuki24:
finishes https://artsyproduct.atlassian.net/browse/AUCT-671

This PR adds a new `calculatedCost` field to the `SaleArtwork` type. The actual logic that does the math isn't implemented yet, but this should allow us to work on GraphQL clients since they need the schema to be updated.

## Query

Screen Shot 2019-10-18 at 1 05 56 PM

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -3258,7 +3258,7 @@ type BuyOrder implements Order {
 }
 
 type CalculatedCost {
-  buyers_premium: BuyersPremiumAmount
+  buyersPremium: BuyersPremiumAmount
   subtotal: SubtotalAmount
 }
 
@@ -10606,9 +10606,9 @@ type SaleArtwork {
   reserve_status: String
   sale_id: String
   sale: Sale
-  calculated_cost(
+  calculatedCost(
     # Max bid price for the sale artwork
-    bid_amount_cents: Int
+    bidAmountCents: Int
   ): CalculatedCost
 
   # Currency symbol (e.g. "$")
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -3028,6 +3028,25 @@ type BuyersPremium {
   percent: Float
 }
 
+type BuyersPremiumAmount {
+  # A formatted price with various currency formatting options.
+  amount(
+    decimal: String = "."
+
+    # Allows control of symbol position (%v = value, %s = symbol)
+    format: String = "%s%v"
+    precision: Int = 0
+    symbol: String
+    thousand: String = ","
+  ): String
+
+  # An amount of money expressed in cents.
+  cents: Float
+
+  # A pre-formatted price.
+  display: String
+}
+
 type BuyOrder implements Order {
   # A globally unique ID.
   __id: ID!
@@ -3236,6 +3255,11 @@ type BuyOrder implements Order {
 
   # Buyer phone number
   buyerPhoneNumber: String
+}
+
+type CalculatedCost {
+  buyers_premium: BuyersPremiumAmount
+  subtotal: SubtotalAmount
 }
 
 enum CancelReasonType {
@@ -10582,6 +10606,10 @@ type SaleArtwork {
   reserve_status: String
   sale_id: String
   sale: Sale
+  calculated_cost(
+    # Max bid price for the sale artwork
+    bid_amount_cents: Int
+  ): CalculatedCost
 
   # Currency symbol (e.g. "$")
   symbol: String
@@ -11507,6 +11535,25 @@ input submitPendingOfferInput {
 type submitPendingOfferPayload {
   orderOrError: OrderOrFailureUnionType
   clientMutationId: String
+}
+
+type SubtotalAmount {
+  # A formatted price with various currency formatting options.
+  amount(
+    decimal: String = "."
+
+    # Allows control of symbol position (%v = value, %s = symbol)
+    format: String = "%s%v"
+    precision: Int = 0
+    symbol: String
+    thousand: String = ","
+  ): String
+
+  # An amount of money expressed in cents.
+  cents: Float
+
+  # A pre-formatted price.
+  display: String
 }
 
 type System {
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1496,6 +1496,30 @@ type BuyersPremium {
   percent: Float
 }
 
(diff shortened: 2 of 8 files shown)


src/schema/v1/sale_artwork.ts
```diff
@@ -370,6 +371,31 @@ export const SaleArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+              cents: bid_amount_cents * 0.2,
+              display: `$${((bid_amount_cents * 0.2) / 100)
+                .toFixed(2)
+                .replace(/\d(?=(\d{3})+\.)/g, "$&,")}`,
+            },
+            subtotal: {
+              cents: bid_amount_cents * 1.2,
+              display: `$${((bid_amount_cents * 1.2) / 100)
+                .toFixed(2)
+                .replace(/\d(?=(\d{3})+\.)/g, "$&,")}`,
```
yuki24: Just hard-coding the amount and doing a simple math so the client can actually see that the value gets updated based on the argument, but I'm not really trying to optimize this code since we will replace it soon.

src/schema/v1/sale_artwork.ts
```diff
@@ -370,6 +371,31 @@ export const SaleArtworkType = new GraphQLObjectType<any, ResolverContext>({
           return saleLoader(sale_id)
         },
       },
+      calculated_cost: {
```
dblandin: The v1 schema has a mix between camelCase and snake_case field names. I think we should stick to the camelCase preference for new additions and it'll also be easier on the eventual V2 upgrade.

yuki24: @alloy @zephraph Should we use camel case in v1? What do you think?

_schemaV2.graphql
```diff
@@ -1496,6 +1496,30 @@ type BuyersPremium {
...
+
+    # Allows control of symbol position (%v = value, %s = symbol)
+    format: String = "%s%v"
+    precision: Int = 0
+    symbol: String
+    thousand: String = ","
+  ): String
+
+  # An amount of money expressed in cents.
+  cents: Float
```
dleve123: Just to discuss something that @yuki24 and I discussed IRL – can we make this an `Int` as cents aren't really divisible?

src/schema/v1/sale_artwork.ts
zephraph: If it's new, yes.

src/schema/v1/sale_artwork.ts
dleve123: Small thing: to save the reader needing to look up the definition of this regex, thoughts on extracting a named variable for it? I think it would make the code a bit more clear.

src/schema/v2/sale_artwork.ts
```diff
@@ -316,6 +317,31 @@ export const SaleArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+            description: "Max bid price for the sale artwork",
+          },
+        },
+        resolve: (_params, { bidAmountCents }, _loaders) => {
+          return {
+            buyersPremium: {
+              cents: bidAmountCents * 0.2,
+              display: `$${((bidAmountCents * 0.2) / 100)
+                .toFixed(2)
+                .replace(/\d(?=(\d{3})+\.)/g, "$&,")}`,
```
dleve123: Similar comment to the above

_schemaV2.graphql
dleve123: I think @yuki24 was trying to explain this to me IRL (and I maybe didn't perfectly understand), but seems like this is a a byproduct of using types generated by the [`money` function](https://github.com/artsy/metaphysics/blob/4a81e2eed5ae9771352b45283e218cb6c4ce92cd/src/schema/v2/fields/money.ts#L77-L103)...I wonder if there's some context...and there is, we use float to handle very large integers that are larger than 64 bits: https://github.com/artsy/metaphysics/pull/480

src/schema/v1/sale_artwork.ts
yuki24: The next time we touch this code we will probably just remove it entirely rather than understanding what it does. All we have to know is it does some kind of formatting to make fake data close to reality, and I don't think it is worth extracting.

AUCT-671

dleve123: Fair enough!

_schema.graphql
```diff
@@ -11473,6 +11501,25 @@ type submitPendingOfferPayload {
...
+    symbol: String
+    thousand: String = ","
+  ): String
+
+  # An amount of money expressed in cents.
+  cents: Float
+
+  # A pre-formatted price.
+  display: String
+}
```
erikdstock: Given that this is repeated twice can we do something like `type SuptotalAmount implements Money` ?

https://graphql.org/learn/schema/#interfaces

src/schema/v1/sale_artwork.ts
alloy: Aye, all new fields should be camelCase, regardless of schema version. We already started that before v2.

yuki24: Awesome, done: https://github.com/artsy/metaphysics/pull/2004/commits/656c439fb50035dde2355b8ac7ccb5c745d664f7

_schema.graphql
yuki24: I've also had a chat about this with @dleve123 - this could be a shared `Amount` type rather than defining a type for each amount field. In fact, that's what I did in the first place and later I noticed there is a `money` helper function so I switched to it.

I think there are definitely questions around this or space for improvement. However, that seems like more a foundational change and I'd like a separate thread for that.

yuki24:
Alright, I've converted snake case keys to camel keys, and I think this is in a good state. There are some questions around the `Amount`-related types, but I think that should be a separate conversation. I'll go ahead and merge this PR.

_schema.graphql
zephraph: Here's a PR that introduces a `Price` type which ultimately this should be replaced with. 

In general it's a bad practice to _copy_ a type. I'm going to recommend to platform that we look into tooling to help with this. cc @alloy

alloy: Tooling to solve what problem specifically?

PR #2019, opened 2019-10-24
Rename Price fields in ListPrice to be Money

zephraph:
Somehow I fudged on #2015 and merged some failing tests. 

Not sure how I missed that. This should fix them though.


kierangillen:
🤑 💸

PR #2021, opened 2019-10-25
Update SaleArtwork#calculatedCost to run calculation via Gravity

dblandin:
Update the `calculatedCost` field on sale artworks to pull calculation
results from the Gravity API.

The following subfields are exposed:

- `bidAmount`: Echos the input provided by the client along with formatted
  results
- `buyersPremium`: The calculated buyer's premium for the bid amount
  provided
- `subtotal`: The combination of the supplied bid amount and the
  calculated buyer's premium

_Note: This does break the API contract defined along with the mock data. I think it's worth making this change now and early to align with defined naming preferences._

**Request**

```graphql
{
  sale(id: "finarte-photographs-4") {
    saleArtwork(id: "franco-fontana-paesaggio-baia-delle-zagare") {
      calculatedCost(bidAmountCents: 100000) {
        bidAmount {
          display
          minor
          currencyCode
          major
        }
        buyersPremium {
          display
          minor
          currencyCode
          major
        }
        subtotal {
          display
          minor
          currencyCode
          major
        }
      }
    }
  }
}
```

**Response**

```json

{
  "data": {
    "sale": {
      "saleArtwork": {
        "calculatedCost": {
          "bidAmount": {
            "display": "€100,000.00",
            "minor": 10000000,
            "currencyCode": "EUR",
            "major": 100000
          },
          "buyersPremium": {
            "display": "€19,237.20",
            "minor": 1923720,
            "currencyCode": "EUR",
            "major": 19237.2
          },
          "subtotal": {
            "display": "€119,237.20",
            "minor": 11923720,
            "currencyCode": "EUR",
            "major": 119237.2
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/src/lib/loaders/loaders_without_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_without_authentication/gravity.ts
@@ -67,6 +67,10 @@ export default opts => {
     saleArtworksFilterLoader: gravityLoader("filter/sale_artworks"),
     saleArtworksLoader: gravityLoader(id => `sale/${id}/sale_artworks`, {}, { headers: true }),
     saleArtworkLoader: gravityUncachedLoader<any, { saleId: string, saleArtworkId: string }>(({ saleId, saleArtworkId }) => `sale/${saleId}/sale_artwork/${saleArtworkId}`, null),
+    saleArtworkCalculatedCostLoader: gravityLoader<any, { saleId: string, saleArtworkId: string, bidAmountMinor: number }>(
+      ({ saleId, saleArtworkId, bidAmountMinor }) =>
+        `sale/${saleId}/sale_artwork/${saleArtworkId}/calculated_cost?bid_amount_cents=${bidAmountMinor}`
+    ),
     saleLoader: batchSaleLoader,
     salesLoader: batchSalesLoader,
     salesLoaderWithHeaders: gravityLoader('sales', {}, { headers: true }),
--- a/src/lib/loaders/loaders_without_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_without_authentication/gravity.ts
@@ -14,6 +14,16 @@ export default opts => {
     multipleLoader: gravityLoader<{ id: string, is_auction: boolean }[]>('sales')
   })
 
+  type GravityCalculatedCostResponse = {
+    display_bid_amount: string
+    bid_amount_cents: number
+    buyers_premium_cents: number
+    display_buyers_premium: string
+    subtotal_cents: number
+    display_subtotal: string
+    currency: string
+  }
+
   return {
     artistArtworksLoader: gravityLoader(id => `artist/${id}/artworks`),
     artistGenesLoader: gravityLoader(id => `artist/${id}/genome/genes`),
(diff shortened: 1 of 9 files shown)


src/lib/loaders/loaders_without_authentication/gravity.ts
```diff
@@ -67,6 +67,10 @@ export default opts => {
     saleArtworksFilterLoader: gravityLoader("filter/sale_artworks"),
     saleArtworksLoader: gravityLoader(id => `sale/${id}/sale_artworks`, {}, { headers: true }),
     saleArtworkLoader: gravityUncachedLoader<any, { saleId: string, saleArtworkId: string }>(({ saleId, saleArtworkId }) => `sale/${saleId}/sale_artwork/${saleArtworkId}`, null),
+    saleArtworkCalculatedCostLoader: gravityLoader<any, { saleId: string, saleArtworkId: string, bidAmountMinor: number }>(
```
dleve123: @dblandin I recall @alloy [mentioning](https://artsy.slack.com/archives/CP9P4KR35/p1572287779181400?thread_ts=1572284881.180100&cid=CP9P4KR35) that this first type parameter could be used to define a type on the Gravity response.

Thoughts on this?

src/schema/v2/sale_artwork.ts
```diff
@@ -318,26 +318,39 @@ export const SaleArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+            bidAmountMinor,
+          })
+
           return {
             buyersPremium: {
-              cents: bidAmountCents * 0.2,
-              display: `$${((bidAmountCents * 0.2) / 100)
-                .toFixed(2)
-                .replace(/\d(?=(\d{3})+\.)/g, "$&,")}`,
+              cents: data.buyers_premium_cents,
```
dleve123: This seems a bit surprising to me. How is it that the tests assert upon the API breaking change of cents/dollars -> minor/major yet we're still returning cents here? Is there some additional layer of data conversation to major/minor?

dleve123: Ah perhaps this seems to be something to due with the `Money` type expecting `cents` as input. Bit mysterious to me perhaps, probably this is the pattern 👍

src/lib/loaders/loaders_without_authentication/gravity.ts
dblandin: Good call! Added. :+1:

src/schema/v2/sale_artwork.ts
dblandin: Yep, that's right. The Money type handles that conversion.

dblandin:
> @yuki24 will have the most context on how to handle this breaking change for the latent (feature flag off) Emission code.

As @yuki24 pointed out [in Slack](https://artsy.slack.com/archives/CA3D02SDN/p1572644874117300?thread_ts=1572641655.113300&cid=CA3D02SDN), Emission is only using the `display` field.

```graphql
buyersPremium {
  display
}
subtotal {
  display
}
```

https://github.com/artsy/emission/blob/2bfaf9330488c575ba090514ed5f46d74049bb56/src/lib/Components/Bidding/Screens/ConfirmBid/PriceSummary.tsx#L59-L64

There is a breaking change in the argument though:

```graphql
bidAmountMinor: Int!
```

We changed the argument name from `bidAmountCents` to `bidAmountMinor` and it's now a required argument. I checked with @yuki24 and since this is still behind a feature flag and unreleased, we're going to roll with this change and follow-up with an update in Emission.

PR #2037, opened 2019-11-05
[v2] Add PageCursors to ArtworkConnectionInterface

alloy:
First one that feels comfortable can take this PR and merge 👍 

cc @mzikherman

### DIFF
--- a/src/schema/v2/fields/pagination.ts
+++ b/src/schema/v2/fields/pagination.ts
@@ -164,7 +164,7 @@ export function connectionWithCursorInfo(
     ...config,
     connectionFields: {
       pageCursors: {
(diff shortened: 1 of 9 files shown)


src/schema/v2/fields/pagination.ts
```diff
@@ -164,7 +164,7 @@ export function connectionWithCursorInfo(
     ...config,
     connectionFields: {
       pageCursors: {
-        type: PageCursorsType,
+        type: new GraphQLNonNull(PageCursorsType),
```
alloy: @mzikherman I made this non-null, because I believe it can never be null.

PR #2040, opened 2019-11-05
[v2] Add KAWS stitching of filterArtworksConnection

alloy:
Alright, got KAWS stitching in V2 and use the filter artworks connection instead. I also automated the schema extension so that it’s always up-to-date with the filter args defined in MP.


alloy:
Gah, on my phone. Can you take a look at the test failure?

PR #2047, opened 2019-11-07
Make sure vortex stitching works with v2

sweir27:
This PR allows us to use vortex stitching with metaphysics v2.

I opted for the "quick and dirty" approach here (duplicating our stitching manifest vs. adding special logic to allow the same stitching manifest to work for both).

Once reaction and volt are migrated (I imagine the latter is probably not _too_ involved), it will be really easy to rip out the `v1` code.

The tests are exact copies of the v1 ones, except they reference `runQuery` from the `v2` folder.


alloy:
Works for me 👌

PR #2048, opened 2019-11-08
Remove v2 schema transforms

ds300:
Doing this for a few reasons:

- `id` fields were being removed from kaws and exchange entities but they were previously used in reaction for implicit cache updates, so that broke certain things (e.g. redirect logic in the order app)
- The other transforms were not needed. We had one that removed an unused type from the gravity schema but we decided that it would be better to leave that noise in the schema for now and remove the transforms entirely than to have to maintain the transforms. Usage of that type is heavily discouraged through field/type naming so not much of a risk there.


alloy:
Perfect!

#mergeongreen

PR #2049, opened 2019-11-08
Partner.locationsConnection & Partner.cities

kierangillen:
- Adds `Partner.locationsConnection`, `Partner.locations` is now removed and had not been used prior to the future `6.0` release.
- Adds `Partner.cities`, in cases we were using the original `Partner.locations` it was just cases where we need a list of the `cities`, this new resolver provides a unique array of the Partners' cities. 

Screen Shot 2019-11-08 at 12 22 34 PM

Screen Shot 2019-11-08 at 12 19 10 PM

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -4685,6 +4685,37 @@ type Location {
   summary: String
 }
 
+# A connection to a list of items.
+type LocationConnection implements LocationsConnectionInterface {
+  # Information to aid in pagination.
+  pageInfo: PageInfo!
+
+  # A list of edges.
+  edges: [LocationEdge]
+  pageCursors: PageCursors!
+  totalCount: Int
+}
+
+# An edge in a connection.
+type LocationEdge implements LocationsEdgeInterface {
+  # The item at the end of the edge
+  node: Location
+
+  # A cursor for use in pagination
+  cursor: String!
+}
+
+interface LocationsConnectionInterface {
+  pageCursors: PageCursors!
+  pageInfo: PageInfo!
+  edges: [LocationsEdgeInterface]
+}
+
+interface LocationsEdgeInterface {
+  node: Location
+  cursor: String
+}
+
 type LotStanding {
   # Your bid if it is currently winning
   activeBid: BidderPosition
@@ -5353,6 +5384,9 @@ type Partner implements Node {
   categories: [PartnerCategory]
   collectingInstitution: String
   counts: PartnerCounts
+
+  # A list of the partners unique city locations
+  cities(size: Int = 25): [String]
   defaultProfileID: String
   hasFairPartnership: Boolean
   href: String
@@ -5360,7 +5394,14 @@ type Partner implements Node {
   isDefaultProfilePublic: Boolean
   isLinkable: Boolean
   isPreQualify: Boolean
-  locations(size: Int = 25): [Location]
+
+  # A connection of locations from a Partner.
+  locationsConnection(
+    after: String
+    first: Int
+    before: String
+    last: Int
+  ): LocationConnection
   name: String
   profile: Profile
 
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -4686,7 +4686,7 @@ type Location {
 }
 
 # A connection to a list of items.
-type LocationConnection implements LocationsConnectionInterface {
+type LocationConnection {
   # Information to aid in pagination.
   pageInfo: PageInfo!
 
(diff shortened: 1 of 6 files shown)


src/schema/v2/location.ts
```diff
@@ -119,9 +123,34 @@ export const LocationType = new GraphQLObjectType<any, ResolverContext>({
...
+    pageInfo: { type: new GraphQLNonNull(PageInfoType) },
+    edges: { type: new GraphQLList(LocationsEdgeInterface) },
+  },
+})
+
+export const locationsConnection = connectionWithCursorInfo({
+  nodeType: LocationType,
+  connectionInterfaces: [LocationsConnectionInterface],
+  edgeInterfaces: [LocationsEdgeInterface],
+})
```
alloy: Are you really in need of these interfaces in client code?

src/schema/v2/partner.ts
```diff
@@ -183,6 +183,35 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
+                  return (
+                    city &&
+                    locationCities.indexOf(city) === pos &&
+                    city.length > 0
+                  )
+                }
+              )
+              return filteredForDuplicatesAndBlanks
+            })
+            .catch(err => console.log(err))
```
alloy: Perhaps better to let these errors bubble up?

kierangillen: 👍

src/schema/v2/partner.ts
```diff
@@ -212,16 +241,43 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
+            args
+          )
+
+          interface GravityArgs {
+            exclude_ids?: string[]
+            page: number
+            published: boolean
+            size: number
+            total_count: boolean
+          }
```
alloy: It’s great that you define types for these! But perhaps define them on the `partnerLocationsLoader` instead?

src/schema/v2/location.ts
```diff
@@ -119,9 +123,34 @@ export const LocationType = new GraphQLObjectType<any, ResolverContext>({
   }),
 })
 
-const Location: GraphQLFieldConfig<void, ResolverContext> = {
+export const Location: GraphQLFieldConfig<void, ResolverContext> = {
```
alloy: Rename this to `LocationField`, we started calling these what they are and the same applies to types being called e.g. `LocationType`.

_schemaV2.graphql
```diff
@@ -5353,14 +5384,24 @@ type Partner implements Node {
   categories: [PartnerCategory]
   collectingInstitution: String
   counts: PartnerCounts
+
+  # A list of the partners unique city locations
+  cities(size: Int = 25): [String]
```
alloy: Is `cities` actually used and is the city data not also provided by the `locationsConnection`?

src/schema/v2/partner.ts
alloy:
```ts
partnerLocationsLoader: gravityLoader<
  { ... }, // response type, if helpful
  { exclude_ids?: string[], … }, // params type
>(id => `partner/${id}/locations`, {}, { headers: true }),
```

_schemaV2.graphql
kierangillen: https://github.com/artsy/emission/blob/master/src/lib/Scenes/Artwork/Components/PartnerCard.tsx#L100 There are a couple cases in Emission like this one where we just need a list of the cities. In this view we wouldn't need to do a pagination query if we can just query for the `Partner.cities`, also we can remove a lot of logic and this util completely - https://github.com/artsy/emission/blob/master/src/lib/utils/filterLocations.tsx

src/schema/v2/partner.ts
kierangillen: Don't the params then need to be spread in the url? This isn't the convention we've been using. I'll take a look into it.

src/schema/v2/location.ts
kierangillen: 👍

src/schema/v2/location.ts
kierangillen: Opting to remove these interfaces.

src/schema/v2/partner.ts
kierangillen: OK moved these types!

src/schema/v2/show.ts
```diff
@@ -414,7 +414,7 @@ export const ShowType = new GraphQLObjectType<any, ResolverContext>({
       location: {
         description:
           "Where the show is located (Could also be a fair location)",
-        type: Location.type,
+        type: LocationField.type,
```
alloy: Export the type separately instead and use that here, because the `type` of a `GraphQLFieldConfig` is a wider union than the exact type that `LocationType` has.

_schemaV2.graphql
```diff
@@ -4685,6 +4685,37 @@ type Location {
   summary: String
 }
 
+# A connection to a list of items.
+type LocationConnection implements LocationsConnectionInterface {
```
alloy: Looks like the schema needs to be re-generated to remove these interfaces?

_schemaV2.graphql
alloy: Hmm, ok, but in that case does this really need a `size` argument?

_schemaV2.graphql
kierangillen: Regenerated.

src/schema/v2/show.ts
kierangillen: Updated to just pass the type here.

alloy:
The `cities` stuff can be followed-up on :+1:

PR #2053, opened 2019-11-12
Remove unused Image type from kaws schema

xtina-starr:
Related to https://github.com/artsy/kaws/pull/172 this removes the unused `Image` type from the KAWS schema.


alloy:
This looks good, but let’s wait with merging until after the Reaction migration QA.

alloy:
Ok, let’s do this

PR #2054, opened 2019-11-12
Adds a test to verify persited queries against the schema

zephraph:
This is a test to help prevent breaking changes to emission for fields that are currently being used (like with what happened for https://github.com/artsy/metaphysics/pull/2049)

The caveat to this test is that we only make additive changes to the query map so there's a lot of queries that are incompatible with the current schema. These are _likely_ v1 changes or early v2 changes. 

Anyway, I just added a known failures file with all the queries that are currently failing.


ashfurrow:
Awesome! I like the idea of a "known failures" list, since that lets us be strategic about future breaking changes. Good stuff 👍

sweir27:
Nice! Out of curiosity, what percentage of queries are in that `knownFailures` list?

zephraph:
@sweir27 when I last checked there were 1336 overall persisted queries. There are ~950 known failures which means only ~30% of them are passing.

zephraph:
@alloy that'll be a little tricky, but I'll play around with it.

zephraph:
Okay, so now it's testing against v1 _and_ v2. Only the v1 queries that were actually passing against V1 will actively be tested. There's a separate ignore list for v1 queries too.

alloy:
Ace, thanks–let’s give this a go! :shipit:

PR #2087, opened 2019-12-03
Add currency conversion to Money type

anandaroop:
https://artsyproduct.atlassian.net/browse/GROW-1679

This gives `Money` fields the ability to do a query-time conversion to USD dollars:

Screen Shot 2019-12-04 at 3 37 16 PM

 The conversion is based on the exchange rate data held by Gravity, which is updated nightly and served from `/api/v1/exchange_rates`.

Because it is based on Gravity's exchange rate data, conversion rates may be up to a day old, which should be fine for our present purposes.

This is my and @jiahaodavid's first time mucking around with currency fields. We are open to suggestion about field naming, proper use of major/minor etc.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -5119,10 +5119,10 @@ type Money {
   display: String
 
   # An amount of money expressed in major units (like dollars).
-  major: Float!
-
-  # An amount of money converted to USD major units (dollars).
-  toUSD: Float!
+  major(
+    # ISO-4217 code of a destination currency for conversion
+    convertTo: String
+  ): Float!
 }
 
 type Mutation {
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -5120,6 +5120,9 @@ type Money {
 
   # An amount of money expressed in major units (like dollars).
   major: Float!
+
+  # An amount of money converted to USD major units (dollars).
+  toUSD: Float!
 }
 
 type Mutation {
--- a/src/lib/loaders/loaders_without_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_without_authentication/gravity.ts
@@ -38,7 +38,11 @@ export default opts => {
     artworkLoader: gravityLoader(id => `artwork/${id}`),
     artworksLoader: gravityLoader("artworks"),
     bidderLoader: gravityLoader(id => `bidder/${id}`),
-    exchangeRatesLoader: gravityLoader("exchange_rates"),
+    exchangeRatesLoader: gravityLoader(
+      "exchange_rates",
+      {},
+      { requestThrottleMs: 1000 * 60 * 60 }
+    ),
     fairArtistsLoader: gravityLoader(
       id => `fair/${id}/artists`,
       {},
--- a/src/lib/loaders/loaders_without_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_without_authentication/gravity.ts
@@ -38,6 +38,7 @@ export default opts => {
     artworkLoader: gravityLoader(id => `artwork/${id}`),
     artworksLoader: gravityLoader("artworks"),
     bidderLoader: gravityLoader(id => `bidder/${id}`),
+    exchangeRatesLoader: gravityLoader("exchange_rates"),
     fairArtistsLoader: gravityLoader(
       id => `fair/${id}/artists`,
       {},
--- a/src/schema/v2/fields/__tests__/money.test.ts
+++ b/src/schema/v2/fields/__tests__/money.test.ts
@@ -151,7 +151,7 @@ describe("major(convertTo:)", () => {
     })
   })
 
-  it("only supports USD", () => {
+  xit("only supports USD", () => {
     const context = mockArtworkContext({
       id: "some-european-artwork",
       price_currency: "EUR",
@@ -170,7 +170,7 @@ describe("major(convertTo:)", () => {
       }
     `
 
-    expect(runQuery(query, context)).rejects.toThrowError(/*Only USD*/)
+    expect(runQuery(query, context)).rejects.toThrowError(/Only USD/)
   })
 
   it("returns null for missing prices", () => {
--- a/src/schema/v2/fields/__tests__/money.test.ts
+++ b/src/schema/v2/fields/__tests__/money.test.ts
@@ -72,14 +72,8 @@ describe(amount, () => {
 })
 
 describe("toUSD", () => {
-  it("converts an exact price from native currency to USD dollars", () => {
-    const mockArtwork = {
-      id: "some-european-artwork",
-      price_currency: "EUR",
-      price_cents: [1000],
-    }
-
-    const context = {
+  const mockArtworkContext = mockArtwork => {
+    return {
       artworkLoader: () => {
         return Promise.resolve(mockArtwork)
       },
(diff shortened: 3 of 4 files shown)


_schemaV2.graphql
```diff
@@ -5120,6 +5120,9 @@ type Money {
 
   # An amount of money expressed in major units (like dollars).
   major: Float!
+
+  # An amount of money converted to USD major units (dollars).
+  toUSD: Float!
```
mzikherman: Wondering if rather than an extra field, `major` and `minor` can get arguments passed to it which might return USD? like `major(convertTo: "USD")` or something.

src/lib/loaders/loaders_without_authentication/gravity.ts
```diff
@@ -38,6 +38,7 @@ export default opts => {
     artworkLoader: gravityLoader(id => `artwork/${id}`),
     artworksLoader: gravityLoader("artworks"),
     bidderLoader: gravityLoader(id => `bidder/${id}`),
+    exchangeRatesLoader: gravityLoader("exchange_rates"),
```
mzikherman: This might benefit from being cached in Metaphysics for a while (since these only change daily). That can be done with invoking: `gravityLoader(path, {}, { requestThrottleMs: 1000 * 60 * 60 * 6 }` to cache for 6 hours, for instance.

src/schema/v2/fields/money.ts
```diff
@@ -137,6 +137,18 @@ export const Money = new GraphQLObjectType<any, ResolverContext>({
...
       },
     },
+    toUSD: {
+      type: new GraphQLNonNull(GraphQLFloat),
+      description: "An amount of money converted to USD major units (dollars).",
+      resolve: async ({ cents, currency }, {}, { exchangeRatesLoader }) => {
+        const factor = currencyCodes[currency.toLowerCase()].subunit_to_unit
+        const major = cents / factor
+
+        const exchangeRates = await exchangeRatesLoader()
```
mzikherman: On the one hand, I'm thinking that while we generally don't want to swallow errors it seems as though an error fetching an exchange rate, or an exchange rate not being found for a particular currency, should return `null` vs. an error?

However, with recent work in Reaction/Force to not crash the app based on _any_ error resolving a leaf node of a GraphQL query, vs. just a very short list of crash-able errors, that may be less important (ie- our UI's _should_ be resilient to these kinds of errors).

src/schema/v2/fields/money.ts
```diff
@@ -137,6 +137,18 @@ export const Money = new GraphQLObjectType<any, ResolverContext>({
         return cents / factor
       },
     },
+    toUSD: {
+      type: new GraphQLNonNull(GraphQLFloat),
+      description: "An amount of money converted to USD major units (dollars).",
+      resolve: async ({ cents, currency }, {}, { exchangeRatesLoader }) => {
```
mzikherman: This resolver might have a special case for amounts already denominated in `USD` to avoid a fetch for the exchange rates.

_schemaV2.graphql
anandaroop: Nice! That sounds like a better and more future-proof api design.

I took your suggestion, but with two caveats:
- Only applied to `major`, since major-to-major conversion is the only supported use case now
- Raises an error if you try to convert to something other than USD, since that's the only currency we need to support presently (and that's what the exchange rate tables are based on). This field design should allow this to be expanded later if need be.

src/lib/loaders/loaders_without_authentication/gravity.ts
anandaroop: Good call, added.

src/schema/v2/fields/money.ts
anandaroop: Good call, incorporated into the new implementation

src/schema/v2/fields/money.ts
anandaroop: Hm I'm not up on those latest error-handling changes. Is there a specific change you'd prefer to see here?

src/schema/v2/fields/__tests__/money.test.ts
```diff
@@ -68,3 +70,165 @@ describe(amount, () => {
...
+          listPrice {
+            ... on Money {
+              major(convertTo: "CAD")
+            }
+          }
+        }
+      }
+    `
+
+    expect(runQuery(query, context)).rejects.toThrowError(/*Only USD*/)
```
anandaroop: This throw is not working as expected. It _does_ throw, but not with my supplied error message (which is why it's commented out here). I think maybe there's something about the promise rejection handling I'm not getting.

throw

src/schema/v2/fields/money.ts
mzikherman: Yea, basically invoking `exchangeRatesLoader` might throw an error, or the corresponding exchange rate might be not found, so you'd have a divisor of `null` when calculating the USD amount.

Basically wondering if in either/both of those cases, you may want to explicitly be defensive and return `null`, vs possibly throwing an error.

But, `null` feels like an odd return type, plus this is a non nullable field.

So I think this might be good as-is!

src/schema/v2/fields/money.ts
```diff
@@ -131,10 +131,34 @@ export const Money = new GraphQLObjectType<any, ResolverContext>({
...
         const factor = currencyCodes[currency.toLowerCase()].subunit_to_unit
         // TODO: Should we round or used a fixed precision?
-        return cents / factor
+        const major = cents / factor
+
+        const needsConversion = !!convertTo && convertTo !== currency
+        if (needsConversion) {
+          const exchangeRates = await exchangeRatesLoader()
+          // from here, very USD specific
+          if (convertTo !== "USD") {
```
mzikherman: This can maybe move above the `if (needsConversion)`, just to save a fetch if an unsupported conversion is requested.

src/schema/v2/fields/__tests__/money.test.ts
mzikherman: Ya, based on that stack trace it looks like a possible issue with the stitching lib, and throwing an error from a union type field. I wouldn't worry too much about it, can timebox and pend the test.

anandaroop: I tried but couldn't fix the test, pending for now!

src/schema/v2/fields/money.ts
anandaroop: Gotcha, yeah I guess raising an error on a problem with loading exchange rates sounds appropriate then. Is there anything I need to know about the new Force error handling with regard to this? Is this the "principal field" work?

mzikherman: Yea, basically we used to throw an error (which server-side would be a 500 error page to the user), on _any_ GraphQL `error` being present....even if the only error for a large artwork page query was one teensy leaf, an exchange rate error'ing for instance.

So we got used to being defensive about those scenarios. And not that it's necessarily bad to be defensive! But, that defensiveness might be better in the UI component. And, with the `principalField` directive and some updates to Reaction middleware, we now only throw away the query if an error occurs in a `principalField` resolver, otherwise we let the components render.

So just for sanity and peace of mind it might have been better to be defensive everywhere- Metaphysics, UI's, etc. However, now we can more properly report leaf errors and let the UI decide, and no longer need to be so overly defensive in the MP layer.

* sorry for that long-winded explanation! I owe a blog post!

PR #2112, opened 2020-01-02
Update emission query map

artsyit:
Greetings human :robot: This PR was created as part of emission's automated deployment.


ashfurrow:
Okay so it appears that the test failures were pulled in while `master` was not building correctly (I can't find where the failure was introduced). The problem is that the Emission deploy script has been amending to this PR rather than opening new ones (this makes sense) but that means that this PR never pulls in the commits that fixed the build. I'm going to close and then trigger an Emission deploy, and hopefully get these queries deployed (they're starting to block MX work).

ashfurrow:
Hmmm, I'd rather pull in the changes manually.

ashfurrow:
Seems like that did the trick! I'm going to bring this up in the FE iOS Practice this Friday, it's a neat little case study.

PR #2113, opened 2020-01-02
Remove alloy as a point-person of MP

dleve123:
Does adding a description after the fact fix the Peril build?


zephraph:
@dleve123 it should, yeah

PR #2115, opened 2020-01-02
Fix Fair exhibition period spec failures (v1 and v2 schemas)

dleve123:
* Removed unused attribute stubs
* Update assertion such that specs expect that year of exhibition period is
  included.
  * exhibition period uses `dateRange` under the hood, which has its own
    thorough unit test suite, so I don't think there's any loss of
    coverage from changing the "case" (from present year to past year) of
    the exhibition period being asserted upon in the spec.
  * Moreover, 2019 is always in the past, so test should be stable now.

### DIFF
--- a/src/schema/v1/__tests__/fair.test.js
+++ b/src/schema/v1/__tests__/fair.test.js
@@ -167,7 +167,6 @@ describe("Fair", () => {
         end_at: "2019-02-17T11:00:00+00:00",
         partner_shows_count: 4,
         name: "Aqua Art Miami 2018",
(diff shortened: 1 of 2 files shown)


dleve123:
A prior commit that was on a poorly named branch related to fixing the build was already merged in:  https://github.com/artsy/metaphysics/commit/7ac201fd03ec318e2b54f81b7f327ba736095f7b – so that should be considered as part of this PR as well!

joeyAghion:
🙏

PR #2117, opened 2020-01-08
#hackathon2020 Update to latest Exchange schema

ashkan18:
to add `commerceMyOrders` and phone number in pickup.


ashkan18:
merging this and going to rely on integrity to make sure things are ok

PR #2119, opened 2020-01-13
new schema from kaws

mbilokonsky:
adds `descriptionMarkdown` and `hasDescriptionMarkdown` to kaws schema.


ashleyjelks:
#mergeongreen

PR #2133, opened 2020-01-23
Attempt to upgrade dd-trace from 0.7.3 to 0.16.3

joeyAghion:
This got complicated fast. First I attempted to upgrade `dd-trace`, but we had patched the 0.7.3 source to fix a serialization issue. From the context I was able to find, I understand that the serialization problem was related to GraphQL traces specifically (which were just removed in https://github.com/artsy/metaphysics/pull/2131), so I removed the patch and upgraded anyway. We also reference `@types/dd-trace`, but that package has been [deprecated saying](https://www.npmjs.com/package/@types/dd-trace) "dd-trace provides its own type definitions," so I removed that as well. Lint-ing failures led me to modify our custom cache-tracing code slightly, but I haven't been able to verify that it does the right thing.

The history of `dd-trace` upgrades has been fraught, so this should be tested carefully in staging:
* The original upgrade from 0.6.0 to 0.7.3 (https://github.com/artsy/metaphysics/pull/1480) was reverted because it caused crash-looping pods upon deploy (https://github.com/artsy/metaphysics/pull/1486).
* The re-upgrade attempt (https://github.com/artsy/metaphysics/pull/1490) was _again_ immediately reverted (https://github.com/artsy/metaphysics/pull/1492) due to different [serialization-related] crashes.
* Eventually it was upgraded to 0.7.3 with a patch in https://github.com/artsy/metaphysics/pull/1496.
* An attempt to upgrade to 0.10.4 and use the built-in memcached plugin (https://github.com/artsy/metaphysics/pull/1659) was reverted because GraphQL traces failed to show up (https://github.com/artsy/metaphysics/pull/1661).
* Cache tracing exhibited memory leaks in tests that were optimized in https://github.com/artsy/metaphysics/pull/1848.

If this is successfully released, we should try out the GraphQL tracing again. We also should consider using the built-in `memcached` support again, but I don't understand the original problems enough to guide that.


src/lib/tracer.ts
```diff
@@ -26,9 +26,9 @@ export function init() {
 const createCommand = (command: string) => <T>(
   promise: Promise<T>
 ): Promise<T> => {
-  const parentScope = tracer.scopeManager().active()
+  const parentScope = tracer.scope().active()
   const span = tracer.startSpan("memcached", {
-    childOf: parentScope && parentScope.span(),
+    childOf: parentScope || undefined,
```
joeyAghion: I had to do this for type reasons, but maybe there's a nicer, more idiomatic way?

joeyAghion:
Yes, typescript questions aside, I think trying this in staging next makes sense to confirm (1) it doesn't damage the tracing we already have and (2) doesn't affect memory/availability. (Given the past difficulty, I wouldn't _bet_ on that being successful, but I don't know other meaningful tests to do first.)

PR #2144, opened 2020-01-29
[PUR-1602] Sort Other Works By section by merchandisability rating

jpotts244:
Update sorting in the Other Works By section of the artist page to be based on desc merchandisability
[Jira Ticket](https://artsyproduct.atlassian.net/browse/PURCHASE-1602)

Paired with @sepans 🎉 

Testing
To test this change we ran metaphysics locally and saw the data sorting change

Before code change:
`artist/jr/artworks?exclude_ids%5B%5D=5b5ca00b5622c23c3700fa8b&page=1&size=8&sort=-published_at`

Screen Shot 2020-01-29 at 3 00 43 PM

After code change:
`artist/jr/artworks?exclude_ids%5B%5D=5b5ca00b5622c23c3700fa8b&page=1&size=8&sort=-merchandisability`

Screen Shot 2020-01-29 at 3 00 05 PM

### DIFF
--- a/src/schema/v2/artwork/artworkContextGrids/ArtistArtworkGrid.ts
+++ b/src/schema/v2/artwork/artworkContextGrids/ArtistArtworkGrid.ts
@@ -45,7 +45,7 @@ export const ArtistArtworkGridType = new GraphQLObjectType<
           options,
           artwork,
         })
-        gravityArgs.sort = "-published_at"
+        gravityArgs.sort = "-merchandisability"
 
         return artistArtworksLoader(artist.id, gravityArgs).then(artworks => {
           if (!artworks) return null
--- a/src/schema/v2/artwork/artworkContextGrids/__tests__/defaultContext.test.ts
+++ b/src/schema/v2/artwork/artworkContextGrids/__tests__/defaultContext.test.ts
@@ -71,6 +71,28 @@ describe("Default Context", () => {
     }
   })
 
+  it("Calls the loader class with the correct arguments", async () => {
+    const artistArtworks = [
+      { id: "artwork1", title: "Artwork 1" },
+      { id: "artwork2", title: "Artwork 2" },
+      { id: "artwork3", title: "Artwork 3" },
+    ]
+    context.artistArtworksLoader = jest.fn(() =>
+      Promise.resolve(artistArtworks)
+    )
+
+    await runAuthenticatedQuery(query, context).then(data => {
+      expect(data.artwork.contextGrids.length).toEqual(3)
+      expect(context.artistArtworksLoader).toHaveBeenCalledWith("andy-warhol", {
+        exclude_ids: ["abc123"],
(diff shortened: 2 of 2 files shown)


src/schema/v2/artwork/artworkContextGrids/ArtistArtworkGrid.ts
```diff
@@ -45,7 +45,7 @@ export const ArtistArtworkGridType = new GraphQLObjectType<
           options,
           artwork,
         })
-        gravityArgs.sort = "-published_at"
```
jpotts244: Any tips or tricks for testing a change like this in the metaphysics tests? 

Looking like the tests mock out responses

sweir27:
This does look good!

If you want to write a test, I think you can do something similar to [here](https://github.com/artsy/metaphysics/blob/master/src/schema/v2/__tests__/city.test.ts#L170-L172) (where you inspect the mocked data loader for the arguments it was called with).

jpotts244:
Thanks for the tips! I just pushed up a test to expose the loader's arguments changes. Let me know if you think it is too bulky or have any feedback @sepans @sweir27 !

src/schema/v2/artwork/artworkContextGrids/__tests__/defaultContext.test.ts
```diff
@@ -71,6 +71,28 @@ describe("Default Context", () => {
...
 
+  it("Calls the loader class with the correct arguments", async () => {
+    const artistArtworks = [
+      { id: "artwork1", title: "Artwork 1" },
+      { id: "artwork2", title: "Artwork 2" },
+      { id: "artwork3", title: "Artwork 3" },
+    ]
+    context.artistArtworksLoader = jest.fn(() =>
+      Promise.resolve(artistArtworks)
+    )
```
sepans: Is this block needed or it is already being mocked in the `beforeEach` block [above](https://github.com/artsy/metaphysics/pull/2144/files#diff-da0029ccfd26c7da1ca996217d7e48d7R60)?

jpotts244: Good question!  Glad you called this out - I am a little insecure about this blurb and let me know if you have any hints and improvements!  

I added this block here to convert the `artistArtworksLoader` function into a mock function with Jest. Without this change I wasn't able to access the arguments/use the `toHaveBeenCalledWith` matcher and would only have access to the results of the function after it had ran.

sepans: Ah I see yes you are right it makes total sense. Sorry I didn't notice you added that `jest.fn` part.

sepans:
Going to steal @sweir27 's authority and merge this 😄 🚀

jpotts244:
thanks @sepans !

PR #2159, opened 2020-02-05
[WIP] [FX-1795] Exclude Collection search results from some Eigen versions

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-1795

Update: closing this in favor of #2160 — since the required change can basically be accomplished by changing a single GraphQL query argument, @ashfurrow had the slick idea of intercepting this at the persisted query layer.

### DIFF
--- a/src/schema/v2/search/SearchResolver.ts
+++ b/src/schema/v2/search/SearchResolver.ts
@@ -94,6 +94,15 @@ export class SearchResolver {
     }
   }
 
+  isForEigenVersionRequiringWorkaround() {
+    const { userAgent } = this.context
+    return (
+      userAgent &&
+      (userAgent.includes("Artsy-Mobile/6.2.0") ||
+        userAgent.includes("Artsy-Mobile/6.2.1"))
+    )
+  }
+
   resolve() {
     if (!this.args.page) {
       delete this.args.page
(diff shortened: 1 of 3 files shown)


src/schema/v2/search/SearchResolver.ts
```diff
@@ -94,6 +94,15 @@ export class SearchResolver {
     }
   }
 
+  isForEigenVersionRequiringWorkaround() {
+    const { userAgent } = this.context
+    return (
+      userAgent &&
+      (userAgent.includes("Artsy-Mobile/6.2.0") ||
+        userAgent.includes("Artsy-Mobile/6.2.1"))
```
joeyAghion: This might benefit from a comment or ideally a time horizon after which the work-around can be removed.

anandaroop: Yeah, I was thinking of adding [another time-bomb spec](https://github.com/artsy/metaphysics/blob/826d125963060e20f24ee8ab7939f9cc967b5e21/src/schema/v1/__tests__/partner_show_events.test.ts#L5-L10), but in this instance I think we will likely  close this out in favor of https://github.com/artsy/metaphysics/pull/2160.

PR #2160, opened 2020-02-05
Modifies persisted queries for Eigen 6.2.0 and 6.2.1

ashfurrow:
Versions 6.2.0 and 6.2.1 of Eigen included a queries for Search that were too permissive. That is, we were asking for the default entities instead of explicitly naming them. This caused a problem with the FX work on Collections, since this work was still in-progress (but was routable from Search).

Since we can't hot fix the iOS app, the solution is to modify Metaphysics' behaviour to exclude collections from the search results. #2159 had tried this through user agent sniffing. We didn't plan on keeping that change permanently, and it's kind of messy.

A better approach would be to modify the queries in the query map. Eigen will continue to ask for persisted queries with specific IDs, and this PR modifies the queries that Metaphysics looks up. A big advantage here is that we can keep this change permanently, and not modify the source code directly.

## Approach

I looked at [Eigen's changelog](https://github.com/artsy/eigen/blob/master/CHANGELOG.yml) to get the specific Emission version numbers bundled with those Eigen releases. Then I found the query hashes for [6.2.0](https://github.com/artsy/emission/blob/v1.21.11/data/complete.queryMap.json#L74-L75) and [6.2.1](https://github.com/artsy/emission/blob/v1.21.32/data/complete.queryMap.json#L74-L75). Finally, I matched those query IDs to the Metaphysics query map, to add the following arguments to the `searchConnection` resolver:

```
, entities: [ARTIST, ARTWORK, FAIR, GENE, SALE]
```

Search on those versions of the app will now exclude Collections in Search (by only including those entities). By default, search was also returning other entities (like tags, for example) but these seem to be the entities well-supported by the app.

## Testing

I checked out the Eigen builds to make sure this works by pointing to Metaphysics locally.

6.2.0:

6 2 0 results

6.2.1:

6 2 1 results

I can follow up testing on staging once this is merged. 

I will follow up with a PR in Emission to modify the query there.


anandaroop:
Whoa, very clever solution 😎 !

This makes sense to me. A few questions —

> Eigen will continue to ask for persisted queries with specific IDs, and this PR modifies the queries that  Metaphysics looks up. A big advantage here is that we can keep this change permanently, and not modify the source code directly.

Makes sense to me that deployed versions of Eigen will continue to query MP via these persisted IDs. 

What happens though when a future Eigen version comes out and fetches search results via these same queries? So long as they are not using updated versions of these queries they would presumably continue to get these narrower results by virtue of the change in this PR. Otoh if the queries _do_ get updated in Emission they would use different IDs and these newer queries would need to include the narrower entity scope explicitly, correct? Which leads to the next question..

> I will follow up with a PR in Emission to modify the query there.

By this you mean modify the graphql queries (presumably the `entities` arg) in the components themselves, not just the ones in the query map, correct? (If I understand persistent queries correctly, in Emission those queries in the query map are not "used" so much as they are just there for reference?)

And finally — I notice all the queries touched here are `mode: AUTOSUGGEST` — does this cover all the usages of search in iOS? (I don't know the Emission search internals, so I guess I'm just confirming that if you finally hit the "Search" key on the on-screen keyboard you are not getting search results via some other mode.)

ashfurrow:
> What happens though when a future Eigen version comes out and fetches search results via these same queries?

Since the queries will be updated, they'll have new IDs. 

> By this you mean modify the graphql queries [...] not just the ones in the query map, correct?

Correct! You can see them here: https://github.com/artsy/emission/pull/2075

> does this cover all the usages of search in iOS?

Yes, Emission only uses "Autosuggest" for its Search feature.

anandaroop:
Fantastic. I think we are on the same page — merging!

PR #2162, opened 2020-02-05
Update emission query map

artsyit:
Greetings human :robot: This PR was created as part of emission's automated deployment.


src/data/complete.queryMap.json
```diff
@@ -1519,8 +1519,8 @@
...
-  "0b90b9c3923995ca0fd4fa8d509f7be9": "query AutosuggestResultsQuery(\n  $query: String!\n  $count: Int!\n) {\n  ...AutosuggestResults_results_1bcUq5\n}\n\nfragment AutosuggestResults_results_1bcUq5 on Query {\n  results: searchConnection(query: $query, mode: AUTOSUGGEST, first: $count, entities: [ARTIST, ARTWORK, FAIR, GENE, SALE]) {\n    edges {\n      node {\n        imageUrl\n        href\n        displayLabel\n        ... on SearchableItem {\n          displayType\n          slug\n        }\n        ... on Node {\n          id\n        }\n        __typename\n      }\n      cursor\n    }\n    pageInfo {\n      endCursor\n      hasNextPage\n    }\n  }\n}\n",
```
anandaroop: @ashfurrow this looks like a bot PR that tries to undo our artisanal changes, yes?

I'd say maybe we should just close it but there's another non-search query impacted down below 🤔

ashfurrow: Argh, gotcha. That's because this isn't merged yet so the query map is regenerating as-is: https://github.com/artsy/emission/pull/2075

We'll need to undo this change (ie: re-do the earlier change) once that PR gets merged. Does that make sense?

anandaroop: Yes, thanks for confirming 👍

PR #2169, opened 2020-02-09
Stub start identity verification mutation

dleve123:
Introduces `startIdentityVerification` mutation.

Goal of this PR is to have the types in place so that front-end development has be unblocked, but doesn't wire up to Gravity yet.

## When starting an identity verification auth'ed

Screen Shot 2020-02-11 at 11 08 36 AM

## When starting an identity verification unauth'ed

Screen Shot 2020-02-11 at 11 08 55 AM

Jira: https://artsyproduct.atlassian.net/browse/AUCT-887


src/schema/v2/startIdentityVerificationMutation.ts
```diff
@@ -0,0 +1,75 @@
...
+  fields: {
+    identityVerificationId: {
+      type: GraphQLString,
+      description: "Primary ID of the identity verification to be started",
+    },
+  },
+})
+
+const FailureType = new GraphQLObjectType<any, ResolverContext>({
+  name: "startIdentityVerificationFailure",
```
damassi: Types should all be `TitleCase`

_schemaV2.graphql
```diff
@@ -7186,6 +7191,33 @@ enum sort {
   ASC
 }
 
+type startIdentityVerificationFailure {
+  mutationError: GravityMutationError
+}
+
+input startIdentityVerificationMutationInput {
```
damassi: eg

src/schema/v2/startIdentityVerificationMutation.ts
dleve123: Done in 263540698f0cfb069ad4ba443f87e375d90590be

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -9,6 +18,14 @@ export default (accessToken, userID, opts) => {
   )
 
   return {
+    startIdentityVerificationLoader: ({
+      identityVerificationId,
+    }: StartIdentityVerificationGravityInput) => {
+      return Promise.resolve<StartIdentityVerificationGravityOutput>({
+        identity_verification_id: identityVerificationId,
+        identity_verification_wizard_url: "https://staging.artsy.net/auctions",
```
dleve123: just wanted any old URL-like-string here. Will be replaced by return value from Gravity.

dleve123:
Assigning myself to ensure that the commits are rebased down before merging. With that said, I'd love an approved review from:

1. At least 1 person on the auctions engineering team (@yuki24 and/or @erikdstock)
1. At least 1 of @zephraph or @damassi 

---

This is my first significant foray into MP, so any educational comments would be highly appreciated!

src/lib/loaders/loaders_with_authentication/gravity.ts
erikdstock: any reason to not have the return value just be a standard (updated) Identity Verification object?

dleve123: Yeah good question. I was thinking about a couple of things:

1. [minor] I wanted to do what was minimal to get this mutation exposed into MP, acknowledging that it seems likely that we'll want to iterate on the shape and structure here.

2. [major] the wizard URL isn't really part of an identity verification in the sense that we aren't going to persist or consistently expose the wizard URL. It's very much a one-time bit of data. As such, it seems not `part of` an identity verification. So instead of modeling out an `IdentityVerification` in MP in this PR, I think it makes the most sense to defer that modeling into we're forced to implement query-side code to support the Reaction interfaces.

src/schema/v2/startIdentityVerificationMutation.ts
```diff
@@ -0,0 +1,95 @@
...
+})
+
+const OutputType = new GraphQLUnionType({
+  name: "StartIdentityVerificationResponseOrError",
+  types: [SuccessType, FailureType],
+})
+
+export const startIdentityVerificationMutation = mutationWithClientMutationId<
+  any,
+  any,
```
yuki24: Are these types intended to be `any`s?

dleve123: These are straight-up copy+pastes from other mutation definitions. I've mostly been using [`createAccountRequestMutation`](https://github.com/artsy/metaphysics/blob/master/src/schema/v2/createAccountRequestMutation.ts#L88-L92) and [`deleteCreditCardMutation`](https://github.com/artsy/metaphysics/blob/master/src/schema/v2/me/delete_credit_card_mutation.ts#L7-L11).

I briefly dived into type-hell attempting to understand the "why" behind these type arguments, but came back with no more knowledge.

If you have any guidance here, I'm all ears!

src/schema/v2/startIdentityVerificationMutation.ts
```diff
@@ -0,0 +1,95 @@
...
+  mutateAndGetPayload: (
+    { identityVerificationId },
+    { startIdentityVerificationLoader }
+  ) => {
+    if (!startIdentityVerificationLoader) {
+      throw new Error("You need to be signed in to perform this action")
+    }
+
+    return startIdentityVerificationLoader(identityVerificationId)
+      .then(result => result)
```
zephraph:
```suggestion
```

This isn't really necessary.

src/schema/v2/__tests__/startIdentityVerificationMutation.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        }
+      }
+    }
+  }
+}
+`
+
+describe("starting an identity verification", () => {
+  it("requires an access token", () => {
+    runQuery(mutation).catch(error => {
```
zephraph: Given that this runs async, you need a way to let jest know when the test ends.  

You can either add `async` before the start of the arrow function on line 25 and an `await` before this `runQuery` statement _or_ you can return `runQuery`.

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -9,6 +14,12 @@ export default (accessToken, userID, opts) => {
   )
 
   return {
+    startIdentityVerificationLoader: (identityVerificationId: string) => {
```
damassi: Note that the loaders in this object are all alphabetical

src/schema/v2/startIdentityVerificationMutation.ts
dleve123: resolved in e8c4f61f

src/schema/v2/__tests__/startIdentityVerificationMutation.test.ts
damassi: (Could also pass in `done` to it, and then call that after `expect`)

src/schema/v2/__tests__/startIdentityVerificationMutation.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      expect(error.message).toEqual(
+        "You need to be signed in to perform this action"
+      )
+    })
+  })
+
+  it("STUB: returns the given identity verification ID and a link to the staging auctions page", async () => {
```
damassi: What is `STUB`?

src/schema/v2/__tests__/startIdentityVerificationMutation.test.ts
dleve123: I'll update to include `async`, but interesting that this test passed without any further changes needed.

dleve123: Done in f5f9dae34b34a21f8483e1e19992c9ba1cffddf0

src/schema/v2/__tests__/startIdentityVerificationMutation.test.ts
erikdstock: I believe this is just intended to communicate in the test docstring that the spec is not real or fully implemented yet..

dleve123: correct - just wanted to have a flag value in the specs that this isn't legit yet

PR #2170, opened 2020-02-09
Add artworkSizes filter to auction results and remove unused arg

ashkan18:
Depends on https://github.com/artsy/diffusion/pull/188

# Change
Adds search by list of new `ArtworkSizes` enum when filtering auction results.


sepans:
Are we going to do a reaction PR with schema update along with this? Should I create it?

ashkan18:
yes @sepans going to remove WIP from this

PR #2171, opened 2020-02-10
[MX-180] Adds Profiles to iOS Search queries

ashfurrow:
Seems like in #2160 (and later, #2165) we missed one entity: `PROFILE`. This one matters because it's the index all the galleries fall under, so we definitely want to surface those results (and the app can definitely handle these).

I will definitely promise to not make a habit of updating the query map for live versions of the app, but I must say that modifying the persisted queries to fix bugs is _very_ handy.


anandaroop:
This looks good, but can we also take a look at lines 1438-1439 in the query map?

(The query map was updated in a couple of places in #2160 and only _some_ needed to be re-applied in #2165 to undo the bot changes — so if this PR was modeled on #2165 it may have missed those initial ones. Lmk if that makes sense!)

ashfurrow:
@anandaroop Great catch, you're right! I'll amend.

anandaroop:
Thanks 🙌

PR #2177, opened 2020-02-12
Update kaws example urls to https

anandaroop:
Found this while debugging some [pesky local development issues](https://artsy.slack.com/archives/C1HH3KNJG/p1581540116042300).


xtina-starr:
@anandaroop good find!

PR #2196, opened 2020-02-22
Revert "Remove auction results for now"

ashkan18:
Reverts artsy/metaphysics#2195

# Why?
Problem is solved in diffusion database. We no longer need this patch.

cc: @sepans


ashkan18:
#mergeongreen

PR #2208, opened 2020-02-28
Set v2 as default endpoint

zephraph:
I swear I've done this before... 

Anyway, updates he endpoint links to point to v2 by default.


damassi:
@zephraph heh, nice timing: https://github.com/artsy/metaphysics/pull/2206

PR #2213, opened 2020-03-02
HOLD => [Consignments] Temporarily rename stitched types

damassi:
This PR unguards stitching from Convection and prefixes all consignments-specific types with `unstable__` in order to signify that 
- a) they're a moving target; and 
- b) things don't clash with unstitched convections code 

I still need to do some testing once I get latest Eigen up and running to ensure old mutations still work; for the time being this is on hold.


jonallured:
Hey @damassi what's the state of this PR? I think it's stale no?

damassi:
Closin!

PR #2214, opened 2020-03-02
Metaphysics developer can see whether a sale should show a lot condition report

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-916

## Todos

 * [x] Add more test coverage
 * [x] Merge and deploy https://github.com/artsy/gravity/pull/12804

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1023,6 +1023,9 @@ type Artwork implements Node & Searchable & Sellable {
   # Is this artwork part of an auction that is currently running?
   isBiddable: Boolean
 
+  # Can a user request a lot conditions report for this artwork?
+  canRequestLotConditionsReport: Boolean
+
   # When in an auction, can the work be bought immediately
   isBuyNowable: Boolean
   isComparableWithAuctionResults: Boolean
@@ -6407,6 +6410,7 @@ type Sale implements Node {
   isPreliminary: Boolean
   isRegistrationClosed: Boolean
   isWithBuyersPremium: Boolean
+  isLotConditionsReportEnabled: Boolean
   liveStartAt(
     format: String
 
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -904,6 +904,88 @@ describe("Artwork type", () => {
     })
   })
 
+  describe("#canRequestLotConditionsReport", () => {
+    const query = `
+      {
+        artwork(id: "richard-prince-untitled-portrait") {
+          canRequestLotConditionsReport
+        }
+      }
+    `
+
+    it("is true if the artwork's sale is an auction, live, and has lot conditions report enabled", () => {
+      context.saleLoader = sinon.stub().returns(
+        Promise.resolve({
+          live: true,
+          is_auction: true,
+          lot_conditions_report_enabled: true,
+        })
+      )
+
+      return runQuery(query, context).then(data => {
+        expect(data).toEqual({
+          artwork: {
+            canRequestLotConditionsReport: true,
+          },
+        })
+      })
+    })
+
+    it("is false if the artwork's sale is not an auction", () => {
+      context.saleLoader = sinon.stub().returns(
+        Promise.resolve({
+          live: true,
+          is_auction: false,
+          lot_conditions_report_enabled: true,
+        })
+      )
+
+      return runQuery(query, context).then(data => {
+        expect(data).toEqual({
+          artwork: {
+            canRequestLotConditionsReport: false,
+          },
+        })
+      })
+    })
+
+    it("is false if the artwork's sale is an auction but has not started yet or is already closed", () => {
+      context.saleLoader = sinon.stub().returns(
+        Promise.resolve({
+          live: false,
+          is_auction: true,
+          lot_conditions_report_enabled: true,
+        })
+      )
+
+      return runQuery(query, context).then(data => {
+        expect(data).toEqual({
+          artwork: {
+            canRequestLotConditionsReport: false,
+          },
+        })
+      })
+    })
+
+    it("is false if the lot conditions report for the sale is not enabled", () => {
+      context.saleLoader = sinon.stub().returns(
+        Promise.resolve({
+          live: true,
+          is_auction: true,
+          lot_conditions_report_enabled: false,
+        })
+      )
+
+      return runQuery(query, context).then(data => {
+        expect(data).toEqual({
+          artwork: {
+            canRequestLotConditionsReport: false,
+          },
+        })
+      })
+    })
+  })
+
   describe("#isBuyNowable", () => {
     const query = `
       {
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -335,6 +335,27 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           return false
         },
       },
+      canRequestLotConditionsReport: {
+        type: GraphQLBoolean,
+        description:
+          "Can a user request a lot conditions report for this artwork?",
+        resolve: ({ sale_ids }, _options, { saleLoader }) => {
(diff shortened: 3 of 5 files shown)


_schemaV2.graphql
```diff
@@ -1023,6 +1023,9 @@ type Artwork implements Node & Searchable & Sellable {
   # Is this artwork part of an auction that is currently running?
   isBiddable: Boolean
 
+  # Is this artwork part of an auction that is currently running?
+  canRequestLotConditionsReport: Boolean
```
yuki24: I'd be curious to hear about this name. The name `canRequestLotConditionsReport` is bit of a convention-beaking name as we have many more `isAdjective` fields, but it reads better from a front-end point of view. Any suggestions or feedback here?

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -904,6 +904,88 @@ describe("Artwork type", () => {
...
+        artwork(id: "richard-prince-untitled-portrait") {
+          canRequestLotConditionsReport
+        }
+      }
+    `
+
+    it("is true if the artwork's sale is an auction, live, and has lot conditions report enabled", () => {
+      context.saleLoader = sinon.stub().returns(
+        Promise.resolve({
+          live_start_at: moment.utc("2020-12-31").format(),
```
dleve123: Are we sure about this? I was under the impression that we wanted to enable the ability for a user to request lot condition reports _even before_ a sale goes live. Can you double check this?

dleve123: Actually, this would mean that we _never_ show the button for LAI sales, which I'm pretty certain isn't right...

The reaction code is using [`is_biddable`](https://github.com/artsy/metaphysics/blob/b5ba5830706151e65f587a3c0597a5d0d16dfff3/src/schema/v1/sale_artwork.ts#L33) to determine whether to display the button or not...

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -904,6 +904,88 @@ describe("Artwork type", () => {
...
+      return runQuery(query, context).then(data => {
+        expect(data).toEqual({
+          artwork: {
+            canRequestLotConditionsReport: false,
+          },
+        })
+      })
+    })
+
+    it("is false if the artwork's sale is not live", () => {
```
dleve123: To clarify above comment, I don't think these are the intended requirements.

src/schema/v2/artwork/__tests__/artwork.test.js
yuki24: The `is_biddable` function you referred to is a function that is used for `SaleArtwork` type. The actual flag we use in Reaction is [the one in the `Artwork` type](https://github.com/artsy/metaphysics/blob/476198a1f1db35261abf55272aa6c4d27f7f44c9/src/schema/v2/artwork/index.ts#L327-L333), which is different. Obviously it is a question, but I think what we should be doing as part of the story is to respect the existing behaviour and add the extra check with `lot_conditions_report_enabled`, which is what's in this PR.

yuki24: Also we only check the presence of the `live_start_at`, which means the flag would be `true` even for sales that haven't started yet.

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -904,6 +904,88 @@ describe("Artwork type", () => {
...
+      return runQuery(query, context).then(data => {
+        expect(data).toEqual({
+          artwork: {
+            canRequestLotConditionsReport: false,
+          },
+        })
+      })
+    })
+
+    it("is false if the artwork is biddable", () => {
```
bhoggard: I don't think this description is correct? Is `biddable` what this test checks?

src/schema/v2/artwork/__tests__/artwork.test.js
yuki24: We have discussed on zoom and here are our conclusions:

 * We want to keep the ability to show this button for all sales (live and OO)
 * `Sale#live` in gravity is a very confusingly named attribute as it really means “open for bidding” and has nothing to do with live auction integration
 * We want to extend but not change the existing rules in Reaction and Emission: show the button if the sale is an auction, is “live” (aka open for bidding) and that sale has the Ohm checkbox enabled.

src/schema/v2/artwork/__tests__/artwork.test.js
yuki24: I'm not sure if it sounds better but just updated it to `"is false if the artwork's sale is an auction but has not started yet"`. Any other suggestions?

bhoggard: That's OK

src/schema/v2/sale/index.ts
```diff
@@ -231,6 +231,11 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLBoolean,
         resolve: ({ buyers_premium }) => !!buyers_premium,
       },
+      isLotConditionsReportEnabled: {
```
dleve123: thoughts on testing this field more directly in the sale test? It's inadvertently tested via default values, but there's seems to be a precedent for even testing [trivial fields, like `timeZone`, directly](https://github.com/artsy/metaphysics/blob/73d40d98f13dab5b2fe9f0f2d6742fe081d351b4/src/schema/v2/sale/__tests__/index.test.js#L26-L42).

yuki24: Having a single `describe` for each field would lead a very verbose test file. I don't think we should deliberately extract it out.

src/schema/v2/artwork/index.ts
```diff
@@ -335,6 +335,27 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           return false
         },
       },
+      canRequestLotConditionsReport: {
+        type: GraphQLBoolean,
+        description:
+          "Can a user request a lot conditions report for this artwork?",
+        resolve: ({ sale_ids }, _options, { saleLoader }) => {
+          if (sale_ids && sale_ids.length > 0) {
+            const sale_id = sale_ids[0]
```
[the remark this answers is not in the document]
yuki24: Out of 2.7M artworks in the database, there were only 49 artworks that have multiple sales, none of which was created in 2018 or 2019. It is probably safe to assume that these were assigned to multiple sales unintentionally. Even if it was possible (e.g. https://www.artsy.net/artwork/hurvin-anderson-sun-shade-to-add), we seem to look at [the first `sale`](https://github.com/artsy/metaphysics/blob/73d40d98f13dab5b2fe9f0f2d6742fe081d351b4/src/schema/v2/artwork/index.ts#L639-L643) for the `sale` field on `Sale`, and checking against all the associated sales would lead to a situation where the check looks at the flag for an already closed sale, which we don't want.

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -904,6 +904,88 @@ describe("Artwork type", () => {
...
+      return runQuery(query, context).then(data => {
+        expect(data).toEqual({
+          artwork: {
+            canRequestLotConditionsReport: false,
+          },
+        })
+      })
+    })
+
+    it("is false if the artwork's sale is an auction but has not started yet", () => {
```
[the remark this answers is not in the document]
yuki24: I thought I should do this in the first place, but I thought we shouldn't spread out the overloaded term.

src/schema/v2/artwork/index.ts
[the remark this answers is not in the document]
dleve123: > we seem to look at the first sale for the sale field on Sale, and checking against all the associated sales would lead to a situation where the check looks at the flag for an already closed sale, which we don't want.

Makes perfect sense – I didn't catch the fact that we already consider the first sale to be main one – 👍

PR #2216, opened 2020-03-03
update shippingInfo to account for Euro zone shipping

oxaudo:
Another piece for https://artsyproduct.atlassian.net/browse/GALL-2598

We might also need to update the string that says `Free domestic shipping only` but waiting for a confirmation on that from @joanna20.


src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -1537,18 +1537,6 @@ describe("Artwork type", () => {
       })
     })
 
-    it("is set to free domestic shipping when domestic_shipping_fee_cents is 0 and international_shipping_fee_cents is present", () => {
```
oxaudo: That was just duplicated hence the remove here.

ansor4:
this looks good to me!

oxaudo:
Ok. A bit confused with how it's trying to merge with master now. Closing this PR and will reopen with the same changes with the clean master.

PR #2220, opened 2020-03-04
Fixes a bug where lot condition report can not be enabled

yuki24:
Apparently `sale.live` is not a thing in Gravity API v1 and this was always evaluating to `false` in staging. Instead, we should be looking at [`auction_state`](https://github.com/artsy/gravity/blob/6f3f75b39fd88dbc7181c71a53f128aefd00e8f5/app/models/domain/sale.rb#L201-L207), which has no ambiguity 😎 I've also added one more test for the `preview` state.


bhoggard:
#mergeongreen

PR #2221, opened 2020-03-05
[Consignments] Reenable Convection Stitching

damassi:
**Related**: https://github.com/artsy/convection/pull/504/

This unwires unstitched mutations related to creating / updating a consignments submission, in favor of stitching. 

Now that convection PR has been merged, if one pulls down this PR and points Eigen at it, can create stitched submission at consignments staging. See https://convection-staging.artsy.net/admin/submissions/31121 for an example. 

The schema conflicts below relate to _old_ consignments `convection` schema, before it was modified to match MP, and can thus be ignored.

This is a phased roll-out:
- [x] Merge convection PR up above
- [x] Merge this PR 
- [ ] Enable `ENABLE_CONSIGNMENTS_STITCHING` on staging
- [ ] Test submissions in iOS app thoroughly 
- [ ] Deploy to prod

There are details in the PR up above, but the main change required was to match the types / schema in Convection to what was manually built using GraphQL loaders and the REST API. This was required so as to avoid breaking changes to our iOS app and enable a code-free transition on the Eigen side of things. 

If everything goes smoothly we can follow-up and delete a bunch of consignments code from MP as it will be entirely replaced by stitching.  

#### This flow has been tested by:

1) Setting `CONVECTION_API_BASE=http://localhost:5000/api` 
1) Setting ENABLE_CONSIGNMENTS_STITCHING=true`
1) Booting Convection
1) Running `yarn sync-remote-schema` to sync Convection's localhost schema
1) Booting MP 
1) Booting Eigen and running the "shake device" action to show settings screen where MP endpoint was then pointed at MP localhost
1) Creating a new Consignments submission, including image upload. 

#### History

Previously, we ran into issues stitching in Convection. I believe this was before we had a pattern for transforming / aliasing type names across services. The problems observed in [this issue](https://github.com/artsy/metaphysics/issues/1017) no longer seem to apply.


src/data/complete.queryMap.json
```diff
@@ -1132,7 +1130,6 @@
...
-  "15aa0a85a06807f72c1b64cbb3c312e8": "mutation updateConsignmentSubmissionMutation(\n  $input: UpdateSubmissionMutationInput!\n) {\n  updateConsignmentSubmission(input: $input) {\n    consignmentSubmission {\n      internalID\n    }\n  }\n}\n",
```
damassi: @ashfurrow - I manually removed this, and a few others... but this seems wrong?

damassi:
@mzikherman / @jonallured - this is ready to go.

src/data/complete.queryMap.json
ds300: yeah there should be no reason to delete entries from this file. It would degrade performance in the app, and the eigen deploy script would only add these entries back so we need to find out what the underlying problem is here.

damassi: @ds300  - tests pass locally, so i think it has to do with the ENV bar above. If this doesn't break the app, lets go ahead and merge and then follow up once ENV var is flipped on in staging and so on.

ds300: it will only get added back in the next time there's a build on eigen/master and then MP master will start failing. I think we should fix the env problem before merging this.

damassi: Once it's merged into master and the ENABLE_CONSIGNMENTS_STITCHING var is flipped on in staging, the query map will regenerate and it will pass. 

@ds300 - You can test this locally by checking out this PR, checking out `src/data/complete.queryMap.json` from master, and then running `yarn jest`.

damassi: Adding to CI worked 👍

PR #2228, opened 2020-03-09
Allow auctionResultsConnection to be filtered by new date fields

jpotts244:
Add the ability to filter the auctionResultsConnection by created date ranges 
Jira Ticket: https://artsyproduct.atlassian.net/browse/PURCHASE-1813

Ex of filtering queries:

`earliestCreatedYear: 2000, latestCreatedYear: 2001`
Screen Shot 2020-03-09 at 2 26 25 PM

`earliestCreatedYear: 1986, latestCreatedYear: 2001`
Screen Shot 2020-03-09 at 2 26 04 PM


_schema.graphql
```diff
@@ -1,19 +1,19 @@
 directive @principalField on FIELD
 
-# Autogenerated input type of AddAssetToConsignmentSubmission
 input AddAssetToConsignmentSubmissionInput {
-  # A unique identifier for the client performing the mutation.
+  # The type of the asset
```
jpotts244: That pre-commit hook that runs generated a really large diff to the two schemas, is this expected? If not, any suggestions on getting around that? I was a little surprised to see such a large diff

sepans: I am guessing `ENABLE_CONSIGNMENTS_STITCHING` env variable is recently set to true on staging but you don't have it in your `.env` file so all the [convection](https://github.com/artsy/convection) stitched schema fields that are already added to master is being removed by `dump-schema`

jpotts244: Ah yes, that was exactly it. Thanks for the context!!

damassi: @jpotts244 - mind updating `.env.example` with this value?

jpotts244: will do ^

zephraph:
This is so good 👏

PR #2235, opened 2020-03-11
update shippingInfo to account for continental Euro shipping

oxaudo:
This is a reopen of #2216 starting from clean master.


oxaudo:
K. MP is next. Merging @ansor4

PR #2236, opened 2020-03-11
Update convection schema

jonallured:
While mobbing with @damassi and @pepopowitz we worked on getting our convection stitching more in line with how we do it for vortex. See https://github.com/artsy/convection/pull/517 for more on that.

So this PR sees us checking in that new schema file for convection and then yanking the pieces that are no longer needed, including the dump-remote-schema script.


package.json
```diff
@@ -17,7 +17,6 @@
     "build": "yarn build:lib && yarn build:index && yarn build:fixtures && yarn build:remote-schemas",
     "ci": "yarn test",
     "dev": "DEBUG=info,warn,error nf start --procfile Procfile.dev -w",
-    "dump-remote-schema": "babel-node --extensions '.ts,.js' ./scripts/dump-remote-schema.ts",
     "dump-schema": "babel-node --extensions '.ts,.js' ./scripts/dump-schema.ts",
```
damassi: @jonallured - mind adding a `"sync-schema:convection": "copy ../_schema.graphql data/convection.graphql"` command here?

jonallured: Might be overkill - we have that command documented over in convection and it makes more sense to me to copy from there rather than from MP. 🤷‍♂

PR #2246, opened 2020-03-19
Update convection schema

jonallured:
This PR gets MP up to date with the changes over on https://github.com/artsy/convection/pull/526. There are breaking changes here, but only on the query operations, not the mutations. That's important because Eigen uses those mutations and drift here would cause errors on shipped versions of the app.


jonallured:
This is green so going to merge.

PR #2250, opened 2020-03-19
Fix convection artist id mapping

jonallured:
This was a miss on my part where I didn't check that the mapping from convection artist ids to gravity artist ids was lined up. Volt couldn't resolve this so once this is in, then I can go back over there and get my stitching working correctly again.


jonallured:
I'm hoping to get this one merged in before doing a deploy tomorrow morning: https://github.com/artsy/volt/pull/4217.

PR #2252, opened 2020-03-19
[Consignments] Delete unstitched code

damassi:
This deletes dead code around creating a submission using our conventional, unstitched `loader` pattern. Now that we have schema stitching in place we hit convection directly. 

- [x] Tested local Eigen submission


_schemaV2.graphql
```diff
@@ -3055,96 +3056,90 @@ type ConditionReportRequest {
 
 # Consignment Submission
 type ConsignmentSubmission {
-  additional_info: String
-  artist_id: String!
+  additionalInfo: String
+  artistId: String!
```
damassi: @jonallured - not 100% sure why this diff is showing up here -- assuming we've already copied `_schema` from Convection to MP from your other work?

jonallured: Yeah, I'm confused too - I guess it's possible this code keeps getting added and removed as people with the ENV var set and not set work in MP? But then I would expect things to break and they haven't so I'm pretty confused, haha. Either way, by merging this we should eliminate that vector of complexity altogether so I think we do this thing! 🙌

jonallured:
Thanks so much for this follow up!!! 🏆

PR #2260, opened 2020-03-24
[AUCT-962] Add identityVerified to me object

bhoggard:
Completes https://artsyproduct.atlassian.net/browse/AUCT-962

The related Gravity API change is already in production.


yuki24:
#mergeongreen

src/schema/v2/me/index.ts
```diff
@@ -151,6 +155,7 @@ const MeField: GraphQLFieldConfig<void, ResolverContext> = {
       "followsAndSaves",
       "lotsByFollowedArtistsConnection",
       "identityVerification",
+      "identityVerified",
```
yuki24: I had to remove this field to make it work otherwise MP wouldn't attempt to load data using `meLoader`.

PR #2272, opened 2020-04-03
Adds sales module to home page

ashfurrow:
This is to support an in-progress Eigen PR: https://github.com/artsy/eigen/pull/3142

It supports a query that looks like this:

Screen Shot 2020-04-03 at 15 32 15

I'm adding it to the `HomePage` type because that's what the app uses for its home screen. We want to show live sales on that screen (instead of, as currently, showing some artworks from a [single "featured" sale](https://github.com/artsy/metaphysics/blob/4bad8cfc0ece757a981f9fda93a3845ffd0c534b/src/schema/v2/home/fetch.ts#L48-L60)).

I've got an `any` on `HomePageSalesModuleType` that I don't like but I'm not familiar enough with Metaphysics to figure out what it's _supposed_ to be.

### DIFF
--- a/src/schema/v2/home/__tests__/home_page_sales_module.test.js
+++ b/src/schema/v2/home/__tests__/home_page_sales_module.test.js
@@ -1,0 +1,38 @@
+import { runQuery } from "schema/v2/test/utils"
+
+describe("HomePageFairsModule", () => {
+  it("works", async () => {
+    const liveSales = [
+      {
+        id: "the-greatest-sale-ever",
+        name: "The Greatest Sale Ever",
+      },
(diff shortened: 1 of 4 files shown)


src/schema/v2/home/__tests__/home_page_sales_module.test.js
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable promise/always-return */
```
ashfurrow, line 1: I should remove this

ashfurrow:
Reviewed over Zoom 👍

PR #2274, opened 2020-04-06
Few updates to support the art keeps going feature

sweir27:
For the "Art Keeps Going" feature, we need to be able to fetch `Collections`, `Auctions`, and `Fairs` by a list of IDs, as well as query for the contents of an `OrderedSet`... all at the root level.

A better solution (for the future) is to have metaphysics be able to serve up the shape of data necessary to render marketing-type pages, but for this iteration we have to settle for manually-entered lists of IDs and query for them this way.

I'll leave some comments inline explaining my reasoning!


_schema.graphql
```diff
@@ -1,23 +1,5 @@
 directive @principalField on FIELD
 
-# Autogenerated input type of AddAssetToConsignmentSubmission
-input AddAssetToConsignmentSubmissionInput {
```
sweir27, line 4: This looks wrong! Will look into it later...

src/data/kaws.graphql
```diff
@@ -141,6 +141,7 @@ enum GroupTypes {
 
 type Query {
   collections(
+    slugs: [String!]
```
sweir27, line 144: This was added as part of [this KAWS](https://github.com/artsy/kaws/pull/225) PR.

src/lib/stitching/kaws/__tests__/stitching.test.ts
```diff
@@ -1,96 +1,13 @@
...
-  },
-  Object {
-    "requestThrottleMs": 3600000,
-  },
-]
-`)
+import { getKawsMergedSchema } from "lib/stitching/kaws/__tests__/testingUtils"
+import { getFieldsForTypeFromSchema } from "lib/stitching/lib/getTypesFromSchema"
+
+describe("KAWS Stitching", () => {
```
sweir27, line 4: I deleted the rest of the tests because they were disabled/the comment suggested they were making real API calls 😳.

This test format follows the one we have for Exchange.

src/lib/stitching/kaws/__tests__/testingUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+import { kawsStitchingEnvironmentV2 } from "../stitching"
+import { GraphQLSchema } from "graphql"
+import { executableKawsSchema } from "../schema"
+import localSchema from "schema/v2/schema"
+
+let cachedSchema: GraphQLSchema & { transforms: any }
+let stitchedSchema: ReturnType<typeof kawsStitchingEnvironmentV2>
+let mergedSchema: GraphQLSchema & { transforms: any }
+
+/** Gets a cached copy of the transformed kaws schema  */
```
sweir27, line 11: This is also adapted from the exchange `testingUtils` file.

src/schema/v2/articlesByInternalID.ts
```diff
@@ -0,0 +1,21 @@
+import Article from "./article"
+import {
+  GraphQLList,
+  GraphQLString,
+  GraphQLFieldConfig,
+  GraphQLNonNull,
+} from "graphql"
+import { ResolverContext } from "types/graphql"
+
+export const ArticlesByInternalID: GraphQLFieldConfig<void, ResolverContext> = {
```
sweir27, line 10: The need here is to fetch a list of articles by a list of IDs.

A better solution is to update positron to accept a list of `ids` as an argument, in which case I would adapt the `articles` type.

If I get time before Wednesday, I'll look into that. For now, this will fetch for each article individually.

src/schema/v2/ordered_set.ts
```diff
@@ -36,6 +38,24 @@ const OrderedSetType = new GraphQLObjectType<any, ResolverContext>({
         })
       },
     },
+    itemsConnection: {
```
sweir27, line 41: We're using an `OrderedSet` to store the list of artworks that are "featured" in the grid on the art keeps going page, but the `ArtworkGrid` needs a connection.

Eventually this could be extended to be a union type similar to `items`. For not it explicitly only resolves for artworks.

src/schema/v2/sales.ts
```diff
@@ -16,13 +21,13 @@ export const SalesConnectionField: GraphQLFieldConfig<void, ResolverContext> = {
...
     //       `nodes` root field and use that instead.
     //
-    // ids: {
-    //   type: new GraphQLList(GraphQLString),
-    //   description: `
-    //     Only return sales matching specified ids.
-    //     Accepts list of ids.
-    //   `,
-    // },
+    id: {
```
sweir27, line 24: Also needed this 😅. I noticed it was removed from v2. I added it back and made the slight adjustment to use `id` instead of `ids`.

src/schema/v2/schema.ts
```diff
@@ -106,7 +110,7 @@ const rootFields = {
   creditCard: CreditCard,
   // externalPartner: ExternalPartner,
   fair: Fair,
-  // fairs: Fairs,
+  fairs: Fairs,
```
sweir27, line 113: `fairs` already accepts a list of `ids` 😄, but it wasn't exposed at the root.

sweir27:
A general observation: we don't have a consistent way of querying for a list of objects by a list of `ids`.

I also know we're moving more towards returning `connections` instead of `lists`, but when you're querying by ids I'm not sure if the connection even makes sense.

src/lib/stitching/kaws/stitching.ts
```diff
@@ -102,10 +103,19 @@ export const kawsStitchingEnvironmentV2 = (
   const filterArtworksArgsV2 = pageable(filterArtworksArgsV2WithoutPageable)
   return {
     // The SDL used to declare how to stitch an object
-    extensionSchema: `
+    extensionSchema: gql`
     extend type Artist {
       marketingCollections(size: Int): [MarketingCollection]
     }
+    extend type Viewer {
```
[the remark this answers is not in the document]
sweir27: `marketingCollections` are already available at the root level! But they need to be exposed under `Viewer` (since it's being used in a relay context!). Other fields get automatically exposed under `Viewer` if they're at the root, but I'm guessing that there's something special about stitched fields preventing that?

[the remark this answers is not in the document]
mzikherman: Sorry my bad! I thought this was to power the old-style "feature" pages which weren't Relay-based.

Yea, the automatic exposure in `Viewer` is handled custom here: https://github.com/artsy/metaphysics/blob/9af073c7566d15e695e0a19a70efec2065086f82/src/schema/v2/schema.ts#L148-L151

_schema.graphql
sweir27: Yep! That fixed.

src/schema/v2/articlesByInternalID.ts
sweir27: Fixed! Turns out positron did accept `ids` (I just missed it when looking yesterday 😅). 

I removed this type altogether.

src/schema/v2/sales.ts
sweir27: Updated.

src/schema/v2/ordered_set.ts
```diff
@@ -36,6 +38,24 @@ const OrderedSetType = new GraphQLObjectType<any, ResolverContext>({
...
+      type: artworkConnection.connectionType,
+      description:
+        "Returns a connection of the items. Only Artwork supported right now.",
+      resolve: ({ id, item_type }, options, { setItemsLoader }) => {
+        // Only ArtworkConnections are supported at this time.
+        if (item_type === "Artwork") {
+          return setItemsLoader(id).then(items => {
+            return connectionFromArraySlice(items, options, {
```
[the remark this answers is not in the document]
sweir27: Fixed!

src/schema/v1/ordered_set.ts
```diff
@@ -27,7 +27,8 @@ const OrderedSetType = new GraphQLObjectType<any, ResolverContext>({
     items: {
       type: new GraphQLList(ItemType),
       resolve: ({ id, item_type }, _options, { setItemsLoader }) => {
-        return setItemsLoader(id).then(items => {
+        return setItemsLoader(id).then(({ body }) => {
```
mzikherman, line 30: ⛳️ can do `body: items` to save a line! #superimportantcomment

src/schema/v2/ordered_set.ts
```diff
@@ -28,14 +31,52 @@ const OrderedSetType = new GraphQLObjectType<any, ResolverContext>({
...
       },
     },
+    itemsConnection: {
+      type: artworkConnection.connectionType,
+      description:
+        "Returns a connection of the items. Only Artwork supported right now.",
+      args: pageable(),
+      resolve: (
+        { id, item_type },
+        { before, after, first, last },
```
mzikherman, line 50: This can just be `options` (you're desctructuring but then re-structuring it a few lines down)

sweir27:
Finally 🍏 gonna merge! Thanks for the review @mzikherman 😄

PR #2279, opened 2020-04-07
[Kaws/Stitching] Fix deprecated _id field

damassi:
Noticed we were getting back a default collection when querying `... on Artist` and saw that we were passing old v1 `_id`. Updated to pass `internalID`.


sweir27:
Nice!! Good catch.

sweir27:
We must not be using this anywhere?

damassi:
@sweir27, nope we are not.

PR #2282, opened 2020-04-07
Readd root fields

erikdstock:
Paired on this with @damassi and @bhoggard.

This PR adds some root fields to v2 and also some new filter arguments to the root `saleArtworksConnection`. Barry and I aren't clear what specs (if any) need to be added or whether there are any other gotchas we should be aware of.


src/schema/v2/filterSaleArtworks.ts
```diff
@@ -0,0 +1,117 @@
+import { map, omit } from "lodash"
```
damassi, line 1: Is this `filterSaleArtworks.ts` file needed? Stuff that was commented out in rootFields should still exist in the file tree, yah?

erikdstock: I've been working on removing this commit, which was accidentally included from our work yesterday ... but because i use asdf version manager, and it does not work unless your .nvmrc uses the full versionname (with patch), through some bug I cannot rebase - when it gets back to the pre-patched commit 1305d30 it blows up. So i'll just add a revert commit, I guess.

PR #2284, opened 2020-04-08
[Consignments] Add Artist.isTargetSupply

damassi:
Adds a new field `isTargetSupply` under our `targetSupply` field namespace. 

#### Query
```gql
{
  artist(id: "andy-warhol") {
    targetSupply {
      isTargetSupply
    }
  }
}
```

#### Response
```json
{
  "data": {
    "artist": {
      "targetSupply": {
        "isTargetSupply": true
      }
    }
  }
}
```

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -701,6 +701,7 @@ type Artist implements Node & Searchable & EntityWithFilterArtworksConnectionInt
   isFollowed: Boolean
   isPublic: Boolean
   isShareable: Boolean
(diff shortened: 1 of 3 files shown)


damassi:
@jonallured - IMO fields are pretty cheap and easy, but I slept on this and agree. If someone needs to know about market supply, they can look directly in `marketSupply`. No need to duplicate.

damassi:
Addressed Jon's feedback, and this is pretty straightforward so going to merge!

PR #2290, opened 2020-04-08
Reflect changes to the viewing room schema in Gravity

iskounen:
The purpose of this PR is to reflect changes that were made to the Gravity GraphQL schema in https://github.com/artsy/gravity/pull/12933. This replaces the `subsectionConnection` that is queryable in a viewing room with a simple array of subsections for simplicity.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -701,7 +701,6 @@ type Artist implements Node & Searchable & EntityWithFilterArtworksConnectionInt
   isFollowed: Boolean
   isPublic: Boolean
   isShareable: Boolean
-  isTargetSupply: Boolean
   displayLabel: String
   location: String
   meta: ArtistMeta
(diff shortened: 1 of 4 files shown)


_schemaV2.graphql
```diff
@@ -701,7 +701,6 @@ type Artist implements Node & Searchable & EntityWithFilterArtworksConnectionInt
   isFollowed: Boolean
   isPublic: Boolean
   isShareable: Boolean
-  isTargetSupply: Boolean
```
iskounen, line 704: @damassi I don't know why this change is appearing in the v2 schema. I saw this after I copy-pasted the new Gravity schema into `gravity.graphql` and ran `yarn dump:local`. Please let me know if removing this is a problem or expected.

P.S. I do have `ENABLE_CONSIGNMENTS_STITCHING=true` in my .env file.

iskounen: For Artsy employees, this was discussed in [a Slack thread](https://artsy.slack.com/archives/CS3H1K4BD/p1586446723395200.). The TL;DR is that this field used to exist as a top-level field in the "artist" type, as well as nested under a "target supply" type, and was removed as a top-level field. We missed a step somewhere and it wasn't removed from the V2 schema until this PR.

PR #2291, opened 2020-04-08
Submission pagination in convection

pepopowitz:
Partially addresses https://artsyproduct.atlassian.net/browse/CSGN-37.
Depends on https://github.com/artsy/convection/pull/568.

This PR updates the schemas to account for new pagination fields on convection submissions.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -3118,6 +3118,28 @@ type ConsignmentOffer {
   state: String
 }
 
+type ConsignmentPageCursor {
+  # first cursor on the page
+  cursor: String!
+
+  # is this the current page?
+  isCurrent: Boolean!
+
+  # page number out of totalPages
+  page: Int!
+}
+
+type ConsignmentPageCursors {
+  around: [ConsignmentPageCursor!]!
+
+  # optional, may be included in field around
+  first: ConsignmentPageCursor
+
+  # optional, may be included in field around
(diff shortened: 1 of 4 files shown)


_schemaV2.graphql
```diff
@@ -701,7 +701,6 @@ type Artist implements Node & Searchable & EntityWithFilterArtworksConnectionInt
   isFollowed: Boolean
   isPublic: Boolean
   isShareable: Boolean
-  isTargetSupply: Boolean
```
pepopowitz, line 704: @damassi This doesn't seem right, but I don't know why it's getting removed when i run `dump:local` to regenerate the schemas.

src/lib/stitching/convection/schema.ts
```diff
@@ -22,6 +22,8 @@ export const executableConvectionSchema = () => {
     Asset: "ConsignmentSubmissionCategoryAsset",
     Category: "ConsignmentSubmissionCategoryAggregation",
     Offer: "ConsignmentOffer",
+    PageCursor: "ConsignmentPageCursor",
```
pepopowitz, line 25: Not sure about this, but I chose to include them as mapped names, as that's the approach Exchange is using. :cargocult:

damassi: Yup, works 👍

damassi:
LGTM 👍 

#mergeOnGreen

PR #2292, opened 2020-04-08
Stitch MP Artwork under Grav ViewingRoomArtwork

mdole:
Turns out that because we effectively [erase Gravity's version](https://github.com/artsy/metaphysics/blob/19246f1b3caa99f7e9f143168621c3bc8680643b/src/lib/stitching/gravity/schema.ts#L26) of the GQL `Artwork` type when merging Gravity's schema into MP, we were losing that type under `ViewingRoomArtwork`. In other words, we couldn't run a query like this:
```graphql
{
  viewingRoom(id: "") {
    title
      edges {
        node {
          internalID
          artwork {
            id
          }
        }
      }
    }
  }
}
```
...because Metaphysics' schema only had `internalID` inside of `node`.

@mzikherman was kind enough to sit down with me and work through this. With this PR, we're now referencing MP's own `Artwork` type, which is actually handy in the long run since Gravity's `Artwork` type was missing some things like `price` that we'll need. 

This PR is dependent on https://github.com/artsy/gravity/pull/12934, which changes `ViewingRoomArtwork`'s `artwork` field to `artworkID` so that we can use it in the fragment that allows us to reference a MP Artwork.

Matt also raised the point that with this method as-is, we'd effectively be making 10 separate queries every time we requested 10 artworks - we should think about batching those requests for efficiency in the future.

Screenshot of this in action: 

Screen Shot 2020-04-08 at 7 24 10 PM

[Slack thread](https://artsy.slack.com/archives/C1HH3KNJG/p1586382589020800)

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -12572,6 +12572,7 @@ type ViewingRoom {
 type ViewingRoomArtwork {
   # Unique ID
   internalID: ID!
+  artwork: Artwork
 }
 
 # The connection type for ViewingRoomArtwork.
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -701,7 +701,6 @@ type Artist implements Node & Searchable & EntityWithFilterArtworksConnectionInt
   isFollowed: Boolean
   isPublic: Boolean
   isShareable: Boolean
-  isTargetSupply: Boolean
   displayLabel: String
   location: String
   meta: ArtistMeta
@@ -8102,6 +8101,7 @@ type ViewingRoom {
 type ViewingRoomArtwork {
   # Unique ID
   internalID: ID!
+  artwork: Artwork
 }
 
 # The connection type for ViewingRoomArtwork.
--- a/src/lib/stitching/gravity/stitching.ts
+++ b/src/lib/stitching/gravity/stitching.ts
@@ -1,0 +1,38 @@
+import gql from "lib/gql"
+import { GraphQLSchema } from "graphql"
+
+export const gravityStitchingEnvironment = (localSchema: GraphQLSchema) => {
+  return {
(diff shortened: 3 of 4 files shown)


iskounen:
> Matt also raised the point that with this method as-is, we'd effectively be making 10 separate queries every time we requested 10 artworks - we should think about batching those requests for efficiency in the future.

This may be easier to address now than it will be in the future. What do you think about creating a resolver for the whole viewing room (instead of an individual viewing room artwork) in the stitching file so that we can make 1 query for an array of artwork IDs?

ds300:
Similar to Adam's suggestion, we could

- update the ViewingRoom gravity model to return a plain array of artworkIDs (as mentioned in the gravity PR)
- in MP stitch in an `artworksConnection` field on the `ViewingRoom` type which makes batched queries based on the artworkIDs array.

I couldn't find an example of exactly this, but there are similar things already in the code base and I think most of the bits of logic we need are already implemented.

mdole:
We hopped on a call earlier to discuss this - we think it makes the most sense to get the grav PR + this PR merged as-is since that will unblock the Eigen work, and we can follow up to implement batching or make other revisions to our types setup before we release to users and are somewhat locked into our schema decisions.

We're thinking of possible future iterations in 3 different "tracks":
1. Replace `artworksConnection` on the `ViewingRoom` type in Gravity with just `artworks` (which would be an array of `Types::ViewingRoomArtworkType`), and then introduce the connection in MP
2. Update the resolver introduced in this PR to do some batch fetching without changing the Gravity types (somewhat unclear on how this would happen exactly, but we're pretty sure it's possible)
3. Move the full `ViewingRoomType` from Gravity into Metaphysics

PR #2302, opened 2020-04-09
Sync schema to reaction on update

damassi:
Following [recent work to sync changes to Eigen](https://github.com/artsy/metaphysics/pull/2288/), lets carry it on to Reaction 🥳 

Ran script locally and seemed to work great, but no actual changes to sync yet: 

Screen Shot 2020-04-09 at 3 48 28 PM

@ds300 - https://github.com/artsy/update-repo is seriously cool! Lets broadcast it.

### DIFF
--- a/scripts/push-schema-changes.js
+++ b/scripts/push-schema-changes.js
@@ -10,8 +10,7 @@ async function main() {
 
     execSync("yarn dump:staging")
 
-    await updateRepo({
-      repo: { owner: "artsy", repo: "eigen" },
+    const updateSchemaAction = {
       branch: "update-schema",
       title: "Update metaphysics schema",
       targetBranch: "master",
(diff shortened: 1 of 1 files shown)


zephraph:
So w/ auto, we set up ssh keys via CircleCI directly. Circle has a way to automatically configure a repo with SSH keys for pushing to that repo. The caveat for that is that it's intended for the _same_ repo that CI is running against so it wouldn't be helpful in this case. 

If you're using an SSH key I believe you don't need a github token to push.

ds300:
Ah i see. We do need both kinds if auth then, ssh for pushing and access token for creating the PR. So let's do what I said before and replace the repo-level ssh key in this job with a user-level key for `artsyit`.

ds300:
ooooh I did some research and looks like you \*can\* push to a repo with a GH_TOKEN after all.

you just check out the repo with https as `https://<GH_TOKEN>@github.com/owner/repo.git`

So we can change `update-repo` to do that and then just merge this as-is  🙌 feel free to take that work on, otherwise I can get around to it on tuesday.

ds300:
🎉 https://github.com/artsy/reaction/pull/3390

damassi:
Yesssss

PR #2306, opened 2020-04-13
Resolve viewing room artworks as a batch instead of individually (GALL-2766)

iskounen:
This PR updates the Gravity stitching library to extend the `ViewingRoom` type with an artwork connection field that will fetch all artworks in a viewing room as a paginated batch.

A resolver for the `ViewingRoomArtwork` type already exist, and its job is to fetch an individual artwork that matches the `artworkID` field. What I am attempting to do is fetch all of the artworks in a viewing room with one query by stitching-in a new connection field on the `ViewingRoom` type that fetches all associated `artworkID`s at once.

### DIFF
--- a/src/lib/stitching/gravity/__tests__/stitching.test.ts
+++ b/src/lib/stitching/gravity/__tests__/stitching.test.ts
@@ -4,42 +4,22 @@ import {
   getGravityStitchedSchema,
 } from "./testingUtils"
 
-it("extends the ViewingRoom type with an artworks field", async () => {
+it("extends the ViewingRoom type with an artworksConnection field", async () => {
   const mergedSchema = await getGravityMergedSchema()
   const artworkConnectionFields = await getFieldsForTypeFromSchema(
     "ViewingRoom",
     mergedSchema
   )
 
-  expect(artworkConnectionFields).toContain("artworks")
+  expect(artworkConnectionFields).toContain("artworksConnection")
 })
 
 it("resolves the artworks field on ViewingRoom as a paginated list", async () => {
   const { resolvers } = await getGravityStitchedSchema()
-  const artworksResolver = resolvers.ViewingRoom.artworks.resolve
+  const artworksResolver = resolvers.ViewingRoom.artworksConnection.resolve
 
   const artworks = await artworksResolver(
-    {
-      artworksConnection: {
-        edges: [
-          {
-            node: {
-              artworkID: "1",
-            },
-          },
-          {
-            node: {
-              artworkID: "2",
-            },
-          },
-          {
-            node: {
-              artworkID: "3",
-            },
-          },
-        ],
-      },
-    },
+    { artworkIDs: ["1", "2", "3"] },
     { first: 2 },
     {
       artworksLoader: () =>
--- a/src/lib/stitching/gravity/__tests__/stitching.test.ts
+++ b/src/lib/stitching/gravity/__tests__/stitching.test.ts
@@ -1,0 +1,69 @@
+import { getFieldsForTypeFromSchema } from "lib/stitching/lib/getTypesFromSchema"
+import {
+  getGravityMergedSchema,
+  getGravityStitchedSchema,
+} from "./testingUtils"
+
+it("extends the ViewingRoom type with an artworks field", async () => {
+  const mergedSchema = await getGravityMergedSchema()
+  const artworkConnectionFields = await getFieldsForTypeFromSchema(
+    "ViewingRoom",
+    mergedSchema
+  )
+
+  expect(artworkConnectionFields).toContain("artworks")
+})
+
+it("resolves the artworks field on ViewingRoom as a paginated list", async () => {
+  const { resolvers } = await getGravityStitchedSchema()
+  const artworksResolver = resolvers.ViewingRoom.artworks.resolve
+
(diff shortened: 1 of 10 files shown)


iskounen:
The `viewingRoom` query result as of 94c5e33...

Screen Shot 2020-04-13 at 11 47 57 AM

src/lib/stitching/gravity/stitching.ts
```diff
@@ -8,12 +9,35 @@ export const gravityStitchingEnvironment = (localSchema: GraphQLSchema) => {
       extend type ViewingRoomArtwork {
         artwork: Artwork
       }
-    `,
 
+      extend type ViewingRoom {
+        artworks: ArtworkConnection
```
mzikherman, line 14: A connection usually accepts pagination arguments. Like `artworks(first: Int, after: Int)`. Also, should rename to `artworksConnection`.

src/lib/stitching/gravity/stitching.ts
```diff
@@ -8,12 +9,35 @@ export const gravityStitchingEnvironment = (localSchema: GraphQLSchema) => {
...
+      }
+    `,
     resolvers: {
+      ViewingRoom: {
+        artworks: {
+          resolve: (parent, _args, context, _info) => {
+            const ids = parent.artworksConnection.edges.map(
```
mzikherman, line 21: How are pagination arguments going to flow from `artworks` (`first`, `last`, `after`, ...) through to the underlying Gravity field?

I would switch the Gravity schema to add `artworkIDs` as a field on the viewing room parent. That field should be the full, un-paginated list of artwork id's.

Then, you can slice into that field using the pagination arguments, and call `artworksLoader`.

src/lib/stitching/gravity/stitching.ts
iskounen: - Thanks for teaching me about the pagination arguments. I've included those.
- I can't rename it to `artworksConnection` at the moment because the `ViewingRoom` type already has a field by that name. I will open a PR in Gravity to refactor that field to be an array of all `ViewingRoomArtwork`s that belong to a viewing room.

iskounen: [Here](https://github.com/artsy/gravity/pull/12951) is the Gravity PR.

I added a new field `viewingRoomArtworks` instead of refactoring the existing `artworksConnection` to avoid breaking the existing schema. My plan is to open a separate set of PRs to Gravity and Metaphysics to remove the existing `artworksConnection` field and rename `viewingRoomArtworks` once it is wired-up correctly and I've had a chance to coordinate with @mdole.

Please let me know if you think I'm being overly cautious with this refactor.

src/lib/stitching/gravity/__tests__/stitching.test.ts
```diff
@@ -0,0 +1,69 @@
+import { getFieldsForTypeFromSchema } from "lib/stitching/lib/getTypesFromSchema"
```
mdole, line 1: yay thanks for adding tests!!

mdole:
looking solid so far; agreed with Matt's comments. I'll take another look once Gravity + this have been updated! happy to pair on that if you want some more 👀

iskounen:
**TODO: 2020/04/14 3:11 PM EDT**

- [ ] Resolve merge conflicts (...again)
- [ ] Update Gravity GraphQL schema from https://github.com/artsy/gravity/pull/12951
- [ ] Include a total count in the artwork connection

iskounen:
**TODO: 2020/04/15 9:46 AM EDT**

- [ ]  Only pass the artwork ids that are needed for the current page to the artworksLoader

src/schema/v2/schema.ts
```diff
@@ -101,6 +102,7 @@ const rootFields = {
   artwork: Artwork,
   // artworkVersion: ArtworkVersionResolver,
   artworksConnection: filterArtworksConnection(),
+  artworks: Artworks,
```
mzikherman, line 105: can prefix this with `_do_not_use_artworks: Artworks` to be even stronger than the deprecation reason.

iskounen:
The `viewingRoom` query result as of 1de942a...

Screen Shot 2020-04-20 at 9 45 19 AM

iskounen:
@mzikherman I'd love to hear your thoughts on https://github.com/artsy/gravity/pull/12963 if you have a minute 🙏

mzikherman:
@iskounen I thought that stitching was working w/ this (and the other Gravity changes)? I'm not clear what other issues may be left that leads us to want to not use stitching (that seems to be the intent of adding REST endpoints).

iskounen:
@mzikherman good call. I decided to ask you about that PR before I read through your commit diff. I think your solution is elegant.

On a separate note, I'm not clear on why we want to discourage future use of `Artworks`. Could you please help me understand that?

mzikherman:
@iskounen no huge reason, just that we want to be mindful of adding root fields. In this case, a root field that returns an artwork connection (and currently takes in `ids`) was added in order to be used in stitching. There isn't any use case for it outside of that context, and so it would be nice to send a signal to the developer that they shouldn't really be using it (outside of a stitching resolver).

mdole:
LGTM! Let's get it on staging 🎉

PR #2308, opened 2020-04-13
Add hasSecondFactorEnabled attribute to Me type

dblandin:
Add an attribute to the Me type to represent 2FA enrollment status.

Depends upon this API update :lock: : https://github.com/artsy/gravity/pull/12947


dblandin:
Merging this in after successfully testing against Gravity staging locally.

The `/api/v1/me` responses from Gravity are cached so this required a `Rails.cache.clear` in a Gravity console.

PR #2313, opened 2020-04-14
Deploy

artsyit:
This is an automatically generated release PR!


ds300:
Hey, sorry about this failure, we're still having some minor issues with the schema sync job after renovate upgraded the node orb. It might be fixed in master soon, but if not feel free to ignore failures in that particular CI job, they're not mission critical. As long as it gets past the 'deploy-staging' step with ✅ we'll be fine.

PR #2314, opened 2020-04-14
[PURCHASE-1870] Add filter to allow filtering for lots with empty date info

jpotts244:
Add new filter allows users to filter out lots where created date information is blank.
diffusion changes: https://github.com/artsy/diffusion/pull/226

This can also be used in combination with other date filters:
`allowEmptyCreatedDates=true`
Screen Shot 2020-04-14 at 3 02 35 PM

`allowEmptyCreatedDates=false`
Screen Shot 2020-04-14 at 3 02 18 PM

when not present - it defaults to true
Screen Shot 2020-04-14 at 3 12 00 PM

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -607,6 +607,10 @@ type Artist implements Node & Searchable & EntityWithFilterArtworksConnectionInt
 
     # Filter auction results by latest created at year
     latestCreatedYear: Int
+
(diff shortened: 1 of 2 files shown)


jpotts244:
not super confident in my changes here just wanted to get your expert eyes on it, no rush :]

PR #2316, opened 2020-04-15
Allow page types to be exposed in search

sweir27:
I noticed that we'd stopped exposing static pages in our search/autocomplete results! Tracked the issue down to here.

This works as expected:
image

(And with local force pointing this way:)
Screen Shot 2020-04-15 at 11 38 40 AM


dblandin:
Assigning back to you to merge when ready :+1:

PR #2326, opened 2020-04-22
Provide isFirstMessage from Message, remove need for initialMessage

zephraph:
Fixes https://artsyproduct.atlassian.net/browse/PURCHASE-1867

This PR does a few things:

1. Exposes the new `is_first_message` field from impulse as `isFirstMessage` on the `Message` object. This is a useful field to simplify the rendering logic that clients have to perform for rendering the first message. 
2. Updates how `Message.body` renders when it's the first message. Previously, [clients had to determine](https://github.com/artsy/eigen/blob/1c5994a58f6ffb7f1fa83bf9c6e23601f66e01a5/src/lib/Components/Inbox/Conversations/Message.tsx#L140) if they should use `initialMessage` from the conversation or the message. This is unnecessary logic for the clients to have to perform because we _never_ want to show the collector what was being displayed as the first message text. Now, clients won't have to do any fiddling when displaying messages. They can just display the first message like any other message. 
3. Deprecates `initialMessage` on conversation. It's not really needed if the first message is that exact same thing.

### DIFF
--- a/src/schema/v2/me/conversation/message.ts
+++ b/src/schema/v2/me/conversation/message.ts
@@ -47,6 +47,11 @@ export const MessageType = new GraphQLObjectType<any, ResolverContext>({
       type: new GraphQLNonNull(GraphQLString),
       resolve: ({ id }) => id,
     },
+    isFirstMessage: {
+      description: "True if message is the first in the conversation.",
+      type: GraphQLBoolean,
+      resolve: ({ is_first_message }) => is_first_message,
+    },
     isFromUser: {
       description: "True if message is from the user to the partner.",
       type: GraphQLBoolean,
@@ -86,7 +91,19 @@ export const MessageType = new GraphQLObjectType<any, ResolverContext>({
       description:
         "Unaltered text if possible, otherwise `body`: a parsed/sanitized version from Sendgrid.",
       type: GraphQLString,
-      resolve: ({ body, original_text }) => {
+      resolve: ({
+        body,
+        original_text,
+        conversation_from_name,
+        conversation_initial_message,
+        is_first_message,
+      }) => {
+        if (is_first_message) {
+          const parts = conversation_initial_message.split(
+            "Message from " + conversation_from_name + ":\n\n"
+          )
+          return parts[parts.length - 1]
+        }
         if (original_text) {
           return original_text
         }
--- a/src/schema/v2/me/conversation/message.ts
+++ b/src/schema/v2/me/conversation/message.ts
@@ -100,7 +100,7 @@ export const MessageType = new GraphQLObjectType<any, ResolverContext>({
       }) => {
         if (is_first_message) {
           if (!conversation_initial_message) {
-            return "Hi, I’m interested in purchasing this work. Could you please provide more information about the piece?"
+            return ""
           }
           const parts = conversation_initial_message.split(
             "Message from " + conversation_from_name + ":\n\n"
--- a/src/schema/v2/me/conversation/message.ts
+++ b/src/schema/v2/me/conversation/message.ts
@@ -100,9 +100,7 @@ export const MessageType = new GraphQLObjectType<any, ResolverContext>({
       }) => {
         if (is_first_message) {
           if (!conversation_initial_message) {
-            throw new Error(
-              "This is the first message, but initial_message not provided"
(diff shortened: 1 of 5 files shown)


src/schema/v2/me/conversation/message.ts
```diff
@@ -86,7 +91,19 @@ export const MessageType = new GraphQLObjectType<any, ResolverContext>({
...
-      resolve: ({ body, original_text }) => {
+      resolve: ({
+        body,
+        original_text,
+        conversation_from_name,
+        conversation_initial_message,
+        is_first_message,
+      }) => {
+        if (is_first_message) {
+          const parts = conversation_initial_message.split(
```
jpotts244, line 102: Is it possible for a conversation's `initial_message` to be `null`?

Just did a quick scan of the impulse schema and it looks like there are no non-null constraints on that column - wondering if that is something we should check here or if it is more of a rare edge case.

zephraph: That's a good question. If it's `null` maybe we should throw an error. In my mind, we _never_ want to show a message if it's got information intended for a partner attached to it.

zephraph: I updated it to throw an exception in that case.

zephraph: cc @starsirius what do you think? Should we do more to solidify that upstream?

starsirius: Looks like we do have some conversations have `null` `initial_message`:
```
Conversation.where(initial_message: nil).count
=> 12610
```
Briefly looked at a few of them, they were created from Eigen. I wonder if Eigen allows submitting inquiries with empty message. 🤔

zephraph: Hmmmm.... I'm not sure what we want to do here then. Unless we try to parse the message for the stuff we send to the gallery partner we're going to end up exposing what we send to them to a collector. Not that that's _terrible_ it's just a bit... awkward?

starsirius: I'm not sure if we want to raise an exception here. Maybe it's ok to just return nothing and let clients handle some kind of fallback?

src/schema/v2/me/conversation/message.ts
```diff
@@ -86,7 +91,24 @@ export const MessageType = new GraphQLObjectType<any, ResolverContext>({
...
+        is_first_message,
+      }) => {
+        if (is_first_message) {
+          if (!conversation_initial_message) {
+            throw new Error(
+              "This is the first message, but initial_message not provided"
+            )
+          }
+          const parts = conversation_initial_message.split(
+            "Message from " + conversation_from_name + ":\n\n"
```
starsirius, line 108: Was this "Message from..." injected from somewhere?

zephraph: I pulled it over from the `initial_message` field already present in metaphysics. Not sure though.

src/schema/v2/me/conversation/message.ts
zephraph: Hmmm... there is no fallback?

starsirius: Yeah when there is no initial message, the email that partners get is just the template with empty message (even the default/auto-populated "Hi, I’m interested in purchasing this work..." message) in it.

Maybe we need some design input about the fallback in this case, but I also think that initial message should be required when collector submits an inquiry.

zephraph: Ahhh, I see. Yes, makes sense.

zephraph: I moved this to be empty as per design's suggestion.

zephraph: Following up w/ a conversation [here](https://artsy.slack.com/archives/C9YNS4X32/p1588085836070400?thread_ts=1588085272.067600&cid=C9YNS4X32)

If the `initial_conversation` is null (or empty), we're going to return null to the client. Eigen will handle this in the same way it'd handle an empty string.

zephraph:
Okay, thanks for the feedback everyone! I'm gonna 🚢 it.

PR #2329, opened 2020-04-23
[WIP] srcSet support

dzucconi:
Sorry, another draft PR here.

Was just looking around and thinking that we should properly support pixel density in places where we can. We're currently fetching a much larger image in things like ArtworkGrids, then using in browser resizing to scale it.

There's nothing really preventing us from doing this properly now with the current schema. Something like:

```
image {
  _1x: resized(width: 300) {
    width
    height
    url
  }
  _2x: resized(width: 600) {
    url
  }
  _3x: resized(width: 900) {
    url
  }
}
```

It's just verbose. We also apparently don't expose the `quality` param either.

-----

So I'd suggest we refactor the schema a bit and add a default `srcSet` field which glues these into a single attribute we can just use out of the box.

We'd want to cover this a bit more with proper specs and do some testing IRL to determine what some sane defaults are for quality settings at various pixel densities. But we can definitely get away with much more aggressive compression at the higher densities.

Anyway reason this is draft is:
* How does Gemini deal with spikey loads? Like if all of a sudden every artwork grid on the site was uncached from Gemini's perspective, would that kill anything?
* How does Gemini deal with storage in general? We'll also have many more versions in the cache if we go down this road.


src/schema/v2/image/resized.ts
```diff
@@ -59,6 +75,28 @@ const ResizedImageUrlType = new GraphQLObjectType<any, ResolverContext>({
...
+      resolve: ({ src, resize: { width, height } }) =>
+        SUPPORTED_PIXEL_DENSITIES.map(({ density, quality }) =>
+          proxy(
+            src,
+            "resize",
+            width && width * density,
+            height && height * density,
+            quality
+          )
+        ).join(", "),
```
dzucconi, line 99:
```gql
{
  artist(id: "gerhard-richter") {
    id
    image {
      width
      resized(height: 200) {
        width
        height
        src
        srcSet
      }
    }
  }
}
```

```json
{
  "data": {
    "artist": {
      "id": "QXJ0aXN0OjRkZjY5MzQ5YmMzY2YxMDAwMTAwMDcyNg==",
      "image": {
        "width": 711,
        "resized": {
          "width": 138,
          "height": 200,
          "src": "https://d7hftxdivxxvm.cloudfront.net?resize_to=fit&width=138&height=200&quality=80&src=https%3A%2F%2Fd32dm0rphc51dk.cloudfront.net%2F8yqS4Lx1H9kfjDhDpk1Q4w%2Flarge.jpg",
          "srcSet": "https://d7hftxdivxxvm.cloudfront.net?resize_to=fit&width=138&height=200&quality=80&src=https%3A%2F%2Fd32dm0rphc51dk.cloudfront.net%2F8yqS4Lx1H9kfjDhDpk1Q4w%2Flarge.jpg, https://d7hftxdivxxvm.cloudfront.net?resize_to=fit&width=276&height=400&quality=65&src=https%3A%2F%2Fd32dm0rphc51dk.cloudfront.net%2F8yqS4Lx1H9kfjDhDpk1Q4w%2Flarge.jpg, https://d7hftxdivxxvm.cloudfront.net?resize_to=fit&width=414&height=600&quality=50&src=https%3A%2F%2Fd32dm0rphc51dk.cloudfront.net%2F8yqS4Lx1H9kfjDhDpk1Q4w%2Flarge.jpg"
        }
      }
    }
  }
}
```

damassi: #superminor - nested implicit returns make things hard to debug later through, sinice one can only easily attach breakpoints to concrete statements such as variables or `return`; otherwise everything is evaluated inline and the debugger goes weird.

PR #2332, opened 2020-04-27
Introduce Sale.userNeedsIdentityVerification field

dleve123:
Problem:

Logic about whether the current user needs to undergo identity
verification is scattered around Reaction and Eigen, increasing risk
that business logic starts to drift and generally is a larger code
surface area than needed.

Solution:

Introduce new field to `Sale` which is the source of truth as to whether
the current user needs to undergo identity verification or not for the
sale in question.

Why introduce this field to `Sale` instead of `me`, `User`, or `Bidder`?

- See this [Slack convo][0].

- Because there are times where the user isn't yet registered to bid, but we
need to know if the current user needs to undergo identity verification
or not, so extending `Bidder` doesn't seem to make sense.

- Moreover, there are even times whether there's no reference to a
current user (`me`). In that case, we want to display that the "current
user" needs to undergo identity verification, so extending `me`/`User`
doesn't seems to make sense.

- We always have reference to a `Sale`, so extending that type with this
field makes the most sense. This is a similar pattern to
`Sale.registrationStatus`.

Jira: https://artsyproduct.atlassian.net/browse/AUCT-983
Related PR: https://github.com/artsy/gravity/pull/12996

[0]:https://artsy.slack.com/archives/CA3D02SDN/p1587581641465800

---

Documentation of the behavior of `Sale.userNeedsIdentityVerification`

```
    userNeedsIdentityVerification
      when the sale doesn't requires identity verification
        ✓ returns false (10ms)
      when the sale does require identity verification
        when there is no current user
          ✓ returns true (5ms)
        when there is a current user
          when the user is registered for the sale
            ✓ returns false bidder.needs_identity_verification is false (5ms)
            ✓ returns true when bidder.needs_identity_verification is true (4ms)
          when the user is not registered for the sale
            ✓ returns false when the user is identity verified (3ms)
            ✓ returns true when the user is not identity verified (4ms)
```

### DIFF
--- a/src/schema/v2/sale/index.ts
+++ b/src/schema/v2/sale/index.ts
@@ -287,24 +287,19 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
           _options,
           { meLoader, meBiddersLoader }
         ) => {
-          if (require_identity_verification) {
-            if (!meLoader || !meBiddersLoader) {
-              return true
+          if (!require_identity_verification) return false
+          if (!meLoader || !meBiddersLoader) return true
+
+          return meBiddersLoader({ sale_id: id }).then(bidders => {
+            if (bidders.length > 0) {
+              const bidder = bidders[0]
+              return bidder.needs_identity_verification
             } else {
-              return meBiddersLoader({ sale_id: id }).then(bidders => {
-                if (bidders.length > 0) {
-                  const bidder = bidders[0]
-                  return bidder.needs_identity_verification
-                } else {
-                  return meLoader().then(({ identity_verified }) => {
-                    return !identity_verified
-                  })
-                }
+              return meLoader().then(({ identity_verified }) => {
+                return !identity_verified
               })
             }
-          } else {
-            return false
-          }
+          })
         },
       },
       saleArtworksConnection: {
--- a/src/schema/v2/sale/index.ts
+++ b/src/schema/v2/sale/index.ts
@@ -278,19 +278,28 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
         resolve: ({ require_identity_verification }) =>
           require_identity_verification,
       },
-      requiresIdentityVerificationFor: {
+      userNeedsIdentityVerification: {
         type: GraphQLBoolean,
+        description:
+          "True if the current client needs to undergo identity verification for this sale, false otherwise",
         resolve: (
-          { require_identity_verification },
+          { require_identity_verification, id },
           _options,
-          { meLoader }
+          { meLoader, meBiddersLoader }
         ) => {
           if (require_identity_verification) {
-            if (!meLoader) {
+            if (!meLoader || !meBiddersLoader) {
               return true
             } else {
-              return meLoader().then(({ identity_verified }) => {
-                return !identity_verified
+              return meBiddersLoader({ sale_id: id }).then(bidders => {
+                if (bidders.length > 0) {
+                  const bidder = bidders[0]
+                  return bidder.needs_identity_verification
+                } else {
+                  return meLoader().then(({ identity_verified }) => {
+                    return !identity_verified
+                  })
+                }
               })
             }
           } else {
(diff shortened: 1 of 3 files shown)


dleve123:
@dzucconi @yuki24 @zephraph – I'd love an approving code review from at least 2 of you as you have been collaborators in the design of this. Assignee to myself so that I can handle Git cleanup before merging.

src/schema/v2/sale/index.ts
```diff
@@ -278,6 +278,35 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
         resolve: ({ require_identity_verification }) =>
           require_identity_verification,
       },
+      userNeedsIdentityVerification: {
+        type: GraphQLBoolean,
+        description:
+          "True if the current user needs to undergo identity verification for this sale, false otherwise",
```
dzucconi, line 284: I don't have a ton of context surrounding this feature, but semantically this feels confusing trying to discern the difference between `requireIdentityVerification` — Like why wouldn't this be wrapped up into that field?

src/schema/v2/sale/index.ts
```diff
@@ -278,6 +278,35 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
...
+                    return !identity_verified
+                  })
+                }
+              })
+            }
+          } else {
+            return false
+          }
+        },
+      },
```
dzucconi, line 309: I would flatten out all the conditionals to make this more legible

```tsx
if (!require_identity_verification) {
  return false;
}

if (!meLoader || !meBiddersLoader) {
  return true;
}

return meBiddersLoader({ sale_id: id }).then((bidders) => {
  if (bidders.length > 0) {
    const [bidder] = bidders;
    return bidder.needs_identity_verification;
  }

  return meLoader().then(({ identity_verified }) => {
    return !identity_verified;
  });
});

```

----

Though I can't really speak to any of the other logic.

src/schema/v2/sale/index.ts
dleve123: Yeah, there's a fair bit of context here! A few things:

1. The sale can require all bidders need to be identity verified or not – `Sale.requireIdentityVerification`
2. A user can be identity verified or not – `user.identity_verified`
3. A user, in the context of the sale, (aka a bidder) can need to undergo identity verification before bidding on the sale, given a variety of conditions. This field is the encapsulation of those conditions. One "core" condition is whether the sale requires identity verification, but there's more to consider (is the user identity verified, is the user a bidder on the sale, if so are they approved to bid regardless).

---

Happy to chat briefly to unpack this more, if you are interested. TL;DR is that `requireIdentityVerification` is related, but very much distinct, from this field!

dzucconi: If this field is the encapsulation of those conditions then would it make more sense to name it something like `sale.isBiddableByUser` or something? — Speaking of the action that's capable of being taken rather than some underlying implementation detail

dleve123: > If this field is the encapsulation of those conditions then would it make more sense to name it something like sale.isBiddableByUser or something?

Hmm - this field will be used to change the Bid button to a Register to Bid button and determine whether a particular identity verification disclaimer will be displayed. `isBiddableByUser` is too general I think -- there are other factors that determine whether the sale is biddable by a user (for example, if the sale is open to bidding), which is distinct from the topic in question.

I think it `userNeedsIdentityVerification` is the right size and thing for what we're trying to express here.

src/schema/v2/sale/index.ts
dleve123: +1 - thanks for the input here, will refactor accordingly.

dleve123: Addressed in https://github.com/artsy/metaphysics/pull/2332/commits/5406a988de80fc9d5544360718b576157a0bc580 – personally, I don't like to go full on guard class, so left a single `if / else` expression in.

src/schema/v2/sale/index.ts
erikdstock: This all makes sense to me- It also feels like a property that could be ported to other types in the future (purchases and offers for example). I wonder if the description field would make more sense as "True if the sale requires identity verification and we cannot prove the current user has verified their identity, false otherwise."

PR #2333, opened 2020-04-28
Read the correct field from gravity

pvinis:
We were using the wrong name coming from gravity.

Using this query
```
{
  show(id: "5e6e2efe4d2aa0000e2be3da") {
    name
    images {
      imageURL
      isDefault
    }
  }
}
```
I would get something like
```
{
  "data": {
    "show": {
      "name": "ART ON PAPER",
      "images": [
        {
          "imageURL": "https://d32dm0rphc51dk.cloudfront.net/7vFrlr0sqJBwP7T478Vvbw/:version.jpg",
          "isDefault": null
        },
        {
          "imageURL": "https://d2v80f5yrouhh2.cloudfront.net/prVTNCjKDNEdVPYoq41prQ/:version.jpg",
          "isDefault": null
        },
        {
          "imageURL": "https://d2v80f5yrouhh2.cloudfront.net/6aAeK7lCF1bQxBeeXxteIA/:version.jpg",
          "isDefault": null
        }
      ]
    }
  }
```
but now I get
```
{
  "data": {
    "show": {
      "name": "ART ON PAPER",
      "images": [
        {
          "imageURL": "https://d32dm0rphc51dk.cloudfront.net/7vFrlr0sqJBwP7T478Vvbw/:version.jpg",
          "isDefault": false
        },
        {
          "imageURL": "https://d2v80f5yrouhh2.cloudfront.net/prVTNCjKDNEdVPYoq41prQ/:version.jpg",
          "isDefault": false
        },
        {
          "imageURL": "https://d2v80f5yrouhh2.cloudfront.net/6aAeK7lCF1bQxBeeXxteIA/:version.jpg",
          "isDefault": true
        }
      ]
    }
  }
```

PS: I wasn't sure how to make a transformation to name the incoming json data from `default` to `is_default` and keep the rest of the code same, so I kept the json same and changed the field we read. In this case, because `default` is a reserved name in js, I would prefer the former, but I guess the latter works too. If there is an obvious place I missed, let me know and I can patch this PR.

PPS: I wasn't sure on who to tag for this, to review and assign. Feel free to swap around if I have the completely wrong people. 😅

### DIFF
--- a/src/schema/v2/image/index.ts
+++ b/src/schema/v2/image/index.ts
@@ -21,7 +21,7 @@ export { normalize as normalizeImageData } from "./normalize"
 
 export const getDefault = images => {
   if (isArray(images)) {
-    return find(images, { is_default: true } as any) || first(images)
+    return find(images, { default: true } as any) || first(images)
   }
   return images
 }
--- a/src/schema/v2/image/index.ts
+++ b/src/schema/v2/image/index.ts
@@ -58,7 +58,7 @@ const ImageType = new GraphQLObjectType<any, ResolverContext>({
     },
     isDefault: {
       type: GraphQLBoolean,
-      resolve: ({ is_default }) => is_default,
+      resolve: ({ default: is_default }) => is_default,
     },
     isZoomable: {
       type: GraphQLBoolean,
--- a/src/schema/v2/image/index.ts
+++ b/src/schema/v2/image/index.ts
@@ -21,7 +21,10 @@ export { normalize as normalizeImageData } from "./normalize"
 
 export const getDefault = images => {
   if (isArray(images)) {
-    return find(images, { default: true } as any) || first(images)
+    return (
+      find(images, img => img.is_default === true || img.default === true) ||
+      first(images)
+    )
   }
   return images
 }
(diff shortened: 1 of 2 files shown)


src/schema/v2/image/index.ts
```diff
@@ -58,7 +58,7 @@ const ImageType = new GraphQLObjectType<any, ResolverContext>({
     },
     isDefault: {
       type: GraphQLBoolean,
-      resolve: ({ is_default }) => is_default,
+      resolve: ({ default: is_default }) => is_default,
```
ds300, line 61: Looks like there is in fact an image type in gravity that has an `is_default` property. It is called `AdditionalImage` . Could they both be passed through this resolver in different contexts? That would explain all the existing code and fixtures using the `is_default` property name. If so, maybe this should be

 ```suggestion
      resolve: (image) => image.is_default || image.default,
```

and then you can update the response typings to include both as optional properties?

zephraph: Yep, agree w/ @ds300 above

pvinis:
#MergeOnGreen

pvinis:
Hm, I wanted to test if this works after the fact. I guess not. :D

PR #2335, opened 2020-04-28
[MX-245] Exposes basis of artist suggestions through schema

ashfurrow:
These were being returned from Gravity but then discarded. This exposes them through a new GraphQL Interface using a conditionally backwards-compatible schema change (ie: we need to make sure to expose any data clients are expecting in the new Interface). I've checked Emission's history to gather them all, and the resolver isn't currently referenced in Reaction, so we should be fine.

This allows us to continue requesting the data in queries as-is, but also expose a new `basedOn` field for `SuggestedArtist` objects. Existing queries work, and we don't need to deprecate anything to get the data 🎉  

The query looks like this:

```graphql
{
  homePage {
    artistModules {
      id
      key
      results {
        __typename
        name
        ... on SuggestedArtist {
          basedOn {
            name
          }
        }
      }
    }
  }
}
```

And we receive results like this:

```json
{
  "data": {
    "homePage": {
      "artistModules": [
        {
          "id": "SG9tZVBhZ2VBcnRpc3RNb2R1bGU6eyJrZXkiOiJTVUdHRVNURUQifQ==",
          "key": "SUGGESTED",
          "results": [
            {
              "__typename": "SuggestedArtist",
              "name": "Dennis Adams",
              "basedOn": {
                "name": "Paul Pfeiffer"
              }
            },
            {
              "__typename": "SuggestedArtist",
              "name": "Zhou Xiaohu",
              "basedOn": {
                "name": "Paul Pfeiffer"
              }
            },
            /* ... */
          ]
        },
        {
          "id": "SG9tZVBhZ2VBcnRpc3RNb2R1bGU6eyJrZXkiOiJUUkVORElORyJ9",
          "key": "TRENDING",
          "results": [
            {
              "__typename": "Artist",
              "name": "Hisao Hanafusa"
            },
            {
              "__typename": "Artist",
              "name": "Harumi Nakashima"
            },
            /* ... */
          ]
        },
        {
          "id": "SG9tZVBhZ2VBcnRpc3RNb2R1bGU6eyJrZXkiOiJQT1BVTEFSIn0=",
          "key": "POPULAR",
          "results": [
            {
              "__typename": "Artist",
              "name": "Pablo Picasso"
            },
            {
              "__typename": "Artist",
              "name": "Banksy"
            },
            /* ... */
          ]
        }
      ]
    }
  }
}
```

Big thanks to @dblandin for his help pairing on this 🙌

### DIFF
--- a/src/schema/v2/home/__tests__/home_page_artist_module.test.js
+++ b/src/schema/v2/home/__tests__/home_page_artist_module.test.js
@@ -10,7 +10,13 @@ describe("HomePageArtistModule", () => {
         homePage {
           artistModule(key: ${key}) {
             results {
+              __typename
               slug
+              ... on SuggestedArtist {
+                basedOn {
+                  slug
+                }
+              }
             }
           }
         }
@@ -42,6 +48,11 @@ describe("HomePageArtistModule", () => {
           birthday: null,
           artworks_count: null,
         },
+        sim_artist: {
+          id: "similar",
+          birthday: null,
+          artworks_count: null,
+        },
       },
     ],
   }
@@ -56,15 +67,17 @@ describe("HomePageArtistModule", () => {
     it("returns trending artists", () => {
       return runAuthenticatedQuery(query("TRENDING"), context).then(
         ({ homePage }) => {
-          expect(homePage.artistModule.results).toEqual([{ slug: "trending" }])
+          expect(homePage.artistModule.results[0].slug).toEqual("trending")
+          expect(homePage.artistModule.results[0].__typename).toEqual("Artist")
         }
       )
     })
 
-    it("returns trending artists", () => {
-      return runAuthenticatedQuery(query("TRENDING"), context).then(
+    it("returns popular artists", () => {
+      return runAuthenticatedQuery(query("POPULAR"), context).then(
         ({ homePage }) => {
-          expect(homePage.artistModule.results).toEqual([{ slug: "trending" }])
+          expect(homePage.artistModule.results[0].slug).toEqual("popular")
+          expect(homePage.artistModule.results[0].__typename).toEqual("Artist")
         }
       )
     })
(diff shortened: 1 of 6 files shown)


src/schema/v2/home/__tests__/home_page_artist_module.test.js
```diff
@@ -82,15 +96,15 @@ describe("HomePageArtistModule", () => {
...
         ({ homePage }) => {
-          expect(homePage.artistModule.results).toEqual([{ slug: "trending" }])
+          expect(homePage.artistModule.results[0].slug).toEqual("trending")
         }
       )
     })
 
-    it("returns trending artists", () => {
-      return runAuthenticatedQuery(query("TRENDING"), context).then(
+    it("returns popular artists", () => {
```
ashfurrow, line 104: This, and the other one like it on line 74, were copy and pasted but never modified.

src/schema/v2/home/home_page_artist_module.ts
```diff
@@ -46,21 +51,66 @@ export const HomePageArtistModuleTypes: {
       return suggestedSimilarArtistsLoader({
         exclude_followed_artists: true,
         exclude_artists_without_forsale_artworks: true,
-      }).then(({ body }) => map(body, "artist"))
+      }).then(({ body }) =>
+        body.map(({ artist, sim_artist }) => ({
```
ashfurrow, line 55: Gravity's response here is to include an object with both an `artist` and a `sim_artist` (the similarity basis for the match). We were throwing it away with the previous implementation (`map(body, "artist")`). This includes it.

src/schema/v2/home/home_page_artist_module.ts
```diff
@@ -46,21 +51,66 @@ export const HomePageArtistModuleTypes: {
...
+      type: GraphQLString,
+      resolve: ArtistType.getFields()["formattedArtworksCount"].resolve,
+    },
+    image: Image,
+    artworksConnection: {
+      type: artworkConnection.connectionType,
+    },
+    basedOn: {
+      type: ArtistType,
+      resolve: ({ sim_artist }) => sim_artist,
```
ashfurrow, line 109: And here is where the similar artist basis gets exposed in the schema.

src/schema/v2/home/home_page_artist_interface.ts
```diff
@@ -0,0 +1,28 @@
+import { GraphQLInterfaceType, GraphQLString } from "graphql"
```
damassi, line 1: This is pretty trivial, but can we do camelCase file names? No for things to not be 1:1, especially in our MPv2 world.

ashfurrow: Yup, no problem 👍 I based it on its siblings but didn’t notice it not match the rest.

PR #2340, opened 2020-04-29
[MX-245] Adds basis for artist suggestions in schema

ashfurrow:
Originally implemented in https://github.com/artsy/metaphysics/pull/2335. We had to revert that change in #2338 because it broke the existing app. It turns out that the schema change wasn't as backwards-compatible as we'd thought. 

## The Problem

Everything would work as expected in a structured query like this:

```graphql
{
  homePage {
    artistModules {
      results {
        name
        ... on SuggestedArtist {
          basedOn {
            name
          }
        }
      }
    }
  }
}
```

But the problem is that Relay is more explicit in how it asks for data, creating queries that explicitly reference the type in a fragment spread:

```graphql
fragment ArtistCard_artist on Artist {
  name
  # ...
}

fragment ArtistRail_rail on HomePageArtistModule {
  results {
    __typename
    ...ArtistCard_artist
  }
}

fragment Home_homePage on HomePage {
  artistModules {
    ...ArtistRail_rail
  }
}

query HomeQuery {
  homePage {
    ...Home_homePage
  }
}
```

The `... on Artist` part of the query is what makes it incompatible. In the app, we were seeing all `null` data get passed through (until we re-ran `yarn relay` to generate new schema-compatible queries). I missed this when testing with simplified queries, but noticed that the app beta broken once #2335 was deployed to staging.

## The Solution

Instead, we added a new field called `basedOn` to _all_ Artist objects. It's a wide-reaching change, but it's totally backwards-compatible since it's purely additive to the schema. 

I also did the same test cleanup as the original PR.
## An Alternative

Alternatively, we could deprecate the existing `homePage.artistModule.results` field and replace it with something that returns the data we need. That is to say, we could add a new GraphQL object type to the schema and return an array of them in a new field that replaces `homePage.artistModule.results`. 

This would allow existing clients to continue using the old field, but had two drawbacks: it's overkill to get access to just one more field, and I had a _really_ hard time naming a replacement for the (then-deprecated) `results` field 😅 `results2` felt bad but everything else felt worse.

So I feel like adding `basedOn` to `Artist` itself is a good compromise, especially with a good field description. 

---

Thanks to @ds300 for talking through this with me 🙌

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -626,6 +626,9 @@ type Artist implements Node & Searchable & EntityWithFilterArtworksConnectionInt
     before: String
     last: Int
   ): AuctionResultConnection
+
+  # In applicable contexts, this is what the artist (as a suggestion) is based on.
+  basedOn: Artist
   bio: String
 
   # The Artist biography article written by Artsy
--- a/src/schema/v2/artist/index.ts
+++ b/src/schema/v2/artist/index.ts
@@ -299,6 +299,11 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
           )
         },
       },
+      basedOn: {
+        type: ArtistType,
+        description:
+          "In applicable contexts, this is what the artist (as a suggestion) is based on.",
+      },
       bio: {
         type: GraphQLString,
         resolve: ({ nationality, years, hometown, location }) => {
--- a/src/schema/v2/home/__tests__/home_page_artist_module.test.js
+++ b/src/schema/v2/home/__tests__/home_page_artist_module.test.js
@@ -11,6 +11,9 @@ describe("HomePageArtistModule", () => {
           artistModule(key: ${key}) {
             results {
               slug
+              basedOn {
+                slug
+              }
             }
           }
         }
@@ -42,6 +45,11 @@ describe("HomePageArtistModule", () => {
           birthday: null,
           artworks_count: null,
         },
+        sim_artist: {
+          id: "similar",
+          birthday: null,
+          artworks_count: null,
+        },
       },
     ],
   }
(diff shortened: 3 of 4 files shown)


dblandin:
Interesting! The solution sounds good to me. I'm still a bit confused about the interface incompatibility.

I would've thought that a resulting query from the fragment spreads you outlined above would be compatible since `Artist` implemented the `HomePageArtist` interface.

```graphql
query HomeQuery {
  homePage {
    ... on HomePage {
      artistModules {
        ... on HomePageArtistModule {
          results {
            ... on Artist {
              __typename
              name
            }
          }
        }
      }
    }
  }
}
```

In new app versions, I think the query would have been updated to also include the basedOn bit:

```graphql
query HomeQuery {
  homePage {
    ... on HomePage {
      artistModules {
        ... on HomePageArtistModule {
          results {
            ... on Artist {
              __typename
              name
            }

            ... on SuggestedArtist {
              __typename
              name

              basedOn {
                name
              }
            }
          }
        }
      }
    }
  }
}
```

Maybe I'm missing something between the specific queries generated in the app and how they hit metaphysics. Either way, this approach seems straightforward and avoids the original issue of needing to create a new `results` field. :+1:

ashfurrow:
@dblandin Yeah this surprised me too. Here's the query you provided ran against that branch:

Screen Shot 2020-05-01 at 09 53 10

I guess that Metaphysics (or maybe GraphQL itself?) isn't marshalling the data from `SuggestedArtist` types into `Artist` despite the fact that they both implement an interface that covers all the data that was requested. You can see that results _do_ continue to work for the `artistModules` that continue to return a concrete `Artist` type. It's a shame!

PR #2347, opened 2020-05-01
Re-orders fair results so that fairs that haven't started yet are at the end

ashfurrow:
We were seeing fairs listed in the app sorted only by start_at, which put some fairs that haven't actually started yet at the front of the list. This places fairs whose `start_at` is in the future at the end of the list.

I've taken a follow-up to re-examine some of the other logic in here, since it's kind of weird!

### DIFF
--- a/src/schema/v2/home/home_page_fairs_module.ts
+++ b/src/schema/v2/home/home_page_fairs_module.ts
@@ -1,6 +1,8 @@
 import Fair from "schema/v2/fair"
 import { GraphQLList, GraphQLObjectType, GraphQLNonNull } from "graphql"
 import { ResolverContext } from "types/graphql"
+import { groupBy } from "lodash"
+import moment from "moment"
 
 export const HomePageFairsModuleType = new GraphQLObjectType<
   any,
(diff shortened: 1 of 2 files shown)


src/schema/v2/home/home_page_fairs_module.ts
```diff
@@ -17,22 +19,36 @@ export const HomePageFairsModuleType = new GraphQLObjectType<
           sort: "-start_at",
           active: true,
```
anandaroop, line 20: Tricky! Looks like the intent here is to show the "freshest" events first, but that gets complicated by the fact that some of these so-called "active" events [haven't actually started yet](https://github.com/artsy/gravity/blob/6b51e8221dd8038f07e184661c4e21d61bd9e431/app/models/util/event_status.rb#L23).

The fix below looks reasonable to me 👍

ashfurrow: Yeah exactly. Another option we considered was to make two API calls to Gravity, one for `status=current` and another for `status=upcoming`, but it seemed like overkill and we were getting back more data than we expected (so there's something nontrivial about how those APIs handle the `status` param). 

Thanks Roop!

PR #2352, opened 2020-05-04
Add more robust way to get last message, deprecate unnecessary or unused fields

zephraph:
Hi friends!

It's me again. Back with new deprecations and different ways to query the same things 😄. 

I'm trying to fix this: 

image

That `About the collector` part isn't what we want users to see 😄 

I was QA'ing PURCHASE-1861 (which I thought fixed this issue) and much to my chagrin there was still a collector preview in the message snippet! Turns out that in #2326 I fixed the _first_ message, but not the _last_ message (which in this case is what's being shown... there's a conversation with only one message so it's both the first and the last). 

Before this change to get the last message you'd do

```graphql
me {
  conversationConnection {
    edges {
      node {
        lastMessage
      }
    }
  }
}
```

That's all well and good, but in  #2326 I added logic to format the messages to make sure this particular case doesn't happen. It's just in the `MessageConnection` that that happened. So, why not just use `MessageConnection`?

```graphql
me {
  conversationConnection {
    edges {
      node {
        messageConnection(last: 1) {
	   	  edges {
			node {
			  body
			}
		  }
		}
      }
    }
  }
}
```

Other than the verbosity of the query, it's a step up because we have one set of handlers for the message formatting/logic checks. 

---

Other things I did:
- Deprecate `ImpulseID`. Everywhere else we use `internalID` to be representative of the underlying system id and we should keep with that pattern
- Deprecate `lastMessage` because we only need one way to do a thing
- Deprecate `lastMessageID` because that can be retrieved from the message.


src/schema/v2/me/conversation/message.ts
```diff
@@ -39,11 +39,9 @@ export const MessageType = new GraphQLObjectType<any, ResolverContext>({
   interfaces: [NodeInterface],
   fields: {
     ...InternalIDFields,
-    // This alias exists specifically because our fork of Relay Classic did not yet properly support using `__id`
-    // instead of `id`, which lead to Relay overwriting `id` fields with the `__id` value. Thus using a completely
-    // different field name works around this. You should probably not use it.
     impulseID: {
       description: "Impulse message id.",
+      deprecationReason: "Prefer internalID",
```
starsirius, line 44: 👏 V2 migration!

zephraph: It continues! 😄

src/schema/v2/me/conversation/index.ts
```diff
@@ -272,13 +276,16 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
     lastMessage: {
       type: GraphQLString,
       description: "This is a snippet of text from the last message.",
+      deprecationReason: "Prefer querying `messagesConnection(last:1)`",
```
starsirius, line 279: Some benefits of using `lastMessage` directly:

- It's explicitly set/updated in Impulse so we don't need to worry about ordering of the messages in a conversation. Messages are created async and we might want to sort by `order: :asc` to make sure the last one is actually the last message.
- It could be less expensive to query for some cases. For example, the inbox list view can query conversation level stuff and get this last message to show a preview, without querying messages for a conversation.

starsirius:
I'm not sure why the "About the collector" is in the last message. _I think_ it's only in the first message (the inquiry message) because it's generated by [Gravity email template](https://github.com/artsy/gravity/blob/1ed9f89eeb7e824a3fee0961ee898c90c4f09d51/app/views/mail_templates/artwork_inquiry_notification_radiation.html.haml#L24).

zephraph:
In this case there's only one message in the conversation so the last message is the first message. I'm assuming that's why it's present.

src/schema/v2/me/conversation/index.ts
zephraph: That sounds reasonable. Let me take another crack at it.

starsirius:
Ohh I missed it in the description. It's the _only_ message in this conversation.

src/schema/v2/me/conversation/index.ts
zephraph: So, we don't currently have a way (that I can see). The only way that I'd know to fix this is to update

https://github.com/artsy/impulse/blob/ab5d41f676ea37c9b9c9fa07536e191c56cdfaf4/app/api/v1/conversations_endpoint.rb#L95-L102

If we updated it to return the id of the initial message then we could check that against the last updated message and return the `initial_message` instead.

Alternatively we could update the data to not include the blurb... but that seems like a much bigger lift. 

Thoughts?

starsirius: Maybe another option is to check the `order` of the last message. If it's order `1` then we know it's the first message as well. We can probably tweak [`lastMessage`](https://github.com/artsy/metaphysics/blob/42a1752e7a46457fb5cac1ad6c02c7723c6660c6/src/schema/v2/me/conversation/index.ts#L272-L277) to return initial message instead in that case.

zephraph: `order` isn't sent down as apart of `last_message`. It's attached to the `conversations/:id` api response as `_embedded.last_message` and it only has a few fields associated with it. 

image

This will require upstream updates to impluse.

zephraph: Ah, okay, yeah I found it. 

`last_message` is provided here: https://github.com/artsy/impulse/blob/629603865d33f4344bdee1c515ea251ded1619e3/app/api/v1/presenters/conversation_presenter.rb#L28

It uses the `MessagePresenter` class here: https://github.com/artsy/impulse/blob/ab5d41f676ea37c9b9c9fa07536e191c56cdfaf4/app/api/v1/presenters/message_presenter.rb. I can look to add `order` there.

zephraph: Okay, I've went around in circles on this a bit and I'm back to thinking maybe how it stands now is the best option. It's not _optimal_ for sure, but I wonder if it's good enough?  

So let me recap. 

`Snippet` returned on `last_message` (which we store) is returning the full message (including details about the collector that we send to the partner). We obviously can't display that to the collector via metaphysics. 

So if the `last_message` is also the first message of the conversation then we need to use `initial_message` instead. Great. Seems simple enough. The challenge is understanding when it's the first message. 

So, how do we do that? Well, we could use `order` as mentioned above... except order has a good bit of `nil`s. This was discussed [here](https://github.com/artsy/impulse/pull/591#discussion_r411803038). In that case, relying on `order` seems to unreliable to me and I'd rather avoid it. 

I've been talking to @jpotts244 and it looks like we could potentially lift some of the work she did for `first_message` up to the conversation message. But again, that's hitting messages and looking for the first message. It feels like if we're doing that lookup anyway, the current approach isn't the worst. Granted, here we're likely doing an extra http call which isn't the best for conversations list. If we paginate that though, it might not be that bad? 

I'll pair w/ @jpotts244 later on lifting some of the first message logic to the conversation so we can identify that message and grab the id from it to compare against the last message id.

mzikherman: Overall, I think we should continue to use the last messsage/snippet provided by Impulse. This means that one HTTP call (Metaphysics -> Impulse) can render an inbox view. That's a common view for users to hit.

Then, when they are in a specific conversation (and we want to render messages, artwork info, etc.) -> that's where it's good to add in the extra Impulse -> Radiation 'enriching', query for the inquireable info, etc.

It might be a premature optimization, but I would hesitate to add Radiation/extra HTTP calls needed to fulfill the top level Inbox view.

So is the whole issue that when `lastMessage` _is_ a relatively short message, and it's the first message, the snippet includes some of that generated inquiry text? I would wonder how Eigen handles that (presumably it's exposed to the same issue?). But, I would maybe just remove the `About this collector...` string (either in MP, or maybe in Force). That's sort of brittle to need to know about here, but it's also unlikely to change.

zephraph: Just a note, this is also an issue on Eigen in prod: 

image

starsirius: I think the `order` on the initial message might be reliable, since we don't suppress the initial inquiry message from Impulse/Radiation. Looking at the data, it seems that the latest conversation without message order 1 was back in 2015. Maybe we just didn't do it back then. 🤔 

Removing `About this collector` is also a pragmatic solution! Maybe it's worth standardizing our own delimiter in emails.

zephraph: That's fair. I'll re-open to impulse PR and we can look at adding that logic (in addendum to what's here). For now though, this is better than what's in production and I've spent significantly longer on this PR than intended 😓. If everyone is 👍 for this solution as is, I'd like to move forward.

starsirius: Yeah let's move forward with this simple approach and we can come back to the `order` approach when we learn more about the missing `order` issue.

zephraph:
Thanks for the feedback all! Gonna move forward with this one and will follow up with improvements in another PR.

src/schema/v2/me/__tests__/conversation/conversation.test.js
```diff
@@ -13,7 +13,7 @@ describe("Me", () => {
           from_name: "Percy",
           _embedded: {
             last_message: {
-              snippet: "Cool snippet",
+              snippet: "Cool snippet About this collector: Percy is a good cat",
```
mzikherman, line 16: 😹

PR #2361, opened 2020-05-07
[MX-291] Adds homePage.marketingCollectionModule.results

ashfurrow:
This PR adds a new module to the top-level `homePage` field so that we can make queries like this:

```graphql
{
  homePage {
    marketingCollectionsModule {
      results {
        title
        artworksConnection(first: 3) {
          edges {
            node {
              slug
            }
          }
        }
      }
    }
  }
}
```

Here it is working locally:

Screen Shot 2020-05-07 at 11 57 02

This is to support MX work to include three collections (New this Week, Highlights at Auction, and Trending Emerging Artists) in the app's home page. I chose to add this `homePage` module, rather than have the app query the collections itself, so that MP is the source of truth for which collections get displayed in the app.

Screen Shot 2020-05-07 at 12 00 28

GraphQL objects must have _at least one_ field prior to being stitched, so I had to add an `_unusedField` field. I marked it as deprecated with a clear explainer and it always returns `null`, but I'm open to other suggestions.

### DIFF
--- a/src/lib/stitching/kaws/stitching.ts
+++ b/src/lib/stitching/kaws/stitching.ts
@@ -119,7 +119,7 @@ export const kawsStitchingEnvironmentV2 = (
       )}): FilterArtworksConnection
     }
     extend type HomePageMarketingCollectionsModule {
-      results: [MarketingCollection]
+      results: [MarketingCollection]!
     }
   `,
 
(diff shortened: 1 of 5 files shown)


src/lib/stitching/kaws/stitching.ts
```diff
@@ -147,6 +150,34 @@ export const kawsStitchingEnvironmentV2 = (
...
+          resolve: (_source, _args, context, info) => {
+            // We hard-code the collections slugs here in MP so that the app can
+            // display different collections based only on an MP change (and not
+            // an app deploy).
+            return info.mergeInfo.delegateToSchema({
+              schema: kawsSchema,
+              operation: "query",
+              fieldName: "marketingCollections",
+              args: {
+                slugs: [
```
sweir27, line 169: If it becomes more trouble to manage these via PRs, we could also consider making it an ENV var! But like the direction of putting this here vs. the app 🎉.

ashfurrow: An ENV var makes sense – I don't think these are expected to really change that often but if they ever need to, I think we can make the move then 👍

ashfurrow:
After speaking with Sarah, I've amended the last commit to include handling any errors from calling KAWS. Errors from any failed request will still bubble up to the Metaphysics response, which feels correct to me. Unfortunately, the app currently considers any `errors` present to indicate that the _entire_ request has failed, so we still have some follow-up work to do (MX already has [MX-319](https://artsyproduct.atlassian.net/browse/MX-319) to improve the error-handling). But the important thing is that the resolver will always adhere to the schema we've defined 👍 

I've verified the behaviour locally by changing my `KAWS_API_BASE` environment variable to a garbage value.

PR #2363, opened 2020-05-07
Feature Type

dzucconi:
```gql
{
  feature(id: "alserkal-art-week") {
    name
    isActive
    description(format: HTML)
    image {
      width
      height
      url(version: WIDE)
    }
  }
}
```

```json
{
  "data": {
    "feature": {
      "name": "Alserkal Art Week",
      "isActive": true,
      "description": "<p>Alserkal Art Week Online</p>\n<p>Discover new exhibitions by Alserkal Avenue Galleries</p>\n",
      "image": {
        "width": 3240,
        "height": 2160,
        "url": "https://d32dm0rphc51dk.cloudfront.net/wsfiG9lAGOzs0Rey0felsQ/wide.jpg"
      }
    }
  },
  "extensions": {
    "requests": {
      "gravity": {
        "requests": {
          "feature/alserkal-art-week": {
            "time": "0s 148.453ms",
            "cache": false,
            "length": "N/A"
          }
        }
      }
    },
    "requestID": "5b64a810-913c-11ea-a822-9d8c9bdc0114"
  }
}
```

### DIFF
--- a/src/lib/loaders/loaders_without_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_without_authentication/gravity.ts
@@ -1,3 +1,4 @@
+import { Gravity } from "types/runtime"
 import factories from "../api"
 import { uncachedLoaderFactory } from "lib/loaders/api/loader_without_cache_factory"
 import gravity from "lib/apis/gravity"
@@ -27,6 +28,8 @@ export default opts => {
   }
 
   return {
+    featureLoader: (id: string) =>
+      gravityLoader(id => `feature/${id}`)(id).then(Gravity.Feature.check),
     createAccountRequestLoader: gravityLoader(
       "account_requests",
       {},
--- a/src/schema/v2/Feature/FeatureImageType.ts
+++ b/src/schema/v2/Feature/FeatureImageType.ts
@@ -1,0 +1,43 @@
+import { GraphQLString, GraphQLObjectType, GraphQLInt } from "graphql"
+import { Gravity } from "types/runtime"
+import { ResolverContext } from "types/graphql"
+import {
+  FeatureImageVersionEnum,
+  FeatureImageVersion,
+  FEATURE_IMAGE_VERSIONS,
+} from "./FeatureImageVersionEnum"
+
+export const FeatureImageType = new GraphQLObjectType<
+  Gravity.Feature,
+  ResolverContext
+>({
+  name: "FeatureImage",
+  fields: {
+    width: {
+      description: "Unscaled (original) image width",
+      type: GraphQLInt,
+      resolve: ({ original_width }) => original_width,
+    },
+    height: {
+      description: "Unscaled (original) image height",
+      type: GraphQLInt,
+      resolve: ({ original_height }) => original_height,
+    },
+    url: {
+      type: GraphQLString,
+      description:
+        "Ask for a version and you might receive it. Versions are not guaranteed to be present.",
+      args: {
+        version: {
+          type: FeatureImageVersionEnum,
+          defaultValue: FEATURE_IMAGE_VERSIONS.SOURCE.value,
+        },
+      },
+      resolve: ({ image_urls }, args) => {
+        const { version } = args as { version: FeatureImageVersion }
+        const url = image_urls[version.key as keyof typeof image_urls]
+        return url
+      },
+    },
+  },
+})
--- a/src/schema/v2/Feature/FeatureType.ts
+++ b/src/schema/v2/Feature/FeatureType.ts
@@ -1,0 +1,38 @@
+import {
+  GraphQLString,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLBoolean,
+} from "graphql"
+import { Gravity } from "types/runtime"
+import { ResolverContext } from "types/graphql"
+import { SlugAndInternalIDFields } from "schema/v2/object_identification"
+import { markdown } from "schema/v2/fields/markdown"
+import { FeatureImageType } from "./FeatureImageType"
+
+export const FeatureType = new GraphQLObjectType<
+  Gravity.Feature,
+  ResolverContext
+>({
+  name: "Feature",
+  description: "A Feature",
+  fields: () => ({
+    ...SlugAndInternalIDFields,
+    name: {
+      type: new GraphQLNonNull(GraphQLString),
+    },
+    isActive: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      resolve: ({ active }) => active,
+    },
+    description: markdown(),
+    image: {
+      type: FeatureImageType,
+      resolve: feature => {
+        // If there is no available image return null here rather than down-tree (!)
+        if (feature.image_versions.length === 0) return null
+        return feature
+      },
+    },
+  }),
+})
--- a/src/schema/v2/fields/markdown.ts
+++ b/src/schema/v2/fields/markdown.ts
@@ -2,35 +2,44 @@ import Format from "schema/v2/input_fields/format"
 import { GraphQLString, GraphQLFieldConfig } from "graphql"
 import { isExisty } from "lib/helpers"
 import marked from "marked"
+import { ResolverContext } from "types/graphql"
 
-export function formatMarkdownValue(value, format) {
+type Format = "html" | "markdown"
+type Value = string | null | undefined
+
+export const formatMarkdownValue = (value: string, format: Format): string => {
   if (format === "html" || format === "markdown") {
     const renderer = new marked.Renderer()
+
     marked.setOptions({
       renderer,
       gfm: true,
-      tables: true,
       breaks: true,
       pedantic: false,
       sanitize: false,
       smartypants: false,
+      tables: true,
     })
+
     return marked(value)
   }
 
   return value
 }
 
-export function markdown(fn?: any): GraphQLFieldConfig<any, any> {
+export const markdown = <T>(
+  fn?: (response: T) => Value
+): GraphQLFieldConfig<T, ResolverContext> => {
   return {
     type: GraphQLString,
     args: {
       format: Format,
     },
     resolve: (obj, { format }, _, { fieldName }) => {
-      const value = fn ? fn(obj) : obj[fieldName]
+      const value: Value = fn ? fn(obj) : obj[fieldName]
 
-      if (!isExisty(value)) return null
+      if (!isExisty(value) || typeof value !== "string") return null
+
       return formatMarkdownValue(value, format)
     },
   }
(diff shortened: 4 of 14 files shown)


package.json
```diff
@@ -92,6 +92,7 @@
     "patch-package": "6.2.0",
     "performance-now": "2.1.0",
     "postinstall-prepare": "1.0.1",
+    "proportional-scale": "^3.0.0",
```
dzucconi, line 95: Tiny one of mine: https://github.com/dzucconi/proportional-scale — just a typesafe and better tested version of the resizing function that's already here, basically.

src/lib/loaders/loaders_without_authentication/gravity.ts
```diff
@@ -27,6 +28,8 @@ export default opts => {
   }
 
   return {
+    featureLoader: (id: string) =>
+      gravityLoader(id => `feature/${id}`)(id).then(Gravity.Feature.check),
```
dzucconi, line 32: ✔️

src/schema/v2/Feature/FeatureImageType.ts
```diff
@@ -0,0 +1,75 @@
...
+          ),
+        }
+
+        const image: TScaledImage = {
+          ...scaled,
+          src: urls["1x"],
+          srcSet: `${urls["1x"]} 1x, ${urls["2x"]} 2x`,
+        }
+
+        return image
```
dzucconi, line 71: I'll extract out and write some specs for this function, which is really the only actual logic in this PR.

src/schema/v2/Feature/FeatureType.ts
```diff
@@ -0,0 +1,34 @@
...
+  fields: () => ({
+    ...SlugAndInternalIDFields,
+    name: {
+      type: new GraphQLNonNull(GraphQLString),
+    },
+    isActive: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      resolve: ({ active }) => active,
+    },
+    description: markdown(),
```
dzucconi, line 28: You can now use this by threading through the type of the response as well 🤷 

```tsx
foo: markdown<Gravity.Foo>(({ bar }) => bar)
```

src/schema/v2/fields/markdown.ts
```diff
@@ -2,35 +2,44 @@ import Format from "schema/v2/input_fields/format"
...
+
     marked.setOptions({
       renderer,
       gfm: true,
-      tables: true,
       breaks: true,
       pedantic: false,
       sanitize: false,
       smartypants: false,
+      tables: true,
```
dzucconi, line 21: This option gets dropped in the 1.0.0 release of marked, if we ever want to upgrade.

src/schema/v2/fields/markdown.ts
```diff
@@ -2,35 +2,44 @@ import Format from "schema/v2/input_fields/format"
...
   return {
     type: GraphQLString,
     args: {
       format: Format,
     },
     resolve: (obj, { format }, _, { fieldName }) => {
-      const value = fn ? fn(obj) : obj[fieldName]
+      const value: Value = fn ? fn(obj) : obj[fieldName]
+
+      if (!isExisty(value) || typeof value !== "string") return null
```
dzucconi, line 41: `isExisty` should be a typeguard?

src/schema/v2/image/ScaledImage.ts
```diff
@@ -0,0 +1,28 @@
...
+export const ScaledImage = new GraphQLObjectType<TScaledImage, ResolverContext>(
+  {
+    name: "ScaledImage",
+    fields: {
+      scale: { type: new GraphQLNonNull(GraphQLFloat) },
+      width: { type: new GraphQLNonNull(GraphQLInt) },
+      height: { type: new GraphQLNonNull(GraphQLInt) },
+      paddingBottom: { type: new GraphQLNonNull(GraphQLString) },
+      src: { type: new GraphQLNonNull(GraphQLString) },
+      srcSet: { type: new GraphQLNonNull(GraphQLString) },
```
dzucconi, line 25: Just created a new, more sensible version of the resized image type with non-nullable properties and dropping the `url` name.

src/schema/v2/Feature/FeatureImageType.ts
```diff
@@ -0,0 +1,75 @@
...
+  GraphQLNonNull,
+} from "graphql"
+import { Gravity } from "@artsy/quartz"
+import { scale } from "proportional-scale"
+import { ResolverContext } from "types/graphql"
+import proxy from "schema/v2/image/proxies"
+import { ScaledImage, TScaledImage } from "../image"
+
+export const FeatureImageType = new GraphQLObjectType<
+  Gravity.Feature,
```
dzucconi, line 14: Does anyone have strong opinions on type naming conventions? I should probably rename this to be `Gravity.TFeature` for clarity?

src/schema/v2/Feature/FeatureImageType.ts
```diff
@@ -0,0 +1,75 @@
...
+        })
+
+        const urls = {
+          "1x": proxy(url, "resize", scaled.width, scaled.height, args.quality),
+          "2x": proxy(
+            url,
+            "resize",
+            scaled.width * 2,
+            scaled.height * 2,
+            args.quality - 25 // Can get away with lower quality for higher DPI
```
dzucconi, line 61: This can probably be an argument as well `dprQualityOffset` or something.

package.json
sweir27: oooh interesting

src/lib/loaders/loaders_without_authentication/gravity.ts
sweir27: 😮 !

src/schema/v2/Feature/FeatureImageType.ts
sweir27: Why `TFeature`?

src/lib/loaders/loaders_without_authentication/gravity.ts
damassi: @dzucconi - mind briefly explaining this form / what it does?

src/schema/v2/Feature/FeatureImageType.ts
damassi: 👌 Yah seems like this can be abstracted away at some point

src/schema/v2/image/proxies/embedly.ts
```diff
@@ -1,25 +1,32 @@
 import urljoin from "url-join"
 import qs from "qs"
 import config from "config"
+import { Mode } from "."
```
damassi, line 4: I _think_ this kind of import path (`.`) will blow up on linux

src/schema/v2/image/ScaledImage.ts
```diff
@@ -0,0 +1,28 @@
...
+} from "graphql"
+import { scale } from "proportional-scale"
+import { ResolverContext } from "types/graphql"
+
+export type TScaledImage = ReturnType<typeof scale> & {
+  src: string
+  srcSet: string
+}
+
+export const ScaledImage = new GraphQLObjectType<TScaledImage, ResolverContext>(
```
dzucconi, line 16: Ah this should actually be ScaledImageType, not ScaledImage, which would let me avoid the TScaledImage naming.

src/lib/loaders/loaders_without_authentication/gravity.ts
dzucconi: @damassi sure — basically just compares the response to the type we're expecting at runtime. If it differs then it throws an error there rather than allowing it to propagate down-tree.

So what we'd want to do actually is write a custom error extension to print out the exact path etc, right now since this just throws a normal runtime error you'd get say this (in this case I took the response and omitted `name` which is considered non-nullable currently):

dzucconi: Or let's say Gravity was feeling creative and decided to put a number there:



Ultimately in practice though what we'd see is errors in Sentry that are much simpler to debug rather than filtering out to different resolvers. We'd simply update the typing to be accurate and fix the bug with this newfound knowledge 🧠

src/schema/v2/image/proxies/embedly.ts
dzucconi: ah good call

src/lib/loaders/loaders_without_authentication/gravity.ts
damassi: Makes sense -- ty 👍

src/schema/v2/Feature/FeatureImageType.ts
```diff
@@ -0,0 +1,75 @@
...
+          ),
+        }
+
+        const image: TScaledImage = {
+          ...scaled,
+          src: urls["1x"],
+          srcSet: `${urls["1x"]} 1x, ${urls["2x"]} 2x`,
```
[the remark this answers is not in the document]
dzucconi: We don't handle DPR pretty much anywhere — generally we just shovel an oversized image into a smaller tag and hope for the best, which leads to images being scaled by the client and oversized images loading on 1x screens.

dzucconi:
@sweir27 any typing is better than no typing. I can manually pull the type over, but it's getting a type one way or the other 😅

src/schema/v2/Feature/FeatureImageType.ts
```diff
@@ -0,0 +1,75 @@
...
+      resolve: ({ image_urls }, args) => image_urls[args.version],
+    },
+    resized: {
+      type: ScaledImage,
+      args: {
+        width: { type: GraphQLInt },
+        height: { type: GraphQLInt },
+        quality: { type: GraphQLInt, defaultValue: 80 },
+        version: {
+          type: new GraphQLNonNull(GraphQLString),
```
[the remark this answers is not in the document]
dzucconi: The thing of it is that they are possibly undefined for literally every field, so: we can't define them via enum since we only know them at runtime and on a per-request basis. We like sort of know what MIGHT be there MOST of the time. Though this varies from model to model. 🤕

src/schema/v2/Feature/FeatureImageType.ts
dzucconi: I'll pull this off for now and encode some better information about the model into the default image type just so we can get this merged now. Will follow up though because it is necessary going forward.

sweir27:
@dzucconi totally 🙈! I'll advocate for manually pulling that type over while we figure out how we want to scale. Feels more incremental/will give us the space to choose the right path.

dzucconi:
@sweir27 — pulled over the typings from quartz now. Should be fine to merge on green.

sweir27:
@dzucconi 👍#mergeongreen

PR #2364, opened 2020-05-07
Preserve empty array in query string for viewing room artworks (GALL-2836)

starsirius:
https://artsyproduct.atlassian.net/browse/GALL-2836

In https://github.com/artsy/metaphysics/pull/2351, we changed the behavior globally. Unfortunately, there are multiple places clients are sending empty array params but expecting APIs (e.g. Gravity) to ignore that and return non-empty results.

It seems not easy to exhaustively fix everywhere so this is a fix just for viewing room artworks.

Tests are trying to catch issues by this workaround in case the underlying implementation changes. Ideally I'd be interested in more integration type of tests but I'm not sure how. So my approach here is:

- Test that we are mapping argument properly for dataloader to use.
- Then test with the mapped argument, the query string conversion that dataloader is relying on converts the argument to desired query string.

Welcome any ideas!

## Before
Screen Shot 2020-05-04 at 9 25 21 AM

## After
Screen Shot 2020-05-07 at 5 29 44 PM

### DIFF
--- a/src/lib/__tests__/helpers.test.ts
+++ b/src/lib/__tests__/helpers.test.ts
@@ -1,10 +1,11 @@
 import {
   exclude,
-  toKey,
   isExisty,
   removeNulls,
-  stripTags,
   resolveBlueGreen,
+  stripTags,
+  toKey,
+  toQueryString,
 } from "lib/helpers"
 
 describe("exclude", () => {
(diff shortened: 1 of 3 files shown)


src/lib/__tests__/helpers.test.ts
```diff
@@ -25,6 +26,26 @@ describe("exclude", () => {
...
 })
 
+describe("toQueryString", () => {
+  it("stringifies regular arraies", () => {
+    expect(
+      toQueryString({ ids: [13, 27], visible: true, availability: "for sale" })
+    ).toBe("availability=for%20sale&ids%5B%5D=13&ids%5B%5D=27&visible=true")
+  })
+
+  it("ignores empty arraies", () => {
```
starsirius, line 36: 🙈 Just document current behavior.

PR #2368, opened 2020-05-08
Adds runtypes

dzucconi:
I'm not pulling over the script (yet) as it's not totally thought out and I still feel like there's value to that package being separate, if only conceptually. If we're going down this route of treating the API as a black-box (which I actually think is much less work than the alternatives, but I'll write up a technical plan for this) then it makes sense to do the error driven codegen and I just foresee difficulties there with that tooling being co-located.


src/types/gravity/artworkResponse.d.ts
```diff
@@ -1,174 +0,0 @@
-// This is generated from taking a few Artwork responses from
```
sweir27, line 1: Was this not being used anywhere?

sweir27:
One question but otherwise looks good! Merge at will.

src/types/gravity/artworkResponse.d.ts
dzucconi: I thought it wasn't but now I'm seeing that it's ambient and we aren't using strict. So I think removing it and running yarn type-check should be safe. But I'm not 100% so I'll put it back — but this is really annoying.

PR #2377, opened 2020-05-12
[MX-325] Limits home page sales to auctions #trivial

ashfurrow:
We were showing all the sales on the home page instead of just the sales that are auctions (a slight difference, but basically we have some BNMO sales that aren't auctions – see the ticket for more info). This PR updates the home page to return the same data as fetched in the Auctions view in the app:

https://github.com/artsy/eigen/blob/dd7c200a809fa81a423c94fab28097cc261ff15c/src/lib/Scenes/Sales/index.tsx#L115

By fixing this on MP, we'll get this to all users without an app update.

### DIFF
--- a/src/schema/v2/home/home_page_sales_module.ts
+++ b/src/schema/v2/home/home_page_sales_module.ts
@@ -14,6 +14,7 @@ export const HomePageSalesModuleType = new GraphQLObjectType<
         // Check for all sales that are currently running
         const gravityOptions = {
           live: true,
+          is_auction: true,
           size: 10,
           sort: "timely_at,name",
         }


ashfurrow:
Weird. This must be the same issue that @icirellik was seeing [on Slack](https://artsy.slack.com/archives/CP9P4KR35/p1589291122194400). I'm restarting the Heroku dyno to see if that fixes it.

ashfurrow:
Hmm okay I followed the breadcrumbs. Let's lay it out.

On Metaphysics PRs, Peril is configured to download and run a file directly from Metaphysics:

https://github.com/artsy/peril-settings/blob/fbcfe31642526ea806f887a600707526c652623b/peril.settings.json#L23-L25

[This file](https://github.com/artsy/metaphysics/blob/master/peril/compareReactionSchema.ts) tries to grab [Force's version of Reaction](https://github.com/artsy/force/blob/8ec1257d340696526c89e441800c63c4e038f04e/package.json#L80) and then download the specific schema that Reaction is using at that tag:

https://github.com/artsy/metaphysics/blob/8bbde1b113ccc9d5edf909fd657452ad166f6d75/peril/compareReactionSchema.ts#L12-L13

The current version is producing a 404: https://github.com/artsy/reaction/raw/v26.36.3/data/schema.graphql

Looks like [Reaction got updated in Force](https://github.com/artsy/force/commit/f7bfa52c40b82c974f2bf19b6a5f4c436820f6a7#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) but the tag was never created (ie: `v26.36.3` is missing from [Reaction's releases](https://github.com/artsy/reaction/releases)). I looked on [Reaction's CI](https://circleci.com/gh/artsy/reaction) but didn't see any failed deploys.

@zephraph any insight here? I think this problem will go away with the next Reaction deploy but wanted to flag.

ashfurrow:
Looks like https://github.com/artsy/force/pull/5551 failed for the same reason.

ashfurrow:
@brainbicycle you can ignore the Peril error above, this should be good to go 👍

zephraph:
That's some good sleuthing @ashfurrow! Yeah, I've noticed an uptick in these failures recently.

My hypothesis of what's happening in these failures is something like this:

- The PR deploys a canary version say... 1.2.3-canary
- When trying to do a master deploy, it tries to publish 1.2.3 to npm... but npm says it's already published because it's doing a partial match. 

I don't necessarily fully understand, because it's not happening all the time. For now, I'm working on https://github.com/artsy/orbs/pull/104 to prove out that hypothesis. If that stabilizes us, I'll see if I can make an upstream change to permanently resolve the issue.

brainbicycle:
Looks good to me and nice detective work on the peril failures 🕵️

PR #2378, opened 2020-05-12
Remove error handling from meLoader

yuki24:
This seems like a pattern we had adapted a long time ago that no longer seems to be valid. WIth this change, we will be able to see the error in the `errors` property in the response:

catch-error

### DIFF
--- a/src/schema/v2/me/index.ts
+++ b/src/schema/v2/me/index.ts
@@ -162,7 +162,7 @@ const MeField: GraphQLFieldConfig<void, ResolverContext> = {
       "unreadNotificationsCount",
     ]
     if (includesFieldsOtherThanSelectionSet(info, fieldsNotRequireLoader)) {
-      return meLoader().catch(() => null)
+      return meLoader()
     }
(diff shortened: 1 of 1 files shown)


yuki24:
The schema update is unexpected... I'll remove and rebase.

damassi:
@yuki24 - can you update the screenshot with the errors key? There's no feedback there, just `null`.

yuki24:
@damassi Not sure if I understand. The current `master` returns just `me: null` (the screenshot on the right) and this PR updates it to look like the screenshot on the left.

damassi:
Just commenting on 
> WIth this change, we will be able to see the error in the errors property in the response
Which i'm not seeing in the screenshot

yuki24:
@damassi You are probably looking at them in the wrong order...

catch-error2

damassi:
Lol @yuki24 - I see whats up. Maybe reverse those screenshots with "before" and "after" -- visually it looked like we were getting less reporting.

mzikherman:
Merging! I think the Peril comment is safely ignored.

PR #2382, opened 2020-05-12
Add formattedEndAt to viewingRoom

damassi:
This adds a new `formattedEndAt` field to our stitched viewingRoom field off of Gravity. 

This doesn't match the comps exactly, but gets close via the [moment.distance](https://github.com/SokratisVidros/moment.distance.js) plugin.

### DIFF
--- a/src/lib/stitching/gravity/__tests__/stitching.test.ts
+++ b/src/lib/stitching/gravity/__tests__/stitching.test.ts
@@ -3,111 +3,182 @@ import {
   getGravityMergedSchema,
   getGravityStitchedSchema,
 } from "./testingUtils"
+import moment from "moment"
 
-it("extends the ViewingRoom type with an artworksConnection field", async () => {
-  const mergedSchema = await getGravityMergedSchema()
-  const artworkConnectionFields = await getFieldsForTypeFromSchema(
-    "ViewingRoom",
-    mergedSchema
-  )
+describe("gravity/stitching", () => {
+  describe("#artworksConnection", () => {
+    it("extends the ViewingRoom type with an artworksConnection field", async () => {
+      const mergedSchema = await getGravityMergedSchema()
+      const artworkConnectionFields = await getFieldsForTypeFromSchema(
+        "ViewingRoom",
+        mergedSchema
+      )
 
-  expect(artworkConnectionFields).toContain("artworksConnection")
-})
+      expect(artworkConnectionFields).toContain("artworksConnection")
+    })
 
-it("extends the ViewingRoom type with a partner field", async () => {
-  const mergedSchema = await getGravityMergedSchema()
-  const artworkConnectionFields = await getFieldsForTypeFromSchema(
-    "ViewingRoom",
-    mergedSchema
-  )
+    it("resolves the artworks field on ViewingRoom as a paginated list", async () => {
+      const { resolvers } = await getGravityStitchedSchema()
+      const { artworksConnection } = resolvers.ViewingRoom
+      const info = { mergeInfo: { delegateToSchema: jest.fn() } }
 
-  expect(artworkConnectionFields).toContain("partner")
-})
+      artworksConnection.resolve(
+        { artworkIDs: ["1", "2", "3"] },
+        { first: 2 },
+        {},
+        info
+      )
 
-it("resolves the artworks field on ViewingRoom as a paginated list", async () => {
-  const { resolvers } = await getGravityStitchedSchema()
-  const { artworksConnection } = resolvers.ViewingRoom
-  const info = { mergeInfo: { delegateToSchema: jest.fn() } }
+      expect(info.mergeInfo.delegateToSchema).toHaveBeenCalledWith({
+        args: { ids: ["1", "2", "3"], first: 2 },
+        operation: "query",
+        fieldName: "artworks",
+        schema: expect.anything(),
+        context: expect.anything(),
+        info: expect.anything(),
+      })
+    })
 
-  artworksConnection.resolve(
-    { artworkIDs: ["1", "2", "3"] },
-    { first: 2 },
-    {},
-    info
-  )
+    it("converts empty artworkIDs argument", async () => {
+      const { resolvers } = await getGravityStitchedSchema()
+      const { artworksConnection } = resolvers.ViewingRoom
+      const info = { mergeInfo: { delegateToSchema: jest.fn() } }
 
-  expect(info.mergeInfo.delegateToSchema).toHaveBeenCalledWith({
-    args: { ids: ["1", "2", "3"], first: 2 },
-    operation: "query",
-    fieldName: "artworks",
-    schema: expect.anything(),
-    context: expect.anything(),
-    info: expect.anything(),
+      artworksConnection.resolve({ artworkIDs: [] }, { first: 2 }, {}, info)
+
+      expect(info.mergeInfo.delegateToSchema).toHaveBeenCalledWith({
+        args: { ids: [null], first: 2 },
+        operation: "query",
+        fieldName: "artworks",
+        schema: expect.anything(),
+        context: expect.anything(),
+        info: expect.anything(),
+      })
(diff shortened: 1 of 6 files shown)


src/lib/stitching/gravity/__tests__/stitching.test.ts
```diff
@@ -3,111 +3,177 @@ import {
...
-    args: { ids: ["1", "2", "3"], first: 2 },
-    operation: "query",
-    fieldName: "artworks",
-    schema: expect.anything(),
-    context: expect.anything(),
-    info: expect.anything(),
+import moment from "moment"
+
+describe("gravity/stitching", () => {
+  describe("#artworksConnection", () => {
```
damassi, line 9: Just organizes the test a bit better, by field

src/lib/stitching/gravity/__tests__/stitching.test.ts
```diff
@@ -3,111 +3,177 @@ import {
...
-  artworksConnection.resolve({ artworkIDs: [] }, { first: 2 }, {}, info)
-
-  expect(info.mergeInfo.delegateToSchema).toHaveBeenCalledWith({
-    args: { ids: [null], first: 2 },
-    operation: "query",
-    fieldName: "artworks",
-    schema: expect.anything(),
-    context: expect.anything(),
-    info: expect.anything(),
+  describe("#formattedEndAt", () => {
```
damassi, line 60: New additions

src/lib/stitching/gravity/stitching.ts
```diff
@@ -74,6 +81,32 @@ export const gravityStitchingEnvironment = (
             })
           },
         },
+        formattedEndAt: {
+          fragment: gql`
+            ... on ViewingRoom {
+              startAt
+              endAt
```
starsirius, line 88: I think `startAt` and `endAt` are local time (though its ISO8601 value _may_  include UTC as the timezone) and there is a separate `timezone` field to specify the local time timezone.

If Communities viewing rooms always specify UTC as the timezone, this will work. But maybe we want to take the timezone into account when calculating the diff.

src/lib/stitching/gravity/stitching.ts
```diff
@@ -74,6 +81,32 @@ export const gravityStitchingEnvironment = (
...
+            }
+            if (endAt < startAt) {
+              return "Closed"
+            }
+
+            const label =
+              "Closes in " +
+              (moment.duration(
+                endAt.diff(startAt)
+              ) as DistancePlugin).distance()
```
starsirius, line 106: Also curious if moment [Time to X](https://momentjs.com/docs/#/displaying/to/) can accomplish similar thing or maybe it's not that close to design?

damassi: The distance plugin covers many more use-cases and models itself after rails. I can copy the source into MP if the dep is a concern. It's not too big. 

I'm not the greatest with dates and happy to pass this off, otherwise we can iterate.

src/lib/stitching/gravity/stitching.ts
damassi: Added [here](https://github.com/artsy/metaphysics/pull/2382/files#diff-573ae00d2f637334b78f6df4edf48571R108) but not sure if this is the correct way.

damassi:
@dzucconi - I'll investigate making this more generic. For stitched services however one can't overwrite fields, so it would have to be added as a new date type in Gravity; or as it currently is in this PR, a prefixed `formatted` would have to be added. Unfortunately, in a stitched context, common `graphql-js` types can't be used, so there's some additional divergence that would take palace there two.  

So in the end there's a bit of complexity in three separate dimensions. It seems like making this addition on the Gravity side would make the most sense. What do you think would be more preferable?

dzucconi:
Ah, ok. I can see why everyone keeps saying 'stitching is weird/annoying/etc'. 😄 

This relative option has apparently never come up before, so there's that.

There's some value in leaning on the existing `date` field if we ever plan on displaying a non-relative formatted date. Though that's easily something that can be dealt with when it comes up since the type will remain the same.

damassi:
> Ah, ok. I can see why everyone keeps saying 'stitching is weird/annoying/etc'

There's a bit of a disconnect between it and `graphql-js` as Apollo prefers the SDL format. Most of our schema has been written in a way that the two don't interface as nicely as one would like. We'd need a stitching primitive in `graphql-js` to handle this well. The same would apply to a federated schema, again SDL focused. Tradeoffs /shrug

src/lib/stitching/gravity/stitching.ts
```diff
@@ -74,6 +81,35 @@ export const gravityStitchingEnvironment = (
...
+              return null
+            }
+            if (endAt < startAt) {
+              return "Closed"
+            }
+
+            const label =
+              "Closes in " +
+              (moment.duration(
+                endAt.tz(timeZone).diff(startAt)
```
starsirius, line 108: Perhaps this should be relative to now rather than `startAt`?

damassi: @starsirius - would you mind pasting a code snippet of what you'd like to see here?

src/lib/stitching/gravity/stitching.ts
```diff
@@ -74,6 +81,35 @@ export const gravityStitchingEnvironment = (
...
+          resolve: ({ startAt: _startAt, endAt: _endAt, timeZone }) => {
+            const startAt = moment(_startAt)
+            const endAt = moment(_endAt)
+
+            timeZone = timeZone || moment.tz.guess()
+
+            if (endAt > moment().add(30, "days")) {
+              return null
+            }
+            if (endAt < startAt) {
```
starsirius, line 101: Probably `if (now > endAt)` ?

And do we need to worry about `if (now < startAt)`?

src/lib/stitching/gravity/stitching.ts
starsirius: I'm not too concerned about the dep as long as it's well-tested and served our needs! 👍

damassi:
@starsirius - Rather than continue back and forth / trial and error around date logic, would you mind pasting what you would like to see in the resolver function? I can update it, or can pass this work back to you in the interest of efficency.

starsirius:
@damassi I didn't mean to dictate the implementation but would be happy to come up with one version. I'll try to paste something here later today! I just realized I was also assuming the requirements. To clarify, are we looking for
- the total time a viewing room will run (i.e. from start to end), or
- when a viewing room will end relative to now?

I was assuming the latter.

damassi:
The latter. The only bit of logic necessary is that if the `endAt` date is greater than 30 days from now, then hide (return null or some flag to hid in UI). According to Joanna VRs will generally run between 7 and 30 days. So for now, basing things off of endAt will contain things nicely within this window of startAt and endAt, relative to now.

starsirius:
I did some research to persist local time (without converting to UTC) in Rails (i.e. Gravity), but couldn't find a satisfying solution. One approach is [this](https://stackoverflow.com/questions/35859549/saving-datetime-in-utc-isnt-accurate-sometimes) but still has some caveats. Maybe it's not worth the effort at this point to introduce that change in Gravity modeling. That is, we will still convert `startAt` and `endAt` to UTC rather than persisting what user literally inputted.

With that, we don't need the time zone information to calculate the time difference because the `startAt` and `endAt` already represent a certain point in time. So I think we can just slightly tweak your solution like below:

```typescript
resolve: ({ startAt: _startAt, endAt: _endAt }) => {
  const startAt = moment(_startAt)
  const endAt = moment(_endAt)
  const now = moment()

  if (now < startAt || endAt > now.clone().add(30, "days")) {
    return null
  }

  if (now > endAt) {
    return "Closed"
  }

  return "Closes in " + (moment.duration(endAt.diff(now)) as DistancePlugin).distance()
}
```

How to persist user-entered local time is probably a bigger topic to discuss. Thanks for the patience here!

PR #2389, opened 2020-05-14
OrderedSet additions to support Feature/campaign pages

dzucconi:
One of the things I did here was just flesh out [the remaining valid item types](https://github.com/artsy/gravity/blob/0e6bd5ac33aa8494403af5579fa5dbfe8d87b931/app/models/domain/ordered_set.rb#L26) (sans `OrderedSet` itself) so as to avoid errors on existing valid features that may have them. We don't (and won't) have to query on them, but the way the union type is set up is such that [it'll throw errors when resolving the type.](https://github.com/artsy/metaphysics/blob/d297587c138b2ea75874b64971a40cc1ec524e15/src/schema/v2/item.ts#L25)

If we want to actually restrict the features to a subset of these types we'll have to alter [the item_type filter param in Gravity](https://github.com/artsy/gravity/blob/master/app/models/domain/ordered_set.rb#L18) to accept an array. Though in the end this doesn't really get us much of anything because any of the types generated wouldn't take this into account anyway. I think it's OK to just not support this in the UI.


src/schema/v2/Feature/FeatureType.ts
```diff
@@ -41,18 +42,26 @@ export const FeatureType = new GraphQLObjectType<
     },
     setsConnection: {
       type: OrderedSetConnection.connectionType,
-      args: pageable(),
+      args: pageable({
+        sort: {
+          type: OrderedSetSortsEnum,
+          defaultValue: OrderedSetSortsEnum.getValue("KEY"),
```
sweir27, line 48: Should this be `KEY_ASC`?

dzucconi: Good catch. Wish the types inferred the valid keys :/

dzucconi:
I'm going to pull off that last commit and switch back to the way I was previously doing the enum value thing (accessing off a constant) — the "correct" way to do this is apparently `OrderedSetSortsEnum.getValue("KEY_ASC")?.value` which is gross IMO.

dzucconi:
OK, now with typesafe `defautlValue`

src/schema/v2/Feature/FeatureType.ts
```diff
@@ -41,18 +45,26 @@ export const FeatureType = new GraphQLObjectType<
     },
     setsConnection: {
       type: OrderedSetConnection.connectionType,
-      args: pageable(),
+      args: pageable({
+        sort: {
+          type: OrderedSetSortsEnum,
+          defaultValue: ORDERED_SET_SORTS.KEY_ASC.value,
```
sweir27, line 51: 👍

sweir27:
Looks gooood!

PR #2391, opened 2020-05-15
[ViewingRoom] Updates formattedTimeAt logic

damassi:
Update to https://github.com/artsy/metaphysics/pull/2382, based on comments. 

Thanks for looking into this @starsirius!


starsirius:
Thanks for the patience, @damassi ! We can revisit when we have a good way to actually store user-entered local time on the backend!

PR #2396, opened 2020-05-18
Update relay types when updating Eigen's copy of MP schema

mdole:
Following up on https://github.com/artsy/eigen/pull/3319, this PR adds a conditional `yarn relay` that is triggered when updating Eigen's copy of MP schema.

@ds300 I'm not sure how to go about testing this, either writing tests or just running them in my terminal. Any advice for me?

### DIFF
--- a/scripts/push-schema-changes.js
+++ b/scripts/push-schema-changes.js
@@ -49,7 +49,7 @@ async function main() {
         owner: "artsy",
         repo: "reaction",
       },
-      ...updateSchemaAction(),
+      ...updateSchemaAction(true),
     })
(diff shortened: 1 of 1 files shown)


ds300:
I think we probably want to do it in both eigen and reaction, no?

As for testing, I'm not too concerned about separate tests for this, as long as the CI build fails if there's an error, which should be the case already.

mdole:
oh yeah! good call. updated in 74626dd; if the rest looks good then feel free to merge!

ds300:
cool let's give it a go

PR #2397, opened 2020-05-18
Exposes new fields on Feature and FeaturedLink

sweir27:
Corresponding Gravity PR: https://github.com/artsy/gravity/pull/13069

Exposes `subheadline` and `callout` on `Feature` and `description` on `FeaturedLink`.


src/types/runtime/gravity/FeaturedLink.ts
```diff
@@ -15,6 +15,7 @@ export const FeaturedLink = Record({
   href: String,
   title: String,
   subtitle: String,
```
sweir27, line 17: @dzucconi shouldn't this be `.Or(Null)`? (Same with `title` and `href`. Looks like they are all nullable on gravity's side).

dzucconi: Yeah probably — one of the things I was thinking about doing with the generation is treat `""` (empty string) as null when specifying types, since that's how it works in practice in JSX conditional rendering. In this case all of the links that were crawled in staging never had actually `null` values. But they frequently had empty string.

sweir27: Interesting. I wonder if that's an issue in torque (like maybe it passes an empty string if no one enters a value?). I'll update to allow nullability in any case!

dzucconi: Oh ha, yeah almost certainly — I see the default is actually nil in Gravity which is surprising since we pass `default: ''` for a ton of things

src/types/runtime/gravity/Feature.ts
```diff
@@ -14,6 +14,8 @@ export const Feature = Record({
   id: String,
   name: String,
   description: String.Or(Null),
+  subheadline: String.Or(Null),
+  callout: String.Or(Null),
```
dzucconi, line 18: So one thing to note here is — nothing in production is using the runtime validation yet since I only enabled it on the stuff we're using for the new feature pages — but if this was deployed before these fields were in the Gravity responses it'd error since it'd expect the keys to at least be defined. I'm going to write a lil wrapper for the checks that ignores the error and just logs it for the time being.

PR #2399, opened 2020-05-18
Supports Markdown on OrderedSets; generic ImageType on Features

dzucconi:
OK, yeah I missed this one :)


dzucconi:
What's this about?

dzucconi:
Impressed this actually broke a spec.

```
    -     "description": "",
    +     "description": null,
```

Which is interesting. Technically a "breaking change" though in practice it wouldn't matter. I'm also sort of vaguely surprised we never implemented a custom string scalar to hack around pretty pervasive reality of our API responses.

_schemaV2.graphql
```diff
@@ -4285,24 +4285,6 @@ type FeaturedLink {
...
-  url(version: FeatureImageVersion): String
-}
-
-enum FeatureImageVersion {
-  LARGE_RECTANGLE
-  WIDE
-  SQUARE
-  SOURCE
-}
-
```
dzucconi, line 4305: I didn't wind up using this (though it would be nice to have enums for all the sizes). Just switching over to the generic image type to inherit the crop/resize fields. This winds up not being a breaking change either which is good.

dzucconi:
So yeah, we can ignore these warnings

```
Type 'FeatureImage' was removed
Type 'FeatureImageVersion' was removed
Field 'Feature.image' changed type from 'FeatureImage' to 'Image'
```

sweir27:
I _think_ we should also add this to `FeaturedLink`'s [`subtitle`](https://github.com/artsy/metaphysics/blob/master/src/schema/v2/featured_link.ts#L26) field.

sweir27:
#mergeongreen

PR #2400, opened 2020-05-18
Enable Datadog GraphQL Tracing

icirellik:
Problem

Datadog tracing is not enabled for graphql and enabling the graphql plugin doesn't "just work" due to a bug in `dd-trace-js`.

For more information read the following issue:
https://github.com/DataDog/dd-trace-js/issues/951

Solution

The solution is to patch our copy of Datadog with the intended fix and upgrade once the bug has been resolved.

metaphysics-dd-trace


joeyAghion:
👍 from me.

Out of curiosity, how do these patches react when someone upgrades the associated library?

icirellik:
@joeyAghion `patch-package` attempts to run them and will notify you if they don't apply cleanly.

PR #2403, opened 2020-05-19
 Introduce `Me.canRequestEmailConfirmation`

dleve123:
Builds off of https://github.com/artsy/metaphysics/pull/2402 (but OK to merge this in before that I think)

## Problem

One feature of email verification is the ability for an existing user to
verify their email from web or iOS via a banner on web and a sash on
iOS.

One important aspect of these UIs is knowing _whether it is
appropriate to display them or not_.

Moreover, we have been bitten a bit in the past by pushing too much of
this "should I display this UI" logic directly to Reaction or Eigen.
This has resulted in:

1. High risk of code duplication and therefore drift between web and iOS
   clients

2. iOS binary lock-in: Important "should I display this UI" logic lives
   forever in old binaries of Eigen, making it hard for our engineering
   team to ensure a particular experience and locking us in to old APIs.

## Solution

* Introduce `me.canRequestEmailConfirmation`, a field which will evolve
  in its implementation, but remain consistent in its contract.

### User confirmed; cannot request email confirmation

user-already-confirmed-cannot-request-email-confirmation

### User not confirmed; can request email confirmation

user-unconfirmed-can-request-email-confirmation

Jira: https://artsyproduct.atlassian.net/browse/AUCT-997

### DIFF
--- a/src/schema/v2/me/index.ts
+++ b/src/schema/v2/me/index.ts
@@ -56,6 +56,12 @@ const Me = new GraphQLObjectType<any, ResolverContext>({
     email: {
       type: GraphQLString,
     },
+    canRequestEmailConfirmation: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      description:
+        "Allow the user to request email confirmation if they do not have a confirmed email",
+      resolve: ({ confirmed_at }) => !confirmed_at,
+    },
     followsAndSaves: {
       type: new GraphQLObjectType<any, ResolverContext>({
         name: "FollowsAndSaves",
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -5484,6 +5484,9 @@ type Me implements Node {
     last: Int
   ): CreditCardConnection
   email: String
+
+  # Allow the user to request email confirmation if they do not have a confirmed email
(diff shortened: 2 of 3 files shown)


dleve123:
@bhoggard @erikdstock @yuki24 – ready for review. Would like 2 approving reviews before I merge.

cc @dzucconi as we discussed [this in Slack](https://artsy.slack.com/archives/C1HH3KNJG/p1589408873060400) -- decided to not introduce an additional `configuration` field for now. Would rather get more alignment on that and figure out of solution before jumping the gun there. Also, there's currently not any "stack of alerts" to model, but if and when that stack appears, we can introduce a field for that at that time.

src/schema/v2/me/index.ts
```diff
@@ -56,6 +56,12 @@ const Me = new GraphQLObjectType<any, ResolverContext>({
     email: {
       type: GraphQLString,
     },
+    canRequestEmailConfirmation: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      description:
+        "Allow the user to request email confirmation if they do not have a confirmed email",
+      resolve: ({ confirmed_at }) => !confirmed_at,
```
bhoggard, line 63: I don't understand why this has a `!`?

PR #2405, opened 2020-05-19
Run relay-compiler when updating emission/eigen

ds300:
Reimplementation of #2396 

This time we call relay-compiler directly instead of using `yarn`, to avoid the engines check (MP is on node 10 while eigen and reaction are on 12). I also moved the code around a little to make sure that TypeScript can check the data we pass to `updateRepo` (cc @damassi)

Gonna self-merge before y'all wake up to make sure it works this time.


mdole:
Nice! Thanks for tweaking that

PR #2429, opened 2020-05-27
running prettier to get the changes from v2 everywhere

pvinis:
A buuuuunch of changes, mostly adding parens around single arg arrow funcs, which prettier v2 changed the default to `true`. It's a nice default, and the reasoning behind it is that for typescript it makes it easy to add a type immediately, instead of needing to wrap in parens first.

I guess it's better to do this in one go instead of file by file and have fmt changes in every PR for the next 3 years :D.

After running `yarn prettier-project`.

PS. have fun seeing my name in every blame from now on 🤦‍♂️.


sweir27:
Merging in! We can play around with it tomorrow and deploy if things look good. 😄

PR #2441, opened 2020-06-03
Update convection schema

jonallured:
This is mostly just an update but I also added to the convection remap so that the new types I added are prefixed with `Consignments`. I wonder if this mapping is really necessary? Like, if we want that prefix on all the types coming from Convection, maybe it should be more like this:

```ts
  const prefix = "Consignments"
  // Return the new modified schema
  return transformSchema(schema, [
    new RenameTypes((name) => {
      const newName = [prefix, name].join('')
      return newName
    }),
  ])

```

What does the hive mind think of that? /cc @damassi @mzikherman @sweir27

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -4832,12 +4832,13 @@ type ConfirmPickupPayload {
 # Consignment Offer
 type ConsignmentOffer {
   commissionPercentWhole: Int
+  createdAt: ISO8601DateTime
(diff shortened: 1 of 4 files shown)


jonallured:
I'm not sure about that Peril error but this is good to go as far as I can tell. Going to merge. 🤷

PR #2448, opened 2020-06-04
Upgrade DataDog Library

icirellik:
This version of `dd-trace` contains the fix that we were previously
applying with a separate patch.


joeyAghion:
Nice

PR #2465, opened 2020-06-12
Restore geneFamilies to v2 schema

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-2043

This updates the relevant MP fields on order to enable the [/categories](https://www.artsy.net/categories) app to move to the v2 schema

After the feedback below, and some thinking about what really needs to be root field here is where I ended up:

- `geneFamiliesConnection` (renamed from `geneFamilies`) is indeed a root level field and is the entry point for the [the query](https://github.com/artsy/force/blob/7edcfd1c33ded850d78e34a59cf756855dd86646/src/desktop/apps/categories/queries/geneFamilies.js) that the updated `/categories` app will issue, so it remains

- `geneFamily` is not (and tbh never was 😬) a root-level field, so the commented root field is removed altogether

- `orderedSets` might justifiably be a root-level field, but the fact that no-one has needed it in the last year makes me skeptical. (We tend not to use that concept so much anymore.) Originally we used the OrderedSet/OrderedItem concept from Gravity to model some illustrated "featured links" for various genes. In this PR I've abstracted that away into a a new `FeaturedGeneLink` type that does _not_ need to be exposed as a root field.


zephraph:
I know this is still WIP, but would you mind providing a little extra context? I  know there were several efforts to re-introduce root fields and I'm trying to get a better sense of the needs surrounding those issues. 

I paired w/ @mdole on https://github.com/artsy/metaphysics/pull/2469 which looked like it required adding `artworkVersion` back as a root field, but we were able to use the `node` root level field to achieve the needs. 

While that's an option, I'm not necessarily suggesting that's required here. Just looking to learn more.

anandaroop:
Hey, thanks for checking in on this @zephraph — I'll check out that PR to better understand the usage of the `node` field.

For context, the goal here is to move Force's `/categories/` app onto the v2 API.

- That app [queries all gene families](https://github.com/artsy/force/blob/master/src/desktop/apps/categories/queries/geneFamilies.js) via the root-level `GeneFamilies` connection. 

- The singular `GeneFamily` is just the node that's on that connection edge — so I guess that may not need to be a root field after all. Thanks for highlighting this!

But the GeneFamilies connection would seem to still want to be a root field, yes?

anandaroop:
I had a think on the discussion above, and ended up with the current implementation; description updated accordingly.

Would love to know what you think @zephraph!

anandaroop:
Gonna go ahead and merge this so I can keep moving on the Force side of things.

PR #2469, opened 2020-06-15
Query for ArtworkVersions on orders using Node [GALL-2963]

mdole:
@zephraph and I paired on this for a while this afternoon. We removed `ArtworkVersions` as a root field from MP v2 and we rely on it for some orders (see Jira ticket linked below for more details). We figured that instead of just re-adding `ArtworkVersions` as a root field, we might be able to instead have it `implement Node` and stitch it into MP's Exchange schema that way.

We got pretty far with that approach, but we got stuck at the end: we could return `id` and `__typename`, but not the actual contents of `ArtworkVersions`. When running a query ourselves, we did that by adding `... on ArtworkVersions`, but we're unsure if it's possible to add that to the `ArtworkVersion` fragment.

@mzikherman any ideas on this? @zephraph anything I missed?

Some screenshots that demonstrate this:

Successfully querying in browser: 
Screen Shot 2020-06-15 at 6 04 36 PM

Returning `null` for fields when we try to use the stitched query:
Screen Shot 2020-06-15 at 6 05 43 PM

[Jira ticket](https://artsyproduct.atlassian.net/browse/GALL-2963)

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1390,7 +1390,7 @@ type ArtworkVersion implements Node {
   artistNames: String
 
   # The artists related to this Artwork Version
-  artists: String
+  artists: [Artist]
 
   # The Image id
   defaultImageID: String
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1385,7 +1385,7 @@ enum ArtworkSorts {
   TITLE_DESC
 }
 
-type ArtworkVersion {
+type ArtworkVersion implements Node {
   # The names for the artists related to this Artwork Version
   artistNames: String
 
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "@babel/preset-env": "7.4.5",
     "@babel/preset-typescript": "7.3.3",
     "@babel/register": "7.4.4",
+    "@graphql-tools/delegate": "^6.0.10",
     "@heroku/foreman": "2.0.2",
     "accounting": "0.4.1",
     "apollo-link": "1.2.1",
--- a/src/lib/stitching/exchange/stitching.ts
+++ b/src/lib/stitching/exchange/stitching.ts
@@ -1,6 +1,7 @@
 import { GraphQLSchema } from "graphql"
 import { amountSDL, amount } from "schema/v1/fields/money"
 import gql from "lib/gql"
+import { toGlobalId } from "graphql-relay"
 
 const orderTotals = [
   "itemsTotal",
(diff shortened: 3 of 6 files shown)


mzikherman:
I pushed a commit, seems to work? Let me know if this seems alright.

Screen Shot 2020-06-15 at 7 19 58 PM

src/schema/v2/artwork_version.ts
```diff
@@ -70,3 +71,5 @@ export const ArtworkVersionResolver: GraphQLFieldConfig<
       ? authenticatedArtworkVersionLoader(id)
       : null,
 }
+
+export default ArtworkVersionResolver
```
mzikherman, line 75: There's some magic inference going on in the [Node type](https://github.com/artsy/metaphysics/blob/b79c3e635927ebf8aa57b7ea6e687b11c52fa0bb/src/schema/v2/object_identification.ts#L143-L148) that makes it easier to have a default export for the field.

mdole: whoa wait I must be losing my marbles - I _swear_ @zephraph and I had added the default export AND the node interface. no clue what happened there, but thanks for catching the issue @mzikherman !

zephraph:
So @mzikherman, we shouldn't _need_ to add that extra `... on ArtworkVersion`. That should be happening via the stitching call. I've got a lead on how we might be able to do it though, I'll let you know how that goes.

mzikherman:
Oh interesting @zephraph that would be cool!

I was thinking it would be since querying `Node` in practice (sans stitching, just regular), you always have to narrow it.

But I guess it's akin to doing something like:

```
{
  artwork(id: "andy-warhol-skull") {
    ... on Artwork {
      title
    }
  }
}
```

Technically it works/is valid, but you shouldn't _need_ that in order to resolve properly (or the stitching lib/our wrapper should be able to handle that for you).

mdole:
Wooo we got it working! @zephraph's solution of using a newer version of `delegateToSchema` and specifying the `returnType` worked great - we then also updated the types for `ArtworkVersion` because the resolvers either had incorrect types or were returning something that didn't match the specified type.

We made a breaking change to the type of `ArtworkVersion {artists}`, but that type was not actually being used anywhere in our code base. Our hypothesis is that it's been broken forever (or at least a long time) and it just never got noticed till now 🤷‍♂️

Screen Shot 2020-06-16 at 1 33 24 PM

_schemaV2.graphql
```diff
@@ -1385,12 +1385,12 @@ enum ArtworkSorts {
...
 }
 
-type ArtworkVersion {
+type ArtworkVersion implements Node {
   # The names for the artists related to this Artwork Version
   artistNames: String
 
   # The artists related to this Artwork Version
-  artists: String
+  artists: [Artist]
```
zephraph, line 1393: Note that while this is a breaking change to the schema, nothing in our codebase is querying `artists` from `ArtworkVersion` so it should be fine.

package.json
```diff
@@ -44,6 +44,7 @@
     "@babel/preset-env": "7.4.5",
     "@babel/preset-typescript": "7.3.3",
     "@babel/register": "7.4.4",
+    "@graphql-tools/delegate": "^6.0.10",
```
zephraph, line 47: `graphql-tools` was broken down into many packages. Longer term, we need to update all the parts of `graphql-tools`, but for now we can incrementally update particular parts.

src/lib/stitching/exchange/stitching.ts
```diff
@@ -230,16 +233,18 @@ export const exchangeStitchingEnvironment = ({
...
-              fieldName: "artworkVersion",
+              fieldName: "node",
               args: {
-                id,
+                id: globalID,
               },
               context,
               info,
               transforms: exchangeSchema.transforms,
+              returnType: ArtworkVersionType,
```
zephraph, line 247: @mzikherman this `returnType` field (only available on the new `delegateToSchema` function) is what automatically narrows the type down from `Node` to `ArtworkVersion`. This way, you don't have to do the `... on ArtworkVersion` verbosity in the query.

zephraph:
I'm gonna go ahead and ship this. 

Thanks for pairing @mdole!

src/lib/stitching/exchange/stitching.ts
mzikherman: oo awesome! So this plus the `graphql-tools` bump? That's sweet.

PR #2471, opened 2020-06-16
Add hasPassword to me query

MounirDhahri:
### Description
Added hasPassword to me

### Motivation
We need to hide the change password option from frontend.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -6129,6 +6129,7 @@ type Me implements Node {
   email: String
   followsAndSaves: FollowsAndSaves
   hasCreditCards: Boolean
+  hasPassword: Boolean!
   hasQualifiedCreditCards: Boolean
   hasSecondFactorEnabled: Boolean!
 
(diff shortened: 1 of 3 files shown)


_schemaV2.graphql
```diff
@@ -6129,6 +6129,7 @@ type Me implements Node {
   email: String
   followsAndSaves: FollowsAndSaves
   hasCreditCards: Boolean
+  hasPassword: Boolean!
```
MounirDhahri, line 6132: @zephraph recommended that we check again the naming here. Please feel free to suggest a different one

MounirDhahri: how about ? hasRegisteredUsingEmail

zephraph: This flag does match gravity and its intended purpose is to help make UI decisions based on if the user has a password or not, so it's sort of aptly named. 

I just recommended thinking about the naming in case we don't want to telegraph that particular information. Given as it's only available through `me` though, that probably doesn't matter.

PR #2472, opened 2020-06-16
Adding `calculatedEndAt` to Viewing Rooms

pvinis:


zephraph:
I followed up briefly w/ @mdole on this, but I'm not sure about creating a type `CalculatedEndAt` would be the best course of action. When creating a type in our schema we should consider it's usage. 

The field that's being queried here is `calculatedEndAt` but the _type_ it's represented at should be something different. If you also had a `caculatedStartAt` field you wouldn't create a `CalculatedStartAt` type... they'd both use the same type because they'd both have the same shape. I feel like the type should likely be more generic.

pvinis:
I got a bit confused by your comment, but after re-reading it a couple of times, I think I understand. You suggest to have a `calculatedEndAt` *field*, but a `DurationType` or something along those lines. Right? 🤔
If that's what the concern is, then it makes sense to me and I will do something like that. I was still playing around, but it is a good suggestion, thanks 👍.

zephraph:
Yeah! Sorry for the confusion. Making the type generic and broadly useful is preferred to having it super specific to one field.

pvinis:
Closed this, going for the #2477 approach.

PR #2475, opened 2020-06-17
Add stitched ArtistSeries field

williardx:
https://artsyproduct.atlassian.net/browse/FX-2005

This PR stitches `ArtistSeries` field from Gravity into Metaphysics.

`DO NOT MERGE` until we have clarity on schema changes from Viewing Room.

collab: @ashleyjelks @anandaroop 

### Questions

- [x] Do we need to write a test for querying `ArtistSeries`? Wouldn't we just be testing stitching at that level?
- [x] Should we leave out whitespace changes to the schema? Do we have a convention we want to follow?
- [x] Updating Gravity's schema pulled in schema changes to another type that didn't seem to be present on master -- need to confirm that this is right

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1251,6 +1251,17 @@ type ArtistRelatedData {
   ): ArtistConnection
 }
 
+type ArtistSeries {
+  description: String
+  featured: Boolean!
+
+  # Unique ID for this artist series
+  internalID: ID!
+  published: Boolean!
+  slug: String!
+  title: String!
+}
+
 enum ArtistSorts {
   sortable_id_asc
     @deprecated(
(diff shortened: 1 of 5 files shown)


sweir27:
To answer some of your questions:
1. I think at this point adding a test wouldn't do much. Once we start extending fields, we should add tests like [these ones](https://github.com/artsy/metaphysics/blob/master/src/lib/stitching/kaws/__tests__/stitching.test.ts#L7).
2. I don't see the whitespace changes 🤔 -- the automated updates seem okay to me!
3. Was that confirmed in the gravity PR?

Would you mind posting a screenshot of a successful query for artist series so it's clear to future devs what functionality this PR provides?

Also I _think_ this is mergeable, but will wait for you to remove that label/let me know!

williardx:
Thanks! Interesting about the whitespace changes 🤔  My editor's version of Prettier seems to have made some changes but our commit hook must have undone them (which is good!).

PR #2477, opened 2020-06-18
Change `formattedEndAt` format and add `formattedStartAt` [GALL-2988]

pvinis:
⚠️ Connected PR on Force https://github.com/artsy/force/pull/5810 ⚠️ 

That field initially returned the whole string that got displayed on force, but now it just gives the duration in a human-friendly string, and leaves the extra copy to the clients.

Example:
- `distanceToEnd`: `2 days`
- force: `Closes in ${distanceToEnd}`
- eigen: `${distanceToEnd} left`

Also for tag and countdown where we have `short`er time cut-offs:
- `tag: distanceToEnd(short: true)`: `null`
- `countdown: distanceToEnd(short: false)`: `20 days`
- tag: no text
- countdown: `Closes in ${distanceToEnd}`

I will follow up with a PR for fix force, and eigen will come later.

Changes:
- Updated `moment` and removed the types, since they come with the package.
- `import moment from "moment-timezone"` when we use `moment.tz.*`.
- Removed `moment.distance` that was added to make this. Now we can do this with just `moment`.

Test using:
```
{
  viewingRooms {
    edges {
      node {
        title
        slug
        status
        distanceToOpen(short: false)
        countdown: distanceToEnd(short: false)
        tag: distanceToEnd(short: true)
        startAt
        endAt
      }
    }
  }
}

```


mdole:
looks good to me! @pvinis two logistical things: first, looks like we need to rebase over master to pull in the latest schema changes. Second, I imagine we want to wait to merge this one until we're also ready to merge the associated Force PR, right? Since this one will break the existing Viewing Room countdown in Force

pvinis:
I'll rebase.

I'm making the PR on Force now, but I'm not sure about the timing. My guess is I need to test Force using a local MP, and then once we have both PRs ready, we can merge at the same time? 🤔 If we merge this one here before the one in Force, we will just get some failing tests because of the copy change, which should be alright since VRs are not public yet? Not sure how we would handle this for something that's public.

mdole:
well I'm not 100% sure whether we have Communities pages live yet, but if we do, those actually _are_ already using this field and displaying it to users - so I would not want to deploy this to prod before having Force also ready to deploy.

These scenarios are tricky for sure. I think what you said is right; we would want to basically merge them at the same time, test on staging, and then deploy both services. What we want to avoid is getting either this or Force on staging and having to leave them in an un-deployable state for a long time

pvinis:
Hold for me to make the last changes.

src/lib/stitching/gravity/stitching.ts
```diff
@@ -89,21 +84,42 @@ export const gravityStitchingEnvironment = (
...
+                h: "%d hour",
+                hh: "%d hours",
+                d: "%d day",
+                dd: "%d days",
+                M: "%d month",
+                MM: "%d months",
+                y: "%d year",
+                yy: "%d years",
+              },
+            })
```
damassi, line 102: @pvinis - the reason this approach wasn't taken was because this (seemingly static / global) change on moment has already happened somewhere in auctions code (if you search for `updateLocale`). Not sure how to get around that other than tracking down the various call-sites and making sure things are still ok. That said, I could be misunderstanding moment!

oxaudo:
As far as I know we do not have any released Communities VRs for now. Updates like that would definitely be complicated down the road so we need to discuss the strategy for sure. But for this specific ticket I would not worry about timing of changes much since it will not affect any actual users.

We only have 2 VRs in production. Both are our own tests.

src/lib/stitching/gravity/stitching.ts
pvinis: I thought the same thing after reading the docs, but I saw we have tests for the auctions part here https://github.com/artsy/metaphysics/blob/master/src/schema/v2/sale/__tests__/index.test.js#L597 for example, and they seem to pass.

So I'm not sure if the docs are strange and we both misunderstood, or if there is something else going on here. 🤔

PR #2482, opened 2020-06-22
Upgrade Node to v12

eessex:
Use node v12

Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-2287

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,10 @@
-FROM node:10.13.0
+FROM node:12.13.0
 
 WORKDIR /app
 
 # Set up deploy user and working directory
 RUN apt-get update && apt-get install -y \
+  build-essential \
   dumb-init \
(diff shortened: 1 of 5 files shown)


zephraph:
There's a reference to node 10 in the circle config, you'll need to update that too.

Dockerfile
```diff
@@ -1,29 +1,31 @@
...
-  mc-data \
-  unzip && \
-  adduser --disabled-password --gecos '' deploy
-
-# Install the packages
-COPY package.json yarn.lock ./
-COPY patches ./patches
-RUN yarn install
+  git \
+  python && \
```
izakp, line 9: Why do we need Python in the app image?

eessex: This is required by the `appmetrics` package, which seems to be doing some node profiling.

PR #2498, opened 2020-06-30
Support includeUnlisted argument for artworksConnection

dblandin:
Allow the client to request unlisted artworks during a query.

Related Gravity PR: https://github.com/artsy/gravity/pull/13206


dblandin:
Closing this as we aren't concerned with listed vs unlisted as much as retrieving unpublished artworks for CMS scoped to a particular Viewing Room.

PR #2507, opened 2020-07-03
Fix markdown formatter args handling

ds300:
We have this `markdown` helper for text fields. It's for text that contains markdown. It has an optional parameter called `format` which can have a value of either `HTML` or `MARKDOWN`. the `HTML` option is used in web clients to avoid the need to convert the markdown to html in the client itself.

So if you request a text field like

```graphql
someObject {
  textField
}
```

you get markdown back

and if you request a text field like

```graphql
someObject {
  textField(format: HTML)
}
```

you get HTML back

but if you request a text field like

```graphql
someObject {
  textField(format: MARKDOWN)
}
```

you _also get HTML back_ !

This PR fixes that.

This won't be a breaking change. I checked the whole of the artsy org and didn't find any places where the 'markdown' format option is explicitly used.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -4922,6 +4922,14 @@ type Feature {
   subheadline(format: Format): String
 }
 
+# An illustrated link chosen to highlight a Gene from a given GeneFamily
+type FeaturedGeneLink {
+  href: String!
+  image: Image
+  internalID: String!
+  title: String!
+}
+
 type FeaturedLink {
   description(format: Format): String
   href: String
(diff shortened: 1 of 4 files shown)


_schemaV2.graphql
```diff
@@ -5405,6 +5413,40 @@ type GeneEdge {
...
+
+# An edge in a connection.
+type GeneFamilyEdge {
+  # A cursor for use in pagination
+  cursor: String!
+
+  # The item at the end of the edge
+  node: GeneFamily
+}
+
```
ds300, line 5449: these seem unrelated but I'm pretty sure they should be in master already. They're definitely in the live staging schema so I think people have been committing without running the git hooks 🙈

src/schema/v2/fields/__tests__/markdown.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        { format: "html" },
+        {} as any,
+        { fieldName: "description" } as any
+      )
+    ).toMatchInlineSnapshot(`
+      "<p>Here&#39;s a <strong>description</strong> with some <em>emphasis</em> !</p>
+      "
+    `)
+  })
+  it("resolves markdown as html", () => {
```
sweir27, line 32: #minor this should probably be `resolves markdown as markdown`

src/schema/v2/fields/markdown.ts
```diff
@@ -8,7 +8,7 @@ type Format = "html" | "markdown"
 type Value = string | null | undefined
 
 export const formatMarkdownValue = (value: string, format: Format): string => {
-  if (format === "html" || format === "markdown") {
+  if (format === "html") {
```
sweir27, line 11: 👍 Very curious if there was a reason behind this original implementation, but the change makes sense to me!

_schemaV2.graphql
anandaroop: Eek, this looks like me. But I'm definitely not skipping any `husky` steps, at least not intentionally. Wondering how this happened 🤔

ds300: 💭 the pre-commit hook uses the staging env vars. Is the gene family stuff gated behind an env var? Because then I could imagine this happening by first adding the schema code in MP and PR-ing the changes *before* adding the env var to staging.

If not I'm happy to help debug!

ds300:
@sweir27 thanks for the feedback 🙌  this should be 🍏

PR #2509, opened 2020-07-06
Expose filteredArtworksConnection under ArtistSeries

williardx:
https://artsyproduct.atlassian.net/browse/FX-2035

---

## Problem

Users need to be able to look up all the artworks associated with a given artist series.

## Solution

After modifying the `/filter/artworks` endpoint to be able to query for artworks associated with an artist series (https://github.com/artsy/gravity/pull/13211), we now need to expose querying for an artist series's artworks in Metaphysics. Once this is merged, a frontend client will be able to request the artworks associated with an artist series.

The bulk of this work is stitching `FilterArtworks` into `ArtistSeries` for v2. We did _not_ stitch for v1 because we are deprecating v1.

Screen Shot 2020-07-06 at 1 45 13 PM

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -663,6 +663,7 @@ type Artist implements EntityWithFilterArtworksConnectionInterface & Node & Sear
     aggregations: [ArtworkAggregation]
     artistID: String
     artistIDs: [String]
+    artistSeriesID: String
     atAuction: Boolean
     attributionClass: [String]
     before: String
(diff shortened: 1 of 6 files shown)


williardx:
Not sure why Peril claims that there are breaking changes with Force here -- the schema changes in this PR don't remove anything.

sweir27:
@williardx I think the peril warning is actually accurate! It looks like metaphysics' schema is out-of-date, per [this PR comment](https://github.com/artsy/metaphysics/pull/2507#discussion_r449607594).

Should be fixed when that is merged in, or feel free to confirm by generating the v2 schema locally and including in this PR.

_schemaV2.graphql
```diff
@@ -992,6 +993,48 @@ type ArtistSeries {
   artists(page: Int, size: Int): [Artist]
   description: String
   featured: Boolean!
+  filterArtworksConnection(
```
anandaroop, line 996: This entire field definition gets created by that tricky `argsToSDL` business in the stitching file, yeah?

williardx: Yep!

PR #2536, opened 2020-07-17
[MX-414][MX-415][MX-421] - Push notification settings

brainbicycle:
https://artsyproduct.atlassian.net/browse/MX-414  
https://artsyproduct.atlassian.net/browse/MX-415
https://artsyproduct.atlassian.net/browse/MX-421

- Exposes new push notification settings in the **me** endpoint
- Allows updating push notification settings in the **updateMyProfile** mutation
- Exposes new push notification setting in authenticated **user** endpoint

Gravity PR that added the attributes:
https://github.com/artsy/gravity/pull/13261


src/schema/v2/me/update_me_mutation.ts
```diff
@@ -123,13 +153,32 @@ export default mutationWithClientMutationId<any, any, ResolverContext>({
...
       collector_level: collectorLevel,
       price_range_min: priceRangeMin,
       price_range_max: priceRangeMax,
+      receive_purchase_notification: receivePurchaseNotification,
+      receive_outbid_notification: receiveOutbidNotification,
+      receive_lot_opening_soon_notification: receiveLotOpeningSoonNotification,
+      receive_sale_opening_closing_notification: receiveSaleOpeningClosingNotification,
+      receive_new_works_notification: receiveNewWorksNotification,
+      receive_new_sales_notification: receiveNewSalesNotification,
+      receive_promotion_notification: receivePromotionNotification,
```
ashfurrow, line 181: I just want to verify my understanding here, based on Joey's comment in the Gravity PR. So because we're passing these through to Gravity, and they default to `undefined` in Metaphysics (because they are optional), then Gravity won't change anything unless these are specified in the mutation to Metaphysics. In other words, if these aren't included in the input to the mutation, then they won't change anything in Gravity, right?

brainbicycle: Good call out, let me manually test and see if I can add a test in here or gravity for that behavior

src/lib/__tests__/helpers.test.ts
```diff
@@ -65,6 +65,16 @@ describe("toKey", () => {
     ).toBe("foo/bar?size=10&sleep=false&sort=asc")
   })
 
+  it("does not include undefined options in path", () => {
```
brainbicycle, line 68: We call this method in our loaders when generating paths, unspecified options are passed through to loader as undefined but ignored when adding params to path:
https://github.com/artsy/metaphysics/blob/master/src/lib/loaders/api/loader_interface.ts#L21

src/schema/v2/me/update_me_mutation.ts
brainbicycle: I tested manually and not specifying one of the optional params in mutation does pass through the param as undefined to the gravity loader but when path is generated they are ignored so this does not modify the param on gravities side as expected. Added a test for this behavior in `toKey` helper function used for generating paths.

PR #2537, opened 2020-07-17
Respect query param order for Viewing Rooms artworksConnection request

dblandin:
By default, when Metaphysics is contructing an API endpoint URL for a data loader request, query params are sorted to increase cache hits by URL path + params.

To display Viewing Room artworks on artsy.net, we're passing a sorted array of IDs from Gravity to fetch artworks via the Metaphysics `artworksConnection` field.

The order of these `?id[]=one&id[]=two&id[]=three` params should be respected which constructing the full Gravity API endpoint URL, otherwise we display the artworks in the wrong order.

This commit adds an argument, passed through from the stitched resolver, to disable the default sorting behavior for Viewing Room artworks.

See `toQueryString` in `src/lib/helpers.ts`:
https://github.com/artsy/metaphysics/blob/3c0ef2c13858474c1128272a83967b05356900c6/src/lib/helpers.ts#L65-L72

Co-authored-by: @oxaudo 
Co-authored-by: @zephraph


src/schema/v2/artworks.ts
```diff
@@ -16,10 +16,11 @@ const Artworks: GraphQLFieldConfig<void, ResolverContext> = {
       type: new GraphQLList(GraphQLString),
     },
   }),
-  resolve: (_root, options, { artworksLoader }) => {
+  // @ts-ignore
```
dblandin, line 19: This works but I'm not sure how to add this additional key to the type properly or if there's a better way to pass this through. Should I add this argument to the root field instead so it can be passed in as args?

I was able to define an interface to resolve the errors in this file, but ran into other errors in the `v2/schema.ts` file.

```typescript
interface ArtworksResolverContext extends ResolverContext {
  respectParamsOrder: Boolean
}
```

```
src/schema/v2/schema.ts:167:3 - error TS2322: Type '{ artworkAttributionClasses: GraphQLFieldConfig<void, ResolverContext, { [argName: string]: any; }>; article: GraphQLFieldConfig<void, ResolverContext, { ...; }>; ... 36 more ...; usersConnection: GraphQLFieldConfig<...>; }' is not assignable to type 'Thunk<GraphQLFieldConfigMap<any, ResolverContext, { [key: string]: any; }>>'.
  Type '{ artworkAttributionClasses: GraphQLFieldConfig<void, ResolverContext, { [argName: string]: any; }>; article: GraphQLFieldConfig<void, ResolverContext, { ...; }>; ... 36 more ...; usersConnection: GraphQLFieldConfig<...>; }' is not assignable to type 'GraphQLFieldConfigMap<any, ResolverContext, { [key: string]: any; }>'.
    Property 'artworks' is incompatible with index signature.
      Type 'GraphQLFieldConfig<void, ArtworksResolverContext, { [argName: string]: any; }>' is not assignable to type 'GraphQLFieldConfig<any, ResolverContext, { [key: string]: any; }>'.
        Property 'respectParamsOrder' is missing in type 'ResolverContextValues & AvailableLoaders' but required in type 'ArtworksResolverContext'.

167   fields: rootFields,
```

dblandin: Moved `respectParamsOrder` over to a field argument which I think feels a bit better than using the context.

oxaudo:
Oh. I think the spec fail here might be related @dblandin :(

dblandin:
>  that feels a bit indirect for what appears to be a direct bug to solve? (ie- rather than opt-in. with a new arg, we shouldn't be sorting array params in general I don't think?

Agreed! That seems a bit trickier to fix, so I'll merge this in for now and follow-up on that.

PR #2540, opened 2020-07-20
Update gravity schema

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-2060

This PR just adds a missing schema bump to support the earlier work for this^ ticket that happened in https://github.com/artsy/gravity/pull/13263


ashleyjelks:
LGTM!

PR #2554, opened 2020-07-22
Forces Gravity request for push notification settings fields.

ashfurrow:
I was QA'ing [MX-414] and I noticed that the fields were being returned as `null` and that Gravity wasn't being hit, unless I requested other fields.

Before:

Screen Shot 2020-07-22 at 3 05 18 PM

After:

Screen Shot 2020-07-22 at 3 05 25 PM

Looks like this is to avoiding hitting Gravity unnecessarily. But in the case of Pulse, we might only be getting these attributes, so I've removed the fields from the list of things we don't need to ask Gravity about.

Follow-up from #2536.

[MX-414]: https://artsyproduct.atlassian.net/browse/MX-414

### DIFF
--- a/src/schema/v2/me/index.ts
+++ b/src/schema/v2/me/index.ts
@@ -226,13 +226,6 @@ const MeField: GraphQLFieldConfig<void, ResolverContext> = {
       "lotsByFollowedArtistsConnection",
       "identityVerification",
       "unreadNotificationsCount",
-      "receivePurchaseNotification",
-      "receiveOutbidNotification",
-      "receiveLotOpeningSoonNotification",
-      "receiveSaleOpeningClosingNotification",
-      "receiveNewWorksNotification",
-      "receiveNewSalesNotification",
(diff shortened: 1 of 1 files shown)


src/schema/v2/me/index.ts
```diff
@@ -226,13 +226,6 @@ const MeField: GraphQLFieldConfig<void, ResolverContext> = {
       "lotsByFollowedArtistsConnection",
       "identityVerification",
       "unreadNotificationsCount",
-      "receivePurchaseNotification",
-      "receiveOutbidNotification",
-      "receiveLotOpeningSoonNotification",
-      "receiveSaleOpeningClosingNotification",
-      "receiveNewWorksNotification",
-      "receiveNewSalesNotification",
-      "receivePromotionNotification",
```
brainbicycle, line 235: Oh man I totally misunderstood this code, I thought this was not required fields in metaphysics query. Good catch 🙇‍♂️ !

PR #2560, opened 2020-07-27
Add isArtsyLicensed to the sale type

yuki24:
Partially addresses https://artsyproduct.atlassian.net/browse/AUCT-1164

This exposed the newly added `is_artsy_licensed` flag on the sale model so our clients can use that in e.g. sale winning emails.

## Screenshot

Screen Shot 2020-07-27 at 3 39 57 PM

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -8029,6 +8029,7 @@ type Sale implements Node {
 
   # A type-specific ID likely used as a database ID.
   internalID: ID!
+  isArtsyLicensed: Boolean!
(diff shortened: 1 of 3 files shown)


_schemaV2.graphql
```diff
@@ -8029,6 +8029,7 @@ type Sale implements Node {
 
   # A type-specific ID likely used as a database ID.
   internalID: ID!
+  isArtsyLicensed: Boolean
```
erikdstock, line 8032: should this be Non-Null?

PR #2580, opened 2020-08-05
route external clients through nginx

izakp:
Connections through ELBs from external clients should be routed through Nginx.

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -139,11 +139,11 @@ spec:
     - port: 443
       protocol: TCP
       name: https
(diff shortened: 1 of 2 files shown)


artsyjian:
We did Prod this morning at 10:12 AM ET when we experienced elevated latency of Metaphysics to see if it helps with that.

eessex:
👍 Kind of surprised that this was not in place already!

izakp:
> Kind of surprised that this was not in place already!

Me too - I wrote the commit in question and we must have missed it when doing our app audit

PR #2582, opened 2020-08-05
Disables GraphiQL in production

dzucconi:
Context here is that Googlebot (and others?) are erroneously crawling and causing undue load on Metaphysics, leading to instability in Force. See `#incident-96` for more details.

Already mentioned in the README is:

So I think we just need to make it known to anyone who was still relying on the hosted instances.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -226,7 +226,7 @@ function startApp(appSchema, path: string) {
 
         return {
           schema: appSchema,
-          graphiql: true,
+          graphiql: !PRODUCTION_ENV,
           context,
           rootValue: {},
           customFormatErrorFn: graphqlErrorHandler(enableSentry, {


dzucconi:
The behavior here is that Metaphysics will return a `400` if it's a `GET` and a `query=` query string is missing.

zephraph:
Could we add context to this PR about why we're making this move for future reference?

joeyAghion:
Thanks @dzucconi.

As menioned, the GraphiQL GUI is served at the root (`/`) so is difficult to manage with anything that depends on paths: logs, firewall rules, `robots.txt`, etc. Its HTML references tons of server-side endpoints that end up being crawled by Google. Absent specific headers, the app renders the entire GUI for each of these requests (or even arbitrary paths like `/foo/bar/baz`). This caused a latency explosion that caused downstream services to become completely unavailable (in staging yesterday, and in production overnight). We put a fuzzy firewall rule in place to mitigate this, but medium term we should rely on disallow rules (rather than returning errors to Google), or ideally just not offer paths that cause the application to become unstable.

Given the app's own recommendation and @dzucconi's observation that other production environments don't tend to expose this, I thought this was a prudent choice. If we need to bring it back in web form, it should be at a specific path so we can log and authorize it in a more targeted way.

PR #2584, opened 2020-08-05
Add registered argument to salesConnection

yuki24:
partially addresses https://artsyproduct.atlassian.net/browse/AUCT-1151
blocked by https://github.com/artsy/gravity/pull/13354

This simply exposes the new `registered` argument [being added to Gravity](https://github.com/artsy/gravity/pull/13354). I also added a new test file `src/schema/v2/__tests__/salesConnection.test.ts` since there didn't seem to be a test for `salesConnection`.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -7792,6 +7792,9 @@ type Query {
 
     # Limit by published status.
     published: Boolean = true
+
+    # Returns sales the user has registered for if true, returns sales the user has not registered for if false.
+    registered: Boolean
     sort: SaleSorts
   ): SaleConnection
 
@@ -9836,6 +9839,9 @@ type Viewer {
 
     # Limit by published status.
     published: Boolean = true
+
+    # Returns sales the user has registered for if true, returns sales the user has not registered for if false.
+    registered: Boolean
     sort: SaleSorts
   ): SaleConnection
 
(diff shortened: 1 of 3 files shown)


src/schema/v2/__tests__/salesConnection.test.ts
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable promise/always-return */
+import { runQuery } from "schema/v2/test/utils"
+import gql from "lib/gql"
+import sinon from "sinon"
+
+describe("artworksConnection", () => {
```
bhoggard, line 6: Were you intending to also test the `registered` flag here?

yuki24: Ah yes I was. Also this was a bad copy-&-paste, I'll fix that as well.

_schemaV2.graphql
```diff
@@ -7757,6 +7757,9 @@ type Query {
 
     # Limit by published status.
     published: Boolean = true
+
+    # Returns sales the user has registered for if true, returns sales the user has not registered for if false.
+    registered: Boolean
```
erikdstock, line 7762: Similar to my comments in https://github.com/artsy/gravity/pull/13354 about the ambiguity of true/false with a query, we could address that here with a `filter: SaleFilter` and `enum SaleFilter { REGISTERED NOT_REGISTERED }`.

However I see this would probably break from precedent as with the `published` flag above, which was actually the original point of my confusion with it. maybe something like `filters: [SaleFilter]` is what I'm getting at, but clearly out of scope.

yuki24: I actually personally had the same confusion as what @erikdstock raised, and would like to use some brain power to make it more straightforward.

yuki24:
I'll just go ahead and merge this PR since I've just merged https://github.com/artsy/gravity/pull/13354, which should unblock us from merging this PR.

PR #2586, opened 2020-08-06
Allow querying for user stats information in vortex

jpotts244:
Metaphysics wrapper to expose new Vortex user stats graphQL endpoint to client
Vortex API work here: https://github.com/artsy/vortex/pull/260

The only client currently accessing the endpoint will be Volt (sketch PR here: https://github.com/artsy/volt/pull/4529)

Allow authenticated partners and admins to query against a user's total purchase counts
Screen Shot 2020-08-09 at 10 09 26 AM

---

Screen Shot 2020-08-09 at 10 08 08 AM

Unauthenticated users/non-partner users will not be authorized to hit this endpoint
Screen Shot 2020-08-07 at 9 34 25 AM

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -470,6 +470,12 @@ type AnalyticsShow {
   entityId: String!
 }
 
+# Statistics for users
+type AnalyticsUserStats {
+  numberOfPurchases: Int!
+  userId: String!
+}
+
 type AnalyticsVisitorsByCountry {
   metric: String!
   name: String!
@@ -10817,6 +10823,9 @@ type Query {
 
   # Find PartnerStats
   analyticsPartnerStats(partnerId: String!): AnalyticsPartnerStats
+
+  # Query UserStats
+  analyticsUserStats(userId: String!): AnalyticsUserStats
 
   # An Article
   article(
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -472,7 +472,7 @@ type AnalyticsShow {
 
 # Statistics for users
 type AnalyticsUserStats {
-  numberOfPurchases: Int!
+  totalPurchaseCount: Int!
   userId: String!
 }
 
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -458,7 +458,7 @@ type AnalyticsShow {
 
 # Statistics for users
 type AnalyticsUserStats {
-  numberOfPurchases: Int!
+  totalPurchaseCount: Int!
   userId: String!
 }
 
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -456,6 +456,12 @@ type AnalyticsShow {
   entityId: String!
 }
 
+# Statistics for users
+type AnalyticsUserStats {
+  numberOfPurchases: Int!
+  userId: String!
+}
+
 type AnalyticsVisitorsByCountry {
   metric: String!
   name: String!
@@ -7341,6 +7347,9 @@ type Query {
   # Find PartnerStats
   analyticsPartnerStats(partnerId: String!): AnalyticsPartnerStats
 
+  # Query UserStats
+  analyticsUserStats(userId: String!): AnalyticsUserStats
+
   # An Article
   article(
     # The ID of the Article
@@ -9442,6 +9451,7 @@ type UpdateViewingRoomSubsectionsPayload {
 }
 
 type User {
+  analytics: AnalyticsUserStats
   cached: Int
 
   # The given email of the user.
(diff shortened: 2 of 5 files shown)


_schema.graphql
```diff
@@ -470,6 +470,12 @@ type AnalyticsShow {
   entityId: String!
 }
 
+# Statistics for users
+type AnalyticsUserStats {
+  numberOfPurchases: Int!
```
jpotts244, line 475: Just to call out if we like the name of this field before we move this PR forward.

`numberOfPurchases` in Vortex is the count that is getting returned from Vortex. @sepans made a good point that now is a good time to see if we like the name - before we merge the change into metaphysics

_schemaV2.graphql
```diff
@@ -456,6 +456,12 @@ type AnalyticsShow {
   entityId: String!
 }
 
+# Statistics for users
+type AnalyticsUserStats {
+  numberOfPurchases: Int!
+  userId: String!
```
sepans, line 462: Now that this `userId` is stitched here it looks more redundant. Also it has to be something like `internalID` to be consistent. So maybe we can remove it from vortex? @jpotts244 @starsirius @zephraph what do you think?

jpotts244: Ah, like just remove it altogether since we are not reading the data from userId?

_schema.graphql
starsirius: **suggestion (non-blocking):** `purchaseCount` or `totalPurchaseCount` seems to be more consistent with current convention.

*format via [Conventional Comments](https://conventionalcomments.org/) + [saved replies](https://github.com/settings/replies)*

_schemaV2.graphql
starsirius: **thought (no-strong-feeling):** I personally prefer including `userId` in the response so it's self-contained.

I know callers pass in `userId` as a param to get the user's stats, so they already have the `userId`. Sometimes I find it useful/convenient if the response also contains it so I can just pass the entire thing down without worrying the `userId` referenced somewhere else.

`AnalyticsUserStats` is stitched under `User` so it feels redundant as we are able to get the `userId` from `User`. But `AnalyticsUserStats` could also be used independently one day.

Re `internalID`, not en expert but I think `userId` might be preferable here because `AnalyticsUserStats` is not a user.

*format via [Conventional Comments](https://conventionalcomments.org/) + [saved replies](https://github.com/settings/replies)*

_schema.graphql
jpotts244: `totalPurchaseCount` sounds like a good idea! Adjusting that here and also in Vortex [here](https://github.com/artsy/vortex/pull/265)

jpotts244:
Hm, seeing this Peril message 
```

The V2 schema in this PR has breaking changes with Force. Remember to update the Force schema if necessary.Argument 'statuses: [ViewingRoomStatusEnum!]' was removed from field 'Viewer.viewingRoomsConnection'
--

```

but there is not a diff in this PR for those fields

_schemaV2.graphql
sepans: Good point I agree `internalID` is not correct. With @starsirius point above that `userId` no longer bothers me that much 😄

zephraph: **nitpick**: `userId` -> `userID`

I agree that `internalID` wouldn't be good to add here in the same vein that @starsirius mentioned. `internalID` is meant to represent the id of that given type within the system that provides it. Often that's a database id. 

I'm fine with `userID` remaining.

_schemaV2.graphql
```diff
@@ -7341,6 +7347,9 @@ type Query {
   # Find PartnerStats
   analyticsPartnerStats(partnerId: String!): AnalyticsPartnerStats
 
+  # Query UserStats
+  analyticsUserStats(userId: String!): AnalyticsUserStats
```
zephraph, line 7351: **nitpick**: `userId` -> `userID`

I'm noticing some inconsistencies in places with this 😅. Compairing `parterId` vs `internalID`.

zephraph:
Looks good @jpotts244! Small nit about the id casing, but I see we already have inconsistencies so I'll let you drive that decision. 

Feel free to merge when you think it's ready.

_schemaV2.graphql
sepans: Don't want to open up this question now that we have an agreement 😄 but I tried this [here](https://github.com/artsy/metaphysics/pull/2586/files?file-filters%5B%5D=.ts#diff-7e804a266de6111194edd60c2104b605R49) to remove that `userId` when it is being stitched and it worked. Just throwing it here if you think it is good to make this change:
```graphql
type UserStats {
  totalPurchaseCount: Int!
}
extend type User {
  analytics: UserStats
}
```

_schemaV2.graphql
jpotts244: Discussed offline -- we will be leaving it as is for now!

PR #2593, opened 2020-08-07
Viewer fix for statuses

pvinis:
This PR adds the `statuses` field in the viewer we created for force for `viewingRooms`

I tried adding a test for `inputFields include statuses`, but I couldn't add this in an easy way.

I tried getting it by using something like `getFieldsForTypeFromSchema` but for inputFields, but there is no "type" that these are attached to.

I did find and try [this](https://gist.githubusercontent.com/craigbeck/b90915d49fda19d5b2b17ead14dcd6da/raw/e50819812a7a8a95b303ac0ea1464e2679e3e4bc/introspection-query.graphql) which worked and gave me the whole schema basically. I'm sure we could use this in a test and get everything and them map it in an object that we can ask like `Viewer.viewingRoomsConnection.inputFields includes statuses`.

### DIFF
--- a/src/lib/stitching/gravity/stitching.ts
+++ b/src/lib/stitching/gravity/stitching.ts
@@ -137,7 +137,7 @@ export const gravityStitchingEnvironment = (
       }
 
       extend type Viewer {
-        viewingRoomsConnection(first: Int, after: String): ViewingRoomConnection
+        viewingRoomsConnection(first: Int, after: String, statuses: [ViewingRoomStatusEnum!]): ViewingRoomConnection
       }
     `,
     resolvers: {
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -13749,7 +13749,11 @@ type Viewer {
 
   # A list of Users
   users(ids: [String]): [User]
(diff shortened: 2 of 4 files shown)


src/lib/stitching/gravity/stitching.ts
```diff
@@ -137,7 +137,7 @@ export const gravityStitchingEnvironment = (
       }
 
       extend type Viewer {
-        viewingRoomsConnection(first: Int, after: String): ViewingRoomConnection
+        viewingRoomsConnection(first: Int, after: String, statuses: [ViewingRoomStatusEnum!]): ViewingRoomConnection
```
mdole, line 140: Wanted to make sure I understood the syntax of `[ViewingRoomStatusEnum!]`; i.e. what exactly can be null, and found this handy reference: https://graphql-ruby.org/type_definitions/lists.html#nullable-lists-with-non-null-items

TL;DR we can pass `nil` for statuses, but if a list is present, it has to include value(s) of type `ViewingRoomStatusEnum`. Sounds right to me!

PR #2596, opened 2020-08-07
Tune staging rolling update parameters.

artsyjian:
- The current staging rolling update settings (`maxSurge`, `maxUnavailable`), along with HPA `minReplicas` setting, allow the situation of having 0 pods during a rolling update. Changing rolling update settings to ensure that there's always at least 1 pod available during deployment.

- True up staging/prod preStop settings according to: https://github.com/artsy/artsy-hokusai-templates/blob/ac33c4db8754143318612497f68361b64ea1abd6/nodejs/hokusai/staging.yml.j2#L68

These changes should eliminate some Pingdom alerts.

---
Kubernetes doc on rolling update parameters: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#max-surge

### DIFF
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -55,6 +55,10 @@ spec:
                   value: https
             initialDelaySeconds: 5
             periodSeconds: 5
+          lifecycle:
+            preStop:
+              exec:
+                command: ["sh", "-c", "sleep 10"]
         - name: metaphysics-nginx
           image: artsy/docker-nginx:1.14.2
           ports:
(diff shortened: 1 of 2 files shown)


hokusai/staging.yml
```diff
@@ -6,8 +6,8 @@ metadata:
 spec:
   strategy:
     rollingUpdate:
-      maxSurge: 0
-      maxUnavailable: 20%
```
izakp, line 10: I understand `maxUnavailable` as inversely, at least 80% pods will be available during rollout as [per these docs](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#max-unavailable)

Is the issue more so the horizontal pod autoscaler's setting `minReplicas: 1` conflicting with deployment totoal replicas during rollout?

artsyjian: I think the problem is when actual pod count is at min which is 1. In that situation, because k8s is not allowed to surge (maxSurge 0), during deployment, k8s must delete the one and only pod first, before it can add a new pod. I am not 100% sure that it is happening but it seems very possible based on all the settings.

PR #2617, opened 2020-08-12
[FX-2123] Add a formatted artwork count to ArtistSeries

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-2123

### Problem

When an artist series has no for-sale works we don't want to simply show "0 available"

### Solution

- Use the count of all works (not just for-sale ones) as a fallback
- Return a consistently formatted string ("42 available" when there are for-sale works vs "42 works" when there are no for-sale works)

Originally I had this logic living in Force and Eigen, but for the sake of consistency and reusability it seemed sensible to have this in MP, since it is a common formatting concern across clients and surfaces. 

So this will be followed up by PRs to Force and Eigen to simply use the new `artworksCountMessage` field.

Screen Shot 2020-08-12 at 4 43 20 PM

Screen Shot 2020-08-12 at 4 43 26 PM

### DIFF
--- a/src/lib/stitching/gravity/__tests__/stitching.test.ts
+++ b/src/lib/stitching/gravity/__tests__/stitching.test.ts
@@ -710,4 +710,48 @@ describe("gravity/stitching", () => {
       expect(formattedDescription).toEqual("**Bold Type**")
     })
   })
+
+  describe("#artworksCountMessage", () => {
+    it("prefers for-sale artworks count", async () => {
+      const { resolvers } = await getGravityStitchedSchema()
+      const { artworksCountMessage } = resolvers.ArtistSeries
+      const value = await artworksCountMessage.resolve({
+        forSaleArtworksCount: 20,
+        artworksCount: 10,
+      })
+      expect(value).toEqual("20 available")
+    })
+
+    it("falls back to general artworks count", async () => {
+      const { resolvers } = await getGravityStitchedSchema()
+      const { artworksCountMessage } = resolvers.ArtistSeries
+      const value = await artworksCountMessage.resolve({
+        forSaleArtworksCount: 0,
+        artworksCount: 10,
+      })
+      expect(value).toEqual("10 works")
+    })
+
+    it("pluralizes correctly", async () => {
+      const { resolvers } = await getGravityStitchedSchema()
+      const { artworksCountMessage } = resolvers.ArtistSeries
+
+      const zero = await artworksCountMessage.resolve({
+        forSaleArtworksCount: 0,
+        artworksCount: 0,
+      })
+      const one = await artworksCountMessage.resolve({
+        forSaleArtworksCount: 0,
+        artworksCount: 1,
+      })
+      const many = await artworksCountMessage.resolve({
+        forSaleArtworksCount: 0,
+        artworksCount: 42,
+      })
+
+      expect(zero).toEqual("0 works")
+      expect(one).toEqual("1 work")
+      expect(many).toEqual("42 works")
+    })
+  })
 })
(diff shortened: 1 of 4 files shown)


src/lib/stitching/gravity/__tests__/stitching.test.ts
```diff
@@ -707,4 +707,55 @@ describe("gravity/stitching", () => {
...
+      const { resolvers } = await getGravityStitchedSchema()
+      const { artworksCountMessage } = resolvers.ArtistSeries
+
+      const value = await artworksCountMessage.resolve({
+        forSaleArtworksCount: 0,
+        artworksCount: 0,
+      })
+
+      expect(value).toEqual(null)
+    })
```
anandaroop, line 759: This is not specified in any design but I figured returning a falsy value and not displaying a count at all was better than "0 works" — since that's what we were trying to avoid in the first place.

sweir27: Ohh interesting. I probably would expect this to say `0 works`! But if that would cause us to then do some logic/checks on the presence of the "0 works" string, then this makes sense.

sweir27:
Nice tests! 

**thought (non-blocking)**: I have slight reservations against making the string `null` if there are `0 works` but it's not a strong opinion. Feel free to merge as-is!

src/lib/stitching/gravity/__tests__/stitching.test.ts
anandaroop: Cool, so I've put in a question to Nicole about whether we should show a "0 works" message or nothing at all for this edgy edge case — and will let that guide the decision 😎.

sweir27: Sounds good!

src/lib/stitching/gravity/stitching.ts
```diff
@@ -197,6 +202,23 @@ export const gravityStitchingEnvironment = (
...
+        artworksCountMessage: {
+          fragment: gql`
+            ... on ArtistSeries {
+              forSaleArtworksCount
+              artworksCount
+            }
+          `,
+          resolve: async ({ forSaleArtworksCount, artworksCount }) => {
+            const artworksCountMessage = forSaleArtworksCount
+              ? `${forSaleArtworksCount} available`
```
williardx, line 214: **suggestion (non-blocking):** Reformat as explicit `if/else` statements since the nested ternaries make it difficult to quickly grasp what's going on. Non-blocking mostly because it's still clear, just requires a little more effort.

anandaroop: Good call

src/lib/stitching/gravity/__tests__/stitching.test.ts
anandaroop: Confirmed that we want to show "0 works" — commit incoming

anandaroop: Fixed in https://github.com/artsy/metaphysics/pull/2617/commits/668c3733457c6075c765deeed9c6ff22754418ca#diff-e9a6ef1ded9b9a8e03c9320518a09a23R749

src/lib/stitching/gravity/stitching.ts
anandaroop: Fixed in https://github.com/artsy/metaphysics/commit/668c3733457c6075c765deeed9c6ff22754418ca#diff-573ae00d2f637334b78f6df4edf48571R215-R221

anandaroop:
Thanks for the comments — both points addressed. This should be g2g 😄

PR #2620, opened 2020-08-14
Make sure salesConection resolver uses a loader without cache if registered is false

yuki24:
addresses https://github.com/artsy/metaphysics/pull/2616#discussion_r470315496

@mzikherman has pointed out that we aren't using the `loaderWithoutCache` properly when the `registered` argument is `false`. I wasn't sure how I should check it in JS as it could be `undefined` or `null` when it's not specified, so I ended up checking the type of the value and pick up the `loaderWithoutCache` if it's a boolean value. I'm still not 100% confidenent, but at least the test should demonstrate that the authenticated loader is properly picked up.

Let me know if there is anything to think about in this PR.

### DIFF
--- a/src/schema/v2/sales.ts
+++ b/src/schema/v2/sales.ts
@@ -63,7 +63,8 @@ export const SalesConnectionField: GraphQLFieldConfig<void, ResolverContext> = {
       paginationArgs
     )
 
-    const loader = registered ? loaderWithoutCache : loaderWithCache
+    const loader =
+      typeof registered === "boolean" ? loaderWithoutCache : loaderWithCache
     const { body: sales, headers } = ((await loader!(
       {
         id: ids,
--- a/src/schema/v2/__tests__/salesConnection.test.ts
+++ b/src/schema/v2/__tests__/salesConnection.test.ts
@@ -12,16 +12,11 @@ describe("salesConnection", () => {
             .stub()
             .withArgs("sales", {
               first: 5,
-              isAuction: true,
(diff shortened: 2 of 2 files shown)


src/schema/v2/sales.ts
```diff
@@ -63,7 +63,8 @@ export const SalesConnectionField: GraphQLFieldConfig<void, ResolverContext> = {
       paginationArgs
     )
 
-    const loader = registered ? loaderWithoutCache : loaderWithCache
+    const loader =
+      typeof registered === "boolean" ? loaderWithoutCache : loaderWithCache
```
mzikherman, line 67: This should be fine I think! I've used `Boolean` to coerce before:

Screen Shot 2020-08-17 at 10 28 02 AM

yuki24: In this case we actually want to distinguish "not specified"  from `true` or `false` as it means we don't filter at all based on the registration status. This brings up @erikdstock's concern about the API not being intuitive enough and we might want to switch to a more enum-based approach, such as:

 * `registered: ALL`: to return any sales or not filter at all
 * `registered: REGISTERED`: to return only the sales the user has registered for
 * `registered: UNREGISTERED`: to return only the sales the user has _not_ registered for

@erikdstock WDYT?

yuki24:
I'll go ahead and merge this PR as this should be good enough to fix the original issue, and we should start a new conversation about how the argument should look as a public API.

PR #2623, opened 2020-08-17
Auct1155 stitch causality

erikdstock:
Since I don't have a whole lot of experience in metaphysics and 0 in schema stitching, I'll wait for at least 1 approving review from someone else before merging this.

Finishes [AUCT-1155](https://artsyproduct.atlassian.net/browse/AUCT-1155) bringing causality's up-to-the-instant lot standings to users. Lots to work on here- so far we have followed [the new graphql microservice doc](https://github.com/artsy/metaphysics/blob/master/docs/adding_a_new_graphql_microservice.md) - most of this is addressed in https://github.com/artsy/causality/pull/812, which is a blocker to this ticket.

The current state stitches causality's lot standings into `Me` and `saleArtwork` into an `AuctionsLotStanding` - so we can query a `me.auctionsLotStandings` with the associated qualitative gravity data below. I don't know if this is going to pose an efficiency problem - perhaps that could be solved by limiting pagination or something? I'm not clear on the best strategies there.
image

✅  Should we prefix fields/types with `Auction` instead of `Causality`? Or have no prefix at all but exclude root fields we don't need yet (eg `sale` might eventually go under a gravity sale, `lot` could go under `saleArtwork`, etc, and we wait until we need them to decide)
✅  Should `lotStandings` be stitched under `me` for now rather than a root-level field? (Probably)
❌  Should we move our method of fetching the gravity user id for constructing a causality jwt, maybe to a causalityTokenLoader? Not for now.
✅  any other types that should be excluded/renamed? Renamed Lot/Sale to AuctionsLotState/AuctionsSaleState

Final things to check on:
- [x] We have to manually add the line `scalar Long` to `causality.graphql` because sangria doesn't include it as an export. Check why and if we can mitigate this. (addressed in artsy/causality#812)
- [x] Check if there is anything we need to do for schema syncing or other instrumentation between these services
- [x] Try to stitch a link to `saleArtwork` into the `AuctionsLotStanding` interface.


src/data/causality.graphql
```diff
@@ -0,0 +1,315 @@
...
+}
+
+"an online (Artsy) bidder, or an offline bidder in the auction"
+union Bidder = ArtsyBidder | OfflineBidder
+
+"DateTime is a scalar value that represents an ISO8601 formatted date and time."
+scalar DateTime
+
+"a long is like an int, figure out why it wasn't included here"
+scalar Long
```
erikdstock, line 14: I think we may need to adjust our schema somehow to make the Long type exist. I manually added this line because sangria-graphql seems to have its [own special scalar type for `Longs`](https://github.com/sangria-graphql/sangria/blob/4ab54d99e9996e72e01e8ecdf2dafa67426c523a/modules/core/src/main/scala/sangria/schema/package.scala#L30-L33)

src/lib/stitching/causality/schema.ts
```diff
@@ -0,0 +1,30 @@
...
+  makeRemoteExecutableSchema,
+  transformSchema,
+  RenameTypes,
+  RenameRootFields,
+} from "graphql-tools"
+import { readFileSync } from "fs"
+
+export const executableCausalitySchema = () => {
+  const threeBodyLink = createCausalityLink()
+  const threeBodyTypeDefs = readFileSync("src/data/causality.graphql", "utf8")
```
erikdstock, line 12: update names

src/schema/v2/system/causality_jwt.ts
```diff
@@ -10,6 +10,19 @@ import { ResolverContext } from "types/graphql"
 
 const { HMAC_SECRET } = config
 
+export const causalityJwt = ({ role, userId, saleId, bidderId }) =>
```
erikdstock, line 13: extracted this method since we are using it outside the resolver now- but not sure where it should go.

src/schema/v2/system/causality_jwt.ts
```diff
@@ -62,67 +70,47 @@ const CausalityJWT: GraphQLFieldConfig<void, ResolverContext> = {
...
-              iat: new Date().getTime(),
-            },
-            HMAC_SECRET
-          )
-        }
-        return jwt.encode(
-          {
-            aud: "auctions",
-            role: "observer",
+          return causalityJwt({
```
erikdstock, line 73: I think I may have lost a logical branch here while rebasing...

src/lib/loaders/loaders_with_authentication/causality.ts
```diff
@@ -0,0 +1,27 @@
+import factories from "../api"
+import config from "config"
+
+const { EXCHANGE_APP_ID } = config
+
+// TODO: Maybe move causality JWT code from causality/link.ts into here.
+// other wise delet this
```
erikdstock, line 7: this file is just copied completely so we may want to delete it.

erikdstock: The exchange version of this file is used for setting up the token auth, which comes from gravity. Causality's JWT is made in the causality_jwt.ts resolver- `causalityJwt` is its own `String` field already used for live auction websocket connections. I extracted the jwt() call to a separate export in that file for now, but maybe we could factor it out into a causalityTokenLoader here and use that in both places.

Or, maybe we should just allow the repetition. We make a causality jwt by saying `meLoader.then( ({id}) => causalityJwt({userId: id, role, bidderId, saleId}) )` where none of these params but `role` absolutely required.

zephraph:
> Should we prefix fields/types with Auction instead of Causality? Or have no prefix at all but exclude root fields we don't need yet (eg sale might eventually go under a gravity sale, lot could go under saleArtwork, etc, and we wait until we need them to decide)

Not exposing the root fields that you're not immediately using is a definite yes. If you decide to prefix then I believe `Auction` would be better than `Causality` generally. It's worth digging into why some setups (such as exchange) pursued prefixing. Is it because they need to not conflict during the stitching process? (I don't know off the top of my head). 

> Should lotStandings be stitched under me for now rather than a root-level field? (Probably)

Probably, yeah. In general we want as few root fields as possible with them mostly reflecting high level domain objects. I'd recommend referring to our graphql schema design playbook: https://github.com/artsy/README/blob/master/playbooks/graphql-schema-design.md#root-fields

> Should we move our method of fetching the gravity user id for constructing a causality jwt?

I'd need more context here, not sure. 

> any other types that should be excluded/renamed?

TBD

src/data/causality.graphql
erikdstock: this has been fixed on the causality end in artsy/causality#812

src/schema/v2/system/causality_jwt.ts
erikdstock: I considered making a causalityTokenLoader to take care of this but there wasn't an obvious, clean way to do it that would meet the needs of both this file and our general causality `link.ts` file. So for now it is just awkwardly exported here.

ds300 (approved):
This looks great to me, nice one!

src/config.ts
```diff
@@ -161,6 +165,7 @@ export default {
     ENABLE_COMMERCE_STITCHING === undefined
       ? true
       : ENABLE_COMMERCE_STITCHING === "true",
+  ENABLE_CAUSALITY_STITCHING: ENABLE_CAUSALITY_STITCHING === "true",
```
damassi, line 168: Don't think this is necessary. These flags were useful in the past when stitching was experimental, and if we needed to we could flip it off and fall back to regular loader patterns. 

Since there aren't loaders for Causality that map 1:1 to what is being stitched, we don't need a flag since setting it to `false` would break consumers.

src/lib/stitching/mergeSchemas.ts
```diff
@@ -56,6 +59,15 @@ export const incrementalMergeSchemas = (
     gravityStitchingEnvironment(localSchema, gravitySchema, version)
   )
 
+  if (ENABLE_CAUSALITY_STITCHING) {
```
damassi, line 62: ✂️

PR #2624, opened 2020-08-17
Adds remaining fair-related fields

sweir27:
This PR adds the remaining fields we need to populate the "More info" page on the new fair landing.

It also makes a few of the fields `markdown`-style, which means they can be formatted into `HTML`. By default (if you don't pass a `format: HTML` argument) it returns the same value that's been inputted so it should be backwards-compatible.

I didn't add tests since we're just passing through fields from gravity so I didn't think they added much, but happy to!

image


src/schema/v2/fair.ts
```diff
@@ -288,6 +285,11 @@ export const FairType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       startAt: date,
+      summary: {
+        type: GraphQLString,
```
sweir27, line 289: Unclear to me whether this ever will be markdown... I don't _think_ so. Easy enough to change later if we learn differently.

PR #2630, opened 2020-08-18
[MX-482] Add displayUrgencyTag to sales

MounirDhahri:
The type of this PR is: **Feature**

This PR resolves [MX-482]

### Description
Added `displayUrgencyTag` to `Sale` 

[MX-434]: https://artsyproduct.atlassian.net/browse/MX-434
[MX-482]: https://artsyproduct.atlassian.net/browse/MX-482

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -40,7 +40,6 @@
     "expect": false,
     "schema": false,
     "sinon": false,
-    "__TEST__": false,
     "expectPromiseRejectionToMatch": false
   },
   "settings": {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -40,6 +40,7 @@
     "expect": false,
     "schema": false,
     "sinon": false,
+    "__TEST__": false,
     "expectPromiseRejectionToMatch": false
   },
   "settings": {
--- a/src/schema/v2/sale/display.ts
+++ b/src/schema/v2/sale/display.ts
@@ -79,10 +79,10 @@ export async function displayTimelyAt({ sale, meBiddersLoader }) {
   // Timed Auction
   if (start_at) {
     const range = moment().add(5, "days")
-    const startAt = moment(start_at)
-    const isInProgress = startAt < moment()
-    const isNearFuture = startAt > moment() && startAt < range
-    const isFuture = startAt > range
+    const currentTime = moment(start_at)
+    const isInProgress = currentTime < moment()
+    const isNearFuture = currentTime > moment() && currentTime < range
+    const isFuture = currentTime > range
     const dateLabel = (saleDate) => {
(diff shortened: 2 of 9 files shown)


src/types/global.d.ts
```diff
@@ -3,3 +3,9 @@ declare module "*.json" {
   const value: any
   export default value
 }
+
```
MounirDhahri, line 6: Global declarations for json imports were broken since we were not referencing to the types correctly. Please read this for more details: https://mariusschulz.com/articles/importing-json-modules-in-typescript

tsconfig.json
```diff
@@ -12,9 +12,11 @@
     "noUnusedParameters": true,
     "rootDir": "./src",
     "pretty": true,
+    "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
     "target": "es5"
   },
+  "include": ["src/types/**/*.d.ts"],
```
MounirDhahri, line 20: we have declared types within `src/types/**/*.d.ts` that were not recognized since they were not included to `tsconfig.json`

src/schema/v2/sale/display.ts
```diff
@@ -100,3 +100,73 @@ export async function displayTimelyAt({ sale, meBiddersLoader }) {
     }
   }
 }
+
+const defaultStartAt = !__TEST__
```
MounirDhahri, line 104: Avoid tests breaking because of date change

src/schema/v2/sale/display.ts
```diff
@@ -100,3 +100,73 @@ export async function displayTimelyAt({ sale, meBiddersLoader }) {
...
+ * Get sale ending urgency tag
+ * @example
+ * 2 days left, 12 hours left
+ * Auction closed
+ */
+
+export const displayUrgencyTag = ({
+  endAt,
+  auctionState,
+  startAt = defaultStartAt,
```
ds300, line 118: what does `startAt` represent the start of? Seems like it's used as 'the current time'. In which case I think you should not use this `defaultStartAt` property, which doesn't ever change, and instead have a `getCurrentTime()` function so that it stays up to date for every request.

MounirDhahri: My intention was to make the function return the urgency tag between any random dates, but  you're right, it makes more sense to call it `currentTime` and to switch `defaultStartAt` to a function (`getCurrentTime()`) that is up to date for every request

.eslintrc.json
```diff
@@ -40,6 +40,7 @@
     "expect": false,
     "schema": false,
     "sinon": false,
+    "__TEST__": false,
```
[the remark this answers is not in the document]
MounirDhahri: nice, I wasn't aware of it. I will adjust it

PR #2632, opened 2020-08-19
[MX-459] - Add endpoint for resolving fairs and partners from profile ids

brainbicycle:
This adds an endpoint to return a partner or fair given a profile id as first step in plan @ds300 outlined here to move this logic out of Eigen and allow us to remove some old Objective-C code for determining what these entities are:
https://github.com/artsy/eigen/pull/3739#issuecomment-675355830

Follow-up to:
https://github.com/artsy/eigen/pull/3739

**Question:** Any thoughts on alternatives for naming here? I find vanityURLEntity a little opaque


src/schema/v2/vanityURLEntity.ts
```diff
@@ -0,0 +1,61 @@
...
+  },
+  resolve: (_root, { id }, { profileLoader, partnerLoader, fairLoader }) =>
+    profileLoader(id).then((profile) => {
+      const ownerType = profile.owner?.type ?? profile.owner_type // what is the difference between these fields?
+      switch (ownerType) {
+        case "Gallery":
+        case "PartnerGallery":
+        case "Institution":
+        case "PartnerInstitution":
+        case "PartnerInstitutionalSeller":
```
ds300, line 43: I think these should be any subclass of `Partner` (and maybe `Partner` itself?)

The full list seems to be:

 - `PartnerArtistDocument`
 - `PartnerAuction`
 - `PartnerBrand`
 - `PartnerDemo`
 - `PartnerGallery`
 - `PartnerInstitution`
 - `PartnerInstitutionalSeller`
 - `PartnerPrivateCollector`
 - `PartnerPrivateDealer`

Although at the very minimum we should cover the ones actually in use

image

which seems to be all except `PartnerArtistDocument`.

ds300 (approved):
I had a couple of comments around using the profileLoader results, but this looks great to me otherwise. Nice job on the tests! 🙇‍♂️

src/schema/v2/vanityURLEntity.ts
brainbicycle: 🙇‍♂️ That is a nifty trick I need to up my ruby/rails fu, was wondering how to find a canonical list of these. I'll update the list.

src/schema/v2/vanityURLEntity.ts
```diff
@@ -34,13 +34,19 @@ const VanityURLEntity: GraphQLFieldConfig<void, ResolverContext> = {
   },
   resolve: (_root, { id }, { profileLoader, partnerLoader, fairLoader }) =>
     profileLoader(id).then((profile) => {
-      const ownerType = profile.owner?.type ?? profile.owner_type // what is the difference between these fields?
-      switch (ownerType) {
+      switch (profile.owner_type) {
         case "Gallery":
-        case "PartnerGallery":
+        case "Partner":
         case "Institution":
```
ds300: i think we don't need `"Gallery"` or `"Institution"` now since `owner_type` doesn't remove the `Partner*` prefix from the type name.

brainbicycle: 👍 Yeah I don't think we need `Partner` either, I'll find examples of each of these types when doing end to end testing on this.

PR #2654, opened 2020-09-01
[MyCollection] Refactor mutations

damassi:
Digging through other mutations for example code I noticed that we were using an interesting `thingOrDelete` pattern which seemed more robust than the more vanilla mutations we were currently using in My Collection. This refactors things per [Alloy's blog post](https://artsy.github.io/blog/2018/10/19/where-art-thou-my-error/). 

Additionally, this DRYs things up and adds a delete mutation. 

```graphql
# Create 

mutation CreateArtwork($input: MyCollectionCreateArtworkInput!) {
  myCollectionCreateArtwork(input: $input) {
    artworkOrError {
      ... on MyCollectionArtworkMutationSuccess {
        artworkEdge {
          node {
            medium
            id
            internalID
            dimensions {
              cm
              in
            }
          }
        }
      }
      ... on MyCollectionArtworkMutationFailure {
        mutationError {
          error
        }
      }
    }
  }
}
```

```graphql
# Delete 

mutation DeleteArtwork($input: MyCollectionDeleteArtworkInput!) {
  myCollectionDeleteArtwork(input: $input) {
    artworkOrError {
      ... on MyCollectionArtworkMutationDeleteSuccess {
        success 
      }
      ... on MyCollectionArtworkMutationFailure {
        mutationError {
          message
        }
      }
    }
  }
}
```

### DIFF
--- a/src/schema/v2/me/__tests__/myCollectionCreateArtworkMutation.test.ts
+++ b/src/schema/v2/me/__tests__/myCollectionCreateArtworkMutation.test.ts
@@ -2,29 +2,61 @@ import { runAuthenticatedQuery } from "schema/v2/test/utils"
 import gql from "lib/gql"
 
 describe("myCollectionCreateArtworkMutation", () => {
-  it("creates an artwork", async () => {
-    const mutation = gql`
-      mutation {
-        myCollectionCreateArtwork(
-          input: {
-            artistIds: ["4d8b92b34eb68a1b2c0003f4"]
-            medium: "Painting"
-            dimensions: "20x20"
-            title: "hey now"
-            year: "1990"
-          }
-        ) {
-          artwork {
-            medium
-          }
-          artworkEdge {
-            node {
+  const mutation = gql`
+    mutation {
+      myCollectionCreateArtwork(
+        input: {
+          artistIds: ["4d8b92b34eb68a1b2c0003f4"]
+          medium: "Painting"
+          dimensions: "20x20"
+          title: "hey now"
+          year: "1990"
+        }
+      ) {
+        artworkOrError {
+          ... on MyCollectionArtworkMutationSuccess {
+            artwork {
               medium
+            }
+            artworkEdge {
+              node {
+                medium
+              }
+            }
+          }
+          ... on MyCollectionArtworkMutationFailure {
+            mutationError {
+              message
             }
           }
         }
       }
-    `
+    }
+  `
+
+  it("returns an error", async () => {
+    const context = {
+      myCollectionCreateArtworkLoader: () =>
+        Promise.reject(
+          new Error(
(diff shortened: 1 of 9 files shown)


src/schema/v2/me/__tests__/myCollectionCreateArtworkMutation.test.ts
```diff
@@ -2,29 +2,61 @@ import { runAuthenticatedQuery } from "schema/v2/test/utils"
...
+      myCollectionCreateArtwork(
+        input: {
+          artistIds: ["4d8b92b34eb68a1b2c0003f4"]
+          medium: "Painting"
+          dimensions: "20x20"
+          title: "hey now"
+          year: "1990"
+        }
+      ) {
+        artworkOrError {
```
pepopowitz, line 16: 👍 Seeing this in practice helps that article click for me.

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
```diff
@@ -33,47 +30,36 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
...
+        year,
+      })
+
+      return response
+    } catch (error) {
+      const formattedErr = formatGravityError(error)
+      if (formattedErr) {
+        return { ...formattedErr, _type: "GravityMutationError" }
+      } else {
+        throw new Error(error)
```
pepopowitz, line 61: **question [potentially blocking]:**

Is there anything to be worried about with throwing a generic `Error` here, given that [the Graphql type wrapping these errors](https://github.com/artsy/metaphysics/pull/2654/files#diff-8d4fff9e1cab233bb45d1369c1325a66R133) seems to assume that any underlying errors are `GravityMutationError`s? 

_Comment format based on [conventionalcomments.org](https://conventionalcomments.org/)_

damassi: This is 🍝 which matches the other examples in codebase. What happens under the hood is if there's no match for that type then it will throw at a lower level and return an exception, versus the graceful handling setup via this design pattern.

damassi:
Yah @pepopowitz - It was def fun to rework this; learned a bit more about how Graphql unions work and had a bit of an "aha" moment once I encountered the `isTypeOf` method where everything clicked. Not so bad...

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
damassi: Looking at this, we could probably wrap this error stuff in a reusable function and DRY up the codebase more. I can do that in a follow-up.

pepopowitz: No rush on my end for that follow-up, I mostly just wanted to make sure I understood the code that was there.

PR #2656, opened 2020-09-01
add Ingress resources

izakp:
Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-2602

Will follow-up with a PR here and in Substance to switch Cloudflare and CoreDNS DNS records.


izakp:
Not sure why these Peril failures but unrelated so merging...

PR #2657, opened 2020-09-02
[MyCollection] Add sorting to connection

damassi:
Related https://github.com/artsy/gravity/pull/13433 

Adds additional sort parameters to our artworks connection.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -6755,6 +6755,7 @@ type Me implements Node {
     before: String
     first: Int
     last: Int
+    sort: ArtworkSorts
   ): MyCollectionConnection
   name: String
   orders(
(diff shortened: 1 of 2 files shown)


_schemaV2.graphql
```diff
@@ -7114,6 +7115,10 @@ type Mutation {
   ): UpdateViewingRoomSubsectionsPayload
 }
 
+type MyCollectionArtworkMutationDeleteSuccess {
```
pepopowitz, line 7118: I'm assuming this was just missed in #2654, and doesn't actually have anything to do with the spirit of this PR, right?

damassi: Hmm not too sure why these changes are here, but they look right!

PR #2661, opened 2020-09-02
Mounts query batching endpoint

dzucconi:
[Hide diff whitespace changes](https://github.com/artsy/metaphysics/pull/2661/files?diff=unified&w=1)

On it's own this doesn't do anything but enable query batching for Relay. The clients will have to mount the batching middleware for this to do be utilized.

[Here's an article about Apollo batching discussing some tradeoffs, but it's essentially the same thing as the Relay variant](https://ednsquare.com/story/improving-performance-with-batching-client-graphql-queries------XxKSNF).

By default, if [a query occurs at the same time as any others](https://github.com/relay-tools/react-relay-network-modern/blob/ef171f17b0407c77a02089c8a3d86f35933cd633/src/middlewares/batch.js#L56), then those queries will be sent as a batch, together. — this can be tuned into a longer interval, if desirable.

So for instance, on initial page load, the global nav actually makes 3 separate queries. Those can be batched into a single network request. One of the nice side-effects of this is that the data loaders are shared across every query because they share the same request context.

-------

My interest in enabling this is in building the fair exhibitor artwork rails, we have to load a very large amount of data, and traditional pagination may not make sense. Having independent queries and then batching them as they enter the screen in groups can help reduce the amount of network requests you'd make if you scrolled quickly down the page.


_schemaV2.graphql
```diff
@@ -7124,7 +7128,8 @@ type MyCollectionArtworkMutationSuccess {
 }
 
 union MyCollectionArtworkMutationType =
-    MyCollectionArtworkMutationFailure
+    MyCollectionArtworkMutationDeleteSuccess
+  | MyCollectionArtworkMutationFailure
```
dzucconi, line 7132: Unrelated, from one of the pre-push scripts. 🤷

src/index.ts
```diff
@@ -26,6 +26,9 @@ import { middleware as requestIDsAdder } from "./lib/requestIDs"
 import { nameOldEigenQueries } from "./lib/modifyOldEigenQueries"
 import { rateLimiterMiddleware } from "./lib/rateLimiter"
 import { graphqlErrorHandler } from "./lib/graphqlErrorHandler"
+// TODO: Update typings
+// @ts-ignore
+import { graphqlBatchHTTPWrapper } from "react-relay-network-modern"
```
dzucconi, line 31: Will need to PR the typings update as this export isn't typed.

src/index.ts
```diff
@@ -181,66 +184,66 @@ function startApp(appSchema, path: string) {
...
-          userAgent,
-        }
+    const graphqlServer = graphqlHTTP((req, res, params) => {
+      const accessToken = req.headers["x-access-token"] as string | undefined
+      const appToken = req.headers["x-xapp-token"] as string | undefined
+      const userID = req.headers["x-user-id"] as string | undefined
+      const timezone = req.headers["x-timezone"] as string | undefined
+      const userAgent = req.headers["user-agent"]
+
+      const { requestIDs = {} } = res.locals ?? {}
```
dzucconi, line 194: This line contains the only change here. `res.locals` is undefined when batched.

dzucconi: The rest of this diff is just moving the middleware instantiation into a variable `graphqlServer` as it's mounted directly, and then inside the batch middleware.

dzucconi: I don't know if this has any potential impact on how we do logging.

dzucconi:
> What if the calling code issues them at different times

Then they just wouldn't be batched. By default they would have to happen within the same tick of the process. Though in looking at what's practical I think a window of a few milliseconds is more appropriate. They would never split authentication of course, this still respects the same auth headers; they are batched on the client, not the server.

> or even (in a pathological case) needs the response from one to issue the next?

In such a case I'd *imagine* that if one request depended on the response of the other, then they couldn't occur at the same time? I'm perhaps not that imaginative though. 😬 

> Even with batching, this seems like a likely performance problem, if not for the client, than for the upstream APIs.

Yeah, I think I tend to agree based on what I'm seeing in development. Though interestingly I'm not seeing highly batched requests take much longer in practice. It probably is just the case that this style of loading for these content rails doesn't make much sense.

On the whole though, enabling this is probably worthwhile. That header optimization alone seems worthwhile, cutting every page request from 3 independent metaphysics queries, to a single one. I could see us just conditionally mounting the batch middleware in Force with an environment variable to see what effect this has in practice.

src/index.ts
```diff
@@ -181,66 +184,66 @@ function startApp(appSchema, path: string) {
...
-            variables: params && params.variables,
-            query: (params && params.query)!,
-          }),
-          validationRules,
-          extensions: ({ document, result }) =>
-            createExtensions(document, result, requestID),
-        }
-      })
-    )
+    app.use("/batch", bodyParser.json(), graphqlBatchHTTPWrapper(graphqlServer))
```
zephraph, line 245: I'm trying to get a better sense of what this is doing. So it receives requests at `/batch` and delegates those to graphqlServer. Is it `/v2/batch` or just `/batch` at the root? If the latter, does it always get serviced by the v2 schema?

src/index.ts
zephraph: Are there any other impacts to `res.locals` being undefined?

src/index.ts
dzucconi: both, since these are mounted on both routes.

src/index.ts
```diff
@@ -181,66 +184,66 @@ function startApp(appSchema, path: string) {
...
-          userID,
-          defaultTimezone,
-          ...loaders,
-          // For stitching purposes
-          exchangeSchema,
-          requestIDs,
-          userAgent,
-        }
+    const graphqlServer = graphqlHTTP((req, res, params) => {
+      const accessToken = req.headers["x-access-token"] as string | undefined
```
damassi, line 188: #sominor Maybe store type info in a type var? 
```
type Header = string | undefined 
```

dzucconi: Oh I concur, tho I'd rather not refactor any of this for the moment 🗡️

src/index.ts
dzucconi: Sorry, I think rather than `res.locals` being undefined. It's just `res.locals.requestIDs`. There's a middleware that shovels them onto the request for logging. And with batch requests, there are no request IDs.

> As of v4.0.0, the batch middleware utilizing request IDs in queries and corresponding results has been renamed legacyBatchMiddleware. The legacy middleware included a request ID with each query included in the batch and expected the server to return each result with the corresponding request ID. The new batchMiddleware simply expects results be returned in the same order as the batched queries.

dzucconi: (But yeah it'd be more appropriate to make this change within that middleware)

zephraph:
Fundamentally I'm 👍 to the addition even if I'm a little wary of it's overuse. I am also slightly concerned with what a higher volume of smaller queries might do with network latency and MP's event loop. I've got nothing to substantiate that concern though so it not be an issue.

It'd be good to be explicit about when batching is preferred to traditional pagination (or whatever the relay equivalent is for whatever is trying to be achieved)

Great work on this though @dzucconi, I'm excited for the potential.

zephraph (approved):

src/index.ts
dzucconi: OK so it actually was being undefined and it was due to a small bug in the batch middleware. Patched the issue along with exporting the type in https://github.com/relay-tools/react-relay-network-modern/pull/116

dzucconi:
I don't love that we have to include the relay runtime with this for no real reason. The batching middleware is simple enough and should maybe just be extracted.

PR #2662, opened 2020-09-02
Backport NoSchemaIntrospectionCustomRule from graphql@15.2.0

joeyAghion:
For reasons described in [PLATFORM-2793](https://artsyproduct.atlassian.net/browse/PLATFORM-2793), it would be nice to be able to require authentication for introspection queries.

I tried to upgrade `graphql` to 15.2.0, which has a built-in rule added (https://github.com/graphql/graphql-js/pull/2600), but encountered tons of failures. For now I'd be satisfied to adapt the rule to our codebase. The goal, ultimately, is for introspection to be enabled in development, but require an authorization header in staging/production.

TODO:
* ~Introspection shouldn't be possible locally in this version of the code, but for some reason still is.~ (This was caused by the `ENABLE_APOLLO` flag being `true` locally.)
* ~Improve typing of rule responses (currently `any`).~ (Fixed, at the expense of some slightly ugly array concatenation.)
* ~Maybe fix tests?~ (What tests? 😝 )


joeyAghion:
Updated to require an `Authorization` header if an `INTROSPECT_TOKEN` is configured and the env is staging or production.

### Migration

```
hokusai [staging|production] env set INTROSPECT_TOKEN=<some new cryptographically random token>
```

We plan to try this^ only in staging until we can surface and fix clients.

.env.example
```diff
@@ -37,6 +37,7 @@ GRAVITY_SECRET=REPLACE
 HMAC_SECRET=https://www.youtube.com/watch?v=F5bAa6gFvLs
 IMPULSE_API_BASE=https://impulse-staging.artsy.net/api
 IMPULSE_APPLICATION_ID=REPLACE
+INTROSPECT_TOKEN=replaceme
```
dzucconi, line 40: Nitpick but I'd change this to `REPLACE` to match the rest of the examples.

dzucconi:
All this looks good to me

PR #2667, opened 2020-09-04
Push schema changes to non-relay clients like Volt, too

joeyAghion:
Today there were several stop-and-go moments while trying to release minor changes to Volt. Metaphysics' schema would undergo modifications, and suddenly the pending Volt release would be blocked on new conflicts (necessitating a manual `rake graphql:schema:update`, PR, merge, build, and release).

@damassi suggested we push schema updates to Volt much like we already do to Force and Eigen.

Volt uses [artemis](https://github.com/yuki24/artemis#the-convention), which expects a JSON-formatted GraphQL schema at a particular file location. I had to hack up the original script to operate in these 2 modes: `.graphql` (with `relay-compiler` and a `data/schema.graphql` file) or `.json` (at an arbitrary other path).

There's some work in progress ([GALL-3230](https://artsyproduct.atlassian.net/browse/GALL-3230)) to start using Relay in Volt, so some of this might be temporary. On the other hand, systems like Pulse and Exchange also use `artemis` and might benefit from this new capability.

* Any and all feedback is welcome, including a total rewrite, since this is way outside my wheelhouse.
* This reference was helpful in figuring out how to map between the different schema formats: https://www.apollographql.com/blog/three-ways-to-represent-your-graphql-schema-a41f4175100d/
* This is hard to test, but I ran the critical commands from a local `node` console and confirmed there were only whitespace changes to Volt's committed schema files, plus a few minor differences that didn't look critical (presumably a result of Volt and Metaphysics using different graphql libraries internally).

### DIFF
--- a/scripts/push-schema-changes.js
+++ b/scripts/push-schema-changes.js
@@ -3,11 +3,19 @@
 const { updateRepo } = require("@artsy/update-repo")
 const { execSync } = require("child_process")
 const path = require("path")
+const { buildSchema, introspectionQuery, graphqlSync } = require("graphql")
+const { readFileSync, writeFileSync } = require("fs")
 
 /**
- * @param {'eigen' | 'force'} repo
+ * @param {string} repo - Name of artsy repo to update
+ * @param {string} [dest=data/schema.graphql] - Path to schema file
+ * @param {boolean} [relay=true] - Runs relay-compiler on schema file (otherwise writes JSON)
  */
-async function updateSchemaFile(repo) {
+async function updateSchemaFile(
+  repo,
+  dest = "data/schema.graphql",
+  relay = true
+) {
   await updateRepo({
     repo: {
       owner: "artsy",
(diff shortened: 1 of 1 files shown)


scripts/push-schema-changes.js
```diff
@@ -3,11 +3,19 @@
...
- * @param {'eigen' | 'force'} repo
+ * @param {string} repo - Name of artsy repo to update
+ * @param {string} [dest=data/schema.graphql] - Path to schema file
+ * @param {boolean} [relay=true] - Runs relay-compiler on schema file (otherwise writes JSON)
  */
-async function updateSchemaFile(repo) {
+async function updateSchemaFile(
+  repo,
+  dest = "data/schema.graphql",
+  relay = true
```
damassi, line 17: Since we're now using multiple arguments this would probably benefit from named args so that the invocation is more readable. Something like:

```js
async function updateSchema({
  repo,
  dest = "data/schema.graphql",
  relay = true
}) {
  ...
}
```
Then down below we can do 
```js
updateSchemaFile({
  repo: "eigen",
  dest: "...",
  relay: false
})
```

scripts/push-schema-changes.js
```diff
@@ -18,16 +26,25 @@ async function updateSchemaFile(repo) {
...
     labels: ["Merge On Green"],
     update: (repoDir) => {
-      execSync(
-        `cp _schemaV2.graphql '${path.join(repoDir, "data/schema.graphql")}'`
-      )
+      const repoDest = path.join(repoDir, dest)
       execSync("yarn install --ignore-engines", { cwd: repoDir })
-      execSync("./node_modules/.bin/relay-compiler", { cwd: repoDir })
-      execSync("./node_modules/.bin/prettier --write data/schema.graphql", {
+      if (relay) {
```
ds300, line 35: nonblocking: I'd probably have skipped the `relay` flag and done this by checking the file extension.

ds300 (approved):
Aye this looks like it'll do the trick 🍏 

I'll merge now so we can test this, and if you want to create a follow up PR to convert to named arguments that's cool. FWIW I'm in favour of named arguments generally, but in this case the function won't escape this one small file so I don't mind either way.

ds300:
Merging this caused a failure due to using `__dirname` to resolve the path for the schema file. I should have caught that ahead of time, apologies.

Added a fixup commit here: https://github.com/artsy/metaphysics/commit/6561e25a0fcfd69472f13aa1564e05a341cab7ca

scripts/push-schema-changes.js
joeyAghion: I was wondering about that and didn't know whether we could assume the `.graphql`/`.json` association or not. Thanks!

joeyAghion:
Thanks for the fix and refactor... much nicer now. The first automatic changes are [here](https://github.com/artsy/volt/pull/4610) so I'm reviewing them carefully.

ds300:
Might be a good idea to use a deterministic json serializer to keep diffs small e.g. https://www.npmjs.com/package/fast-json-stable-stringify

joeyAghion:
Now that the follow-up PRs are happening ([e.g.](https://github.com/artsy/volt/pull/4611/files)), the diffs do seem small and the JSON stable. I chalk the big initial diff up to:
* differences in how the schema-printing of legacy `commentDescription`s handled missing comments (`null` vs. `""`), and
* differences between artemis's output using [`JSON.pretty_generate`](https://github.com/yuki24/artemis/blob/bc892448e4cc357065b4069c650afd949730a08d/lib/tasks/artemis.rake#L38) vs. `prettier`'s output.

PR #2669, opened 2020-09-04
Update Pulse's schema automatically too

joeyAghion:
Like Volt, Pulse uses `artemis` to store and use a copy of Metaphysics' schema. Unlike Volt, it doesn't use `prettier`, so I had to make that step optional.


ds300 (approved):
🙇

PR #2673, opened 2020-09-08
Fix broken links in README

starsirius:
Instead of the _relative_ links from [README](https://github.com/artsy/README/blob/master/README.md), we should use absolute links.


sepans:
💯

PR #2683, opened 2020-09-09
Enable includeArtworksByFollowedArtists flag on SaleArtworksConnection

MounirDhahri:
Enable `includeArtworksByFollowedArtists` flag on SaleArtworksConnection

By default, I am setting the flag to `true` to keep the existing behavior on Force

### DIFF
--- a/src/schema/v2/sale_artworks.ts
+++ b/src/schema/v2/sale_artworks.ts
@@ -55,7 +55,7 @@ const SaleArtworkCounts = {
   resolve: (artist) => artist,
 }
 
-const SaleArtworksConnectionType = connectionWithCursorInfo({
+export const SaleArtworksConnectionType = connectionWithCursorInfo({
   name: "SaleArtworks",
   nodeType: ArtworkType,
   edgeType: SaleArtworkType,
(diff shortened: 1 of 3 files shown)


src/schema/v2/sale_artworks.ts
```diff
@@ -115,7 +115,7 @@ export const SaleArtworksConnectionField: GraphQLFieldConfig<
     _root,
     {
       // artistIDs,
-      // includeArtworksByFollowedArtists,
+      includeArtworksByFollowedArtists = true,
```
ashfurrow, line 118: We should use the arguments definition to expose default variables in the schema, for example:

```tsx
includeArtworksByFollowedArtists: {
  type: GraphQLBoolean,
  defaultValue: true
},
```

But I don't think that `true` is a good default here, due to this TODO comment above in the file:

https://github.com/artsy/metaphysics/blob/e1c227edc16ac0f38ab28db73c32c26bd8b9d5d7/src/schema/v2/sale_artworks.ts#L74-L77

I've [asked on Slack](https://artsy.slack.com/archives/CP9P4KR35/p1599681100120300) for suggestions here, since I'm not sure what exactly the best solution is.

src/schema/v2/sale_artworks.ts
```diff
@@ -114,21 +112,36 @@ export const SaleArtworksConnectionField: GraphQLFieldConfig<
...
+    this type was only accessible under `Me` and includeArtworksByFollowedArtists
+    was always true. A TODO (see below) was left to clean this up, but before it
+    could be cleaned up, the type was already being used elsewhere. We need to keep
+    the schema backwards-compatible while also allowing non-`Me` uses to have undefined
+    for this argument. So we use this workaround.
+
+    See: https://github.com/artsy/metaphysics/blob/e1c227edc16ac0f38ab28db73c32c26bd8b9d5d7/src/schema/v2/sale_artworks.ts#L74-L77
+     */
+    let includeArtworksByFollowedArtists = requestIncludeArtworksByFollowedArtists
+    if (info.parentType.name === "Me") {
```
damassi, line 138: Nice, so that's how it's done 👍

damassi (approved):

ashfurrow:
Thanks for the review, Chris 👍 Re-assigning back to @MounirDhahri for merging.

MounirDhahri:
Thanks for taking care of this @ashfurrow 🙏

PR #2685, opened 2020-09-09
Add diffusion token loader

jonallured:
Update: this does in fact work but I was having issues because I didn't know that I'd have to pass BOTH headers in the Playground, like so:

```json
{
  "X-Access-Token":"shhhh",
  "Authorization":"Bearer shhhh"
}
```

More chatter here: https://artsy.slack.com/archives/CS3H1K4BD/p1600113637322100


damassi:
#mergeOnGreen

PR #2689, opened 2020-09-10
fix bug in Sale.displayTimelyAt

erikdstock:
Bug discovery and discussion [here 🔒 ](https://artsy.slack.com/archives/CP9P4KR35/p1599767367137800).
cc @joeyAghion @starsirius 

This PR fixes 2 issues I discovered in how we calculate our `displayTimelyAt` property on the `Sale` type. It was happening in places we do not convert a JSON string timestamp into a `Moment` before doing a time comparison. In the process of fixing the first bug I noticed we using values that were already moments instead of normal strings, so I converted them all and discovered the second issue.

### DIFF
--- a/src/schema/v2/sale/__tests__/index.test.ts
+++ b/src/schema/v2/sale/__tests__/index.test.ts
@@ -580,110 +580,129 @@ describe("Sale type", () => {
       [
         {
           auction_state: "open",
-          live_start_at: moment().subtract(1, "days"),
-          registration_ends_at: moment().subtract(2, "days"),
+          live_start_at: moment().subtract(1, "days").toISOString(),
+          registration_ends_at: moment().subtract(2, "days").toISOString(),
         },
         "in progress",
       ],
-      [{ end_at: moment().subtract(1, "days") }, null],
+      [
+        {
+          end_at: moment().subtract(1, "days").toISOString(),
+          start_at: moment().subtract(10, "days").toISOString(),
+        },
+        null,
+      ],
       [
         {
           auction_state: "open",
-          live_start_at: moment().subtract(2, "days"),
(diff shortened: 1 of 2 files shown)


src/schema/v2/sale/__tests__/index.test.ts
```diff
@@ -580,110 +580,129 @@ describe("Sale type", () => {
...
         },
         "in progress",
       ],
-      [{ end_at: moment().subtract(1, "days") }, null],
+      [
+        {
+          end_at: moment().subtract(1, "days").toISOString(),
+          start_at: moment().subtract(10, "days").toISOString(),
+        },
+        null,
```
erikdstock, line 593: Even though the bug was present, this test was incorrectly passing (expected string is L593 here) because `start_at` was not present- the bug does not manifest if the `start_at` value is not present - see [below](https://github.com/artsy/metaphysics/pull/2689/files#diff-6a8cbee18d0733cd05693bceea3c8cd5R83)

src/schema/v2/sale/__tests__/index.test.ts
```diff
@@ -580,110 +580,129 @@ describe("Sale type", () => {
       [
         {
           auction_state: "open",
-          live_start_at: moment().subtract(1, "days"),
-          registration_ends_at: moment().subtract(2, "days"),
+          live_start_at: moment().subtract(1, "days").toISOString(),
+          registration_ends_at: moment().subtract(2, "days").toISOString(),
```
starsirius, line 584: Good to get the type right! 👍

starsirius (approved):
Great catch! Thanks for fixing it (and the tests)! @erikdstock 💯

PR #2697, opened 2020-09-15
Update .env.example with DIFFUSION_APP_ID

damassi:
Updates `.env.example` with `DIFFUSION_APP_ID`. Noticed this was missing.


jonallured:
Good call, thanks!! 🙌

PR #2699, opened 2020-09-15
[Artwork] Add metric field to artwork, as well as individual dimensions

damassi:
Related https://github.com/artsy/gravity/pull/13468

Adds `metric`, `height`, `width` and `depth` fields to the Artwork type. 

Screen Shot 2020-09-15 at 1 06 30 PM


jonallured (approved):
LGTM!! 🎉

PR #2714, opened 2020-09-23
Shows all fairs on home, regardles of mobile image

ashfurrow:
We were previously only showing fairs on the home page when they had a `mobile_image` set, but this is causing issues (see: [Slack thread](https://artsy.slack.com/archives/CN8S32FJR/p1600876254022800) and a [recent waves change](https://github.com/artsy/waves/pull/275)).

Since we launched a new home feed early this spring, the app doesn't actually use this `mobile_image`. Making this change will affect older versions of the app, which will now show a grey circle on the old home screen (not a huge UX problem but wanted to bring it up).

Tested this locally:

Screen Shot 2020-09-23 at 16 41 03


sweir27:
Thank you!

PR #2718, opened 2020-09-24
[My Collection]: Add edition number + size to my collection mutation params

ashleyjelks:
Adds the parameters `editionSize` and `editionNumber` to myCollection's create and update mutations.

https://artsyproduct.atlassian.net/browse/CSGN-379

Screen Shot 2020-09-24 at 4 36 12 PM
Screen Shot 2020-09-24 at 4 36 58 PM


jonallured (approved):
Nice one! It would be cool if these payloads could be computed by a shared util function but no biggie.

#mergeongreen

PR #2719, opened 2020-09-24
enhancement: MIGRATION create Me#pendingIdentityVerificationId field 

erikdstock:
Depends on https://github.com/artsy/gravity/pull/13492
Part of [AUCT-1142].
**MIGRATION - requires setting `FORCE_URL` in staging and production**

Promotes User#pending_identity_verification to a `me` field so we can link a user to their pending identity verification flow. The intent is to allow a user trying to register for a sale to navigate directly to identity verification from the sale without having to leave artsy to check their email.

This PR represents a departure (to my knowledge) from metaphysics conventions as it is now constructing a canonical identity verification flow link, so maybe something to think about.

[AUCT-1142]: https://artsyproduct.atlassian.net/browse/AUCT-1142


erikdstock:
Added, then removed these fields from v1 because I misinterpreted some client code & thought that my use case was still on the v1 schema.

_schemaV2.graphql
```diff
@@ -6987,6 +6987,7 @@ type Me implements Node {
     state: CommerceOrderStateEnum
   ): CommerceOrderConnectionWithTotalCount
   paddleNumber: String
+  pendingIdentityVerificationId: String
```
zephraph, line 6990: In general we'd like to keep client logic as simple as possible. I think providing the URL (so that every client doesn't need to manually construct it) would be a good addition. It also makes it where we only need to go to one place to make an updated. 

You could also just have an `IndentityVerification` or `PendingIdentityVerification` object that contains both the id and the URL so that either are retrievable if necessary. If you think that the id would only be used for generating the URL though, then only passing the URL would be good.

_schemaV2.graphql
```diff
@@ -6344,6 +6344,9 @@ type HomePageSalesModule {
 }
 
 type IdentityVerification {
+  # Verification flow entry point
+  flowURL: String
+
```
zephraph, line 6349: 🙌

src/config.ts
```diff
@@ -92,6 +93,7 @@ const mustHave = {
   CONVECTION_API_BASE,
   CONVECTION_APP_ID,
   DELTA_API_BASE,
+  FORCE_URL,
```
zephraph, line 96: @erikdstock it may be worth considering if we can put any defaults in place here to have it work as folks would expect. If we have a good way to detect if the running instance is staging vs prod that'd be ideal because then we could default it to `staging.artsy.net` & `artsy.net` respectively.

zephraph: Eh, looking through it I haven't seen a good way to do that. I'm fine w/ skipping it.

zephraph (approved):

src/config.ts
erikdstock: :+1: https://github.com/artsy/metaphysics/pull/2719/commits/84f4904bef6c2304c7b893ba29b2a09ca0f2c5a4

src/config.ts
```diff
@@ -170,7 +169,10 @@ export default {
       ? true
       : ENABLE_COMMERCE_STITCHING === "true",
   ENABLE_HEAPDUMPS,
-  FORCE_URL,
+  FORCE_URL:
+    FORCE_URL || NODE_ENV === "production"
+      ? "https://artsy.net"
+      : "https://staging.artsy.net",
```
zephraph, line 175: I'm not sure `NODE_ENV` is a good thing to switch off of. Even in staging, we run `NODE_ENV` as production. We could just default it to `https://www.artsy.net` if it's undefined, otherwise leave it out and people can figure out what they need if/when it errors.

PR #2721, opened 2020-09-25
[My Collection] Expose edition size + set in Artwork query

ashleyjelks:
Adds `editionSize` and `editionNumber` fields  to `Artwork` query object.

Screen Shot 2020-09-25 at 4 20 57 PM

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -411,7 +411,6 @@ describe("Artwork type", () => {
     })
   })
 
-  describe("#edition_sets", () => {
     const query = `
       {
         artwork(id: "richard-prince-untitled-portrait") {
@@ -435,11 +434,11 @@ describe("Artwork type", () => {
           availability: "permanent collection",
         },
         {
-          availabilitiy: "for sale",
+          availability: "for sale",
           price: "$1,000",
         },
         {
-          availabilitiy: "not for sale",
+          availability: "not for sale",
         },
         { forsale: true },
       ]
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -411,6 +411,7 @@ describe("Artwork type", () => {
     })
   })
 
+  describe("#edition_sets", () => {
     const query = `
       {
         artwork(id: "richard-prince-untitled-portrait") {
@@ -434,11 +435,11 @@ describe("Artwork type", () => {
           availability: "permanent collection",
         },
         {
-          availability: "for sale",
+          availabilitiy: "for sale",
           price: "$1,000",
         },
         {
-          availability: "not for sale",
+          availabilitiy: "not for sale",
         },
         { forsale: true },
       ]
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -213,6 +213,18 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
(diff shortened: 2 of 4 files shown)


src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -411,6 +411,49 @@ describe("Artwork type", () => {
     })
   })
 
+  describe("#edition_size", () => {
```
damassi, line 414: 🐫 pls

ashleyjelks: I was following the syntax of all the others tests here...but maybe its just a holdover from before v2?

src/schema/v2/artwork/index.ts
```diff
@@ -213,6 +213,27 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      editionSize: {
+        type: GraphQLString,
+        resolve: ({ edition_sets }) => {
+          if (edition_sets && edition_sets.length === 1) {
+            return edition_sets[0]!.edition_size
+          }
```
damassi, line 221: Can maybe do 

```
if (editionSets?.length) {
  return edition_sets[0].edition_size
} 
```
Or even a one liner (since null / undefined is automatically handled):
```jsx
resolve: ({ edition_sets }) => edition_sets?.[0]?.edition_size
```

ashleyjelks: 💡 !

src/schema/v2/artwork/index.ts
```diff
@@ -213,6 +213,27 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+        resolve: ({ edition_sets }) => {
+          if (edition_sets && edition_sets.length === 1) {
+            const editionNumber =
+              edition_sets[0]!.available_editions?.[0] ?? null
+
+            if (editionNumber) {
+              return Number(editionNumber)
+            }
+          }
+        },
```
damassi, line 235: Same here, can lean heavily on the `?`: 
```
resolve: ({ edition_sets }) => edition_sets?.[0]?.available_editions?.[0]
```

src/schema/v2/artwork/__tests__/artwork.test.js
damassi: Yup likely a hold over! 

Here's my thinking about this kind of stuff: A general rule of thumb I tend to follow is that evolution is always ok -- a similar example is how we used to `snake_case.js` all of our file names way back when, but when we started writing more idiomatic JS we started `camelCasing.js` them -- however, most files in a given directory are still `snake_case`, and rather than renaming everything, we just create all new files in `camelCase` and accept that newer code is more important than old and that mixing some pattterns is ok within reason. 

Same applies here -- even tho old props from v1 were `snake_case` new props are camel, and in an effort to align new code and other find-and-replacable bits 1:1 with each other, its an acceptable trade off to ignore older patterns within a file.

src/schema/v2/artwork/index.ts
ashleyjelks: Good catch, not using the nullable coalescing operator...however I don't want to call `Number()` on null so I think a check here for a non-null value is needed first?

damassi (approved):
Couple of minor comments but otherwise LGTM 👍 merge when ready

src/schema/v2/artwork/index.ts
damassi: That makes sense 👍 could maybe still leverage this existential form and assign to a variable that you can do an `if (var)` on to wrap as a Number.

PR #2728, opened 2020-09-29
Add a placeholder response for inquiry questions to unblock ui work

zephraph:
This is a pretty critical piece of pre-work to quickly unblock UI work for the first inquiry project so that we can begin parallelizing asap.

This adds a `inquiryQuestions` field to the `Artwork` type. This field will be used to power a modal after the user taps contact gallery that will send structured questions to gravity (and ultimately to our partners) so that they can provide more consistent, faster responses (all while reducing the overhead for collectors in deciding how to ask about basic questions). 

We'll follow up on hooking this up to gravity soon.

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -1250,6 +1250,42 @@ describe("Artwork type", () => {
       })
     })
   })
+  describe("#inquiryQuestions", () => {
+    const query = `
+        {
+          artwork(id: "richard-prince-untitled-portrait") {
+            isInquireable
+            inquiryQuestions {
+              internalID
+              question
+            }
+          }
+        }
+      `
+    beforeEach(() => {
+      artwork.inquireable = true
+      artwork.ecommerce = false
+    })
+
+    it("returns available inquiry questions", () => {
+      return runQuery(query, context).then((data) => {
+        expect(data.artwork.inquiryQuestions).toEqual([
+          {
+            internalID: "price_and_availability",
+            question: "Price & Availability",
+          },
+          {
+            internalID: "shipping",
+            question: "Shipping",
+          },
+          {
+            internalID: "condition_and_provenance",
+            question: "Condition & Provenance",
+          },
+        ])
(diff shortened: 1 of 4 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -334,6 +335,30 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           return normalizeImageData(size ? _.take(sorted, size) : sorted)
         },
       },
+      inquiryQuestions: {
+        type: new GraphQLList(InquiryQuestionType),
```
xtina-starr, line 339: Do you think it's worth adding a comment that this is fixture/placeholder data for now? Outside of the context of this PR it's not super obvious.

zephraph: Will do

zephraph:
I'm adding a test in just a sec.

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -1250,6 +1250,42 @@ describe("Artwork type", () => {
       })
     })
   })
+  describe("#inquiryQuestions", () => {
```
xtina-starr, line 1253: 🎉 🎉 🎉

xtina-starr (approved):
This looks great!

src/schema/v2/artwork/index.ts
```diff
@@ -334,6 +335,36 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+            /**
+             * TODO: This is temporary, fixture data hardcoded to enable the
+             * first inquiry project to be broken down into parallel tasks. Will
+             * be followed w/ a PR to hook it up to a soon to be built gravity
+             * hook.
+             */
+            return [
+              {
+                id: "price_and_availability",
+                question: "Price & Availability",
```
starsirius, line 353: Nice! I can imagine we can extend it to include something like a type so that UI knows it's going to render a checkbox, etc.

starsirius (approved):

PR #2737, opened 2020-09-30
Fix the update mutation too

jonallured:
Related: https://github.com/artsy/metaphysics/pull/2734

I believe we've identified this payload computation is a place we could improve things, but I just wanted to fix it for now. Going to mark this MOG but happy to revisit if there are ideas about a better approach.

/cc @artsy/csgn-devs @joeyAghion

### DIFF
--- a/src/schema/v2/me/myCollectionUpdateArtworkMutation.ts
+++ b/src/schema/v2/me/myCollectionUpdateArtworkMutation.ts
@@ -79,7 +79,7 @@ export const myCollectionUpdateArtworkMutation = mutationWithClientMutationId<
 
     try {
       const response = await myCollectionUpdateArtworkLoader(artworkId, {
-        artist_ids: artistIds,
+        artists: artistIds,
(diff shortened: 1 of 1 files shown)


src/schema/v2/me/myCollectionUpdateArtworkMutation.ts
```diff
@@ -79,7 +79,7 @@ export const myCollectionUpdateArtworkMutation = mutationWithClientMutationId<
 
     try {
       const response = await myCollectionUpdateArtworkLoader(artworkId, {
-        artist_ids: artistIds,
+        artists: artistIds,
```
damassi, line 82: @jonallured - sorry for removing MOG, just want to discuss a bit. Can you explain this change a bit?

damassi: Discussion here: https://artsy.slack.com/archives/CS3H1K4BD/p1601490192118100

PR #2742, opened 2020-10-01
CSGN-356: Support creating images for My Collection artworks

jonallured:
This PR uses an update I made over on Gravity here:

https://github.com/artsy/gravity/pull/13512

That PR allows one to pass `source_bucket` and `source_key` when creating an `AdditionalArtwork` record. Those arguments will then be passed to Gemini which will fetch and process the corresponding images.

https://artsyproduct.atlassian.net/browse/CSGN-356

/cc @artsy/csgn-devs

### DIFF
--- a/src/lib/loaders/loaders_with_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_with_authentication/gravity.ts
@@ -183,6 +183,11 @@ export default (accessToken, userID, opts) => {
       },
       { method: "POST" }
     ),
+    myCollectionCreateImageLoader: gravityLoader(
+      (id) => `artwork/${id}/image`,
+      {},
+      { method: "POST" }
+    ),
     myCollectionUpdateArtworkLoader: gravityLoader(
       (id) => `my-collection/artworks/${id}`,
       {
--- a/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
+++ b/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
@@ -42,6 +42,9 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
     editionSize: {
       type: GraphQLString,
     },
+    externalImageUrls: {
+      type: new GraphQLList(GraphQLString),
+    },
     height: {
       type: GraphQLString,
     },
@@ -68,11 +71,12 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
       costMinor,
       editionSize,
       editionNumber,
+      externalImageUrls = [],
       ...rest
     },
-    { myCollectionCreateArtworkLoader }
+    { myCollectionCreateArtworkLoader, myCollectionCreateImageLoader }
   ) => {
-    if (!myCollectionCreateArtworkLoader) {
+    if (!myCollectionCreateArtworkLoader || !myCollectionCreateImageLoader) {
       return new Error("You need to be signed in to perform this action")
     }
 
@@ -84,6 +88,18 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
         edition_size: editionSize,
         edition_number: editionNumber,
         ...rest,
+      })
+
+      const artworkId = response.id
+      const regex = /https:\/\/(?<sourceBucket>.*).s3.amazonaws.com\/(?<sourceKey>.*)/
+
+      await externalImageUrls.forEach(async (url) => {
+        const { sourceBucket, sourceKey } = url.match(regex).groups
+
+        await myCollectionCreateImageLoader(artworkId, {
+          source_bucket: sourceBucket,
+          source_key: sourceKey,
+        })
       })
 
       return response
--- a/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
+++ b/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
@@ -93,14 +93,22 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
       const artworkId = response.id
       const regex = /https:\/\/(?<sourceBucket>.*).s3.amazonaws.com\/(?<sourceKey>.*)/
 
-      await externalImageUrls.forEach(async (url) => {
-        const { sourceBucket, sourceKey } = url.match(regex).groups
+      const createImagePromises = externalImageUrls
+        .map((url) => {
+          const match = url.match(regex)
 
-        await myCollectionCreateImageLoader(artworkId, {
-          source_bucket: sourceBucket,
-          source_key: sourceKey,
+          if (!match) return
+
+          const { sourceBucket, sourceKey } = url.match(regex).groups
+
+          return myCollectionCreateImageLoader(artworkId, {
+            source_bucket: sourceBucket,
+            source_key: sourceKey,
+          })
         })
-      })
+        .filter(Boolean)
+
+      await Promise.all(createImagePromises)
 
       return response
     } catch (error) {
--- a/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
+++ b/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
@@ -93,20 +93,24 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
       const artworkId = response.id
       const regex = /https:\/\/(?<sourceBucket>.*).s3.amazonaws.com\/(?<sourceKey>.*)/
 
-      const createImagePromises = externalImageUrls
+      const imageSources = externalImageUrls
         .map((url) => {
           const match = url.match(regex)
 
           if (!match) return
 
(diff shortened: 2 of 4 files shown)


src/schema/v2/me/myCollectionCreateArtworkMutation.ts
```diff
@@ -86,6 +90,19 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
         ...rest,
       })
 
+      const artworkId = response.id
+      const regex = /https:\/\/(?<sourceBucket>.*).s3.amazonaws.com\/(?<sourceKey>.*)/
+
+      externalImageUrls = externalImageUrls || []
```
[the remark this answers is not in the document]
pepopowitz: I think you can assign a default value in a destructuring assignment. So your destructured arguments to the function (lines 66-71) would look like 

```
{
  artistIds,
  costCurrencyCode,
...
  externalImageUrls = [],
  ...rest
}
```


I'd verify this works, but I think it should.

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
```diff
@@ -86,6 +90,19 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
...
+      await externalImageUrls.forEach(async (url) => {
```
pepopowitz, line 97: potentially blocking: 

Does this do what we think it does? Iterating over async calls is always challenging, and while this code might "succeed", I don't know if it's doing what we think it's doing. I think it's probably just spinning off the calls to the `myCollectionCreateImageLoader` and moving on, never waiting for any to succeed. 

This article describes the difficulty of iterating over async calls with `forEach`: https://medium.com/dailyjs/the-pitfalls-of-async-await-in-array-loops-cf9cf713bfeb

I usually approach this type of code with `promise.all` (also described in that article).

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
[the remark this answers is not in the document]
jonallured: Good call, thanks!! 🙌

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -183,6 +183,11 @@ export default (accessToken, userID, opts) => {
       },
       { method: "POST" }
     ),
+    myCollectionCreateImageLoader: gravityLoader(
+      (id) => `artwork/${id}/image`,
+      {},
```
damassi, line 188: (nonblocking) Just noting that all of the other MC endpoints require this auth object here:
```
 user_id: userID,
private: true,
```

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
damassi: Promise.all will perform all of the requests in parallel too! We should update.

damassi: What happens when an error occurs in the two awaits?

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
```diff
@@ -86,6 +90,18 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
...
+      const regex = /https:\/\/(?<sourceBucket>.*).s3.amazonaws.com\/(?<sourceKey>.*)/
+
+      await externalImageUrls.forEach(async (url) => {
+        const { sourceBucket, sourceKey } = url.match(regex).groups
+
+        await myCollectionCreateImageLoader(artworkId, {
+          source_bucket: sourceBucket,
+          source_key: sourceKey,
+        })
+      })
```
damassi, line 103: Can we add some tests around this image behavior?

damassi (commented):
Can we add some tests around this image behavior? It should be fairly trivial to mock / assert since we're handling everything through loaders.

jonallured:
Oh man, I'm really glad you suggested tests @damassi, I had all kinds of trouble in here, haha! So with my updates, I am much more defensive about these external urls and am also properly awaiting for all the promises to resolve.

I spent a bunch of time in the tests tinkering with extracting shared setup. What I was trying to do there was reduce the noise and boost the signal of these tests and that was a bit time consuming because I'm not as familiar with these testing tools. But I think it turned out pretty well! Any further thoughts?

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
jonallured: This is now addressed in the tests - the whole thing blows up even though some images were created. 😬 Open to other ideas, but hoping I can convince yall to merge this and then address edge cases in future iterations? 🤞

src/lib/loaders/loaders_with_authentication/gravity.ts
jonallured: Unless I'm missing something, I believe these additional params are not required for this particular endpoint.

damassi (approved):
This looks solid in terms of testing @jonallured 👍 The following feedback is all nonblocking! 

One bit of meta, in-general feedback is that this file breaks some of the our MP testing conventions by being a bit too DRY and therefore somewhat more difficult to reason about. 

Looking at something like the [microfunnel.tests.ts](https://github.com/artsy/metaphysics/blob/master/src/schema/v2/artist/targetSupply/__tests__/index.test.ts) file one can find something fairly representative to the app as a whole. For each test, we duplicate the query and things are roughly linear. Mocked loaders are defined on a test by test basis, or in the immediate outer context. Adding a new test is as simple as copying the previous one and making adjustments, and there's no need to hunt around and piece together a bunch of variables. 

I think moving the variables from outside of the top-level `describe` block into more local sections of the test -- with these variables testing things specific to things in and around that particular test) would improve code long-term as we move off of this project and on to another team. But this feedback is all non-blocking! Only an effort to harmonize with some of the more dumb/linear testing patterns seen throughout the app. 

On the whole though this is well done and totally mergable 👍

jonallured:
This is def something I've struggled with throughout my career - the tension between making tests as dumb as possible and eliminating noise in tests. I think at this point my theory is that boosting the signal of tests is more important than making them dumb. These tests around GQL requests and responses are particularly challenging just because the inputs and outputs are so verbose.

Thanks for linking to that microfunnel testing file - that was helpful for me to see and is a good example of what I would consider too much noise. I read that test file over and over trying to figure out the different between two tests - it turns out the difference was the id of the artist!

https://github.com/artsy/metaphysics/blob/5e3587518ea274ef6eed5e046d2eeb0f8d480f31/src/schema/v2/artist/targetSupply/__tests__/index.test.ts#L27-L69

Compare that with a refactor like this:

```ts
  describe("isInMicrofunnel", () => {
    it("returns false if artist not in microfunnel", async () => {
      const id = "andy-warhol"

      const query = computeQuery(id)
      const mockLoader = jest.fn().resolveWithValue({ id })
      const context = { artistLoader: mockLoader }
      const response = await runQuery(query, context)

      expect(response.artist.targetSupply.isInMicrofunnel).toEqual(false)
    })

    it("returns true if artist in microfunnel", async () => {
      const id = "alex-katz"

      const query = computeQuery(id)
      const mockLoader = jest.fn().resolveWithValue({ id })
      const context = { artistLoader: mockLoader }
      const response = await runQuery(query, context)

      expect(response.artist.targetSupply.isInMicrofunnel).toEqual(true)
    })
  })
```

I think it's easier to see that the ids of the artist are different between these two tests and that is what we're trying to convey to Future Us - the id of the artist matters. It's still unclear to me WHY this matters or where in the stack we're making this choice, but at least now I have something to start investigating.

So that's my pitch - being able to compare two tests and see the signal of what's different is worth a bit of abstraction. There are no right answers, but this is something I think about a lot, haha.

pepopowitz:
This is (another) example of readability & maintainability being subjective. I side with Jon's approach of boosting what makes the test unique (to the point that [I've written about it before](https://www.stevenhicks.me/blog/2018/01/chekhovs-gun-and-better-unit-tests/)), but I've come to accept that not everyone wants to read code/tests the way I want to read them. 

For me the answer to differences in preference is "accept that the code won't always read like you want it to read"....but I also know that I'm comfortable with a higher level of inconsistency than a lot of people, and that is a definite side-effect of accepting that code won't always read like you want it to read. 🤷 🤷 🤷

damassi:
Yah its more a matter of harmonizing with the rest of the codebase, but that of course shouldn't preempt evolution if necessary. But either way this test file is now a clear outlier for future devs to have to grok. 

The more subjective point is that less abstraction in tests is actually more, but that of course can be argued, as Jon has done above.

damassi:
That blog post is 💯 @pepopowitz -- can't go wrong with Chekov 

Screen Shot 2020-10-03 at 12 07 02 AM

PR #2761, opened 2020-10-08
Update exchange schema

sepans:
Change `myOrder` `state` param to `states`

This query is only being used on production behind a flag so safe to synchronously update it.


starsirius (approved):
👍

PR #2764, opened 2020-10-08
Add more images to my collection artworks

jonallured:
This PR follows on #2742 and adds support for adding additional images when editing a My Collection artwork.

After some chatter here:

https://artsy.slack.com/archives/C01B2P6LJUU/p1602188267269900

I set out to add the `externalImageUrls` prop to the `myCollectionUpdateArtworkMutation` and did extract a func to compute the image sources so that the business logic around matching and parsing these URLs could be encapsulated. Then I updated both mutations to use the func.

Then I updated the tests like so:

* leave some of the tests right where they were for the create mutation - these serve as integration tests
* extract some of the existing tests to focus on the newly extracted helper func
* copy/paste those integration tests into the update mutation test file - more integration coverage

So now the business logic of this helper func can evolve but the outer context doesn't have to care and is just focused on how mutations behave in certain situations. 😎


damassi (approved):
Looks great @jonallured 👍

PR #2765, opened 2020-10-08
Create shared env

emmadickson:
Supports https://artsyproduct.atlassian.net/browse/PLATFORM-2640


.gitignore
```diff
@@ -1,5 +1,5 @@
 .env
-.env.staging
+.env.*
```
joeyAghion, line 2: I believe this will cause `.env.example` to be ignored.

.nvmrc
```diff
@@ -1 +1 @@
-v12
+12.4
```
joeyAghion, line 1: This doesn't match the Dockerfile: https://github.com/artsy/metaphysics/blob/7cc291d16db51c7cc03c12b35840eb5f0dbe3010/Dockerfile#L1

scripts/setup.sh
```diff
@@ -0,0 +1,28 @@
...
+  source ~/.nvm/nvm.sh
+  nvm install
+fi
+
+echo 'Install yarn and memchached'
+brew bundle
+
+echo 'Using yarn to install required node packages'
+yarn install
+yarn add dotenv-cli
```
joeyAghion, line 19: I think this command will _update_ the `package.json` file and result in changes that must be committed. If it's necessary, I believe the right thing to do is to add it to the project's `devDependencies` so the `yarn install` command above will take care of installing it along with other packages.

scripts/setup.sh
```diff
@@ -0,0 +1,28 @@
...
+
+echo 'Copy the example env to the end file'
+cp .env.example .env
+
+echo 'Aws fetch the shared config file'
+aws s3 cp s3://artsy-citadel/dev/.env.metaphysics .env.shared
+
+
+echo 'Now run: dotenv -e .env -e .env_shared yarn dev'
```
joeyAghion, line 28: Ideally the `yarn dev` command can just take care of this. Otherwise there's so much indirection (from `dotenv` to `yarn dev` to `nf` to `Procfile.dev`).

_schemaV2.graphql
```diff
@@ -7487,6 +7487,7 @@ input MyCollectionUpdateArtworkInput {
   depth: String
   editionNumber: String
   editionSize: String
+  externalImageUrls: [String]
```
joeyAghion, line 7490: Intentional?

package.json
```diff
@@ -64,6 +64,7 @@
     "dataloader": "1.3.0",
     "dd-trace": "0.21.0",
     "debug": "2.6.9",
+    "dotenv-cli": "^4.0.0",
```
joeyAghion, line 67: Shouldn't be necessary as a non-dev dependency.

scripts/setup.sh
```diff
@@ -0,0 +1,27 @@
...
+yarn install
+
+echo 'Copy the example env to the end file'
+cp .env.example .env
+
+echo 'Aws fetch the shared config file'
+aws s3 cp s3://artsy-citadel/dev/.env.metaphysics .env.shared
+
+
+echo 'Now run: dotenv -e .env_shared yarn dev'
```
joeyAghion, line 27: Does this respect `.env` by default, or do both have to be specified?

joeyAghion:
(Not sure what's up with the build failure. Maybe you need to `rebase` on `master`?)

Please also update the README with the new recommended setup steps.

_schemaV2.graphql
emmadickson: no i removed this

scripts/setup.sh
emmadickson: I've replaced this with a more manual process since dotenv was being finicky

joeyAghion:
I pushed another commit that fixes `hokusai dev` mode by specifying `PORT=3000` in the `yarn dev` command. This makes local development more closely match how the container runs in other environments, and is the same whether running locally or in docker.

**Developers: note that the local app will now listen on `3000` instead of `5000`.**

The README should be updated to reflect this^ along with more automated set-up.

README.md
```diff
@@ -49,12 +49,9 @@ To get yourself set up with all the project's dependencies:
...
-# Installed system tools (like cache servers)
-brew bundle
-# Get set up with a default env
-cp .env.oss .env
+# Run the setup script
+
+source ./scripts/setup.sh
 ```
 
 For Artsy staff wanting to expand on your `.env`, you can use `hokusai staging env get` to see staging's env vars. Or better yet, use [copy_env][] to match
```
joeyAghion, line 57: I think we can drop this section, since it overlaps/conflicts with what's done by the setup script.

README.md
```diff
@@ -77,7 +74,28 @@ With your dependencies set up, you can run Metaphysics by running:
 yarn dev
 ```
 
-Which will start the server on http://localhost:5001
+This will pull the needed environment variables into .env.shared. If you need to override any of these values
```
joeyAghion, line 77: I think this is fine to explain, but this is a confusing location because the command above (`yarn dev`) doesn't actually do what this describes.

README.md
```diff
@@ -77,7 +74,28 @@ With your dependencies set up, you can run Metaphysics by running:
...
+
+Be sure that memcached is no longer running before starting hokusai by running
+
+```brew services stop memcached
+
+```
+
+To test run
+
+```hokusai test
```
joeyAghion, line 96: There's a whole `Testing` section below in this README so I don't think this is necessary.

PR #2771, opened 2020-10-13
[FX-2357] Supports a default srcSet on resized and cropped types

dzucconi:
Closes: [FX-2357](https://artsyproduct.atlassian.net/browse/FX-2357)

This, instead of having to alias multiple versions in queries and do the math in our heads. The `srcSet` output can/should be tweaked probably. We probably want to adjust the quality and possibly add a `3x`.


src/schema/v2/image/resized.ts
```diff
@@ -8,46 +8,92 @@ import {
...
-    factor,
-    width,
-    height,
-    url,
+    factor: scaled.scale,
+    width: scaled.width || null,
+    height: scaled.height || null,
+    url: url1x,
+    src: url1x,
+    srcSet: `${url1x} 1x, ${url2x} 2x`,
```
dzucconi, line 85: The only thing I might suggest is that — this can be done lazily with individual resolvers for each field. I'm assuming these functions and string interpolation are trivial here, so I just compute the full object upfront. Either way works for me.

src/schema/v2/image/cropped.ts
```diff
@@ -10,39 +9,69 @@ import {
...
+>({
   name: "CroppedImageUrl",
   fields: {
     width: {
-      type: GraphQLInt,
+      type: new GraphQLNonNull(GraphQLInt),
     },
     height: {
-      type: GraphQLInt,
+      type: new GraphQLNonNull(GraphQLInt),
```
dzucconi, line 56: Tried to tighten up some of these returns that we know are non-nullable

src/schema/v2/image/__tests__/resized.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        factor: 1,
+        width: null,
+        height: null,
+        url: url,
+        src: url,
+        srcSet: `${url} 1x`,
+      })
+    })
+  })
+})
```
dzucconi, line 124: These are the same specs that existed but I padded them out with a few more examples and the new srcSet stuff. [I know the scale function itself to be correct.](https://github.com/dzucconi/proportional-scale/blob/master/src/scale.test.ts)

src/schema/v2/image/resized.ts
damassi: Don't think there'd be a creeping performance need for making this lazy, right? (not sure what `proxy` does under the hood tho)

dzucconi: Nah, no need that I can see just... trying to be aware? It's all just string interpolation.

PR #2782, opened 2020-10-20
fix: followed fairs galleries schema errors

admbtlr:
A couple of errors relating to queries for retrieving followed fairs and galleries. Not sure how no-one ever noticed this with the galleries query before, since it means that fairs will also be returned when requesting followed galleries.


MounirDhahri:
Nice catch!

MounirDhahri (approved):

MounirDhahri:
I added @sweir27 to check this out as well since she also worked on fairs and galleries and they might be using it as you described  👀

PR #2783, opened 2020-10-20
Add `active` arg to Me.bidders

erikdstock:
Not blocked by, but depends on https://github.com/artsy/gravity/pull/13571 to work.
Completes PURCHASE-2194

This PR adds an optional `active: Boolean` arg to the Me.bidders field, allowing a user to query her bidders limiting them to only active (published, unended) sales. Omitting it or setting `active: false` would match today's behavior and `active: true` opts into the filter.

### DIFF
--- a/src/schema/v2/me/__tests__/bidders.test.js
+++ b/src/schema/v2/me/__tests__/bidders.test.js
@@ -52,5 +52,32 @@ describe("Me", () => {
         }
       )
     })
+
+    it("calls the gravity endpoint with the `filter` param if included", () => {
+      const query = `
+        {
+          me {
+            bidders(active: true) {
+              internalID
+            }
+          }
+        }
+      `
+      const response = () =>
+        Promise.resolve([{ id: "Foo ID" }, { id: "Bar ID" }])
+      const meBiddersLoader = jest.fn(response)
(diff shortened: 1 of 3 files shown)


src/schema/v2/me/__tests__/bidders.test.js
```diff
@@ -52,5 +52,32 @@ describe("Me", () => {
...
+        {
+          me {
+            bidders(active: true) {
+              internalID
+            }
+          }
+        }
+      `
+      const response = () =>
+        Promise.resolve([{ id: "Foo ID" }, { id: "Bar ID" }])
```
zephraph, line 67: I don't inherently know the shape of these objects, but would it make sense to mock one of these to be representative of a non-active state to ensure it's filtered out as expected?

zephraph (approved):
Pretty straight forward, looks good to me. One question on the test, but otherwise 👍

src/schema/v2/me/__tests__/bidders.test.js
erikdstock: The filtering happens at the gravity layer so all we need to test here is that the argument is properly forwarded.

PR #2785, opened 2020-10-20
feat: Add `SubmitInquiryRequestMutation` for artwork inquiry

sepans:
[PURCHASE-2133]

Sample query/response

```graphql
mutation sendinquiry {
  submitInquiryRequestMutation(input: {
    inquireableID: "yaacov-agam-color-pyramids",
    inquireableType: "Artwork"
    questions: [
      {questionID: "price_and_availability"},
      {questionID:"shipping_quote", details: "{ \"address\": \"112 Main st Brooklyn NY 11211\"}"}]
  }) {
    
    inquiryRequest {
      inquireable {
        __typename
        ... on Artwork {
          title
          artists {
            name
            birthday
          }
          partner {
            name
          }
        }
      }
      inquirer {
        name
        email
      }
      shippingLocation {
        address
        city
        country
      }
      inquiryQuestions
    }
  }
}
```

Response: 
```json
{
  "data": {
    "submitInquiryRequestMutation": {
      "inquiryRequest": {
        "inquireable": {
          "__typename": "Artwork",
          "title": "Color Pyramids",
          "artists": [
            {
              "name": "Yaacov Agam",
              "birthday": "1928"
            }
          ],
          "partner": {
            "name": "Ai Bo Gallery"
          }
        },
        "inquirer": {
          "name": "Sepand",
          "email": "sepand.ansari+2@artsymail.com"
        },
        "shippingLocation": {
          "address": "112 Main st Brooklyn NY 11211",
          "city": null,
          "country": ""
        },
        "inquiryQuestions": [
          "price_and_availability",
          "shipping_quote"
        ]
      }
    }
  }
}
```

[PURCHASE-2133]: https://artsyproduct.atlassian.net/browse/PURCHASE-2133

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -6472,10 +6472,10 @@ type InquiryRequest {
   inquireable: InquiryItemType
   inquirer: User
   inquiryQuestions: [String]
-  inquiryShippingLocation: Location
 
   # A type-specific ID likely used as a database ID.
   internalID: ID!
+  shippingLocation: Location
 }
 
 enum Intents {
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -6449,6 +6449,8 @@ type ImageURLs {
   normalized: String
 }
 
+union InquiryItemType = Artwork | Show
+
 type InquiryQuestion {
   # A globally unique ID.
   id: ID!
@@ -6456,6 +6458,24 @@ type InquiryQuestion {
   # A type-specific ID.
   internalID: ID!
   question: String!
+}
+
+input InquiryQuestionInput {
+  details: String
+  questionID: String!
+}
+
+# A request to inquire on an artwork
+type InquiryRequest {
+  # A globally unique ID.
+  id: ID!
+  inquireable: InquiryItemType
+  inquirer: User
+  inquiryQuestions: [String]
+  inquiryShippingLocation: Location
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
 }
 
 enum Intents {
@@ -7363,6 +7383,11 @@ type Mutation {
   startIdentityVerification(
     input: startIdentityVerificationMutationInput!
   ): startIdentityVerificationMutationPayload
+
+  # Create an artwork inquiry request
+  submitInquiryRequestMutation(
+    input: SubmitInquiryRequestMutationInput!
+  ): SubmitInquiryRequestMutationPayload
   unpublishViewingRoom(
     input: UnpublishViewingRoomInput!
   ): UnpublishViewingRoomPayload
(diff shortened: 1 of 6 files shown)


src/schema/v2/me/conversation/submit_inquiry_request_mutation.ts
```diff
@@ -0,0 +1,136 @@
...
+      },
+    },
+    inquirer: {
+      type: UserType,
+      resolve: ({ inquirer }, _, { userByIDLoader }) => {
+        const { id } = inquirer
+        return userByIDLoader(id)
+      },
+    },
+    inquiryShippingLocation: {
```
zephraph, line 66: small nit: given that this appears in the context of an inquiry request, I think we can omit the `inquiry` prefixes as that's somewhat implied. This could just be `shippingLocation`.

xtina-starr (approved):
This looks sooooo good!!!

src/schema/v2/me/conversation/submit_inquiry_request_mutation.ts
```diff
@@ -0,0 +1,136 @@
...
+  ) => {
+    if (inquireableType === "Artwork") {
+      if (!submitArtworkInquiryRequestLoader) return null
+      const locationQuestion = questions?.find(
+        (question) => question.questionID === "shipping_quote"
+      )
+      let inquiryShippingLocation
+      if (locationQuestion) {
+        try {
+          inquiryShippingLocation = JSON.parse(locationQuestion.details)
```
xtina-starr, line 121: Silly questions: this should support inquiry shipping location objects that look like this right?

```js
inquiry_shipping_location: {
    city: "Greenville",
    coordinates: [42.4153596, -74.0220769],
    country: "United States",
    postal_code: "12083",
    state: "New York",
    state_code: "NY"
}
```

sepans: I believe so. The MP part just parses the stringified object coming from the client and passes it as a Hash to gravity. So as long as gravity accepts ☝️ format it should be fine. But if any data massaging is needed to be done in MP to convert client data structure to what gravity expects, we can follow up and add it here.

_schemaV2.graphql
```diff
@@ -9884,6 +9909,29 @@ type SubmissionEdge {
...
+  inquireableID: String!
+
+  # The type of inquireable object (Artwork or Show)
+  inquireableType: String!
+
+  # Optional inquiry message
+  message: String
+
+  # List of structured inquiry questions
+  questions: [InquiryQuestionInput]
```
zephraph, line 9925: Something we should follow up w/ on both the mutation and the query... we should make these questions non-nullable. What I mean by that is to have the schema represented by `questions: [InquiryQuestionInput]!`. That generally just means we'll _always_ at least return an empty array. If we do that it'll mean less type faffing on the frontend. 

That should be a follow up though, not necessary in this pr.

zephraph (approved):
This is great @sepans, thanks for picking it up!

_schemaV2.graphql
sepans: 👍 Makes sense. In the back of my mind, I thought this mutation needs to be backward compatible and support the cases where only `message` is present to support force (and I even added tests for that 😄)  but that's not the case.

PR #2787, opened 2020-10-20
feat: add vortex schema and regenerate stitched schema

ansor4:
Bump MP schema for new field in vortex. It looks like it's been awhile since vortex's schema has been updated, but all of the changes are additions of new fields, deprecation reasons, or comments.

Relevant code for lastUpdatedAt all the way at the bottom!

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -195,6 +195,8 @@ type AnalyticsPartnerInquiryCountTimeSeriesStats {
 # Inquiry stats of a partner
 type AnalyticsPartnerInquiryStats {
   inquiryCount: Int!
+
+  # Inquiry response time in seconds
   inquiryResponseTime: Int
   partnerId: String!
   period: AnalyticsQueryPeriodEnum!
@@ -208,6 +210,8 @@ type AnalyticsPartnerInquiryStats {
 # Sales stats of a partner
 type AnalyticsPartnerSalesStats {
   orderCount: Int!
+
+  # Order response time in seconds
   orderResponseTime: Int
   partnerId: String!
   period: AnalyticsQueryPeriodEnum!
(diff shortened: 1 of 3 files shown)


_schemaV2.graphql
```diff
@@ -8116,6 +8176,9 @@ type Query {
     statuses: [ViewingRoomStatusEnum!] = [live]
   ): ViewingRoomsConnection
 
+  # Last updated timestamp
+  analyticsLastUpdatedAt: AnalyticsDateTime
```
jpotts244, line 8180: Oh yeah

_schema.graphql
```diff
@@ -317,6 +333,50 @@ type AnalyticsPartnerStats {
...
+  # Pageviews time series
+  timeSeries(
+    cumulative: Boolean = false
+  ): [AnalyticsPartnerStatsPageviewsTimeSeries!]
+  totalCount: Int!
+  uniqueVisitors: Int!
+}
+
+# Pageviews time series data of a partner
+type AnalyticsPartnerStatsPageviewsTimeSeries {
```
jpotts244, line 374: Does this mean these fields haven't been accessible to any client yet?

I wonder if these additions/depreciations going unnoticed means Volt is still expecting to use the deprecated fields? Like `PageviewStats` for example.

ansor4: yeah i'm not sure what's going on here actually, it could be a change in the stitching code. Would be good to confirm.

ansor4: On second read and a bit more thought, I think it's ok for Volt to continue using these for now. This could be another #future-friday initiative, or we could even fold it in with converting Volt's use of artemis to relay.

jpotts244: Ah didnt see your second comment there ^^
cool! ok so they dont block requests/raise errors?

jpotts244 (approved):
Just wanted to make sure the schema changes we non-breaking, will merge after Matt's volt deploy!

_schema.graphql
ansor4: yeah exactly adding a `deprecation_reason` is graphql-ruby's way of marking a field deprecated, which only affects tools and DX, like introspection queries _(where it'll be grouped under "deprecated fields" or something)._ But deprecated fields are still able to be queried.

Link to the deprecation section of the latest version of the spec: https://spec.graphql.org/June2018/#sec-Field-Deprecation

jpotts244: Thanks for the spec!

PR #2788, opened 2020-10-20
feat: Add me.labFeatures

bhoggard:
https://artsyproduct.atlassian.net/browse/CX-714

@ansor4, wanted to make sure this approach makes sense to you. We would like to start using an array, so that we don't have to add a new field to `me` every time we add a new lab feature.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -6986,6 +6986,9 @@ type Me implements Node {
       reason: "Payment Request was deprecated. The field was kept for legacy client support. [Will be removed in v2]"
     )
 
+  # List of lab features for this user
+  labFeatures: [String!]!
+
   # Sale Artworks search results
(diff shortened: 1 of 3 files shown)


ds300:
this seems like it's compatible with the graph-based feature toggles @ansor4 was talking about. We would just need to flatten the graph into a set and put it in this array without affecting consumers. But maybe there's a more forward-thinking api we could adopt already, so I'll hold off until @ansor4 has a chance to comment.

ds300 (approved):
🙌

MounirDhahri (approved):

MounirDhahri:
🎉

ansor4 (approved):
This makes sense to me! Agree with @ds300 that we would want to flatten the underlying feature graph into an array, clients shouldn't need to know how features are structured.

PR #2795, opened 2020-10-23
fix: disable inquiryQuestions for non inquirable artworks

MounirDhahri:
The type of this PR is: **Chore**

This PR resolves [CX-704]

### Description

- Check if an artwork (sale artwork) is inquirable before calling the gravity loader.

**PS: the artwork used in the picture below is from an auction**
Group 2 (2)

Screenshot 2020-10-26 at 11 28 31

[CX-704]: https://artsyproduct.atlassian.net/browse/CX-704

Issue: https://artsyproduct.atlassian.net/browse/CX-746

Co-authored-by: David Sheldrick <djsheldrick@gmail.com>

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -340,12 +340,12 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         description:
           "Structured questions a collector can inquire on about this work",
         resolve: (
-          { inquireable, id },
+          { sale_ids, id },
           _params,
           { inquiryRequestQuestionsLoader }
         ) => {
           // Sale artworks are not inquirable
-          if (!inquireable) {
+          if (!sale_ids.length) {
             return inquiryRequestQuestionsLoader({
               inquireable_id: id,
               inquireable_type: "Artwork",
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -339,11 +339,19 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         type: new GraphQLList(InquiryQuestionType),
         description:
           "Structured questions a collector can inquire on about this work",
-        resolve: ({ id }, _params, { inquiryRequestQuestionsLoader }) => {
-          return inquiryRequestQuestionsLoader({
-            inquireable_id: id,
-            inquireable_type: "Artwork",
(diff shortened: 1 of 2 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -339,11 +339,19 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
-            inquireable_id: id,
-            inquireable_type: "Artwork",
-          })
+        resolve: (
+          { inquireable, id },
+          _params,
+          { inquiryRequestQuestionsLoader }
+        ) => {
+          // Sale artworks are not inquirable
+          if (!inquireable) {
```
ds300, line 348: shouldn't this be the inverse? i.e. if the artwork is inquireable then we should fetch the inquiry questions?

ds300 (commented):
I think you have the logic the wrong way round? but otherwise this change looks good to me.

@zephraph this change was necessary (rather than a mere perf optimisation) because the gravity endpoint for inquiry questions will return the empty string rather than the empty json array for sale artworks, which caused a json parse error in metaphysics. Would it be worth also fixing that?

src/schema/v2/artwork/index.ts
MounirDhahri: I took a look at this again and tested with an `artwork` and a `saleArtwork` and they were both returning `inquirable` as `true`. It looks like the only way to differentiate between the two is to check for the `sale_ids` array

MounirDhahri:
PS: I updated the screenshots to include an artwork with inquiryQuestions query as well

src/schema/v2/artwork/index.ts
```diff
@@ -340,12 +340,12 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
           "Structured questions a collector can inquire on about this work",
         resolve: (
-          { inquireable, id },
+          { sale_ids, id },
           _params,
           { inquiryRequestQuestionsLoader }
         ) => {
           // Sale artworks are not inquirable
-          if (!inquireable) {
+          if (!sale_ids.length) {
```
ds300: i dunno, this seems a bit indirect now. Was there something wrong with using `inquireable`?

ds300:
gonna merge this since it's blocking our QA today, we can follow up with the gravity change if needed.

PR #2805, opened 2020-10-28
feat: stitch in money-typed price fields in AuctionsLotState type (PURCHASE-2171)

starsirius:
https://artsyproduct.atlassian.net/browse/PURCHASE-2171

This stitches in `floorSellingPrice` and `onlineAskingPrice` fields in `AcutionsLotState` type, and uses the recommended Metaphysics `Money` type. The field names were removed in https://github.com/artsy/causality/pull/852 so we can reuse them here.

Migration steps:

- [x] (artsy/causality#848) Add the new raw cents fields (e.g. `floorSellingPriceCents`) and deprecate the Causality `Money` type fields (e.g. `floorSellingPrice`).
- [x] (https://github.com/artsy/ohm/pull/707) Migrate Ohm to use the new cents fields.
- [x] (https://github.com/artsy/causality/pull/852) Remove the Causality `Money` type fields (e.g. `floorSellingPrice`).
- [x] (This PR) Update MP to stitch in MP `Money` type fields.
- [ ] Update Eigen to use the new fields.

Screen Shot 2020-11-30 at 10 06 18 PM

### DIFF
--- a/src/integration/__tests__/__fixtures__/knownV2Failures.json
+++ b/src/integration/__tests__/__fixtures__/knownV2Failures.json
@@ -948,5 +948,12 @@
   "ebb240b91abb54dde55e46da86a0a885": null,
   "acf634b76c571c63c354f9203e7757f1": null,
   "e435b0066fae9cdb0cddd2146c1e383f": null,
-  "810fe4219c35f8c1198fcbd6bdbed29e": null
+  "810fe4219c35f8c1198fcbd6bdbed29e": null,
+  "042c99463b706a2a8a35d4b0a5337c71": null,
+  "7114b24c69da6fdf8a952cc998dded17": null,
+  "5feb6d845487b8b8d696e0cce082eb2b": null,
+  "491575e8d9b9f682de1cb678fd351c4b": null,
+  "e089beca2667c383e7e2c99c42e8750f": null,
+  "bd37ef982226143b0d68d972b5fee276": null,
+  "5d6a47112815cab926fb8f2036b94039": null
 }
--- a/src/lib/stitching/causality/stitching.ts
+++ b/src/lib/stitching/causality/stitching.ts
@@ -1,5 +1,25 @@
 import gql from "lib/gql"
+import { formatMoney } from "accounting"
 import { GraphQLSchema } from "graphql"
+import {
+  moneyMajorResolver,
+  symbolFromCurrencyCode,
+} from "schema/v2/fields/money"
+
+const resolveLotCentsFieldToMoney = (centsField) => {
+  return async (parent, _args, context, _info) => {
+    const { internalID, [centsField]: cents } = parent
+    const { currency } = await context.saleArtworkRootLoader(internalID)
+    const major = await moneyMajorResolver({ cents, currency }, {}, context)
+
+    return {
+      major,
+      minor: cents,
+      currencyCode: currency,
+      display: formatMoney(major, symbolFromCurrencyCode(currency)),
+    }
+  }
+}
 
 export const causalityStitchingEnvironment = ({
   causalitySchema,
@@ -21,6 +41,11 @@ export const causalityStitchingEnvironment = ({
 
       extend type AuctionsLotStanding {
         saleArtwork: SaleArtwork
+      }
+
+      extend type AuctionsLotState {
+        floorSellingPrice: Money
+        onlineAskingPrice: Money
       }
     `,
 
@@ -44,6 +69,26 @@ export const causalityStitchingEnvironment = ({
               info,
             })
           },
+        },
+      },
+      AuctionsLotState: {
+        floorSellingPrice: {
+          fragment: gql`
+            ... on AuctionsLotState {
+              internalID
+              floorSellingPriceCents
+            }
+          `,
+          resolve: resolveLotCentsFieldToMoney("floorSellingPriceCents"),
+        },
+        onlineAskingPrice: {
+          fragment: gql`
+            ... on AuctionsLotState {
+              internalID
+              onlineAskingPriceCents
+            }
+          `,
+          resolve: resolveLotCentsFieldToMoney("onlineAskingPriceCents"),
         },
       },
       Me: {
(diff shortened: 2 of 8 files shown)


src/lib/stitching/causality/stitching.ts
```diff
@@ -22,9 +28,57 @@ export const causalityStitchingEnvironment = ({
...
+            const { currency } = await context.saleArtworkRootLoader(internalID)
+            const { units: minor } = floorSellingPrice
+            const major = await moneyMajorResolver(
+              { cents: minor, currency },
+              {},
+              context,
+              info
+            )
+
+            // return { cents: minor, currency }
```
starsirius, line 62: I was hoping there is an easy way to keep resolving this node as a `Money` type.

src/lib/stitching/causality/stitching.ts
```diff
@@ -22,9 +28,57 @@ export const causalityStitchingEnvironment = ({
...
+
+            return {
+              currencyCode: currency,
+              minor,
+              major,
+              display: formatMoney(major, symbolFromCurrencyCode(currency)),
+            }
```
starsirius, line 69: One option is to resolve all fields manually. With some [extraction](https://github.com/artsy/metaphysics/pull/2805/files#diff-c7c0cff0902f5e31e018c4c77fe43c6c9647323cb8621d5e96ec57222491ab8cR168), this doesn't feel _too_ bad, but still pretty manual.

src/lib/stitching/causality/stitching.ts
```diff
@@ -22,9 +28,57 @@ export const causalityStitchingEnvironment = ({
...
+
+            // return info.mergeInfo.delegateToSchema({
+            //   schema: localSchema,
+            //   operation: "query",
+            //   fieldName: "_do_not_use_money",
```
starsirius, line 74: Another option I noticed currently in MP is to add a new root field, purely for internal stitching use, so that we can use `delegateToSchema`. An example of this pattern is the [artist series image](https://github.com/artsy/metaphysics/blob/0b4ebf1dd09e6a515f8294dd0e480c2465e65a4a/src/lib/stitching/gravity/stitching.ts#L271). But I'm not sure if we should keep adding root fields for this purpose.

src/lib/stitching/causality/stitching.ts
starsirius: Paging @alloy and curious if you have insights or recommendations.

erikdstock:
My first choice here would be to remove/rename causality's money fields completely and, when accessing anything like this (related to currency) via metaphysics, to have the same types and behavior as other metaphysics money types. Though maybe this is isn't good stitching practice.
I don't know if we'll still want to keep the Money type around for client applications that still use the Money type talking directly to causality's graphql api but one target might be to work towards deprecating it completely... 🤔

starsirius:
Apologies for the slowness here.

Yes, it makes sense to me that Causality provides the raw data and MP handles the formatting/display (and we have a centralized interface). Clients that talk to Causality directly then need to handle it themselves (or switch to use MP in the future). I propose that we

Expose a `floorSellingPriceCents` raw field from Causality.
Remove `floorSellingPrice` field from Causality, and instead, stitch that in MP with MP `Money` type.

Looks like currently only Ohm and Eigen My Bids (unreleased) are using it (and `onlineAskingPrice`), and I'll migrate them in a few steps.

src/integration/__tests__/__fixtures__/knownV2Failures.json
```diff
@@ -948,5 +948,12 @@
...
   "e435b0066fae9cdb0cddd2146c1e383f": null,
-  "810fe4219c35f8c1198fcbd6bdbed29e": null
+  "810fe4219c35f8c1198fcbd6bdbed29e": null,
+  "042c99463b706a2a8a35d4b0a5337c71": null,
+  "7114b24c69da6fdf8a952cc998dded17": null,
+  "5feb6d845487b8b8d696e0cce082eb2b": null,
+  "491575e8d9b9f682de1cb678fd351c4b": null,
+  "e089beca2667c383e7e2c99c42e8750f": null,
+  "bd37ef982226143b0d68d972b5fee276": null,
+  "5d6a47112815cab926fb8f2036b94039": null
```
starsirius, line 958: Some persisted My Bids queries became invalid because we migrated the `floorSellingPirce` and `onlineAskingPrice` fields to MP's `Money` type. The feature hasn't been released yet so _I think_ it's safe to ignore here. @zephraph would like to see if you think it's ok.

They all failed for the same reason:

Screen Shot 2020-11-30 at 10 29 10 PM

src/lib/stitching/causality/stitching.ts
```diff
@@ -1,5 +1,25 @@
 import gql from "lib/gql"
+import { formatMoney } from "accounting"
 import { GraphQLSchema } from "graphql"
+import {
+  moneyMajorResolver,
+  symbolFromCurrencyCode,
+} from "schema/v2/fields/money"
+
+const resolveLotCentsFieldToMoney = (centsField) => {
```
starsirius, line 9: Extract the resolver to a function. I'm still interested in a better way to be able to automatically resolve to the `Money` type. Since that's more implementation details, I'll leave it for a separate PR.

starsirius:
Finally this is ready for review again. Thanks for the patience here. The corresponding UI changes is in https://github.com/artsy/eigen/pull/4182.

erikdstock (commented):
lookin good here ✌️

src/schema/v2/fields/money.ts
```diff
@@ -113,6 +113,30 @@ const money = ({ name, resolve }) => ({
   }),
 })
 
+export const moneyMajorResolver = async (
+  { cents, currency },
+  { convertTo = null },
+  { exchangeRatesLoader }
+) => {
+  const factor = currencyCodes[currency.toLowerCase()].subunit_to_unit
```
ashleyjelks, line 121: Does MP not use null coalescing operators?

ashleyjelks (approved):

starsirius:
Merging and giving it a try. Thanks for the review!

PR #2806, opened 2020-10-29
Ensure we don't try to parse details if they don't exist

zephraph:
Tiny bug fix that we noticed when reviewing some of the functionality of the inquiry request mutation.


xtina-starr (approved):
👍

PR #2815, opened 2020-11-03
feat: Update remaining My Collection apis

jonallured:
Ok I think this is the last one! I've been on a quest to retire all the My Collection-specific endpoints in favor of leveraging existing Gravity endpoints and unless I'm missing something I believe this wraps it. What this PR does is clean up the artwork image loader, switch calls to use the existing artwork loader and then updates the loader for drawing the list of artworks to use the existing collection artworks loader. It's a lot but I broke it into a commit for each thing.

To ensure this doesn't break anything, I booted MP locally with these changes and then pointed my simulator at it. I verified the My Collection interactions and saw that things were working.

https://artsyproduct.atlassian.net/browse/GRO-70

/cc @artsy/grow-devs @artsy/collector-experience


joeyAghion (approved):
🙏

PR #2818, opened 2020-11-05
fix: use symbol for sale.buyersPremium not currency

bhoggard:
Metaphysics is returning this:

```json
{
  "data": {
    "sale": {
      "isWithBuyersPremium": true,
      "buyersPremium": [
        {
          "amount": "USD0",
          "percent": 0.25
        },
        {
          "amount": "USD250,000",
          "percent": 0.2
        },
        {
          "amount": "USD4,000,000",
          "percent": 0.125
        }
      ]
    }
  }
```

when it should be returning:

```json
{
  "data": {
    "sale": {
      "isWithBuyersPremium": true,
      "buyersPremium": [
        {
          "amount": "$0",
          "percent": 0.25
        },
        {
          "amount": "$250,000",
          "percent": 0.2
        },
        {
          "amount": "$4,000,000",
          "percent": 0.125
        }
      ]
    }
  }

### DIFF
--- a/src/schema/v2/sale/__tests__/index.test.ts
+++ b/src/schema/v2/sale/__tests__/index.test.ts
@@ -15,7 +15,7 @@ describe("Sale type", () => {
     id: "foo-foo",
     _id: "123",
     collect_payments: true,
-    currency: "$",
+    currency: "USD",
     is_artsy_licensed: false,
     is_auction: true,
     is_preliminary: false,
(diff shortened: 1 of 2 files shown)


pepopowitz (approved):
👍 

Do you think it's worth adding a test that covers this?

dzucconi (approved):

bhoggard:
There already one, and I don't understand why it was passing:

https://github.com/artsy/metaphysics/blob/master/src/schema/v2/sale/__tests__/index.test.ts#L444-L477

dzucconi:
Looks like that spec was passing because currency was being set as symbol: https://github.com/artsy/metaphysics/blob/8651ca947223f818f51259a99b18c835960f8c43/src/schema/v2/sale/__tests__/index.test.ts#L18

bhoggard:
Thanks, I'll fix it!

PR #2823, opened 2020-11-09
fix: don't suppress all myCollectionArtwork errors

brainbicycle:
I was looking into why myCollection calls were failing in Eigen and saw an error in gravity logs for an improper sort type that wasn't showing up in metaphysics. It looks like we were suppressing all errors because the myCollectionArtwork loader in some circumstances can return a 404 when a user does not have a collection. This PR instead checks for that particular error and throws for all other errors for hopefully a bit easier debugging.


ds300 (approved):
👍

PR #2828, opened 2020-11-13
chore: Remove collector info from lastMessage and clean up code

lilyfromseattle:
[PURCHASE-2211]

Information about the collector meant for the gallery was appended to the `lastMessage` and was thus being displayed to the user in the inbox. This PR resolves `lastMessage` to `initial_message` (which has no collector info) if the last message is the only message. Paired with @zephraph 

Before:

Screen Shot 2020-11-13 at 3 59 14 PM

After:

Screen Shot 2020-11-13 at 3 58 43 PM

[PURCHASE-2211]: https://artsyproduct.atlassian.net/browse/PURCHASE-2211

### DIFF
--- a/src/schema/v2/me/conversation/index.ts
+++ b/src/schema/v2/me/conversation/index.ts
@@ -266,20 +266,20 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
       deprecationReason:
         "This field is no longer required. Prefer the first message from the MessageConnection.",
       type: new GraphQLNonNull(GraphQLString),
-      resolve: ({ initial_message, from_name }) => {
+      resolve: ({ initial_message }) => {
         if (!initial_message) return ""
-        const parts = initial_message.split(
-          "Message from " + from_name + ":\n\n"
-        )
-        return parts[parts.length - 1]
+        return initial_message
       },
     },
     lastMessage: {
       type: GraphQLString,
       description: "This is a snippet of text from the last message.",
-      resolve: ({ _embedded = {} }) => {
+      resolve: ({ initial_message, _embedded = {} }) => {
         const lastMessage = _embedded.last_message || {}
-        return lastMessage.snippet.split("About this collector")[0].trim()
+        if (lastMessage.order == 1) {
+          return initial_message
+        }
+        return lastMessage.snippet
       },
     },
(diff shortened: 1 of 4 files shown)


xtina-starr:
This doesn't handle case when a user doesn't pass a custom message right? That would be outside the scope of this PR?

zephraph:
@xtina-starr, that's a good point. Metaphysics doesn't inject anything when the collector doesn't provide a custom message. If impulse auto-adds some default boilerplate message it'll send that down to the user, otherwise the message will be null/empty.

src/schema/v2/me/conversation/index.ts
```diff
@@ -266,20 +266,20 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
...
     },
     lastMessage: {
       type: GraphQLString,
       description: "This is a snippet of text from the last message.",
-      resolve: ({ _embedded = {} }) => {
+      resolve: ({ initial_message, _embedded = {} }) => {
         const lastMessage = _embedded.last_message || {}
-        return lastMessage.snippet.split("About this collector")[0].trim()
+        if (lastMessage.order == 1) {
+          return initial_message
```
starsirius, line 280: 👍

starsirius (approved):
Looks good to me! As we agreed on Slack to deprecate partner show inquiries, we can work on a plan to clean up the code.

I think previously the UI requires collectors to fill out the initial message to submit an inquiry. The new inquiry experience doesn't require that so we should expect to have more conversations without the initial message. The partner side should be fine, and I'm curious how that'd affect collector side of experience. For example, the preview on the list view:

Screen Shot 2020-11-17 at 1 41 56 PM

zephraph:
@starsirius we _shouldn't_ see that w/ this change. So far as I'm aware `initial_message` specifically excluded that. Still, it would be good to verify.

zephraph:
@starsirius, @lilyfromseattle and I went through and double checked the behavior in metaphysics. Can confirm that that message would not carry over. It'll be empty when displayed in the UI now (which while maybe not ideal design wise is fine to move forward with).

starsirius:
@zephraph Thank you for verifying! Yeah we can have a separate ticket for addressing the empty message preview. 👍

PR #2835, opened 2020-11-20
Feature: Add sizeScore to artwork and editionSet

bhoggard:
completes https://artsyproduct.atlassian.net/browse/CX-788


pepopowitz (approved):
Looks great! Thank you!

PR #2836, opened 2020-11-20
feat(stitching): always enable convection/commerce schema stitching

dblandin:
Remove the environment variables which control convection/commerce schema stitching. We always want this behavior enabled and not having these environment variables set has caused occasional headache during local development.

Jira: Addresses [CX-650]

cc/ @pvinis 

[CX-650]: https://artsyproduct.atlassian.net/browse/CX-650


damassi:
Thank you!

PR #2858, opened 2020-12-01
chore(fx-2514): enables jest eslint plugin

dzucconi:
Closes: [FX-2514](https://artsyproduct.atlassian.net/browse/FX-2514)

OK, so I tend to use focused specs quite a bit when working, and frequently forget to unfocus them (and thanks to @damassi for always catching me). The jest plugin takes care of this with the recommended settings and will error.

We had eslint disabled over the specs — I think because of how common the `promise/always-return` rule was in violation. But, this is an easy automated refactoring: just use an async function. It's still disabled file-wide in a lot of places and I just left those alone. The rest of the errors I dealt with. Some much more valid/problematic. In general I think having eslint enabled over specs will help catch a lot of things.

Right now the only things that are coming up as warns are disabled specs with skip. It's nice to have those highlighted in your editor as well.

I'd like to enable this in Force as well, unsure what will be involved with that.

### DIFF
--- a/src/lib/__tests__/cache.test.ts
+++ b/src/lib/__tests__/cache.test.ts
@@ -1,4 +1,3 @@
-/* eslint-disable promise/always-return */
 import zlib from "zlib"
 import config from "config"
 import cache, { client, cacheKey } from "lib/cache"
@@ -23,16 +22,19 @@ function parseCacheResponse(data, cacheCompressionDisabled) {
 
 describe("Cache with compression enabled", () => {
   config.CACHE_COMPRESSION_DISABLED = true
-  expect(config.CACHE_COMPRESSION_DISABLED).toBe(true) // check that the config is mocked
+
+  // FIXME:
+  // eslint-disable-next-line jest/no-standalone-expect
+  expect(config.CACHE_COMPRESSION_DISABLED).toBe(true)
 
   describe("when successfully connected to the cache", () => {
     describe("#get", () => {
       beforeEach(async () => await cache.set("get_foo", { bar: "baz" }))
 
-      it("parses the data and resolves the promise", () => {
-        return cache.get("get_foo").then((data) => {
-          expect(data.bar).toBe("baz")
-        })
+      it("parses the data and resolves the promise", async () => {
+        const data = await cache.get("get_foo")
+
+        expect(data.bar).toBe("baz")
       })
     })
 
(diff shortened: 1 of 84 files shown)


src/lib/__tests__/cache.test.ts
```diff
@@ -40,39 +43,41 @@ describe("Cache with compression enabled", () => {
...
     describe("#set", () => {
       describe("with a plain Object", () => {
-        it("sets the cache and includes a timestamp", async (done) => {
+        it("sets the cache and includes a timestamp", async () => {
+          expect.assertions(2)
+
           await cache.set("set_foo", { bar: "baz" })
 
-          client.get(cacheKey("set_foo"), (_err, data) => {
+          await client.get(cacheKey("set_foo"), (_err, data) => {
```
damassi, line 63: Does the eslint error instruct to use `await` if this is forgotten?

src/lib/__tests__/date.test.ts
```diff
@@ -15,7 +15,7 @@ import {
 } from "lib/date"
 
 describe("date formatting", () => {
-  describe(timeRange, () => {
+  describe("timeRange", () => {
```
damassi, line 18: def think this form is a bit weird and unexpected and clever, but its also valid since it will read the `function.name` prop. Did this throw a warning / error?

src/lib/__tests__/cache.test.ts
dzucconi: Oh, I may have just misunderstood this one. I guess it's just the callback not returning a promise.

In general eslint will error on thenables that aren't returned with `Promise should be returned to test its fulfillment or rejectioneslintjest/valid-expect-in-promise` but yeah this isn't one of those.

src/lib/__tests__/date.test.ts
dzucconi: `Title must be a string (eslintjest/valid-title)`

damassi: Non blocking, especially after all of this work was done, but i think that rule could be disabled. I can't remember where i read it but there's some interesting introspection benefits around passing the function in versus a string, but i also wonder about eslintjest setting this rule as a default -- in like a what do they know kind of way /shrug

damassi (approved):
I've dreamt of switching all of these things to `await` for years -- thanks for all of the updates 👍

src/lib/__tests__/date.test.ts
dzucconi: I'd be curious to know what the benefits are, if you ever dig it up. I'm inclined to roll with the string form since it's just more consistent and the default behavior of the recommended settings. There's a bit of discussion I dug up here https://github.com/jest-community/eslint-plugin-jest/issues/431 for context as well.

PR #2867, opened 2020-12-04
update causality schema

erikdstock:
This is a followup to https://github.com/artsy/causality/pull/861 (already in staging, not production) which fixed some `ID`-typed args that had been changed to `Strings`. I don't think we are currently using any of these argument-containing fields in metaphysics.

Please let me know if this isn't the right way to go about this schema upgrade 🤝

### DIFF
--- a/src/data/causality.graphql
+++ b/src/data/causality.graphql
@@ -104,10 +104,10 @@ type LotStanding implements Node {
   "whether this user has the leading bid"
   isHighestBidder: Boolean!
 
-  "whether this user has the leading bid"
+  "Current lot state"
   lotState: Lot! @deprecated(reason: "prefer `lot`")
 
-  "whether this user has the leading bid"
+  "Current lot state"
   lot: Lot!
 }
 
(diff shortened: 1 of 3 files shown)


src/data/causality.graphql
```diff
@@ -236,7 +236,7 @@ type Query {
   ): LotStandingConnection!
 
   "Returns an increment policy with specific `id`."
-  incrementPolicy(id: String!): IncrementPolicy
+  incrementPolicy(id: ID!): IncrementPolicy
```
starsirius, line 239: These are not showing up in the resulting MP schema updates above. I think it's because they are [not permitted](https://github.com/artsy/metaphysics/blob/b1a67f53608f29777009006f5d1ca98d1d6054f5/src/lib/stitching/causality/schema.ts#L15) when merging schemas. But let me know if I am wrong.

starsirius:
Merging! I think from the MP perspective, it's just a description update. We'll see client updates submitted automatically shortly, e.g. Force, Eigen, and Volt.

PR #2871, opened 2020-12-09
 [DoNotMerge] Refactor consign pages and add realized price / low estimate multiplier

williardx:
https://artsyproduct.atlassian.net/browse/GRO-54

This PR does two things:

- Adds a new `realizedToEstimate` field on the `Artwork` type that only has non-null values for a very small number of artworks as defined by the data in `staticCSVToJSONData`. We're only using this field to show on the `/consign` landing page artworks that were sold at multiples over their low estimate.
- Refactors `ArtistTargetSupplyMicrofunnel` and `getMicrofunnelData` (renamed to `getArtistMicrofunnelMetadata`) to remove the deprecated `artworks` field in favor of using `artworksConnection`. This was some tech debt that created multiple codepaths to the same data and caused some confusion in understanding the code.

**DoNotMerge** until https://github.com/artsy/force/pull/6775 is ready.


damassi (approved):
LGTM 👍

PR #2878, opened 2020-12-10
feat: add auction result type

MounirDhahri:
The type of this PR is: Feature

This PR helps resolve https://artsyproduct.atlassian.net/browse/CX-740

### Description
- Expose auctionResult type

Screenshot 2020-12-10 at 18 55 44

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -8651,6 +8651,12 @@ type Query {
     width: String
   ): FilterArtworksConnection
 
+  # An auction result
+  auctionResult(
+    # The ID of the auction result
+    id: String!
+  ): AuctionResult
+
   # A city-based entry point for local discovery
   city(
     # A point which will be used to locate the nearest local discovery city within a threshold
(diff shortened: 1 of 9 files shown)


_schemaV2.graphql
```diff
@@ -10941,6 +10947,12 @@ type Viewer {
     width: String
   ): FilterArtworksConnection
 
+  # An auction result
+  auctionResult(
+    # The ID of the auction result
+    id: String!
+  ): AuctionResult
+
```
brainbicycle, line 10955: Should this be here twice?

brainbicycle (commented):
This looks good to me 🙌 , added David as reviewer as he is more aware of gotchas here. Is this ready to go besides the test failure?

_schemaV2.graphql
ds300: yeah. if memory serves we have this `viewer` root field which is just a duplication of the top-level `Query` type because it is/(used to be?) awkward to do certain kinds of query in relay on the top level.

ds300 (approved):
looks all good from here :+1:

MounirDhahri:
#squashongreen

PR #2880, opened 2020-12-11
add stitched sellingPrice field

erikdstock:
Part of [PURCHASE-2294]
#minor
This PR updates our stitched causality schema to include the `sellingPrice` field, which is the primary field we want for representing the selling price of an auction lot.

[PURCHASE-2294]: https://artsyproduct.atlassian.net/browse/PURCHASE-2294


erikdstock:
#squashongreen

PR #2893, opened 2020-12-15
fix(my collection) Create images synchronously to avoid gravity race condition

ds300:
We have discovered a race condition in the `POST /artwork/:id/image` gravity endpoint that is causing corrupt data. This PR adds a workaround while we track down the root cause in gravity, in order to unblock work in eigen.


jonallured (approved):
I'm sad this is required but it's a very pragmatic step forward - good thinking!! ❤️

PR #2897, opened 2020-12-16
feat(classifications): add new artwork classification (attribution class) values

dblandin:
We're removing "made-to-order", "reproduction", "editioned multiple", "non-editioned multiple", and "ephemera" from our list of available artwork classifications.

We're adding "open edition" and "unknown edition"

This mapping is currently used by the `artworkAttributionClasses` root field and the `attributionClass` Artwork field.

We should not remove the deprecated values from the mapping until the data migration is complete.

Jira 🔒 : [FX-2465]

### Old + New Values

Screen Shot 2020-12-16 at 5 05 04 PM

### With `EXCLUDE_DEPRECATED_ATTRIBUTION_CLASSES=true`

Screen Shot 2020-12-16 at 5 04 38 PM

[FX-2465]: https://artsyproduct.atlassian.net/browse/FX-2465


src/schema/v2/artworkAttributionClasses.ts
```diff
@@ -1,12 +1,23 @@
...
 const ArtworkAttributionClasses: GraphQLFieldConfig<void, ResolverContext> = {
   type: new GraphQLList(AttributionClass),
   description: "List of all artwork attribution classes",
-  resolve: (_root, _args, _context) => Object.values(attributionClasses),
+  resolve: (_root, _args, _context) => {
+    let values = Object.values(attributionClasses)
+
+    if (EXCLUDE_DEPRECATED_ATTRIBUTION_CLASSES) {
+      values = values.filter((attributionClass) => !attributionClass.deprecated)
+    }
```
anandaroop, line 17: Nice, so we'll toggle this flag to true once the data migration has run 👍

anandaroop (approved):
LGTM

PR #2899, opened 2020-12-17
update readme for introspection setup

pvinis:
After pairing with @annacarey, I noticed that this is only documented on slack [here](https://artsy.slack.com/archives/C02BC3HEJ/p1599250831072100).

I added this in the readme here so it's fast to find when any artsy dev sets up a new app/machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,14 @@ You will need to set up headers with both:
 
 _If you're new to GraphQL, you can checkout [Artsy's GraphQL Workshop](https://github.com/artsy/graphql-workshop)._
 
+### Introspection Setup
+
+In order to get docs for the schema on MP in your playground of choice (Postman, Insomnia, Altair, etc), you need to send the following header:
+```
+Authorization: Bearer <secret>
+```
+and replace `<secret>` with the contents of `Metaphysics INTROSPECT_TOKEN` from 1Password.
+
 ### Sample Queries
 
 Once you have the GraphiQL client running against your local service,
--- a/README.md
+++ b/README.md
@@ -97,11 +97,22 @@ _If you're new to GraphQL, you can checkout [Artsy's GraphQL Workshop](https://g
 
 ### Introspection Setup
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -95,6 +95,14 @@ You will need to set up headers with both:
 
 _If you're new to GraphQL, you can checkout [Artsy's GraphQL Workshop](https://github.com/artsy/graphql-workshop)._
 
+### Introspection Setup
+
+In order to get docs for the schema on MP in your playground of choice (Postman, Insomnia, Altair, etc), you need to send the following header:
+```
+Authorization: Bearer <secret>
+```
+and replace `<secret>` with the contents of `Metaphysics INTROSPECT_TOKEN` from 1Password.
```
dblandin, line 104: The source of truth for this value would be in our Kubernetes cluster and depends on whether you're targeting staging or production. We might also include some hokusai commands in case the 1Password value drifts.

```
$ hokusai staging env get INTROSPECT_TOKEN
$ hokusai production env get INTROSPECT_TOKEN
```

README.md
```diff
@@ -95,6 +95,14 @@ You will need to set up headers with both:
 
 _If you're new to GraphQL, you can checkout [Artsy's GraphQL Workshop](https://github.com/artsy/graphql-workshop)._
 
+### Introspection Setup
```
dblandin, line 98: What do you think about mentioning that schema documentation from staging or production endpoints is only available for internal use.

In development (where `INTROSPECT_TOKEN` is not usually set), you don't need to send this header to fetch schema docs.

dblandin (commented):
Added a couple comments and to summarize:

- Runtime schema introspection is available without an `Authorization` header during development
- Runtime schema introspection from staging/production endpoints is for internal use only
- If you're an Artsy Engineer, you can find the introspect token via hokusai or 1Password.

pvinis:
thanks devon! it should be more complete now.

PR #2900, opened 2020-12-17
feat(my collection): create/update edition information

ds300:
This implements the ability for create/update mutations on 'my collection' artworks to set/update edition information.

### DIFF
--- a/src/schema/v2/me/__tests__/myCollectionUpdateArtworkMutation.test.ts
+++ b/src/schema/v2/me/__tests__/myCollectionUpdateArtworkMutation.test.ts
@@ -332,7 +332,7 @@ describe("myCollectionUpdateArtworkMutation", () => {
       )
     })
 
-    it("does not create an edition set if you don't specify either", async () => {
+    it("resets the edition set if you don't specify either", async () => {
       const mutation = computeMutationInput({
         editionNumber: null,
         editionSize: null,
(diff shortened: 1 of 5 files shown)


src/schema/v2/me/__tests__/myCollectionUpdateArtworkMutation.test.ts
```diff
@@ -189,4 +201,146 @@ describe("myCollectionUpdateArtworkMutation", () => {
...
+          editionSetId: editionedArtwork.edition_sets[0].id,
+        },
+        {
+          edition_size: "50",
+          available_editions: null,
+        }
+      )
+    })
+
+    it("does not create an edition set if you don't specify either", async () => {
```
pepopowitz, line 335: [not-sure-if-blocking]  

Is this the right thing to do if there's no editionNumber/editionSize on edit?

I'm worried it will translate to an experience where I edit an artwork in my collection and remove the edition number/size, but the system doesn't actually remove it when I save.

Maybe there are other places you're anticipating handling that scenario?

pepopowitz (commented):
This is great, thank you!!!!

I'm worried about a scenario where I can't actually remove edition info from my artwork when I edit.

I'm going to reassign to you so that you can merge when you think the scenario is appropriately handled!

src/schema/v2/me/__tests__/myCollectionUpdateArtworkMutation.test.ts
ds300: Very good catch!!!! will fix tomorrow

PR #2904, opened 2020-12-18
Update artwork mediums

dblandin:
While this field is still new use and unused in production, this PR makes a few changes to the schema:

- Rename `Query#artworkMediumTypes` to `Query#artworkMediums`
- Rename internal `MediumType` type to `ArtworkMedium`
- Remove ID fields from `ArtworkMedium/MediumType` type.

The name of the artwork field is still `mediumType`.

Rebase and merge please.

cc/ @dzucconi - This will introduce some required changes to your work in Force.


mzikherman (approved):
🔥

dzucconi:
We need to expose an `artworkMedium(type: MEDIUM)` field for querying individual mediums, as well.

dblandin:
> We need to expose an artworkMedium(type: MEDIUM) (or whatever) field for querying individual mediums, as well.

@dzucconi Ahh, I thought that UI might be handled by `Artwork#mediumType` but maybe not?

dblandin:
Going to go ahead and merge this in. Let's follow-up on the `artworkMedium(type: MEDIUM)` addition if needed.

dzucconi:
@dblandin ah no you're right that's fine. I didn't know it was there. Thanks

PR #2906, opened 2020-12-21
Update copy for artwork classifications

dblandin:
- chore: Mark AttributionClass#info as deprecated
    This attribute doesn't seem to be used in any of our clients and came up while we were updating copy for new classification options.
- chore: update copy for artwork classifications
    Introduce minor copy updates for shortDescription.

@oxaudo I couldn't find any client using the `info` field. Think it's alright to deprecate until we're confident in removing it?

Note: Rebase and merge

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1669,20 +1669,18 @@ type Attachment {
 type AttributionClass {
   # A globally unique ID.
   id: ID!
-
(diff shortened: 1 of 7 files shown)


oxaudo:
I'm cool with deprecating it. I don't know where it was intended to be used but seems like 'shortDescription' was preferred use already!

oxaudo:
@dblandin - test fail seems related here.

PR #2912, opened 2020-12-22
Allow for query of partner top viewed viewing room data

jpotts244:
Builds out the Metaphysics interface to allow for querying of the new Vortex partner analytics of a partner's most viewed viewing rooms.

Associated Vortex changes: https://github.com/artsy/vortex/pull/291 

Takes a partner id, period, and the new viewing room query enum (defined in vortex) and returns a list of Viewing Rooms with the number of user views it has received during the given time period. Also allows us to pull additional viewing room attributes from Gravity's GraphQL schema.

Screen Shot 2021-01-13 at 2 50 42 PM

This data will drive Most Viewed VR in CMS partner analytics 
Screen Shot 2021-01-13 at 2 01 04 PM

Paired with some cool friends on this - shoutout to Sarah, Matt, and Justin for the help here! 🎉

### DIFF
--- a/src/lib/stitching/mergeSchemas.ts
+++ b/src/lib/stitching/mergeSchemas.ts
@@ -80,7 +80,9 @@ export const incrementalMergeSchemas = (localSchema, version: 1 | 2) => {
   if (version === 1) {
     useStitchingEnvironment(vortexStitchingEnvironmentv1(localSchema))
   } else {
-    useStitchingEnvironment(vortexStitchingEnvironmentv2(localSchema))
+    useStitchingEnvironment(
+      vortexStitchingEnvironmentv2(localSchema, gravitySchema)
+    )
   }
 
   // Always stitch kaws
--- a/src/lib/stitching/vortex/stitching.ts
+++ b/src/lib/stitching/vortex/stitching.ts
@@ -13,10 +13,13 @@ const getMaxPrice = (thing: { listPrice: any }) => {
   return thing.listPrice.minor || thing.listPrice.maxPrice.minor
 }
 
-export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
+export const vortexStitchingEnvironment = (
+  localSchema: GraphQLSchema,
+  gravitySchema: GraphQLSchema
+) => ({
   // The SDL used to declare how to stitch an object
   extensionSchema: gql`
-    union AnalyticsRankedEntityUnion = Artwork | Show | Artist
+    union AnalyticsRankedEntityUnion = Artwork | Show | Artist | ViewingRoom
     extend type AnalyticsPricingContext {
       appliedFiltersDisplay: String
     }
@@ -250,6 +253,7 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
         fragment: gql`... on Partner {
           internalID
         }`,
+
         resolve: async (source, _, context, info) => {
           const args = { partnerId: source.internalID }
           return await info.mergeInfo.delegateToSchema({
@@ -307,7 +311,7 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
       entity: {
         fragment: gql`
           ... on AnalyticsRankedStats {
-            rankedEntity{
+            rankedEntity {
               __typename
               ... on AnalyticsArtwork {
                 entityId
@@ -318,17 +322,27 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
               ... on AnalyticsArtist {
                 entityId
               }
+              ... on AnalyticsViewingRoom {
+                entityId
+              }
             }
           }
         `,
         resolve: (parent, _args, context, info) => {
           const removeVortexPrefix = (name) => name.replace("Analytics", "")
           const typename = parent.rankedEntity.__typename
-          const fieldName = removeVortexPrefix(typename).toLowerCase()
+          const typenameWithoutPrefix = removeVortexPrefix(typename)
+          // Data massaging to allow for proper casing for query: viewingRoom, show, artwork, etc
+          const fieldName =
+            typenameWithoutPrefix.charAt(0).toLowerCase() +
+            typenameWithoutPrefix.slice(1)
           const id = parent.rankedEntity.entityId
+          const schema =
+            fieldName == "viewingRoom" ? gravitySchema : localSchema
+
           return info.mergeInfo
             .delegateToSchema({
-              schema: localSchema,
+              schema: schema,
               operation: "query",
               fieldName,
               args: {
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -292,7 +292,7 @@ type AnalyticsPartnerStats {
     @deprecated(reason: "Use pageview for refactored time series bucket code")
   partnerId: String!
 
-  # Artworks, shows, or artists ranked by views. Capped at 20 by the underlying sql query.
+  # Artworks, shows, viewing rooms, or artists ranked by views. Capped at 20 by the underlying sql query.
   rankedStats(
     # Returns the elements in the list that come after the specified cursor.
     after: String
@@ -478,9 +478,9 @@ enum AnalyticsQueryPeriodEnum {
   SIXTEEN_WEEKS
 }
 
-union AnalyticsRankedEntityUnion = Artist | Artwork | Show
+union AnalyticsRankedEntityUnion = Artist | Artwork | Show | ViewingRoom
 
-# Top artworks, shows, or artists from a partner
+# Top artworks, shows, viewing rooms, or artists from a partner
 type AnalyticsRankedStats {
   entity: AnalyticsRankedEntityUnion
   period: AnalyticsQueryPeriodEnum!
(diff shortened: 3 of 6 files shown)


src/schema/v2/schema.ts
```diff
@@ -160,6 +161,7 @@ const rootFields = {
   salesConnection: SalesConnectionField,
   searchConnection: Search,
   show: Show,
+  viewingRoom: ViewingRoom,
```
jpotts244, line 164: Added viewRoom to be accessible as a node in the local schema

jpotts244: Might need a pluralized loader to query for a list of viewing rooms? Looks like there is a pattern here for single lookups and list lookups - doing some research

jpotts244: Ah another issue here, looks like the request needs to be authenticated to get viewing rooms data here. In this current set up, the gravity loader is in the public space

jpotts244:
opening this up for early feedback, main issue here we require authentication to view viewing rooms which is causing our new query to 404. Need to revisit a solution in the new year, so far it has been challenging here!

src/lib/stitching/vortex/stitching.ts
```diff
@@ -318,14 +318,27 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
...
+          // TO CLEAN BEFORE MERGE ---------------
+          // This fieldName line below is a point of failure here since
+          // the type name comes through as "AnalyticsViewingRoom"
+          // and the toLowerCase changes the case to "viewingroom"
+          // we need to match the gravity schema case of "viewingRoom"
+          const typenameWithoutPrefix = removeVortexPrefix(typename)
+          const fieldName =
+            typenameWithoutPrefix.charAt(0).toLowerCase() +
+            typenameWithoutPrefix.slice(1)
+          // -------------------------------------------------
```
mdole, line 339: Is this comment outdated? I think as-written it should work - going from `AnalyticsViewingRoom` -> `VieiwngRoom` -> `viewingRoom`, right?

src/schema/v2/schema.ts
mdole: Oooo interesting - @jpotts244 `viewingRoom` should actually be available as a top-level field already that can be queried with an ID, it's just stitched in from Gravity.

That said, I definitely don't fully understand how loading a VR object by ID in Metaphysics works. I can go open up a staging console and query and we're fine, but I'm less familiar with how this would work when Vortex is hitting Metaphysics.

Happy to chat about this when you're back, or maybe @ansor4 understands what's going on here better than I do?

src/schema/v2/viewingRoom.ts
```diff
@@ -0,0 +1,40 @@
+import { GraphQLString, GraphQLObjectType, GraphQLFieldConfig } from "graphql"
+import cached from "./fields/cached"
+import { InternalIDFields } from "./object_identification"
+import { ResolverContext } from "types/graphql"
+
+export const ViewingRoomType = new GraphQLObjectType<any, ResolverContext>({
```
mdole, line 6: yeah I _feel_ like we shouldn't need this since we define this type in Gravity's GraphQL fields and stitch it in...would def like to learn more about why this would/would not be necessary because I'm not sure if we've set up this Vortex flow with a Gravity GQL type before! E.g. shows and artists are queried via REST endpoints w/ Gravity loaders, so we might actually be able to do less (or at least different) stuff with ViewingRooms since they exist as GQL fields.

This might also relate to the authentication you discussed above - the Gravity REST endpoint I _think_ is used for partners, while the GQL fields are what we use for displaying in Eigen/Force. We might not need to use loaders at all, and therefore not have to worry about auth, if we use the `ViewingRoom` field provided by Gravity GQL.

mdole (commented):
Nice work! Let's chat when you're back about using REST vs. Gravity GQL - my hope is that we can simplify since we already have a `viewingRoom` type from Gravity's GQL schema stitched into MP. I'm not 100% positive though since I'm not super familiar with the Vortex/Metaphysics relationship!

src/lib/stitching/vortex/stitching.ts
jpotts244: Correct! old comment here, this should be working with the new addition of likes 335 - 338 - will remove

src/schema/v2/viewingRoom.ts
jpotts244: Addressed offline! 

Matt is exactly right here - Viewing Rooms are accessed via Gravity’s Graphql interface where the other objects in this pattern were accessed by gravity data loaders. So I was barking up (🐕 ) the wrong tree in these early comments

src/lib/stitching/vortex/stitching.ts
```diff
@@ -318,17 +322,27 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
               ... on AnalyticsArtist {
                 entityId
               }
+              ... on AnalyticsViewingRoom {
+                entityId
+              }
             }
           }
         `,
         resolve: (parent, _args, context, info) => {
```
jpotts244, line 331: Working on some tests for this work locally, it is proving to be very challenging 😅

src/lib/stitching/mergeSchemas.ts
```diff
@@ -80,7 +80,9 @@ export const incrementalMergeSchemas = (localSchema, version: 1 | 2) => {
   if (version === 1) {
     useStitchingEnvironment(vortexStitchingEnvironmentv1(localSchema))
   } else {
-    useStitchingEnvironment(vortexStitchingEnvironmentv2(localSchema))
+    useStitchingEnvironment(
+      vortexStitchingEnvironmentv2(localSchema, gravitySchema)
```
jpotts244, line 84: Think it is OK to only integrate the `gravitySchema` for `vortexStitchingEnvironmentv2`? or should we add this implementation to `vortexStitchingEnvironmentv1` as well? Looks like these functions are explicitly versioned.

I remember we chatted about it when we were mini-mobbing but I dont recall the outcome.

mdole: I'm honestly not sure what v1 is left over from, but relatively sure v2 is what we're consuming with Partner Analytics so seems safe to just use v2 IMO

mdole (approved):
Wohoo! This looks great to me, nice work @jpotts244.

Thinking about order of operations here: seems like the safest thing to do would be make sure the Vortex work is merged & deployed before this is, right? Though we shouldn't technically be consuming this new field until Volt is hooked up, so maybe it's 6-a-one half-dozen-the-other 🤷‍♂️

jpotts244:
Yes! Thanks @mdole, We are waiting behind vortex being merged + deployed, I assigned this to myself so I can shepherd it around

src/lib/stitching/vortex/stitching.ts
```diff
@@ -342,6 +356,9 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
               response.__typename = removeVortexPrefix(typename)
               return response
             })
+            .catch(() => {
+              return null
+            })
```
jpotts244, line 361: Added lines 359-361 above to add some protection in case we are trying to query for viewing rooms that were recently deleted.  I am a little not confident in this as a solution - what do yall think? One thing that is tricky for me is that we are fetching objects from different places (viewing room data is coming from GravityQL, shows/artist/artwork data is coming in from Gravity data loaders. 
Looking like queries that hit the gravity loaders are catching errors in case of 404s at the loader level.


Here's the experience WITHOUT this catch when trying to fetch analytics data for a viewing room that was recently deleted:

Screen Shot 2021-01-15 at 10 19 06 AM

Here is the response with the catch added here. We return `null` for the entity object if the ViewingRoom cannot be fetched from GravityQL. Looking like Volt is prepared to handle this pattern already. (show viewing analytics data for deleted shows for example.

Screen Shot 2021-01-15 at 10 21 35 AM


TL;DR I don't want to an error to bubble up to Sentry if a VR is missing! Would prefer we just return null for the object and continue on our way.  Would love to hear your feedback here!

jpotts244: Discussed with the crew - since this is existing behavior we will resolve the root cause of this exception as part of a separate ticket: https://artsyproduct.atlassian.net/browse/PX-3637

src/lib/stitching/vortex/stitching.ts
```diff
@@ -318,17 +322,27 @@ export const vortexStitchingEnvironment = (localSchema: GraphQLSchema) => ({
...
         `,
         resolve: (parent, _args, context, info) => {
           const removeVortexPrefix = (name) => name.replace("Analytics", "")
           const typename = parent.rankedEntity.__typename
-          const fieldName = removeVortexPrefix(typename).toLowerCase()
+          const typenameWithoutPrefix = removeVortexPrefix(typename)
+          // Data massaging to allow for proper casing for query: viewingRoom, show, artwork, etc
+          const fieldName =
+            typenameWithoutPrefix.charAt(0).toLowerCase() +
+            typenameWithoutPrefix.slice(1)
```
ansor4, line 338: kinda makes you wish strings were mutable in js huh? well, this looks like it's the best way to me!

ansor4 (approved):
Awesome job!

PR #2916, opened 2020-12-23
feat(my collection): allow marking works as editions without requiring extra fields

brainbicycle:
**New behavior:** We are hoping to support users creating artworks marked as editions without requiring them to provide **editionSize** or **editionNumber**

**Old behavior:** Editions were only created when a user passed either editionSize or editionNumber. There was a small bug here where if only editionSize was passed creation would fail validation because of explicit null being passed in the field. This instead doesn't provide the field when it isn't present.

To support the new behavior this PR:
- adds a new field to the `myCollectionCreateArtworkMutation` `isEdition` to explicitly ask for an edition to be created. 
- only passes provided fields to edition creation to prevent validation error from explicit null
- To stay backwards compatible this also keeps logic to create an edition if either **editionSize** or **editionNumber** is passed. 
-  adds a helper field on artwork `isEdition` 
- adds similar fields and logic to `myCollectionUpdateArtworkMutation`, slight difference in that this mutation needs to be able to set individual fields to empty values

TODO:
- [x] Similar logic in update mutation

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -223,6 +223,12 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      isEdition: {
+        type: GraphQLBoolean,
+        resolve: ({ edition_sets }) => {
+          return edition_sets && edition_sets.length >= 1
+        },
+      },
       editionSize: {
         type: GraphQLString,
         resolve: ({ edition_sets }) => edition_sets?.[0]?.edition_size,
--- a/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
+++ b/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
@@ -1,4 +1,4 @@
-import { GraphQLString, GraphQLList, GraphQLInt } from "graphql"
+import { GraphQLString, GraphQLList, GraphQLInt, GraphQLBoolean } from "graphql"
 import { mutationWithClientMutationId } from "graphql-relay"
 import { ResolverContext } from "types/graphql"
 import { GraphQLNonNull } from "graphql"
@@ -56,6 +56,9 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
     depth: {
       type: GraphQLString,
     },
+    isEdition: {
+      type: GraphQLBoolean,
+    },
     editionNumber: {
       type: GraphQLString,
     },
@@ -92,6 +95,7 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
       artistIds,
       costCurrencyCode,
       costMinor,
+      isEdition,
       editionSize,
       editionNumber,
       externalImageUrls = [],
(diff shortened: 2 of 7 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -223,6 +223,12 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      isEdition: {
+        type: GraphQLBoolean,
+        resolve: ({ edition_sets }) => {
+          return edition_sets && edition_sets.length >= 1
```
brainbicycle, line 229: Would love some eyes on this to make sure this is valid logic 👀

src/schema/v2/me/myCollectionUpdateArtworkMutation.ts
```diff
@@ -104,22 +130,36 @@ export const myCollectionUpdateArtworkMutation = mutationWithClientMutationId<
...
-            edition_size: editionSize,
-            available_editions: editionNumber ? [editionNumber] : null,
+
+        if (isEdition === false) {
+          await deleteArtworkEditionSetLoader({ artworkId, editionSetId })
+        } else {
+          const payload = {
+            edition_size: editionSize ? editionSize : null,
+            // TODO: Is there a better way to clear out edition number?
+            available_editions: editionNumber ? [editionNumber] : [""],
```
brainbicycle, line 155: This is a bit weird, I can't find a way of clearing out `avaliable_editions` which populates `edition_number` cleanly. Passing null fails gravity validation, passing empty array does not clear out edition_number. Passing an array with empty string is close to what we are after making 'edition_number' an empty string but ideally I could pass something here to make the field null.

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
```diff
@@ -122,12 +126,19 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
 
       const artworkId = response.id
 
-      if (editionNumber || editionSize) {
+      const providedEditionField = editionNumber || editionSize
```
admbtlr, line 129: nitpick: `providedEditionField` sounds like you're going to use the content of the field; maybe `hasEditionField`? Or actually I think you could just skip the const and munge together the boolean without sacrificing readability.

src/schema/v2/artwork/index.ts
admbtlr: Looks OK to me, but there may be hidden weirdness I don't know about?

admbtlr (approved):
LGTM!

src/schema/v2/me/myCollectionUpdateArtworkMutation.ts
ashfurrow: I agree this is a bit weird. I'm happy to ship this as is and follow-up with a Gravity PR to allow `avaliable_editions` to be `null`. Might make a good KS topic!

ashfurrow (approved):
Looks good 👍 Test coverage is nice, well done.

PR #2923, opened 2021-01-13
feat(PX-3603): do not return certificate of authenticity data if certificate is false

oxaudo:
There is a request not to display the note that COA is not included when it's marked as false. I think this will solve that for all clients.

We are still waiting for confirmation from Product that there is no additional concerns with that removal hence the WIP message here.


sweir27 (approved):
Awesome! This seems 👍 to me. When you have the go-ahead from product, feel free to merge. 😄

oxaudo:
This is cleared to go. Merging.

PR #2924, opened 2021-01-13
feat: Stitch addressConnection into Me query and merge gravity schema

annacarey:
Co-authored-by: Sepand Ansari <sepand.ansari@artsymail.com>

This PR merges gravity schema that includes new `userAddress` type and `userAddressConnection`. The PR also stitches in the `addressConnection` query. 

Jira ticket [2291](https://artsyproduct.atlassian.net/browse/PURCHASE-2291)

### DIFF
--- a/src/lib/stitching/gravity/stitching.ts
+++ b/src/lib/stitching/gravity/stitching.ts
@@ -173,6 +173,11 @@ export const gravityStitchingEnvironment = (
           },
         },
         addressConnection: {
+          fragment: gql`
+          ... on Me {
+            __typename
+          }
+          `,
           resolve: (_parent, args, context, info) => {
             return info.mergeInfo.delegateToSchema({
               schema: gravitySchema,
--- a/src/lib/stitching/gravity/stitching.ts
+++ b/src/lib/stitching/gravity/stitching.ts
@@ -83,6 +83,12 @@ export const gravityStitchingEnvironment = (
     extensionSchema: gql`
       extend type Me {
         secondFactors(kinds: [SecondFactorKind]): [SecondFactor]
+        addressConnection(
+          first: Int
+          last: Int
+          after: String
+          before: String
+        ): UserAddressConnection
       }
 
       extend type ViewingRoom {
@@ -161,6 +167,18 @@ export const gravityStitchingEnvironment = (
               operation: "query",
               fieldName: "_unused_gravity_secondFactors",
               args: args,
+              context,
+              info,
+            })
+          },
+        },
+        addressConnection: {
+          resolve: (_parent, args, context, info) => {
+            return info.mergeInfo.delegateToSchema({
+              schema: gravitySchema,
+              operation: "query",
+              fieldName: "_unused_gravity_userAddressConnection",
+              args: { ...args, userId: context.userID },
               context,
               info,
             })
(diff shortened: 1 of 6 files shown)


src/lib/stitching/gravity/stitching.ts
```diff
@@ -166,6 +172,23 @@ export const gravityStitchingEnvironment = (
...
+          ... on Me {
+            __typename
+          }
+          `,
+          resolve: (_parent, args, context, info) => {
+            return info.mergeInfo.delegateToSchema({
+              schema: gravitySchema,
+              operation: "query",
+              fieldName: "_unused_gravity_userAddressConnection",
+              args: { ...args, userId: context.userID },
```
annacarey, line 186: @zephraph, @sepans and I were wondering if there is a best practice here to use `userId` from context (which we're assuming we'll always have access to on `Me`) or to query for `userId` in the above fragment and in this line, use `parent.userId` instead in `args`. Wondering about best practice or performance considerations?

src/lib/stitching/gravity/stitching.ts
```diff
@@ -166,6 +172,23 @@ export const gravityStitchingEnvironment = (
             })
           },
         },
+        addressConnection: {
+          fragment: gql`
+          ... on Me {
+            __typename
```
sepans, line 178: Another question for @zephraph:
If we remove this fragment, `me` returns `null` unless there is something else being queried. Is this expected? (screenshots 👇🏽 )

<details><summary>Screenshots</summary>

When there is another field requested under `Me` (anything like `initials` here)

Screen Shot 2021-01-13 at 7 15 02 PM

When that field (`initials`) is removed: 😱 

Screen Shot 2021-01-13 at 7 15 19 PM

When a fragment with just `__typename` is added:

Screen Shot 2021-01-13 at 7 16 28 PM




</details>

sepans (approved):
Looks great 🚀 🚀 

Thanks @annacarey 🙌🏽  Amazing to see how quickly you learned some fancy GraphQL schema work.

Just want to know your thoughts on those 2 questions @zephraph otherwise, it looks good!

src/lib/stitching/gravity/stitching.ts
zephraph: A good question. For the performance, performing the extra query is likely negligible. In many cases it's likely the user whose querying this value will be querying other user data from `Me` anyway in which case the initial call would happen either way. 

I think how you have it is fine.

src/lib/stitching/gravity/stitching.ts
zephraph: This seems likely to be an issue w/ the version of stitching we have. It could be that somehow the stitched resolvers are embedded in the `Me` resolver and if that doesn't fire then the child resolvers don't have a chance to fire. That said, I haven't specifically seen this before. Maybe @mzikherman has?

zephraph (approved):

src/lib/stitching/gravity/stitching.ts
mzikherman: My understanding of this was that you needed to request _something_ besides the purely stitched-in field (otherwise GraphQL seems to short-circuit the resolving entirely).

Rather than request some random ad-hoc fields (which might result in an unneeded loader call), can use the reserved `__typename` which definitely won't trigger any.

PR #2928, opened 2021-01-16
feat: fetch lot states from causality for watchedLotConnection

erikdstock:
This PR updates our stubbed `Me.watchedLotConnection` field (#2925) to query actual data from causality which will be combined to fulfill the `{saleArtwork: Gravity, lot: Causality}` interface. In the process I made a new authenticated loader which actually calls the Causality GraphQL endpoint from the server. I don't think it is possible to use schema delegation for this part of the query (extracting the field selection from the `watchedLotConnection` query) so instead I am
a) calling the causality graphql endpoint directly and requesting every possible field
b) passing them to the stitching layer (where the causality Lot type will be available)
c) resolving them there.

All of this code 👆 lives in the single schema/v2/lot.ts file, and then I import it to the causality stitching file and add it to our extension schema code. My reasoning here is that the initial resolver and extension code is highly entangled and will not be easy to understand in the future - in fact in doing this I think this pattern may be more readable in some ways.*

No tests. Not sure if there is a better way to handle it. Some error handling including forwarding errors returned from causality.

*Note: We found a bug with our sale artwork resolution (basically all sale artwork fields come back `null` and the SaleArtworkType resolvers don't seem to be executing) and as a result will needed to change course on our previous approach of defining the lot type entirely in the causality stitching file. In essence we moved back to the change kicked off in [this comment](https://github.com/artsy/metaphysics/pull/2925#discussion_r557709973) with a new code organization.

image

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -9088,9 +9088,14 @@ type Location {
 
 scalar Long
 
+# A unified auction lot with data from our auctions bidding engine.
 type Lot {
   internalID: String
+
+  # The current auction state of the lot.
   lot: AuctionsLotState!
+
+  # The associated SaleArtwork
   saleArtwork: SaleArtwork
 }
 
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -11657,9 +11657,6 @@ type PublishViewingRoomPayload {
 }
 
 type Query {
-  # Returns a lot with specific `id`.
-  _unused_auctionsLot(id: ID!): AuctionsLotState
-
   # Lot standings for a user
   _unused_auctionsLotStandingConnection(
     after: String
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -9088,37 +9088,6 @@ type Location {
 
 scalar Long
 
-# A unified auction lot with data from our auctions bidding engine.
-type Lot {
-  internalID: String
-
-  # The current auction state of the lot.
-  lot: AuctionsLotState!
-
-  # The associated SaleArtwork
-  saleArtwork: SaleArtwork
-}
-
-# A connection to a list of items.
-type LotConnection {
-  # A list of edges.
-  edges: [LotEdge]
-  pageCursors: PageCursors!
-
-  # Information to aid in pagination.
-  pageInfo: PageInfo!
-  totalCount: Int
-}
-
-# An edge in a connection.
-type LotEdge {
-  # A cursor for use in pagination
-  cursor: String!
-
-  # The item at the end of the edge
-  node: Lot
-}
-
 type LotStanding {
   # Your bid if it is currently winning
   active_bid: BidderPosition
@@ -9562,12 +9531,6 @@ type Me implements Node {
     size: Int
   ): [Artist]
   type: String
-  watchedLotConnection(
-    after: String
-    before: String
-    first: Int
-    last: Int
-  ): LotConnection!
 }
 
 # A message in a conversation.
@@ -11694,6 +11657,9 @@ type PublishViewingRoomPayload {
 }
 
 type Query {
+  # Returns a lot with specific `id`.
+  _unused_auctionsLot(id: ID!): AuctionsLotState
+
   # Lot standings for a user
   _unused_auctionsLotStandingConnection(
     after: String
--- a/src/lib/loaders/loaders_with_authentication/causality.ts
+++ b/src/lib/loaders/loaders_with_authentication/causality.ts
@@ -1,0 +1,44 @@
+import urljoin from "url-join"
+import fetch from "node-fetch"
+import config from "config"
+import { causalityJwt } from "schema/v2/system/causality_jwt"
+
+const { CAUSALITY_API_BASE } = config
+const uri = urljoin(CAUSALITY_API_BASE, "graphql")
+
+interface GraphQLArgs {
+  query: string
+  variables: any
+}
+
+export default (_accessToken, userID) => {
+  const causalityLoader = ({
+    query,
+    variables,
+  }: GraphQLArgs): ReturnType<typeof fetch> => {
+    const token = causalityJwt({
+      userId: userID,
+      role: "observer",
+      saleId: null,
+      bidderId: null,
+    })
+
+    const body = JSON.stringify({
+      query,
+      variables,
+    })
+
+    return fetch(uri, {
+      method: "POST",
+      body,
+      headers: {
+        Authorization: `Bearer ${token}`,
+        "Content-Type": "application/json",
+      },
+    })
+  }
+
+  return {
+    causalityLoader,
+  }
+}
--- a/src/lib/loaders/loaders_with_authentication/causality.ts
+++ b/src/lib/loaders/loaders_with_authentication/causality.ts
@@ -2,6 +2,7 @@ import urljoin from "url-join"
 import fetch from "node-fetch"
 import config from "config"
 import { causalityJwt } from "schema/v2/system/causality_jwt"
+import { GraphQLError } from "graphql"
 
 const { CAUSALITY_API_BASE } = config
 const uri = urljoin(CAUSALITY_API_BASE, "graphql")
@@ -12,10 +13,10 @@ interface GraphQLArgs {
 }
 
 export default (_accessToken, userID) => {
-  const causalityLoader = ({
+  const causalityLoader = async ({
     query,
     variables,
-  }: GraphQLArgs): ReturnType<typeof fetch> => {
+  }: GraphQLArgs): Promise<Record<string, unknown>> => {
     const token = causalityJwt({
       userId: userID,
       role: "observer",
@@ -28,7 +29,7 @@ export default (_accessToken, userID) => {
       variables,
     })
 
-    return fetch(uri, {
+    const response = await fetch(uri, {
       method: "POST",
       body,
       headers: {
@@ -36,6 +37,19 @@ export default (_accessToken, userID) => {
         "Content-Type": "application/json",
       },
     })
+
+    const json = await response.json()
+    const { data: causalityData, errors: causalityErrors } = json
+
+    // If the causality request failed for some reason, throw its errors.
+    if (causalityErrors) {
+      const errors = causalityErrors.reduce((acc, error) => {
+        return acc + " " + error["message"]
+      }, "From causality: ")
+      throw new GraphQLError(errors)
+    } else {
+      return causalityData
+    }
   }
 
   return {
--- a/src/lib/stitching/causality/__tests__/stitching.test.ts
+++ b/src/lib/stitching/causality/__tests__/stitching.test.ts
@@ -4,6 +4,9 @@ import { addMockFunctionsToSchema } from "graphql-tools"
 import { useCausalityStitching } from "./testingUtils"
 import gql from "lib/gql"
 import schema from "schema/v1/schema"
+import v2Schema from "schema/v2/schema"
+import { runAuthenticatedQuery } from "schema/v2/test/utils"
+import { getFieldsForTypeFromSchema } from "lib/stitching/lib/getTypesFromSchema"
 
 describe("causality/stitching", () => {
   describe("extending types", () => {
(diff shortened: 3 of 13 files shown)


_schema.graphql
```diff
@@ -11679,15 +11679,6 @@ type PublishViewingRoomPayload {
 }
 
 type Query {
-  # Lot standings for a user
-  _unused_auctionsLotStandingConnection(
```
erikdstock, line 11683: I don't know why this is getting deleted from the schema - in fact `auctionsLots` should be getting added here. See below.

src/lib/loaders/loaders_with_authentication/causality.ts
```diff
@@ -0,0 +1,44 @@
...
+const { CAUSALITY_API_BASE } = config
+const uri = urljoin(CAUSALITY_API_BASE, "graphql")
+
+interface GraphQLArgs {
+  query: string
+  variables: any
+}
+
+export default (_accessToken, userID) => {
+  const causalityLoader = ({
```
erikdstock, line 15: a causality graphql data loader that just creates the causality token and sends a `fetch` request to it. This might not be necessary or good to expose to other resolvers so we could define it elsewhere if we want.

src/lib/stitching/causality/stitching.ts
```diff
@@ -209,29 +212,94 @@ export const causalityStitchingEnvironment = ({
           },
         },
         watchedLotConnection: {
-          resolve: async (_parent, _args, { saleArtworksAllLoader }) => {
-            const watchedSaleArtworksReq = await saleArtworksAllLoader({
+          resolve: async (_parent, args, context, _info) => {
+            const { saleArtworksAllLoader, causalityLoader } = context
+            // fetch sale artworks from gravity
+            const { first = 10, ...rest } = args
```
erikdstock, line 218: Fetching a default of 10 watched sale artworks from gravity. We define this in the stitching file rather than our local schema because the new `Lot` type relies on a causality type and we cannot reference it until the causality schema has been stitched in - for more context and approach see #2925.

How this resolver works:
Watched lots are determined by gravity, their states are fetched from causality, and they are combined, filtering out any sale artworks with missing lot states (this should never happen in production but is common in staging due to the presence of copied data that has not been synced to causality).

See all of this logic in the following lines of this resolver function.

src/lib/stitching/causality/stitching.ts
```diff
@@ -209,29 +212,94 @@ export const causalityStitchingEnvironment = ({
...
+                    ),
+                  },
+                },
+              ]
+            }, [])
 
             return {
-              totalCount: watchedSaleArtworks.length,
-              edges: nodes.map((node) => ({ node })),
+              totalCount,
```
erikdstock, line 301: This minimal relay connection interface may behave unpredictably due to the lots + total count being derived from gravity but filtered based on available causality lot states. I'm not sure if there are any other gotchas to worry about here.

erikdstock (commented):
Some comments and concerns

erikdstock:
Adding some simple error forwarding when (for some reason) a request to causality fails:

image

erikdstock:
So I've pushed up some final changes and think this is ready to merge. I considered removing the new causality loader since stitching should work for most cases but after finishing found it added complexity to introduce a new pattern for setting up the client.

My assumption is that throwing an error as I am now doing will cause sentry to receive these errors as well - so I did not do anything special for sentry reporting.

I did not add tests but could by moving the entire resolver function into its own file.

src/schema/v2/lot.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+  },
+})
+
+/**
+ * Fetch relevant sale artworks and use them to fetch lot states
+ * from causality. Then pass the lot states through to the stitching layer
+ * via the graphql request context.
+ */
+const resolveWatchedLotConnection = async (_parent, args, context) => {
```
erikdstock, line 43: Lot resolution and stitching code is all going in this file, with the relevant pieces brought into either the local schema or the stitched causality extension schema later. This keeps highly coupled code in the same file and reduces noise in confusing stitching files. This is the resolver...

src/schema/v2/lot.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+/**
+ * An extension schema to add the lot type at the causality stitching stage.
+ */
+export const stitchingLotExtensionSchema = gql`
+  # A unified auction lot with data from our auctions bidding engine.
+  extend type Lot {
+    # The current auction state of the lot.
+    lot: AuctionsLotState!
```
erikdstock, line 143: Here we prepare to extend our Lot type with the `AuctionsLotState` type, which is not yet available to use in the more typical declarative sense at the top of this file, but will be available in causality/stitching.ts.

src/schema/v2/lot.ts
```diff
@@ -0,0 +1,179 @@
...
+  Lot: {
+    lot: {
+      resolve: (root, _args, context, _info) => {
+        const { internalID } = root
+
+        // resolve lot if available via context (eg watchedLotConnection resolver)
+        const lotState = context.lotDataMap?.[internalID]
+        if (lotState) {
+          return lotState
+        }
```
erikdstock, line 162: and here we complete the loop by resolving that Lot.lot field using the context api (again in causality/stitching.ts)

damassi:
@erikdstock - since this covers such a critical path (auctions, money) shouldn't there be tests?

src/schema/v2/lot.ts
```diff
@@ -0,0 +1,179 @@
...
+export const auctionLotConnection = connectionWithCursorInfo({
+  nodeType: LotType,
+})
+
+export const watchedLotConnection = {
+  description: "A list of lots a user is watching",
+  type: auctionLotConnection.connectionType,
+  args: pageable(),
+  resolve: resolveWatchedLotConnection,
+}
```
erikdstock, line 179: And here is a `watchedLotConnection` ready to be exported to our local schema (as `Me.watchedLotConnection`)

src/lib/stitching/causality/stitching.ts
```diff
@@ -24,37 +28,15 @@ const resolveLotCentsFieldToMoney = (centsField) => {
...
-
-      # An edge in a connection.
-      type LotEdge {
-        # A cursor for use in pagination
-        cursor: String!
-
-        # The item at the end of the edge
-        node: Lot
-      }
+      ${version === 2 ? stitchingLotExtensionSchema : ""}
```
damassi, line 39: We don't need to worry about v2 / v1 stuff if we're only writing code for v2 schema

src/lib/stitching/causality/stitching.ts
```diff
@@ -86,19 +61,7 @@ export const causalityStitchingEnvironment = ({
...
-        saleArtwork: {
-          resolve: ({ saleArtwork }) => saleArtwork,
-        },
-        lot: {
-          resolve: ({ lot }) => {
-            return lot
-          },
-        },
-      },
+      ...(version === 2 ? stitchedCausalityLotResolver : {}),
```
damassi, line 64: ditto

erikdstock, line 64: simplified addition to our stitched resolver.

src/schema/v2/lot.ts
```diff
@@ -0,0 +1,179 @@
...
+    first,
+    ...rest,
+  }
+  const params = convertConnectionArgsToGravityArgs(connectionOptions)
+  delete params.page
+
+  const { body, headers } = await saleArtworksAllLoader(params)
+  const watchedSaleArtworks: any[] = body
+  const totalCount = parseInt(headers["x-total-count"] || "0", 10)
+  const saleArtworkIds = watchedSaleArtworks.map((sa) => sa._id)
```
damassi, line 60: (non-blocking tip) - This line is a pretty contrived example since its a one-liner, but have noticed this elsewhere. It's much better to be verbose and use real words with variables, versus abbreviate, as abbreviations make it hard to reason about code through time, especially when another dev picks it up. 

```ts
const saleArtworkIds = watchedSaleArtworks.map((saleArtwork) => saleArtwork._id)
```

src/lib/stitching/causality/stitching.ts
```diff
@@ -24,37 +28,15 @@ const resolveLotCentsFieldToMoney = (centsField) => {
...
-
-      # An edge in a connection.
-      type LotEdge {
-        # A cursor for use in pagination
-        cursor: String!
-
-        # The item at the end of the edge
-        node: Lot
-      }
+      ${version === 2 ? stitchedCausalityLotExtensionSchema : ""}
```
erikdstock, line 39: simplified addition to our extension schema.

erikdstock (commented):
some final notes on new approach.

src/schema/v2/lot.ts
```diff
@@ -0,0 +1,179 @@
...
+  }).then((res: FetchResponse) => res.json())
+
+  const { data: causalityData, errors: causalityErrors } = lotStatesResponse
+
+  // If the causality request failed for some reason, throw its errors.
+  if (causalityErrors) {
+    const errors = causalityErrors.reduce((acc, ce) => {
+      return acc + " " + ce["message"]
+    }, "From causality: ")
+    throw new GraphQLError(errors)
```
damassi, line 107: Is this necessary? 

As I understand things, if a loader fails in any way it rejects the promise and then the error handling built up within MP takes it and processes it accordingly.

src/schema/v2/lot.ts
erikdstock: yeah, sorry. this is a common variable naming pattern in scala but maybe scala is not the best language to take naming patterns from.

src/lib/stitching/causality/stitching.ts
erikdstock: This almost definitely causes issues if it is left off because we will otherwise be extending/resolving types that do not exist on v1.

src/schema/v2/sale_artwork.ts
```diff
@@ -92,7 +92,13 @@ export const SaleArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
       ...SlugAndInternalIDFields,
       cached,
-      artwork: { type: Artwork.type, resolve: ({ artwork }) => artwork },
+      artwork: {
+        type: Artwork.type,
+        resolve: (root) => {
+          const { artwork } = root
+          return artwork
+        },
+      },
```
damassi: This file probably doesn't need to change

erikdstock:
> @erikdstock - since this covers such a critical path (auctions, money) shouldn't there be tests?

@damassi are you free to pair on tests later today? I'm not super versed in causality but would be happy to work on tests with someone.

src/schema/v2/sale_artwork.ts
erikdstock: yep, cleaned up

src/schema/v2/lot.ts
erikdstock: the problem is that i am using `node-fetch` with a graphql endpoint-  the request succeeds with an `errors` array + empty `data`.

src/lib/stitching/causality/stitching.ts
damassi: Since we're not using V1 schema anymore, this shouldn't matter. Check out stitching for [convection](https://github.com/artsy/metaphysics/tree/master/src/lib/stitching/convection) or [gravity](https://github.com/artsy/metaphysics/tree/master/src/lib/stitching/gravity) as an example.

src/schema/v2/lot.ts
erikdstock: however i could move the throwing code into the causality loader I wrote so it isn't just returning a fetch `Response` ... 🤔

damassi:
For sure, if @sepans isn't available I can def help out! 

I don't mean to belabor the point, but I do want to call out that as far as tests go: whether one understands how to write them or not we should always be sure to recognize code context. Our auctions code, which intersects with large amounts of money, is arguably the most sensitive at Artsy and there shouldn't ever be a point where a PR like this gets merged without full coverage, caveat or not. We have the #dev-help channel, and plenty of folks who would be willing to help explain / pair. 

So yeah, let me know if someone on Purchase can't help and and we can pair.

src/lib/stitching/causality/stitching.ts
damassi: This comment still stands: https://github.com/artsy/metaphysics/pull/2928#discussion_r564747657

src/schema/v2/lot.ts
damassi: > the request succeeds with an errors array + empty data

This is idiomatic GraphQL, why break out of this pattern?

src/lib/stitching/causality/stitching.ts
mzikherman: Yea @damassi , by default we still apply the stitching and transformations to V1 and V2. You need to manually do this to avoid updating V1 schema. We do this in several places throughout our stitching infra, and prob will be happening more as future code is likely to refer or extend types which themselves don't even exist in V1.

src/schema/v2/lot.ts
```diff
@@ -0,0 +1,179 @@
...
+import { connectionWithCursorInfo } from "./fields/pagination"
+import gql from "lib/gql"
+import { convertConnectionArgsToGravityArgs } from "lib/helpers"
+import { connectionFromArray } from "graphql-relay"
+
+export const LotType = new GraphQLObjectType<any, ResolverContext>({
+  name: "Lot",
+  description:
+    "A lot in an auction containing merged Sale artwork and Lot state data.",
+  interfaces: () => {
```
mzikherman, line 19: Do you need to fetch a lot via the `node(id: ...` structure? If not, I'd leave this out until needed.

src/schema/v2/lot.ts
mzikherman: #minor i would name this `watchedLotConnectionResolver` to stick more with the pattern (resolver at the end).

src/schema/v2/lot.ts
mzikherman: @damassi I think the issue is that, the loader here will actually never fail, unless Causality is itself down. So if you're 'manually' calling a GraphQL endpoint in this way, I think you'd need to also 'manually' see if there are unrecoverable `errors` present in _that_ response, and bubble those up.

mzikherman (approved):
Looks good, but tests would definitely be 👍

src/lib/stitching/causality/stitching.ts
damassi: 👍 I'm just a bit unsure why, for example, convection stitching didn't include this. We're only stitching stuff into v2, there was no stitching in v1, same with Gravity, but I could be wrong about this.

mzikherman: It's definitely stitched in both places. Check out the [V1 Schema](https://github.com/artsy/metaphysics/blob/54d603e12c745ff67116c65107d307f1ac06f3d8/_schema.graphql), lots of mentions of consignment stuff

src/schema/v2/lot.ts
erikdstock: I know this is resolved but just to follow up here I moved this error array handling into the loader itself so we are no longer dealing with a 'fetch' response

src/schema/v2/lot.ts
```diff
@@ -95,41 +95,31 @@ const resolveWatchedLotConnection = async (_parent, args, context) => {
...
+  }, {})
+
+  // place lotData in context for retrieval at stitching stage
+  context.lotDataMap = lotDataMap
+
+  // create nodes from sale artworks
+  const availableLotDataIds = Object.keys(lotDataMap)
+  const nodes = watchedSaleArtworks.reduce((acc, saleArtwork) => {
+    if (!availableLotDataIds.find((id) => id === saleArtwork._id)) {
+      console.warn(`lot state for ${saleArtwork._id} not found - skipping`)
```
damassi, line 113: If this log will remain in prod it would be helpful to add a breadcrumb back to the location:
```ts
console.warn('[metaphysics @ schema/v2/lot] Warning: lots state for ...) 
```

damassi (approved):
Thanks for adding tests 👍 This looks good to go from my side.

src/schema/v2/lot.ts
erikdstock: I'll do this for now but if there is a better way to log/track things like this I'm interested to hear it. This condition should be relatively common in staging and never (as far as i know) in production.

damassi: You can use this one, but i'd still keep a bread crumb - they're always useful: https://github.com/damassi/metaphysics/blob/master/src/lib/loggers.ts#L3

src/schema/v2/lot.ts
mzikherman: Relay generally just needs an `id` field (don't need to specify any interface)

src/lib/stitching/causality/__tests__/stitching.test.ts
```diff
@@ -156,6 +159,111 @@ describe("causality/stitching", () => {
...
+                  }
+                }
+              }
+            }
+          }
+        }
+      `
+
+      // stub out the loaders necessary to assemble this data
+      const context = {
```
mzikherman, line 194: Nice!

src/schema/v2/lot.ts
```diff
@@ -0,0 +1,180 @@
...
+            internalID
+          }
+        }
+      `,
+      resolve: (root, _args, context, _info) => {
+        const {
+          saleArtwork: { internalID },
+        } = root
+
+        // resolve lot if available via context (eg watchedLotConnection resolver)
```
erikdstock, line 159:
```suggestion
        // resolve lot if available via context (see watchedLotConnectionResolver)
```

PR #2929, opened 2021-01-19
add pricePaid to artworks

bhoggard:
Jira ticket: https://artsyproduct.atlassian.net/browse/CX-945

This is my first time using the `Money` type, so I need help figuring out why `display` isn't working.

Related Gravity PR: https://github.com/artsy/gravity/pull/13766

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -48,7 +48,7 @@ import ArtworkMedium from "schema/v2/artwork/artworkMedium"
 // Mapping of category ids to MediumType values
 import artworkMediums from "lib/artworkMediums"
 import { LotStandingType } from "../me/lot_standing"
-import { amount, symbolFromCurrencyCode } from "schema/v2/fields/money"
+import { amount, Money, symbolFromCurrencyCode } from "schema/v2/fields/money"
 import { capitalizeFirstCharacter } from "lib/helpers"
 import { ResolverContext } from "types/graphql"
 import { listPrice } from "schema/v2/fields/listPrice"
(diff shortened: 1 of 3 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -731,6 +731,17 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
         },
       },
+      pricePaid: {
+        type: Money,
+        description:
+          "The price paid for the artwork in a user's 'my collection'",
+        resolve: (artwork) => {
+          return {
+            cents: artwork.price_paid_cents,
+            currency: artwork.price_paid_currency || "USD",
```
bhoggard, line 741: I'm not sure the `||` part is needed here. If we fix the `display` problem, it may not be needed.

pepopowitz:
What do you mean when you say "display is not working"? Is it coming back empty? 

I can't spot anything immediately that looks problematic. I'm happy to pair on this tomorrow to debug it, though.

bhoggard:
It's nil. Yes, let's pair tomorrow.

zephraph:
Hey @bhoggard, so here's the issue: 

https://github.com/artsy/metaphysics/blob/d28cf3fed4a0e0f8972c32e74f3233fa272ee97f/src/schema/v2/fields/money.ts#L154+L157

The money type doesn't actually provide a resolver for `display`. Given that display isn't returned here there's nothing to return. Now _why_ there's no resolver is a good question. Given how it's defined if you don't manually pass a display value then it will be nil. 

cc @starsirius, any idea why we don't have a mechanism to build up the currency display? 

Here's one usage that uses a helper to help build that out: https://github.com/artsy/metaphysics/blob/d28cf3fed4a0e0f8972c32e74f3233fa272ee97f/src/lib/stitching/causality/stitching.ts#L19

In another place we have it manually defined: https://github.com/artsy/metaphysics/blob/d28cf3fed4a0e0f8972c32e74f3233fa272ee97f/src/schema/v2/auction_result.ts#L156-L191

starsirius:
Looking at the [initial PR](https://github.com/artsy/metaphysics/pull/2015/files#r338195024), I think it's... not yet implemented. 😅  Maybe it's time we add a default display resolver?

ds300:
since `display` is not used yet, I wonder whether we should use the `amount` helper to define it: https://github.com/artsy/metaphysics/blob/9562c6309f06a161aac8e353424b2cd1a4490835/src/schema/v2/fields/money.ts#L28 since this would give us parameters for formatting in different locales.

PR #2939, opened 2021-01-22
feat(geodata): exposes cities field; all cities queryable via slug

dzucconi:
So this is being done to support the [`/shows`](https://www.artsy.net/shows) app. We need to be able to query the list of featured cities, all cities, as well as have landing pages for every city with the available shows.


src/lib/loaders/loaders_without_authentication/geodata.ts
```diff
@@ -0,0 +1,21 @@
...
+    geodataLoaderWithoutAuthenticationFactory: geodataLoader,
+  } = factories(opts)
+
+  return {
+    geodataCitiesLoader: () =>
+      geodataLoader<Response>("partner-cities/cities.json")().then(compact),
+    geodataFeaturedCitiesLoader: () =>
+      geodataLoader<Response>("partner-cities/featured-cities.json")().then(
+        compact
+      ),
```
dzucconi, line 19: Awkward way to do the compaction here. Not sure why this data might have nulls in it but:

anandaroop (commented):
Exposing Gravity's city data in MP makes sense to me 👍 

Having it replace the local source-of-truth for the six City Guide cities... would be mostly fine, but there is one case to consider, noted below.

> For example: New York's center goes from [40.713, -74.006] to [40.71, -74.01], which just moves it about 6 blocks uptown. Why/how do these values change, anyway?

They're both based on Google geocoder results, just with different precision — 3 significant digits vs 2.  

- Gravity keeps 2 sig digs, because that [helps with the deduping](https://github.com/artsy/gravity/pull/11301) of the master list of cities that it computes every night.

- MP keeps 3 sig digs in its City Guide master list, because that helps with positioning of the map viewport in the City Guide UX — specifically [we had to nudge Hong Kong](https://github.com/artsy/metaphysics/pull/1595) 

If we can make sure that Hong Kong remains acceptable after unifying the source of truth for the `city` field, then I think all is good. Happy to pair on this.

dblandin:
__non-blocking__: One small naming thought: could the `allCities` and `featuredCities` fields collapse into a single `cities` root field with a `featured: Boolean` argument?

```graphql
{
  cities(featured: true) {
    ...City
  }
}
```

I realize that these sets are backed by two different JSON documents but we might be able to switch the source based on a boolean argument rather than maintain two different root fields.

dzucconi:
@dblandin definitely good idea

src/schema/v2/city/index.ts
```diff
@@ -23,141 +20,159 @@ import {
...
-          type: GraphQLInt,
-          description:
-            "Only used when status is CLOSING_SOON or UPCOMING. Number of days used to filter upcoming and closing soon shows",
-        },
-        includeStubShows: {
-          type: GraphQLBoolean,
-          description: "Whether to include local discovery stubs",
+  fields: () => {
+    const { ShowsConnection } = require("../show")
+    const { fairConnection } = require("../fair")
```
anandaroop, line 50: **non-blocking:** Curious why these imports changed to this style?

.env.test
```diff
@@ -8,6 +8,7 @@ EMBEDLY_ENDPOINT=https://i.embed.ly.test/1/display
 GALAXY_API_BASE=https://galaxy-staging-herokuapp.com
 GEMINI_API_BASE=https://media.artsy.net
 GEMINI_ENDPOINT=https://gemini.cloudfront.test
+GEODATA_API_BASE=http://artsy-geodata.s3-website-us-east-1.amazonaws.com
```
anandaroop, line 11: I left this breadcrumb, now that this data is being leaned on more:
https://artsy-geodata.s3.amazonaws.com/partner-cities/README.txt

anandaroop (approved):
Looks good 👍

src/schema/v2/city/index.ts
dzucconi: Apparently dependencies get resolved in a different order when running tests (?) and there was some circular dependencies when running the specs, so had to thunk/defer the require in order to get anything to load.

.env.test
dzucconi: Thanks!

PR #2941, opened 2021-01-26
Add checklistItem type to partner level

jpotts244:
Co-authored-by: mdole <mdole7@gmail.com>

Addresses: https://artsyproduct.atlassian.net/browse/PX-3631

Development work to allow us to more easily query for artworks that will eventually populate a partner's home checklist in CMS

Eg of query structure:
Screen Shot 2021-01-26 at 9 46 16 AM

### DIFF
--- a/src/schema/v2/checklistItems.ts
+++ b/src/schema/v2/checklistItems.ts
@@ -1,0 +1,20 @@
+import { artworkConnection } from "./artwork"
+import { GraphQLObjectType } from "graphql"
+import { ResolverContext } from "types/graphql"
+import { pageable } from "relay-cursor-paging"
+import { artworksArgs, artworksConnectionResolver } from "./partner"
+
+export const ChecklistItemsType = new GraphQLObjectType<any, ResolverContext>({
+  name: "ChecklistItemsType",
+  fields: () => {
+    return {
+      artworkItemsConnection: {
+        description:
+          "Artworks missing priority metadata for a partner's checklist",
+        type: artworkConnection.connectionType,
+        args: pageable(artworksArgs),
+        resolve: artworksConnectionResolver,
+      },
+    }
+  },
+})
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2285,6 +2285,19 @@ type CaptureHoldPayload {
   holdOrError: InventoryHoldOrErrorUnion!
 }
 
+type ChecklistItemsType {
+  # Artworks missing priority metadata for a partner's checklist
+  artworkItemsConnection(
+    after: String
+    before: String
+    exclude: [String]
+    first: Int
+    forSale: Boolean
+    last: Int
+    sort: ArtworkSorts
+  ): ArtworkConnection
+}
+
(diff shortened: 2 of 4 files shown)


src/schema/v2/checklistItems.ts
```diff
@@ -0,0 +1,20 @@
...
+import { GraphQLObjectType } from "graphql"
+import { ResolverContext } from "types/graphql"
+import { pageable } from "relay-cursor-paging"
+import { artworksArgs, artworksConnectionResolver } from "./partner"
+
+export const ChecklistItemsType = new GraphQLObjectType<any, ResolverContext>({
+  name: "ChecklistItemsType",
+  fields: () => {
+    return {
+      artworkItemsConnection: {
```
jpotts244, line 11: Re Oksana: possibly renaming re thinking: artworks missing metadata connection

src/schema/v2/checklistItems.ts
```diff
@@ -0,0 +1,20 @@
+import { artworkConnection } from "./artwork"
+import { GraphQLObjectType } from "graphql"
+import { ResolverContext } from "types/graphql"
+import { pageable } from "relay-cursor-paging"
+import { artworksArgs, artworksConnectionResolver } from "./partner"
+
+export const ChecklistItemsType = new GraphQLObjectType<any, ResolverContext>({
```
sweir27, line 7: My comments were similar to what @oxaudo mentioned in knowledge share! To reiterate:
- We could either expose fields here that are specific to the checklist (i.e. `missingMetadataArtworksConnection`) _or_ make it more generic (`artworksConnection`) to allow for different permutations. This current naming feels somewhere in between. 😄 
- If we go with the latter, more generic version, I'm not sure I understand the need to have this `checklistItems` type at all, if `artworksConnection` already exists on the partner level, as it would just be nesting an existing field without adding anything additional.

FWIW, it would not be too difficult to change this API in the future given it won't be installed on any kind of native device. So it might be prudent to do the simplest thing here and then make it more generic when we have another use-case.

Another thought on `ChecklistItemsType`-- right now, it's just a field that exposes other fields, but the client still has to declare _which_ fields it wants returned. That means the logic for _which_ checklist items to expose or query for still lives on the client-- I think that's why I'm having trouble understanding what it's adding in its current form.

Another architecture (not suggesting necessarily for right now, but a potential direction!) could be to have metaphysics hold the logic for which checklist items are returned. In that case, the interface would be slightly more generic and might look something like:
```
partner(id: "blah") {
  checklistItems {
    __typename
    ... on MissingMetadataArtworks {
      artworksConnection {
        .....
      }
    }
    ... on AddToFairShow {
       .....
    }
  }
}
```

You could even filter `checklistItems` by passing in partner type or any other data that might be specific to the client/affect which results are returned, but overall the query would be generic from the client's perspective.

Let me know if I can explain my thinking more!

jpotts244:
Closing for now re feedback from knowledge share and here: https://artsy.slack.com/archives/C9ZJYFDNF/p1611764966206100

PR #2945, opened 2021-01-27
Chore: updates Exchange schema with createInquiryOfferOrderWithArtwork

ashleyjelks:
Updates Exchange schema with latest mutation `createInquiryOfferOrderWithArtwork`

cc @artsy/purchase-devs


starsirius:
You'd need to do the following to update MP schema:

1. Copy the exchange schema to [src/data/exchange.graphql](https://github.com/artsy/metaphysics/blob/master/src/data/exchange.graphql) (instead of the root directory). We'll see the proper diff with the mutation we added in exchange.
2. Update MP schema with `yarn dump:local`. We'll see updates to MP's schema file.

Happy to pair on this if you run into blockers!

ashleyjelks:
> You'd need to do the following to update MP schema:
> 
> 1. Copy the exchange schema to [src/data/exchange.graphql](https://github.com/artsy/metaphysics/blob/master/src/data/exchange.graphql) (instead of the root directory). We'll see the proper diff with the mutation we added in exchange.
> 2. Update MP schema with `yarn dump:local`. We'll see updates to MP's schema file.
> 
> Happy to pair on this if you run into blockers!

Welp, I did that extremely incorrectly! 🙈  Just made the updates. Should be good to go now. Thanks @starsirius !

starsirius (approved):
Looks good to me! Thanks, @ashleyjelks! Feel free to merge when ready!

PR #2947, opened 2021-01-28
Fix casing to allow for filtering artworks by forSale argument

jpotts244:
Co-authored-by: mdole <mdole7@gmail.com>

Fixes a bug where we are trying to grab the value of `args.for_sale`, the thing is this will always return `undefined`, and not allow us to properly filter by works that are `forSale=true/false`

Bug - artworks regardless of `isForSale` status are returned in the query response:
Screen Shot 2021-01-28 at 2 32 58 PM

Fix - artworks where only `isForSale=true` are returned in the query response:
Screen Shot 2021-01-28 at 2 33 19 PM

### DIFF
--- a/src/schema/v2/partner.ts
+++ b/src/schema/v2/partner.ts
@@ -121,7 +121,7 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
             page,
             size,
             sort: args.sort,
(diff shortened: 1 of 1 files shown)


mdole (approved):
Let's ship it!!

PR #2951, opened 2021-01-29
feat: add percentage performance over mid estimate to auction result

MounirDhahri:
This PR resolves: https://artsyproduct.atlassian.net/browse/CX-955

**Description:**
- Add percentage performance over mid estimate to auction result

TODO: 
- [x] Make sure the formula is correct

### DIFF
--- a/src/schema/v2/auction_result.ts
+++ b/src/schema/v2/auction_result.ts
@@ -141,6 +141,39 @@ const AuctionResultType = new GraphQLObjectType<any, ResolverContext>({
         }
       },
     },
+    performance: {
+      type: new GraphQLObjectType({
+        name: "AuctionLotPerformance",
+        fields: {
+          mid: {
+            type: GraphQLString,
+            description: "Percentage performance over mid-estimate",
+            resolve: ({
+              hammer_price_cents,
+              high_estimate_cents,
+              low_estimate_cents,
+            }) => {
+              if (
+                hammer_price_cents &&
+                high_estimate_cents &&
+                low_estimate_cents
+              ) {
+                return (
+                  Math.round(
+                    ((hammer_price_cents -
+                      (low_estimate_cents + high_estimate_cents) / 2) /
+                      hammer_price_cents) *
+                      100
+                  ) + "%"
+                )
+              }
+              return null
+            },
+          },
+        },
+      }),
+      resolve: (lot) => lot,
+    },
     estimate: {
       type: new GraphQLObjectType<any, ResolverContext>({
         name: "AuctionLotEstimate",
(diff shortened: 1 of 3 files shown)


src/schema/v2/auction_result.ts
```diff
@@ -141,6 +141,39 @@ const AuctionResultType = new GraphQLObjectType<any, ResolverContext>({
...
+              if (
+                hammer_price_cents &&
+                high_estimate_cents &&
+                low_estimate_cents
+              ) {
+                return (
+                  Math.round(
+                    ((hammer_price_cents -
+                      (low_estimate_cents + high_estimate_cents) / 2) /
+                      hammer_price_cents) *
```
pepopowitz, line 165: I'd expect the mid-estimate to be the denominator here, since we're calculating the percentage _over mid estimate_. 

If the estimate was between 10 and 30, the mid-estimate was 20. If the realized price was 100, I'd expect that to be significantly over mid-estimate (500%). With the current calculation, with 100 in the denominator, the percentage would instead come out quite muted.

src/schema/v2/auction_result.ts
```diff
@@ -141,6 +141,39 @@ const AuctionResultType = new GraphQLObjectType<any, ResolverContext>({
...
+                      100
+                  ) + "%"
```
pepopowitz, line 167: Also, I think intermediate named variables could be helpful here from a readability/maintenance perspective. 

```suggestion
                const midEstimate = (low_estimate_cents + high_estimate_cents) / 2)
                const delta = hammer_price_cents - midEstimate
                return (
                  Math.round(
                    (delta / midEstimate) *
                      100
                  ) + "%"
```

(My formatting here is probably not correct)

pepopowitz (changes requested):
I think the denominator is incorrect in this calculation.

src/schema/v2/auction_result.ts
MounirDhahri: thanks Steve

src/schema/v2/auction_result.ts
```diff
@@ -141,6 +141,35 @@ const AuctionResultType = new GraphQLObjectType<any, ResolverContext>({
...
+          mid: {
+            type: GraphQLString,
+            description: "Percentage performance over mid-estimate",
+            resolve: ({
+              hammer_price_cents,
+              high_estimate_cents,
+              low_estimate_cents,
+            }) => {
+              if (
+                hammer_price_cents &&
```
brainbicycle, line 157: based on conversation in slack think we want to update this to use `realized_price_cents` rather than `hammer_price_cents` 
https://artsy.slack.com/archives/C01B2P6LJUU/p1611936317214600

brainbicycle (approved):

brainbicycle:
LGTM, merge when ready!

PR #2952, opened 2021-01-29
feat: Adds offerable_from_inquiry flag for Inquiry Checkout

jo-rs:
This PR resolves [\[PURCHASE-2413\]](https://artsyproduct.atlassian.net/browse/PURCHASE-2413)

Adds `offerable_from_inquiry` flag for `EditionSet` and `Artwork` to be used with Inquiry Checkout. Relates to [this](https://github.com/artsy/gravity/pull/13786) gravity PR.

[PURCHASE-2413]: https://artsyproduct.atlassian.net/browse/PURCHASE-2413

### DIFF
--- a/src/schema/v2/sellable.ts
+++ b/src/schema/v2/sellable.ts
@@ -21,6 +21,12 @@ export const Sellable = new GraphQLInterfaceType({
       description: "Whether a user can make an offer on the work",
       resolve: ({ is_offerable }) => is_offerable,
     },
+    isOfferableFromInquiry: {
+      type: GraphQLBoolean,
+      description:
+        "Whether a user can make an offer on the work through inquiry",
+      resolve: ({ is_offerable_from_inquiry }) => is_offerable_from_inquiry,
+    },
     isForSale: {
       type: GraphQLBoolean,
       resolve: ({ is_for_sale }) => is_for_sale,
--- a/src/types/gravity/artworkResponse.d.ts
+++ b/src/types/gravity/artworkResponse.d.ts
@@ -39,6 +39,7 @@ export interface GravityArtwork {
   inquireable: boolean
   acquireable: boolean
   offerable: boolean
+  offerable_from_inquiry: boolean
   published_at: string
   can_share: boolean
   can_share_image: boolean
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1351,6 +1351,9 @@ type Artwork implements Node & Searchable & Sellable {
 
   # Whether a user can make an offer on a work
   isOfferable: Boolean
+
+  # Whether a user can make an offer on the work through inquiry
+  isOfferableFromInquiry: Boolean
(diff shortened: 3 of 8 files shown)


src/schema/v2/sellable.ts
```diff
@@ -21,6 +21,12 @@ export const Sellable = new GraphQLInterfaceType({
       description: "Whether a user can make an offer on the work",
       resolve: ({ is_offerable }) => is_offerable,
     },
+    isOfferableFromInquiry: {
+      type: GraphQLBoolean,
+      description:
+        "Whether a user can make an offer on the work through inquiry",
+      resolve: ({ is_offerable_from_inquiry }) => is_offerable_from_inquiry,
```
starsirius, line 28: I see this is following the pattern above but I wonder if we want the field from Gravity `offerable_from_inquiry` instead of `is_offerable_from_inquiry`, like the [artwork type](https://github.com/artsy/metaphysics/pull/2952/files#diff-54b84e0c164089fc202e3439d8ded89ab4925442fb613f4e36a9c22acdb66cb8R397).

And actually I am curious if the resolver in an interface is used. 🤔

starsirius:
I asked on Slack to get clarity for the [resolver on an interface field](https://github.com/artsy/metaphysics/pull/2952#discussion_r568671039). Going to merge this for now. Thanks, @jo-rs! 🥇

src/schema/v2/sellable.ts
sepans: 🐈 The test passes so it should work just fine but does `... => is_offerable_from_inquiry` need to be changed to  `resolve: ... => offerable_from_inquiry`?

src/types/gravity/artworkResponse.d.ts
```diff
@@ -39,6 +39,7 @@ export interface GravityArtwork {
   inquireable: boolean
   acquireable: boolean
   offerable: boolean
+  offerable_from_inquiry: boolean
```
anandaroop, line 42: Hey @jo-rs, checking about something here —

I have an open PR (https://github.com/artsy/metaphysics/pull/2964) where I actually delete this `artworkResponse.d.ts` file because as far as I can tell it is unused and no longer in sync with Gravity.

So I wanted to confirm whether we know for a fact this file is in use? Or if this was more of a global search-and-replace kinda thing? Either way my PR will need an update — thanks!

jo-rs: I was just following form so: not sure if it's in use 🙁

starsirius: I'm not sure either if it's used. I'm 👍  to remove it if it's already outdated and wait for 😱 , if any.

anandaroop: Yeah, seems safe to delete. I'll rebase and do so 👍

Thanks!

PR #2953, opened 2021-01-29
feat(show): exposes online exclusive flag

dzucconi:
#trivial


src/schema/v2/show.ts
```diff
@@ -392,6 +392,13 @@ export const ShowType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLBoolean,
         resolve: ({ fair }) => isExisty(fair),
       },
+      isOnlineExclusive: {
+        description: "Does the show exist soley online",
+        type: new GraphQLNonNull(GraphQLBoolean),
```
dzucconi, line 397: At some point I need to convert *all* booleans to non-nullability

PR #2963, opened 2021-02-01
feat(shows): expands args on top level shows connection

dzucconi:
For now this just cleans up the top-level showsConnection and adds some of the supported options.

The query criteria for online exclusives according to Force is:

```
  has_location: false,
  displayable: true,
  at_a_fair: false,
```

------

As an aside it does seem like we might want to actually model this or at least remove the `at_a_fair: false` constraint, considering that we have online exclusive fairs on the site.

### DIFF
--- a/src/schema/v2/shows.ts
+++ b/src/schema/v2/shows.ts
@@ -1,14 +1,18 @@
-import { GraphQLFieldConfig, GraphQLList, GraphQLString } from "graphql"
+import {
+  GraphQLBoolean,
+  GraphQLFieldConfig,
+  GraphQLList,
+  GraphQLString,
+} from "graphql"
 import { ResolverContext } from "types/graphql"
 import { createPageCursors } from "./fields/pagination"
 import { ShowsConnection } from "./show"
-import { pageable } from "relay-cursor-paging"
+import { CursorPageable, pageable } from "relay-cursor-paging"
 import { convertConnectionArgsToGravityArgs } from "lib/helpers"
 import { connectionFromArray } from "graphql-relay"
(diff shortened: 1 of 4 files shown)


src/lib/loaders/loaders_without_authentication/gravity.ts
```diff
@@ -210,7 +210,7 @@ export default (opts) => {
     setLoader: gravityLoader((id) => `set/${id}`),
     setsLoader: gravityLoader("sets", {}, { headers: true }),
     showLoader: gravityLoader((id) => `show/${id}`),
-    showsLoader: gravityLoader("shows"),
+    showsLoader: gravityLoader("shows", {}, { headers: true }),
```
anandaroop, line 213: Any reason to not use the `showsWithHeadersLoader` that already exists on L214?

src/schema/v2/shows.ts
```diff
@@ -1,37 +1,72 @@
...
-      id: _options.ids,
+  resolve: async (
+    _root,
+    args: {
+      ids?: string[]
+      hasLocation?: boolean
+      sort?: TShowSorts
+      displayable?: boolean
+      atAFair?: boolean
+    } & CursorPageable,
```
anandaroop, line 45: 🙏

anandaroop (commented):
Just one q, otherwise lgtm.

src/lib/loaders/loaders_without_authentication/gravity.ts
dzucconi: lol, guess I missed it. Probably best to just pull off 248a1f3 in that case and swap to use that

PR #2966, opened 2021-02-01
feat: Display info from new COA checkboxes to user on artwork page

annacarey:
Added new copy about certificate of authentication to provide more detail around COA—whether the COA was issued by an authorized authenticating body and/or by the gallery. These correspond to new checkbox inputs in the artwork form in CMS that allows galleries to provide this further detail.

Case for w/ issued by gallery checked:
Screen Shot 2021-02-02 at 1 34 55 PM
Case for w/ issued by authenticating body checked:
Screen Shot 2021-02-02 at 1 29 24 PM
Case for w/ both checked:
Screen Shot 2021-02-03 at 2 36 03 PM

[Jira ticket](https://artsyproduct.atlassian.net/browse/PX-3608)

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -2331,7 +2331,8 @@ describe("Artwork type", () => {
       return runQuery(query, context).then((data) => {
         expect(data.artwork.certificateOfAuthenticity).toEqual({
           label: "Certificate of authenticity",
-          details: "Included (issued by authorized authenticating body and gallery)",
+          details:
+            "Included (one issued by gallery; one issued by authorized authenticating body)",
         })
         expect(data.artwork.hasCertificateOfAuthenticity).toBe(true)
       })
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -2297,6 +2297,40 @@ describe("Artwork type", () => {
         expect(data.artwork.hasCertificateOfAuthenticity).toBe(true)
       })
     })
+    it("is set to proper object when ertificate_of_authenticity and coa_from_authenticating_body are true", () => {
+      artwork.certificate_of_authenticity = true
+      artwork.coa_by_authenticating_body = true
+      return runQuery(query, context).then((data) => {
+        expect(data.artwork.certificateOfAuthenticity).toEqual({
+          label: "Certificate of authenticity",
+          details: "Included (issued by authorized authenticating body)",
+        })
+        expect(data.artwork.hasCertificateOfAuthenticity).toBe(true)
+      })
+    })
+    it("is set to proper when object certificate_of_authenticity and coa_from_gallery are true", () => {
+      artwork.certificate_of_authenticity = true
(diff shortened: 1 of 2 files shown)


src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -2297,6 +2297,40 @@ describe("Artwork type", () => {
         expect(data.artwork.hasCertificateOfAuthenticity).toBe(true)
       })
     })
+    it("is set to proper object when ertificate_of_authenticity and coa_from_authenticating_body are true", () => {
```
sweir27, line 2300: #minor #non-blocking we are totally not consistent about this, but I find test files to be slightly more readable when there's a blank line between subsequent tests/blocks. (Usually when I'm in a file I try to update it!)

Also #superminor typo here `ertificate_of_authenticity`.

sweir27 (approved):
Seems straightforward 👍 from me!

sweir27:
@annacarey it's usually helpful to include a screenshot of the query/response here. I do trust your tests but it's nice to have the confirmation (and those are usually extra helpful down the line for putting the change in context).

src/schema/v2/artwork/__tests__/artwork.test.js
annacarey: OOOH thank you good call on both!

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -2297,6 +2299,45 @@ describe("Artwork type", () => {
...
+      return runQuery(query, context).then((data) => {
+        expect(data.artwork.certificateOfAuthenticity).toEqual({
+          label: "Certificate of authenticity",
+          details: "Included (issued by authorized authenticating body)",
+        })
+        expect(data.artwork.hasCertificateOfAuthenticity).toBe(true)
+      })
+    })
+
+    it("is set to proper when object certificate_of_authenticity and coa_from_gallery are true", () => {
```
oxaudo, line 2315: #superminor: probably should be `to proper object when`....

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -2297,6 +2299,45 @@ describe("Artwork type", () => {
...
+      return runQuery(query, context).then((data) => {
+        expect(data.artwork.certificateOfAuthenticity).toEqual({
+          label: "Certificate of authenticity",
+          details: "Included (issued by gallery)",
+        })
+        expect(data.artwork.hasCertificateOfAuthenticity).toBe(true)
+      })
+    })
+
+    it("is set to proper when object certificate_of_authenticity, coa_from_authenticating_body, and coa_from_gallery are true", () => {
```
oxaudo, line 2327: #superminor: copy paste from above -:)

oxaudo (approved):
Had one super minor note just to say something:)

Please merge when you get a clear from Arielle.

src/schema/v2/artwork/__tests__/artwork.test.js
annacarey: thanks for catching that and taking a look. fixed! just waiting on final A-OK from Arielle and then will merge!

PR #2980, opened 2021-02-05
fix: fix typo in show sort

MounirDhahri:
While testing the featured fairs in the app and in Web, I noticed that the featured fairs are not showing up. When I was to debug the app on staging  I was getting "unavailable show sort mode". After checking the list of available sorts on gravity and comparing it to metaphysics it turned out that we have a typo on one of the sorts. An other thing that I also realized, is that we have a sort mode called `PARTNER_ASC` that was only available on metaphysics but not on gravity, so I removed it.

Screenshot 2021-02-05 at 19 02 51
Screenshot 2021-02-05 at 19 03 17


src/schema/v2/sorts/show_sorts.ts
```diff
@@ -3,11 +3,10 @@ import { GraphQLEnumType } from "graphql"
 const SHOW_SORTS = {
   END_AT_ASC: { value: "end_at" },
   END_AT_DESC: { value: "-end_at" },
-  FEATURED_ASC: { value: "-featured,-start_at" },
-  FEATURED_DESC: { value: "-featured,-end_at" },
+  FEATURED_ASC: { value: "featured" },
+  FEATURED_DESC: { value: "-featured" },
   NAME_ASC: { value: "name" },
   NAME_DESC: { value: "-name" },
-  PARTNER_ASC: { value: "fully_qualified_name" },
```
MounirDhahri, line 10: This is the list of available sorts on gravity.
https://github.com/artsy/gravity/blob/62c498f581ed283ead1a1ba79980e8c759c2d392/app/api/v1/fairs_shows_endpoint.rb#L17

 `fully_qualified_name` doesn't exist there.

brainbicycle (approved):
LGTM, thanks for looking into. I think it makes sense to limit to just fixing the featured sort for now. @dzucconi it looks like the featured sort changes were causing an error when loading fair rail in Eigen, this PR is to just revert those to old values but let us know if we are missing anything here!

dzucconi:
Yuck; apparently we have two different show sorts: https://github.com/artsy/gravity/blob/62c498f581ed283ead1a1ba79980e8c759c2d392/app/models/domain/partner_show.rb#L149-L150 These one's are invalid for other types of shows... 😆  this is fine to merge; will have to untangle this elsewhere.

PR #2988, opened 2021-02-10
Feat: un-deprecate and add params to partner artworksConnection

mdole:
Following up on https://github.com/artsy/gravity/pull/13790, this PR allows us to use the new `imageCountLessThan`, `publishedWithin` (seconds) and `artworkID`. We can thus execute a query like this:

```graphql
{ 
    artworkMetadataChecklistItems: artworksConnection(first: 5 publishedWithin: 7776000 imageCountLessThan:2 artworkID:["5fadb09499a3f8000e1c8afe","5fcfef6acdefca0012db87b2"] exclude:"5fcfef6acdefca0012db87b2") {
      edges {
        node {
          internalID
          title
          image {
            href
          }
          artistNames
          slug
        }
      }
    }
  }
}
```

...and get back artworks that meet all of the criteria in the params. Woohoo!

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -8106,11 +8106,20 @@ type Partner implements Node {
   # A connection of artworks from a Partner.
   artworksConnection(
     after: String
+
+    # Return only artwork(s) included in this list of IDs.
+    artworkIDs: [String]
     before: String
     exclude: [String]
     first: Int
     forSale: Boolean
+
+    # Return artworks with less than x additional_images.
+    imageCountLessThan: Int
     last: Int
+
+    # Return artworks published less than x seconds ago.
+    publishedWithin: Int
     sort: ArtworkSorts
   ): ArtworkConnection
   cached: Int
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -8112,7 +8112,7 @@ type Partner implements Node {
     forSale: Boolean
     last: Int
     sort: ArtworkSorts
-  ): ArtworkConnection @deprecated(reason: "Use `filterArtworksConnection`")
+  ): ArtworkConnection
   cached: Int
   categories: [PartnerCategory]
 
--- a/src/schema/v2/partner.ts
+++ b/src/schema/v2/partner.ts
@@ -33,13 +33,25 @@ import { connectionWithCursorInfo } from "./fields/pagination"
 import { deprecate } from "lib/deprecation"
 
 const artworksArgs: GraphQLFieldConfigArgumentMap = {
-  forSale: {
-    type: GraphQLBoolean,
+  artworkIDs: {
+    type: new GraphQLList(GraphQLString),
+    description: "Return only artwork(s) included in this list of IDs.",
   },
-  sort: ArtworkSorts,
   exclude: {
     type: new GraphQLList(GraphQLString),
   },
+  forSale: {
+    type: GraphQLBoolean,
+  },
+  imageCountLessThan: {
+    type: GraphQLInt,
+    description: "Return artworks with less than x additional_images.",
+  },
+  publishedWithin: {
+    type: GraphQLInt,
+    description: "Return artworks published less than x seconds ago.",
+  },
+  sort: ArtworkSorts,
 }
 
 export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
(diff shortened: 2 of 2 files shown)


_schemaV2.graphql
```diff
@@ -8106,13 +8106,22 @@ type Partner implements Node {
   # A connection of artworks from a Partner.
   artworksConnection(
     after: String
+
+    # Return only artwork(s) included in this list of IDs.
+    artworkID: [String]
```
annacarey, line 8111: should this be `artworkIDs`?

_schemaV2.graphql
```diff
@@ -8106,13 +8106,22 @@ type Partner implements Node {
...
     forSale: Boolean
+
+    # Return artworks with less than x additional_images.
+    imageCountLessThan: Int
     last: Int
+
+    # Return artworks published less than x seconds ago.
+    publishedWithin: Int
     sort: ArtworkSorts
-  ): ArtworkConnection @deprecated(reason: "Use `filterArtworksConnection`")
```
annacarey, line 8115: Nice!

src/schema/v2/partner.ts
```diff
@@ -33,13 +33,25 @@ import { connectionWithCursorInfo } from "./fields/pagination"
 import { deprecate } from "lib/deprecation"
 
 const artworksArgs: GraphQLFieldConfigArgumentMap = {
-  forSale: {
-    type: GraphQLBoolean,
+  artworkID: {
```
annacarey, line 36: same as above re: pluralizing

src/schema/v2/partner.ts
```diff
@@ -109,27 +120,35 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
           )
 
           interface GravityArgs {
+            artwork_id?: string[]
```
annacarey, line 123: `artwork_ids`?

src/schema/v2/partner.ts
```diff
@@ -109,27 +120,35 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
             sort: args.sort,
-            for_sale: args.forSale,
+            total_count: true,
           }
 
           if (args.exclude) {
             gravityArgs.exclude_ids = flatten([args.exclude])
           }
+          if (args.artworkID) {
+            gravityArgs.artwork_id = flatten([args.artworkID])
```
annacarey, line 150: Did you make it singular because that's how it's coming in from gravity?

annacarey (changes requested):
Looks good, just one minor comment/question about pluralizing `artworkIDs`. I like this approach. Seems like it offers us more flexibility around how we define "incomplete" when it comes to metadata.

_schemaV2.graphql
mdole: @annacarey yeah I hear you on this 😅 we did this as `artworkID` because that's what it is in the Gravity endpoint: https://github.com/artsy/gravity/blob/dd5fd7149177b26b138c6d7fd1f27cf3aed3f13d/app/api/v1/partners_artworks_endpoint.rb#L96

That said, we don't _have_ to follow that if we don't want to. My thinking here was to prioritize consistency, but it definitely does read weird. I don't think it's too confusing if this is plural and the gravity version is singular, I'll make that adjustment

src/schema/v2/partner.ts
mdole: this one will stay `artwork_id` though since that's the actual arg we want to pass to Gravity

_schemaV2.graphql
jpotts244: 🤩 

Cool! Ive never un-deprecated a field before in graphql, it is just as easy as removing the `@deprecated` decorator ? Neat

src/schema/v2/partner.ts
jpotts244: Tricky! Maybe an opportunity for a feature friday to pluralize that parameter someday, I would expect it to be dealing with a singular id string as written.

_schemaV2.graphql
mdole: yeah, the `@deprecated` decorator just means "you shouldn't use this" and doesn't really enforce that in any way, so should be as simple as this

jpotts244 (approved):

sweir27:
Woop woop! Nice work, excited to give this a spin. 🙃

PR #2990, opened 2021-02-10
Stitch conversations into orders.

erikdstock:
This PR
~~a) closes off our `commerceCreateInquiryOfferOrderWithArtwork` to calls except through `createInquiryOffer` (could be moved into its own pr)~~ Saving for a move into its own PR in the future.
b) [purchase-2378] Allows us to resolve whether an order (in practice only a CommerceOfferOrder) has an associated conversation.

```
      isInquiryOrder: Boolean!
      # conversation: Conversation # Delaying for now because of difficulties testing and possible flaws in the design pointed out by @sepans 
```

I removed the addition of the `conversation` field because we aren't sure it will work for partners (see comment from @sepans below) but here is a comment from earlier on tests - leaving it in the description even though it no longer applies:
> Unfortunately I had a hard time getting tests to work end-end. At this point i'm testing the positive and negative cases for both new properties but on `conversation` I was unable to get `delegateToSchema` to work with `addMockFunctionsToSchema`. I can go into more details for those interested, but this works as-is and the tests cover most cases (except for that bit of implementation detail which is arguably significant).

cc @artsy/purchase-devs


src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -193,6 +198,7 @@ export const exchangeStitchingEnvironment = ({
               }),
               // how to process the data result at path
               (result) => {
+                console.log("got result", { result })
```
erikdstock: this piece is pretty key - after `delegateToSchema` I consistently logged `result: null` here when using mock functions like:
```
Me: () => {
           conversation: (_root, _params) => {		         //   console.log("Me mock")
            return {
              conversation: () => {
                console.log("conversation mock")
                return {
                  items: [
                    {
                      item: { __typename: "Artwork", title: "Conversation Art" },
                    },
                  ],
                }
              },
            }
          },
```
and
```
Query: () => ({
           me: () => ({
             conversation: {
               items: [
                  {
                    item: { __typename: "Artwork", title: "Conversation Art" },
                  },
                ],
              },
            }),
```

erikdstock (commented):
regarding my other attempts in 7d779, see the comments below. 

The key issue is that I think my working test is actually mocking all calls to return a `Conversation` where as I want to specify that `delegateToSchema` is working and delegating to `me conversation`. Normally I might say this is an implementation detail but the resolver and delegate call are not trivial logic in this case - use of `WrapQuery` to transform the call and so on.

I also tried creating a second `addMockFunctionsToSchema`  and putting all these attempts to mock `me.conversations` there, but passing the `schema` variable (the local schema, i think, because `delegateToSchema` uses `localSchema`) instead of merged schema. Basically I could not get anything to come out the other end.

src/schema/v2/me/conversation/index.ts
```diff
@@ -355,6 +357,8 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
         "The artworks and/or partner shows discussed in the conversation.",
       resolve: (conversation) => {
         const results = []
+        debugger
```
sepans, line 360: 🙈

erikdstock:
```suggestion
```

src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -134,6 +134,73 @@ export const exchangeStitchingEnvironment = ({
...
+      fragment: gql`
+        fragment CommerceOrderConversation on CommerceOfferOrder {
+          impulseConversationId
+        }
+      `,
+      resolve: async (order, _args, context, info) => {
+        const { impulseConversationId } = order
+        if (!impulseConversationId) return null
+
+        return info.mergeInfo.delegateToSchema({
```
sepans, line 159: Aaaah, I just noticed we don't have root-level conversation query so the only way to get it is from `me`. This works well for the collector side but I am wondering if it works in volt where the logged-in user is not the owner of the conversation?

Can we get the `conversationLoader` from `context` and use that for getting the conversation instead of `deligateToSchema` (which in the end calls the same loader because impulse is REST) and leave authorization for impulse?

src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -166,6 +180,7 @@ export const exchangeStitchingEnvironment = ({
       buyerDetails: OrderParty
       sellerDetails: OrderParty
       creditCard: CreditCard
+      isInquiryOrder: Boolean!
```
sepans, line 183: question:
Did you intentionally add `isInquiryOrder` under `CommerceBuyOrder` and `CommerceOrder` or it was needed to make stitching work?
I am not strongly against it though, maybe we start accepting `BUY` inquiry orders soon.

src/schema/v2/me/conversation/index.ts
```diff
@@ -401,8 +404,10 @@ const Conversation: GraphQLFieldConfig<void, ResolverContext> = {
       description: "The ID of the Conversation",
     },
   },
-  resolve: (_root, { id }, { conversationLoader }) =>
-    conversationLoader ? conversationLoader(id) : null,
+  resolve: (_root, { id }, { conversationLoader }) => {
+    console.log("conversationresolver", conversationLoader)
```
erikdstock, line 408:
```suggestion
```

sepans (approved):
Looks great. One question about having `isInquiryOrder` in `BUY` orders. If it is needed totally ok.

src/lib/stitching/exchange/v2/stitching.ts
erikdstock: Yes, it's required because it shared the CommerceOrder interface.

PR #2991, opened 2021-02-11
fix: use en-dash for estimate

MounirDhahri:
**Description:**
Use **en-dash** (&ndash;) on auction results estimate range

### DIFF
--- a/src/schema/v2/__tests__/auction_result.test.ts
+++ b/src/schema/v2/__tests__/auction_result.test.ts
@@ -28,9 +28,6 @@ const mockAuctionResult = {
     centsUSD: 200000,
     display: "€200.000",
   },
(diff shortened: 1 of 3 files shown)


src/schema/v2/__tests__/auction_result.test.ts
```diff
@@ -28,9 +28,6 @@ const mockAuctionResult = {
     centsUSD: 200000,
     display: "€200.000",
   },
-  estimate: {
```
MounirDhahri, line 31: This was here but it wasn't needed for the resolver since we are determining the estimate using `low_estimate_cents` and `high_estimate_cents`

ds300 (approved):
🙏🏽

PR #2994, opened 2021-02-12
Add conversation to Order type (with new root conversation field)

erikdstock:
This PR follows up on https://github.com/artsy/metaphysics/pull/2990 by stitching in `conversation` under a `CommerceOrder` and delegating it to a new root `conversation` field. The previous PR used `me.conversation` but we realized this field would not work for queries from volt which do not use `me`. Still WIP, but if we like this approach it should also hopefully be easier to test.

[PURCHASE-2375]

cc @artsy/purchase-devs 

[PURCHASE-2375]: https://artsyproduct.atlassian.net/browse/PURCHASE-2375

### DIFF
--- a/src/lib/stitching/exchange/__tests__/stitching.test.ts
+++ b/src/lib/stitching/exchange/__tests__/stitching.test.ts
@@ -344,4 +344,108 @@ describe("resolving a stitched conversation", () => {
       data: { commerceOrder: { isInquiryOrder: false } },
     })
   })
+
+  it("resolves conversation field on CommerceOfferOrder", async () => {
+    const allMergedSchemas = await incrementalMergeSchemas(schema, 2)
+    const query = gql`
+      {
+        commerceOrder(id: 4200) {
+          conversation {
+            items {
+              item {
+                ... on Artwork {
+                  title
+                }
+              }
+            }
+          }
+        }
+      }
+    `
+
+    // Mock the resolvers for just an OfferOrder with a conversation id.
+    // The part we are testing is the step that goes from a order
+    // to the conversation.
+    addMockFunctionsToSchema({
+      preserveResolvers: false,
+      schema: allMergedSchemas,
+      mocks: {
+        Query: () => ({
+          commerceOrder: (_root, _params) => {
+            return {
+              __typename: "CommerceOfferOrder",
+              impulseConversationId: "conversation-id",
+            }
+          },
+        }),
+        // FIXME: This mock bypasses our stitched resolver which delegates to
+        // me.conversation
+        Conversation: () => {
+          return {
+            items: [
+              {
+                item: { __typename: "Artwork", title: "Conversation Art" },
+              },
+            ],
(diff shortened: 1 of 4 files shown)


erikdstock:
After discussing this with the purchase team i'm going to proceed. In the process I will:
- check for test optimizations (some TODOs/FIXMEs left there)
- Move conversation types out of the `schema/me` directory since it is no longer specific to that field.
- clean up

src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -146,6 +146,60 @@ export const exchangeStitchingEnvironment = ({
...
+          //     }),
+          //     // how to process the data result at path
+          //     (result) => {
+          //       return result.conversation
+          //     }
+          //   ),
+          // ],
+        })
+      },
+    },
```
xtina-starr, line 202: Any reason this is commented out?

erikdstock: yes, it's left over from a draft PR which used a previous reverted and unreverted commit. this is allowing is to pass `order.conversation` to `me.conversation` but now we are just delegating to the new root conversation field.

src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -181,7 +201,7 @@ export const exchangeStitchingEnvironment = ({
       sellerDetails: OrderParty
       creditCard: CreditCard
       isInquiryOrder: Boolean!
-
+      conversation: Conversation
```
erikdstock, line 204:
```suggestion
      conversation: Conversation
      
```

src/schema/v2/schema.ts
```diff
@@ -126,6 +127,11 @@ const rootFields = {
   city: City,
   cities,
   // collection: Collection,
+  _do_not_use_conversation: {
```
erikdstock, line 130: Pattern reused from [`_do_not_use_image`](https://github.com/artsy/metaphysics/pull/2994/files#diff-f127b0c6324834cfd9f05bc73b0eab077a7e30ff7bf4b9313d107cf36bf442fcR151)

src/lib/stitching/exchange/__tests__/stitching.test.ts
```diff
@@ -344,4 +344,116 @@ describe("resolving a stitched conversation", () => {
...
+      allMergedSchemas,
+      query,
+      {},
+      {
+        conversationLoader: jest.fn(() =>
+          Promise.resolve({
+            items: [
+              {
+                item_type: "Artwork",
+                properties: {
```
erikdstock, line 394: After many tries i finally got this loader to test the full graphql request handling stack.

sepans (approved):
Looks great ❤️ 🚀

erikdstock:
#squashongreen

xtina-starr (approved):
🎉

PR #3002, opened 2021-02-17
Don't show "not open yet" fairs in home module

pvinis:
We don't want the "not open yet" fairs to show up in the home module, so we fix that here.

With this change, I reworked the tests a bit so that the old tests are not skipped anymore, and added two helper deps `jest-extended` and `nanoid`, to use the `.toStartWith` and `.toSatifsyAll` matchers, and to generate ids for the mock objects, respectively.

This resolves [CX-951]

[CX-951]: https://artsyproduct.atlassian.net/browse/CX-951


package.json
```diff
@@ -90,6 +90,7 @@
     "memcached": "2.2.2",
     "moment": "2.26.0",
     "moment-timezone": "0.5.31",
+    "nanoid": "3.1.20",
```
ds300, line 93: should this be a dev dependency?

pvinis: makes sense. yup.

src/schema/v2/home/__tests__/home_page_fairs_module.test.js
```diff
@@ -86,55 +113,33 @@ describe("HomePageFairsModule", () => {
...
-    const results = fairModule.homePage.fairsModule.results
-    expect(results[0].slug).toEqual("current-fair")
-    expect(results[1].slug).toEqual("future-fair")
+    const fairsModule = await runFairsQuery(query, (options) => ({
+      body: options.active ? runningFairs : pastFairs,
+    }))
+    const results = fairsModule.homePage.fairsModule.results
+    expect(results).toHaveLength(8)
+    expect(results).not.toIncludeAnyMembers(pastFairs)
+    expect(results).toSatisfyAll(isRunningFair)
```
pepopowitz, line 122: ah, this is a nice assertion. I like the addition of jest-extended. 👍

src/schema/v2/home/__tests__/home_page_fairs_module.test.js
```diff
@@ -1,40 +1,77 @@
...
+
+// adding this dep explicitly so the tests run when we change the main file
+;() => {
+  require("../home_page_fairs_module")
+}
+
+const mockRunningFair = () => {
+  const id = nanoid()
+  return {
+    id: `running-fair-${id}`,
```
pepopowitz, line 14: Maintainability concern: 

Given each of these three fake fairs have a different prefix in their name (`running-fair-`, `future-fair-`, `past-fair`), do we need to append a unique ID? Dynamically generating these IDs and names seems like it would cause confusion when tests fail and we can't find the exact ID or name in the test code.

The only assertions I see that check uniquely identifying features of each fake fair are checking the prefix _without_ the appended ID, which makes me feel more confident they're not needed. 

Maybe there is some other advantage to generating IDs with nanoid -- if so, let me know!

src/schema/v2/home/home_page_fairs_module.ts
```diff
@@ -19,32 +19,38 @@ export const HomePageFairsModuleType = new GraphQLObjectType<
...
               size: 8 - runningFairs.length,
             }
-            return fairsLoader(newOptions).then(({ body: closedFairs }) =>
-              runningFairs.concat(closedFairs)
+
+            return fairsLoader(newOptions).then(
+              ({ body: unfilteredClosedFairs }) => {
+                const closedFairs = filter(unfilteredClosedFairs, (fair) => {
+                  const endAt = moment.utc(fair.end_at)
+                  return now.isAfter(endAt)
```
pepopowitz, line 50: Minor and probably more a question for my understanding, but just to be sure: 

Does `closed` fairs include fairs that aren't in the past? I'm basically guessing what "closed" means, but it is surprising to me that we'd also have to filter out the ones that are closed *and* not in the past.

pepopowitz (commented):
This is great! Very happy to see the tests turned back on, and the snapshot removed in favor of more direct assertions.

Of all my feedback, the only thing I'd ask for consideration of is the removal of `nanoid`. I don't see a need for another dependency, even if it's just for testing -- though I might not have considered all it's being used for.

src/schema/v2/home/home_page_fairs_module.ts
pvinis: That's the same question I had. During my tests, the query loader seemed to give me more than I ask for. The first query asks for running and gets running and upcoming. The second asks for closed and gets closed and upcoming. I don't know if that's a gravity issue/particularity, but I had to filter on the closed query too :(

src/schema/v2/home/__tests__/home_page_fairs_module.test.js
```diff
@@ -1,40 +1,77 @@
...
+import moment from "moment"
+import { range } from "lodash"
+
+// adding this dep explicitly so the tests run when we change the main file
+;() => {
+  require("../home_page_fairs_module")
+}
+
+const mockRunningFair = () => {
+  const id = nanoid()
```
dzucconi, line 12: `Math.random().toString(26).slice(2)` 🤷

pvinis: Hm ok, I could use this to replace nanoid.

src/schema/v2/home/__tests__/home_page_fairs_module.test.js
pvinis: The idea was that I can do `fairs = [mockRunningFair(), mockRunningFair()]` and get two different fairs that are both running. Then I could test the output using the first ones as a thing to compare with.

In this case here there was not need for this, true. I guess it could be done either with the `Math.random` that @dzucconi suggested or even just using a global (for the file) counter.

I might have been a bit too overzealous 😬.

pvinis:
Simplified the id situation. Thanks for the suggestions, everyone!

src/schema/v2/home/__tests__/home_page_fairs_module.test.js
```diff
@@ -1,40 +1,79 @@
...
-        name: "Zâ“ˆONAMACO FOTO & SALÃ“N DEL ANTICUARIO 2017",
-        published: true,
-        subtype: null,
-        summary: "",
-        layout: null,
-        display_vip: false,
-        has_full_feature: true,
-      },
-    ]
+  it("works", async () => {
```
brainbicycle, line 73: could we update the test description here? Maybe "shows the correct number of fairs"?

src/schema/v2/home/__tests__/home_page_fairs_module.test.js
```diff
@@ -86,55 +115,33 @@ describe("HomePageFairsModule", () => {
...
-        id: "zonamaco-foto-and-sal-n-del-anticuario-2017",
-        name: "Zâ“ˆONAMACO FOTO & SALÃ“N DEL ANTICUARIO 2017",
-      },
-      {
-        id: "past-fair-2017",
-        name: "I Should Not Show Up in the Snapshot",
-      },
-    ]
-
+  it("doesn't return fairs that haven't opened yet", async () => {
```
brainbicycle, line 127: 🎉

brainbicycle (approved):

brainbicycle:
This looks good to me, I am okay to merge when ready!

PR #3004, opened 2021-02-18
Fix broken conversation query

erikdstock:
This PR fixes:
- [x] an issue where querying `isInquiryOrder` or `conversation` would only work when nested under a `...on CommerceOfferOrder {}` spread, otherwise would always be false
- [x] An issue where order.conversation could not be resolved.

This was causing issues with our resolvers, and unfortunately this new approach breaks testing. It also includes a fix for our root _do_not_use_conversation field which was incorrectly not accepting the proper arguments (obscured by a mocked conversationLoader in test).

Mobbed with @starsirius @sepans and @starsirius to get this working, and it is - but tests are not. We agreed to temporarily disable the tests and circle back soon.

**The breaking schema changes are not yet used so should not be a problem.**

### DIFF
--- a/src/lib/stitching/exchange/__tests__/stitching.test.ts
+++ b/src/lib/stitching/exchange/__tests__/stitching.test.ts
@@ -277,13 +277,17 @@ describe("createInquiryOfferOrder", () => {
   })
 })
 
-describe("resolving a stitched conversation", () => {
+// FIXME: These tests don't work
+// eslint-disable-next-line jest/no-disabled-tests
+describe.skip("resolving a stitched conversation", () => {
   it("resolves isInquiryOrder field on CommerceOfferOrder", async () => {
     const allMergedSchemas = await incrementalMergeSchemas(schema, 2)
     const query = gql`
       {
         commerceOrder(id: 4200) {
-          isInquiryOrder
+          ... on CommerceOfferOrder {
+            isInquiryOrder
+          }
         }
       }
     `
@@ -317,7 +321,9 @@ describe("resolving a stitched conversation", () => {
     const query = gql`
       {
         commerceOrder(id: 4200) {
-          isInquiryOrder
+          ... on CommerceOfferOrder {
+            isInquiryOrder
+          }
         }
       }
     `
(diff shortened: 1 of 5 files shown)


src/lib/stitching/exchange/__tests__/stitching.test.ts
```diff
@@ -277,13 +277,17 @@ describe("createInquiryOfferOrder", () => {
   })
 })
 
-describe("resolving a stitched conversation", () => {
+// FIXME: These tests don't work
+// eslint-disable-next-line jest/no-disabled-tests
+describe.skip("resolving a stitched conversation", () => {
```
erikdstock, line 282: Sadly we had to disable these tests. I have a _theory_ about why they don't work (mocked resolver is mocking a `CommerceOrder` but our stitched fragment needs to work off of a `CommerceOfferOrder`, one of the types that implements it, causing mocker confusion), but this whole fix is necessary due to an over-reliance on tests in the first place so I don't have much faith in what i've written below at this point.

src/schema/v2/me/conversation/index.ts
```diff
@@ -404,8 +404,8 @@ const Conversation: GraphQLFieldConfig<void, ResolverContext> = {
       description: "The ID of the Conversation",
     },
   },
-  resolve: (_root, { id }, { conversationLoader }) => {
-    return conversationLoader ? conversationLoader(id) : null
+  resolve: (_root, args, { conversationLoader }) => {
+    return conversationLoader ? conversationLoader(args.id) : null
```
erikdstock, line 408:
```suggestion
  resolve: (_root, { id }, { conversationLoader }) => {
    return conversationLoader ? conversationLoader(id) : null
```

src/schema/v2/schema.ts
```diff
@@ -128,8 +128,7 @@ const rootFields = {
   cities,
   // collection: Collection,
   _do_not_use_conversation: {
-    type: Conversation.type,
-    resolve: Conversation.resolve,
+    ...Conversation,
```
erikdstock, line 131: the root issue of the args problem here - I copied an example below for `_do_not_use_image` and did not include `Conversation.args` as a result.

erikdstock (commented):
Ready for review & merge.

erikdstock:
Talked this over and going to pull the trigger since we mobbed on this earlier today.

PR #3012, opened 2021-02-23
feat: Add orderConnection to Conversation

erikdstock:
This will be a first step in implementing [PURCHASE-2449] to interleave order updates with conversations - We add an `orderConnection` to conversation. See below for a possible query shape.

[PURCHASE-2449]: https://artsyproduct.atlassian.net/browse/PURCHASE-2449

cc @artsy/purchase-devs 
image

### DIFF
--- a/src/schema/v2/me/conversation/index.ts
+++ b/src/schema/v2/me/conversation/index.ts
@@ -405,7 +405,9 @@ const Conversation: GraphQLFieldConfig<void, ResolverContext> = {
     },
   },
   resolve: (_root, { id }, { conversationLoader }) => {
-    return conversationLoader ? conversationLoader(id) : null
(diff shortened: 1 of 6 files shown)


src/schema/v2/me/conversation/index.ts
```diff
@@ -405,7 +405,9 @@ const Conversation: GraphQLFieldConfig<void, ResolverContext> = {
     },
   },
   resolve: (_root, { id }, { conversationLoader }) => {
-    return conversationLoader ? conversationLoader(id) : null
+    return conversationLoader
+      ? conversationLoader(id, { "expand[]": "conversation_orders" })
```
starsirius, line 409: I think this can be removed now?

PR #3020, opened 2021-02-24
feat: Add missing_priority_metadata to partner type

annacarey:
Added `missing_priority_metadata` from gravity as a param for `artworksConnection` on `partner`.

Query works as shown below:
```graphql
query missingMetadata {
  partner(id: "david-zwirner") {
    name
     artworksConnection(first: 5 missingPriorityMetadata: true){
        edges {
          node {
            title
            listPrice {
              __typename
            }
            isPriceHidden
            artist {
              name
            }
            images {
              imageURL
            }
        }
      }
    }
  }
}
```
Screen Shot 2021-02-24 at 11 30 51 AM

[Jira ticket](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=36&projectKey=PX&modal=detail&selectedIssue=PX-3755)

PR #3023, opened 2021-02-25
[PX-3747] Add data loader for PartnerArtworksAll + add custom handling for additional artwork data

jpotts244:
Have partner artworks connection accept new `shallow` argument to fetch additional artwork information.

If `shallow: false` make an additional request to the `partnerArtworksAllLoader` to fetch more data on the artwork (note the `isPriceHidden: false`)

Screen Shot 2021-02-26 at 12 01 15 PM

If `shallow:true` or not present in the parameters, fall back on default `partnerArtworksLoader` behavior (note the `isPriceHidden: null`)
Screen Shot 2021-02-26 at 12 01 25 PM

Also encaptulates @annacarey's new work to support a new MP filter: https://github.com/artsy/metaphysics/pull/3020

Mobbing with @sweir27 @annacarey  @mdole @oxaudo

### DIFF
--- a/src/schema/v2/__tests__/partner.test.js
+++ b/src/schema/v2/__tests__/partner.test.js
@@ -1,5 +1,5 @@
 /* eslint-disable promise/always-return */
-import { runQuery } from "schema/v2/test/utils"
+import { runQuery, runAuthenticatedQuery } from "schema/v2/test/utils"
 import gql from "lib/gql"
 
 describe("Partner type", () => {
@@ -204,6 +204,27 @@ describe("Partner type", () => {
 
   describe("#artworksConnection", () => {
     let artworksResponse
+    const partnerArtworksLoader = jest.fn(() => {
+      return Promise.resolve({
+        body: artworksResponse,
+        headers: {
+          "x-total-count": artworksResponse.length,
+        },
+      })
+    })
+
+    const partnerArtworksAllLoader = jest.fn(() => {
+      return Promise.resolve({
+        body: artworksResponse,
+        headers: {
+          "x-total-count": artworksResponse.length,
+        },
+      })
+    })
+
+    const partnerLoader = jest.fn(() => {
+      return Promise.resolve(partnerData)
+    })
 
     beforeEach(() => {
       artworksResponse = [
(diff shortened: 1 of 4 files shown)


src/schema/v2/partner.ts
```diff
@@ -150,6 +165,24 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
             gravityArgs.artwork_id = flatten([args.artworkIDs])
           }
 
+          if (args.shallow === false && partnerArtworksAllLoader) {
```
jpotts244, line 168: Wanted to get your feedback on ways to handle the case when a non partner user is trying to hit this code path. We added this check to the `if` statement above to fix the type error we were running into yesterday during mobbing - the fact that partnerArtworksAllLoader can be undefined (`Cannot invoke an object which is possibly 'undefined'`). Not entirely sure when it would be undefined, we had the theory that it would be undefined if you were not an authenticated partner/admin.


Right now if you are not an authorized partner/admin you'll get a 403 error, we were thinking maybe we should direct those users to the public `partnerArtworksLoader` as they were not getting an 403 when trying to fetch info from here previous to our shallow changes.

src/schema/v2/partner.ts
```diff
@@ -150,6 +165,24 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
+              const gravityArtworkArgs = {
+                artwork_id: artworkIds,
+              }
+
+              return partnerArtworksAllLoader(id, gravityArtworkArgs).then(
+                ({ body }) => {
+                  return connectionFromArraySlice(body, args, {
+                    arrayLength: artworkIds.length,
+                    sliceStart: 0,
+                  })
```
jpotts244, line 180: Any issues with leaving `arrayLength` and `sliceStart` as these are?

src/schema/v2/partner.ts
sweir27: To answer a few of the questions:
- Correct, if you query this endpoint without a user access token, this loader will be `undefined`. Might help to add a comment specifying that you're adding the check to guard against that case.
- Re: requiring partner-or-admin-level authorization:
  - you could handle it gracefully in code, as we do when [fetching for a causality jwt](https://github.com/artsy/metaphysics/blob/master/src/schema/v1/causality_jwt.ts#L104-L109).
  - This ☝️ might be a little heavy-handed for our use-case. Some other options are to `catch` that particular error and use the loader that doesn't require authorization if it's a 403, and/or to add a comment (or in the description) to the `shallow` param indicating that it only works for partners or admins.

src/schema/v2/partner.ts
sweir27: Does it work if you use the same `offset` and `parseInt(headers["x-total-count"] || "0", 10)` from the response below? That way clients will still be able to paginate and use basic connection features.

(Make sure that we're referencing the `headers` from the promise response from `partnerArtworksLoader` and not from `partnerArtworksAllLoader`. I think this is why it was broken on our first attempt)

jpotts244: Thank you Sarah, using headers from the correct promise works like a charm :]

src/schema/v2/partner.ts
jpotts244: Thanks Sarah, that makes a lot of sense. 

The authenticated loader is `undefined` if the requester is not passing in an access token. I think we were trying to test for regular old user making a request here and was surprised to see the loader be present! I can adjust and make the comments/docs more explicit, we shouldnt be affecting the current behavior if `shallow` is not present

mdole (approved):
Whew! This looks great! Nothing I can think of to change; nice job @jpotts244 @annacarey driving this forward 🙏

src/schema/v2/__tests__/partner.test.js
```diff
@@ -218,108 +239,236 @@ describe("Partner type", () => {
...
+                hasNextPage: false,
+              },
             },
           },
-        },
+        })
+      })
+
+      describe("when shallow is false", () => {
+        it("does not return deep artwork data", async () => {
```
annacarey, line 389: i think `"does not return deep artwork data"` is a little misleading here so i'd want to say that it does not call `partnerArtworksAllLoader` but it does here. I'm not sure how in the test to make sure the context doesn't have access to that loader. in real life, it wouldn't w/o the authentication

annacarey: (either way i think it's still good to still test that the data is returned properly so that we are making sure nothing is broken too)

src/schema/v2/__tests__/partner.test.js
```diff
@@ -218,108 +239,236 @@ describe("Partner type", () => {
...
+                    },
+                  },
+                ],
+              },
+            },
+          })
+        })
+      })
+      describe("when shallow is true", () => {
+        it("does not return deep artwork data", async () => {
```
annacarey, line 431: same as above re: the copy in this block and testing that `partnerArtworksAllLoader` is not called. here, it won't be anyway because `shallow` is true.

annacarey (commented):
Just a few questions about the tests!

src/schema/v2/__tests__/partner.test.js
mdole: @annacarey wait, with `shallow: false`, we DO expect it to be calling `partnerArtworksAllLoader`, yes? Because it's fetching "deep" data?

mdole: isn't our `it` just wrong here, and should instead read `it("returns deep artwork data", ...`

mdole: Re: your original comment, I agree that `it("calls partnerArtworksCallLoader", ...` would be more descriptive of what we're actually testing

annacarey: i think no because it's under the `query is not authenticated` block here so our query should not have access to `partnerArtworksAllLoader` in the context. that is my understanding

mdole: Ah ok, after reading this a few more times I think I get the problem. We're running the UNauthenticated query with `shallow: false`, so we expect `partnerArtworksAllLoader` to not be called. When we try this out in a console, what happens? is this the scenario where we get a full-on error back from Metaphysics, or does it just load "shallow" data?

annacarey: I'm not sure how to run an unauthenticated query in console?

annacarey: Closing the loop on this! I decided to specify the contexts for each of the describe blocks with the authenticated query vs. unauthenticated query. The unauthenticated queries does not have access to `partnerArtworksAllLoader` so the tests now reflect that change. Thanks for your help on this @mdole!

mdole (approved):
woohoo! this looks great - thanks @annacarey for taking some time to talk through how we expect all of the loaders to be called in tests as well.

PR #3029, opened 2021-03-02
fix: [Causality Stitching] Make `lot` private

damassi:
I just realized we could make `auctionsLot` private, since we're simply pointing at causality now.


erikdstock (approved):
You've got no quarrel with me 🤠

PR #3030, opened 2021-03-02
WIP: Add an orderEvents history to commerce orders

erikdstock:
Very WIP/proof of concept for now, putting it up for early comments:

This PR creates a new `OrderEvent` type in our exchange stitching layer that converts an order's offers into a list of events. This is mostly an intermediate step as all of the data is already available on `Order.offers` (a relay connection), but soon we will want to add certain order changes as events as well - for example transitions of `state` when an order is accepted or declined, which are currently not exposed by exchange and only available via the `state_history` property.

paired with @sepans and @starsirius, cc @artsy/purchase-devs 

image

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -8059,9 +8059,16 @@ enum OrderedSetSorts {
 }
 
 type OrderEvent {
-  amount: String
+  amount(
+    decimal: String
+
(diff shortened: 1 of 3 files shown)


erikdstock:
Closing this POC PR in favor of #3039 .

PR #3039, opened 2021-03-10
update exchange schema

erikdstock:
This PR updates the exchange schema to expose `orderHistory` which combines offers with other state changes on an order.


sepans (approved):
🚀

erikdstock:
#mergeongreen

PR #3045, opened 2021-03-13
Migrate MyBids logic to MP

damassi:
Tech overview of how things work in Eigen, courtesy of @erikdstock. Many many thanks 🙏:
https://www.notion.so/artsy/My-Bids-Query-strategy-followups-4c2341e458f347a788b19cf8043f1e7f

Since we're now building out a complete MyBids flow on desktop we needed to move all of the logic from Eigen into a shared place, MP. Part of this effort was also to describe exactly what we need in the UI so that we can declaratively render our bids component and avoid any kind of transformation of our data on the client.

This still needs some more thought, but it works. One major factor here was the need to circumvent stitching for `lotStandingsConnection`, which I think is an ok tradeoff given the amount of shared complexity that will be removed across our apps. 

This might still evolve tho... This current approach is the result of many different experiments trying to work with stitching, and there was just no good way to get intermediate lot connection results from causality via schema, short of a direct query using the loader, without exotic convoluted sdl type hacks. Now we're just straight up saying: fetch x, y, z, combine the results and map to types, right in a single resolver. Then the UI renders out the data exactly as returned, requiring no logic.  It's straightforward. 

One caveat to this work is: we're not using connections because the base of our data (causality) isn't returning a real one, with a max result set of 25 and no pagination, etc. So this better represents the real structure of our data -- a straight-forward list. Once the bidding engine is migrated to Gravity we can revist all of this and it will likely be simplified much much further.  

```graphql
{
  me {
    myBids {
      closed {
        ... 
      }
      active {
        isWatching
        sale {
          bidder {
            qualifiedForBidding
          }
        }
        saleArtworks {
          position
          lot {
            bidCount
            floorSellingPrice {
              display
            }
            internalID
            onlineAskingPrice {
              display
            }
            reserveStatus
            saleId
            sellingPrice {
              display
            }
            soldStatus
          }
          slug
        }        
      }
    }
  }
}
```

cc @artsy/grow-devs


src/schema/v2/me/myBids.ts
```diff
@@ -0,0 +1,259 @@
...
+    },
+    lots: {
+      type: GraphQLList(Lot2),
+    },
+  }),
+})
+
+export const MyBids: GraphQLFieldConfig<void, ResolverContext> = {
+  type: new GraphQLObjectType({
+    name: "MyBids2",
```
damassi, line 35: fixme

src/schema/v2/sale_artwork.ts
```diff
@@ -244,10 +245,20 @@ export const SaleArtworkType = new GraphQLObjectType<any, ResolverContext>({
           )
         },
       },
+      isWatching: {
+        type: GraphQLBoolean,
+        description: "Is this sale artwork being watched by a user",
+        resolve: (saleArtwork) => {
+          return Boolean(saleArtwork.isWatching)
```
damassi, line 252: Fixme

src/schema/v2/sale/index.ts
```diff
@@ -131,6 +131,13 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
           return null
         },
       },
+      bidder: {
+        type: Bidder.type,
+        resolve: async ({ id }, _options, { meBiddersLoader }) => {
+          const bidders = await meBiddersLoader({ sale_id: id })
+          return first(bidders)
+        },
+      },
```
damassi, line 140: fixme: this is the slowest part of the request cycle

src/schema/v2/lot.ts
```diff
@@ -24,3 +25,45 @@ export const Lot = new GraphQLObjectType<any, ResolverContext>({
...
+      resolve: resolveLotCentsFieldToMoney("sellingPriceCents"),
+    },
+    sellingPriceCents: {
+      type: GraphQLInt,
+    },
+    soldStatus: {
+      type: GraphQLString,
+    },
+  }),
+})
```
damassi, line 69: Once bidding engine is moved to Gravity, this will be the type that we'll use. Correct @erikdstock? Thoughts? I remember you brought up some questions around naming; perhaps we can address that here.

src/schema/v2/me/myBids.ts
```diff
@@ -0,0 +1,259 @@
...
+    const me = await meLoader()
+
+    // Fetch all auction lot standings from a given user
+    const causalityPromise = causalityLoader({
+      query: gql`
+        query LotStandingsConnection($userId: ID!, $first: Int) {
+          lotStandingConnection(userId: $userId, first: $first) {
+            edges {
+              node {
+                isHighestBidder
```
damassi, line 89: fixme, need to add

src/schema/v2/lot.ts
erikdstock: I don't think anything is final about future schema at this point... Because there are still some very open questions around websockets/graphql subscriptions among other things in part. My 0.02 would be that this work makes querying individual lot state from causality possible, as well as collections (that a user has watched or bidded upon), and this meets all needs except for live auctions (which have their own unique client implementations) - so this leaves us with a basically complete parity to the gravity-only implementation, and we will have liberty to select the most logical next step when that time comes.

src/schema/v2/me/myBids.ts
```diff
@@ -0,0 +1,259 @@
...
+
+    // Combine ids from categories and dedupe
+    const combinedSaleIds = _.uniq([
+      ...causalitySaleIds,
+      ...registeredSaleIds,
+      ...watchedSaleIds,
+    ])
+
+    // Fetch all sales to format into a list
+    let combinedSales = await Promise.all(combinedSaleIds.map(saleLoader))
```
erikdstock, line 156: non-blocking: `.map(loader)` is a bit hard to parse vs `map(id => loader(id))` - i would probably do it too though.

src/schema/v2/me/watchedLotConnection.ts
```diff
@@ -31,6 +31,12 @@ export const WatchedLotConnection: GraphQLFieldConfig<void, ResolverContext> = {
 
     const { body, headers } = await saleArtworksAllLoader(gravityArgs)
 
+    // TODO: Move this property to gravity JSON response
+    body.forEach((saleArtwork) => {
+      saleArtwork.isWatching = true
+      return saleArtwork
```
erikdstock, line 37: Do we have any other examples of introducing an `isWatching` property that doesn't already exist on json type? What I mean here is that it's one thing to take a field like a time or `sellingPriceCents` and add a resolver to get formatted strings, etc - In this case we are just adding a new property that is very tied to the context of the request. Likewise for moving it into gravity - `SaleArtwork` has no knowledge of users so that json may not belong there. Sorry I feel like i could have typed this all in one sentence!

erikdstock (commented):
Looks good so far, only one real question (adding a user-specific property to `SaleArtwork`). I think it might be worth extracting logic to reduce the number of intermediate values as well - either into individual functions or if that doesn't work (too much overlap) a MyBids class with mostly `private get` members. Like with my code before there is just so much to keep in ones head when you try to read this.

src/schema/v2/me/myBids.ts
```diff
@@ -0,0 +1,259 @@
...
+        saleLoader
+      )
+    ) {
+      return null
+    }
+
+    /**
+     * Causality's connection is faux with a max cap of 25, so we'll only ever
+     * be able to return a non-pagable single list. Increased to an arbitrarily
+     * high number to account for registered sales and watched lots.
```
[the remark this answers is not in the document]
damassi: Ah, i'll update the comment for accuracy. In any case, we can revise connections when bidding engine moves 👍

damassi:
>  I think it might be worth extracting logic to reduce the number of intermediate values as well

My thinking here is there are a fixed number of steps required in order to achieve the desired data structure. Moving stuff to functions might hide some of the complexity behind additional names, but thats what the variable names and comments do in this case. I think if there was more of an "api" to the resolver (other than the single output) then it would make more sense to relocate, but its mostly just: fetch from n * api, fetch ids, fetch artworks for ids, partition into active / inactive; there's no way to get around these steps and hiding them behind containing would likely just increase complexity. (The story is in the comments.)

_schemaV2.graphql
```diff
@@ -7074,6 +7074,20 @@ type Lot implements Node {
   slug: ID!
 }
 
+type Lot2 {
```
zephraph, line 7077: I don't want to get too hung up on naming decisions, but can you talk a little about Lot vs Lot2 and MyBids vs MyBids2? I think my biggest concern is that there's not a lot of contextual information of why this exists and how it differentiates.

src/schema/v2/sale/index.ts
zephraph: Yep, that sounds like a good idea. Batching from MP's side isn't that bad, but we'd have to enhance the gravity API first to support batching for that endpoint: https://github.com/artsy/gravity/blob/master/app/api/v1/me_bidders_endpoint.rb

src/schema/v2/sale_artwork.ts
zephraph: What needs to be fixed?

_schemaV2.graphql
damassi: Oh yeah! i actually need to update these before merge to something more reasonable. (Can explain a bit more tomorrow.)

jonallured (commented):
I confess that aspects of this are over my head but I do agree with @erikdstock that the main body of this work could use some extraction. I read it a few times and came up with a proposal:

```ts
const FETCH_COUNT = 99
const me = await meLoader()

const causalitySaleIds = fetchCausalitySaleIds(FETCH_COUNT, me.id)
const registeredSaleIds = fetchRegisteredSaleIds(FETCH_COUNT)
const watchedSaleIds = fetchWatchedSaleIds(FETCH_COUNT)

const combinedSaleIds = _.uniq([
  ...causalitySaleIds,
  ...registeredSaleIds,
  ...watchedSaleIds,
])

const loadedSales = loadSales(combinedSaleIds)
const enrichedSales = enrichSales(loadedSales)
const sortedSales = sortSales(enrichedSales)

return sortedSales
```

This is pseudo code that I wrote in a markdown file so could have errors (it did 😝 ), but hopefully expresses the idea! I think it's worth doing this so that the overall routine is easier to grok and then the details of the steps can be encapsulated within individual functions. What do you think?

damassi:
> What do you think?

Majority vote wins 😄 

Will refactor.

src/schema/v2/sale_artwork.ts
damassi: Needed to wire this up to both the `myBids` and `watchedLotConnection`. All done 👍

src/schema/v2/me/myBids.ts
damassi: Needed to add the field to `saleArtwork`. Done.

src/schema/v2/me/watchedLotConnection.ts
damassi: We do do this occasionally. I've updated the field description with a note about where context lies and added a todo. I think all of this will resolve automatically once we migrate bidding engine.

_schemaV2.graphql
damassi: Updated to `CausalityLotState` per [this slack convo](https://artsy.slack.com/archives/C01PBEBRB0V/p1616105736019600)

src/schema/v2/lot.ts
damassi: Updated to `CausalityLotState` while things are in flight.

src/schema/v2/sale/index.ts
damassi: Added a note 👍

damassi:
Addressed all feedback! Going to merge and get this to staging and if anyone has any further comments I can take care of it in a follow-up.

damassi:
#mergeOnGreen

PR #3056, opened 2021-03-19
feat: Stitch PriceInsights and MarkerPriceInsights from Vortex, not Diffusion

bhoggard:
https://artsyproduct.atlassian.net/browse/CX-1100


src/lib/stitching/vortex/schema.ts
```diff
@@ -34,7 +34,11 @@ export const executableVortexSchema = ({
         ]
       : []),
     new RenameTypes((name) => {
-      return `Analytics${name}`
+      if (name === "PriceInsights" || name === "MarketPriceInsights") {
```
bhoggard, line 37: These values were not prefixed with `Analytics` when they were coming from Diffusion.

pepopowitz: 2. The message `Field 'MarketPriceInsights.updatedAt' changed type from 'ISO8601DateTime' to 'AnalyticsDateTime'` suggests we should do something similar with DateTime....but it doesn't seem to be as simple as the `BigInt` case. I'm not sure why `ISO8601` was being prefixed when it came from diffusion, but to avoid issues with clients using these types, I think we want that type name not to change.

pepopowitz:
```suggestion
      if (["PriceInsights", "MarketPriceInsights", "BigInt"].includes(name)) {
```

3. (non-blocking) With more than two items in that comparison list, it might also be more readable as checking for the presence in an array.

pepopowitz: 4. These messages also suggest there are other types we need to exclude in the rename, but I'd want to take a closer look to be sure: 

```
Type 'PriceInsightSort' was removed
Type 'PriceInsightConnection' was removed
Type 'PriceInsightEdge' was removed
Field 'marketPriceInsights' was removed from object type 'Query'
Field 'priceInsights' was removed from object type 'Query'
```

pepopowitz (commented):
Nice! Let's take a closer look tomorrow at which types need to be excluded from the `Analytics` prefix to preserve existing clients that use these types.

_schemaV2.graphql
```diff
@@ -7241,32 +7430,32 @@ enum MarketingGroupTypes {
 # Market Price Insights
 type MarketPriceInsights {
   annualLotsSold: Int
-  annualValueSoldCents: BigInt
+  annualValueSoldCents: AnalyticsBigInt
```
pepopowitz, line 7433: Are we sure these type diffs (all the types that now have an `Analytics` prefix) won't break existing clients that are querying MarketPriceInsights? Two places I'd think might be affected --- 

1. Non-update mobile apps. Will they have any issues if the underlying types returned by the queries are named differently?
2. Any existing typescript that is referencing individual types from these queries. I don't know if any exist, but if they do, they would need to change. Is that a better outcome than updating our stitching code to keep the type names all the same?

bhoggard:
Merging this to test on staging.

PR #3057, opened 2021-03-20
chore: update v2 schema

dblandin:
Looks like V2 schema became out-of-sync following #3054.

This is usually updated automatically in a pre-commit hook.

Ran the following command locally to sync things back up:

```
$ node_modules/.bin/nf run --env .env.shared,.env yarn dump:local
```

cc/ @sweir27


dblandin:
#squashonmerge

PR #3064, opened 2021-03-24
fix: Fix total count partner endpoints

annacarey:
Implemented `totalCount` for `artworksConnection` and `showsConnection` on the partner by calculating the count from the `total-count` header and adding to the result returned from MP. For the case with `artworksConnection`, I calculated `totalCount` twice for each case (when authenticated and `shallow=false` and for our default case).

```gql
{
     artworksConnection(first: 25) {
      totalCount
    }
    showsConnection(first: 30) {
      totalCount
    }
  }
}
```

Screen Shot 2021-03-24 at 12 44 10 PM

### DIFF
--- a/src/schema/v2/__tests__/partner.test.js
+++ b/src/schema/v2/__tests__/partner.test.js
@@ -252,6 +252,29 @@ describe("Partner type", () => {
           partnerLoader,
         }
       })
+
+      it("loads the total count", async () => {
+        const query = gql`
+          {
+            partner(id: "bau-xi-gallery") {
+              artworksConnection(first: 3) {
+                totalCount
+              }
+            }
+          }
+        `
+
+        const data = await runAuthenticatedQuery(query, context)
+
+        expect(data).toEqual({
+          partner: {
+            artworksConnection: {
+              totalCount: 3,
+            },
+          },
+        })
+      })
+
       describe("when shallow is false", () => {
         it("calls partnerArtworksAllLoader", async () => {
           const query = gql`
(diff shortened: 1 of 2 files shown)


src/schema/v2/__tests__/partner.test.js
```diff
@@ -462,10 +507,10 @@ describe("Partner type", () => {
     })
 
     it("returns shows", async () => {
-      const query = `
+      const query = gql`
```
annacarey, line 510: the next few changes can be ignored. just added the `gql` here for the formatting

src/schema/v2/partner.ts
```diff
@@ -214,15 +214,20 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
-                      arrayLength: parseInt(
-                        headers["x-total-count"] || "0",
-                        10
-                      ),
-                      sliceStart: offset,
-                    })
+                    return {
+                      totalCount,
+                      ...connectionFromArraySlice(body, args, {
+                        arrayLength: parseInt(
```
sweir27, line 224: Looks like we're repeating the `totalCount` logic here. Could we make this `arrayLength: totalCount`?

src/schema/v2/partner.ts
```diff
@@ -231,10 +236,15 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
-              return connectionFromArraySlice(body, args, {
-                arrayLength: parseInt(headers["x-total-count"] || "0", 10),
-                sliceStart: offset,
-              })
+              const totalCount = parseInt(headers["x-total-count"] || "0", 10)
+
+              return {
+                totalCount,
+                ...connectionFromArraySlice(body, args, {
+                  arrayLength: parseInt(headers["x-total-count"] || "0", 10),
```
sweir27, line 244: (ditto here)

src/schema/v2/partner.ts
```diff
@@ -231,10 +236,15 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
 
           return partnerArtworksLoader(id, gravityArgs).then(
             ({ body, headers }) => {
-              return connectionFromArraySlice(body, args, {
-                arrayLength: parseInt(headers["x-total-count"] || "0", 10),
-                sliceStart: offset,
-              })
+              const totalCount = parseInt(headers["x-total-count"] || "0", 10)
```
annacarey, line 239: I calculated `totalCount` and added it to the result twice: one for each of our two cases: the default here and the second case above when the partner is authenticated and `shallow=false`. I think it's fine but wanted to call out in case you see a way to just do it once.

src/schema/v2/partner.ts
annacarey: Oh yes good call!

sweir27:
Thanks for the fix + the great tests!

PR #3065, opened 2021-03-25
refactor: Rename SaleArtwork.lot to lotState

damassi:
Just a small rename to better reflect the actual function -> lot = sale artwork -> lot state = saleArtwork lot state.


damassi:
#mergeOnGreen

PR #3067, opened 2021-03-27
Revert "feat: Stitch PriceInsights and MarkerPriceInsights from Vortex, not Diffusion"

bhoggard:
Reverts artsy/metaphysics#3056


bhoggard:
#mergeongreen

damassi:
@bhoggard - can you update the description with the reason why a revert was necessary?

bhoggard:
See this slack thread: 

https://artsy.slack.com/archives/C01B2P6LJUU/p1616792127262000

QA showed the results did not match after changing the stitching to Vortex, so this needs to be reverted.

PR #2894, opened 2020-12-15
chore: Bring back me.saleRegistrations as a connection

jo-rs:
Part 1 of 2 for resolving [PURCHASE-2185](https://artsyproduct.atlassian.net/browse/PURCHASE-2185).

Commenting `me.saleRegistrations` field back in to allow for move from `v1` to `v2` on `force`, and turning it into a connection.

Follow up on [this](https://github.com/artsy/force/pull/6810) PR.

PR #3072, opened 2021-04-02
fix: Sidestep batch loader to fix SaleRegistrationConnection

jo-rs:
This PR relates to [PURCHASE-2185].

It sidesteps an issue that was raised after merging [this](https://github.com/artsy/metaphysics/pull/2894) PR in which the batch loader was causing the result to have an unexpected shape. Once merged, the part pertaining to `force` can be brought back.

Note: The PR mentioned above was working locally, but not on Staging due to the `ENABLE_RESOLVER_BATCHING` environment variable not being set locally.

Done alongside @zephraph 

cc @artsy/purchase-devs 

[PURCHASE-2185]: https://artsyproduct.atlassian.net/browse/PURCHASE-2185

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -8311,6 +8311,7 @@ type Partner implements Node {
   collectingInstitution: String
   counts: PartnerCounts
   defaultProfileID: String
+  distinguishRepresentedArtists: Boolean
 
   # Artworks Elastic Search results
   filterArtworksConnection(
@@ -8860,6 +8861,7 @@ type Profile {
   bio: String
   cached: Int
   counts: ProfileCounts
+  displayArtistsSection: Boolean
   href: String
   icon: Image
 
@@ -8874,6 +8876,7 @@ type Profile {
   isPubliclyVisible: Boolean
   isPublished: Boolean
   name: String
+  profileArtistsLayout: String
 
   # A slug ID.
   slug: ID!
--- a/src/schema/v2/me/sale_registrations.ts
+++ b/src/schema/v2/me/sale_registrations.ts
@@ -44,7 +44,7 @@ export const SaleRegistrationConnection: GraphQLFieldConfig<
     const { page, size, offset } = convertConnectionArgsToGravityArgs(
       paginationArgs
     )
-    const { body: sales, headers } = (await (salesLoaderWithHeaders({
+    const response: BodyAndHeaders = await salesLoaderWithHeaders({
       is_auction,
       live,
       published,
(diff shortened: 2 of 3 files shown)


zephraph:
Yeah, there's multiple issues here that raise some complexity. Largely the batch-loaders don't support multiple arguments at this point. If we had better typing, part of this issue would've been communicated sooner. In general batched loaders shouldn't return a different type if it goes through batching or not... but currently it is, which I consider to be a bug. There's probably a lot more here that would just take some dedicated thought on how the API behaves vs what we'd actually like to see. 

With that said, for this issue, it's not worth digging into all those things. Given we don't really want to batch anything here _and_ that we need headers, it's better to just use the specified loader.

_schemaV2.graphql
```diff
@@ -8874,6 +8876,7 @@ type Profile {
   isPubliclyVisible: Boolean
   isPublished: Boolean
   name: String
+  profileArtistsLayout: String
```
starsirius, line 8879: Are these schema changes intentional?

sepans:
Nice catch! 🚀 

I once tried to flip the ENV variables one by one to see if I can reproduce the problem locally but I guess I missed `ENABLE_RESOLVER_BATCHING`

_schemaV2.graphql
zephraph: I suspect they're not

src/schema/v2/me/sale_registrations.ts
```diff
@@ -38,24 +38,21 @@ export const SaleRegistrationConnection: GraphQLFieldConfig<
...
-    ) as any)) as BodyAndHeaders
+    const { body: sales, headers } = (await (salesLoaderWithHeaders({
+      is_auction,
+      live,
+      published,
+      sort,
+      page,
+      size,
+      total_count: true,
+    }) as any)) as BodyAndHeaders
```
damassi, line 55: This form is pretty confusing to read, can we type it like:

```tsx
const response: BodyAndHeaders = await salesLoaderWithHeaders(...)
const { body, headers } = response
...
```

_schemaV2.graphql
jo-rs: Umm, no 🤔 
I wonder why that changed. Good catch, @starsirius!

jo-rs (commented):
~~Umm why didn't the revert push any changed?🤔~~

src/schema/v2/me/sale_registrations.ts
jo-rs: Yessir!

_schemaV2.graphql
jo-rs: I just reverted the file but they got automatically changed again on the pre-commit hook

zephraph: Hmmm, maybe you need to rebase?

jo-rs: Super messy git state 🤦 
Should be fine now. Sorry, folks.

starsirius (approved):
🎉

zephraph (approved):

PR #3073, opened 2021-04-03
POC: conversation event connection

erikdstock:
Future friday project to see what it would take to combine events with messages at the metaphysics layer. We tried to do this at the graphql stitching layer, but creating a union type at that layer makes it really difficult (impossible?) to use the normal stitching helpers like `delegateToSchema` (how do you delegate a fragment on a union). Eventually we decided  to just fetch both, concatenate and sort by date on the client (https://github.com/artsy/eigen/pull/4528). This would move that responsibility into metaphysics, and the `eventConnection` field would effectively deprecate `messagesConnection`

paired with @damassi a bit on this.
cc @artsy/purchase-devs 

unanswered questions:
- do we need to define a node resolver for events
- how would this work with polling or subscriptions
- exchange event fetcher not implemented

### DIFF
--- a/src/schema/v2/me/conversation/conversationEventConnection.ts
+++ b/src/schema/v2/me/conversation/conversationEventConnection.ts
@@ -1,0 +1,201 @@
+import { pageable } from "relay-cursor-paging"
+import { convertConnectionArgsToGravityArgs } from "lib/helpers"
+import { connectionFromArraySlice } from "graphql-relay"
+// import { connectionWithCursorInfo } from "../fields/pagination"
+// import { Lot } from "../lot"
+import {
+  GraphQLBoolean,
+  GraphQLEnumType,
+  GraphQLFieldConfig,
+  GraphQLID,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLString,
+  GraphQLUnionType,
+} from "graphql"
+import { ResolverContext } from "types/graphql"
+import { connectionWithCursorInfo } from "schema/v2/fields/pagination"
+import { MessageType } from "./message"
+import { NodeInterface } from "schema/v2/object_identification"
+import date from "schema/v2/fields/date"
+import { get, merge, sortBy } from "lodash"
+// interface ConversationOrderEvent {
+//   id: string
+//   createdAt: string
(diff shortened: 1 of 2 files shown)


erikdstock:
closing for now, but tracking as possible future work.

src/schema/v2/me/conversation/conversationEventConnection.ts
```diff
@@ -0,0 +1,201 @@
...
+      const messages = message_details.map((message) => {
+        return merge(message, {
+          conversation_initial_message: initial_message,
+          conversation_from_name: from_name,
+          conversation_from_address: from_email,
+          conversation_id: id,
+        })
+      })
+      const orderEvents = orderEventStub.map((event, index) => {
+        return { ...event, id: `event-${index}` }
```
erikdstock, line 185: From discussions with @lilyfromseattle 
- Would we need to filter/mutate events before returning them (example: multiple approval events) or should exchange or client do it?*
- if we add an `id` do we need to do anything else for relay cursors/pagination to work?

* Could new state changes avoid the need to filter events?

src/schema/v2/me/conversation/conversationEventConnection.ts
```diff
@@ -0,0 +1,201 @@
...
+      })
+      const conversationEvents = sortBy(
+        [...messages, ...orderEvents],
+        (event) => new Date(event.createdAt!)
```
erikdstock, line 189: And here we interlace them.

src/schema/v2/me/conversation/conversationEventConnection.ts
```diff
@@ -0,0 +1,201 @@
...
+      )
+      debugger
+      return {
+        ...connectionFromArraySlice(conversationEvents, options, {
+          arrayLength: total_count + orderEvents.length,
+          sliceStart: offset,
+        }),
+        totalCount: total_count + orderEvents.length,
```
erikdstock, line 197: the messages give us a total count and we currently assume the exchange query is giving us all order events, so we just add them together. I don't know if what i have above is the best way to do it or if there are other properties that need to be tweaked aside from `totalCount`.

erikdstock (commented):
Jogging my memory on how this whole thing worked 🥫

PR #3085, opened 2021-04-08
fix: Merge watched and bidded-on artworks

damassi:
This turned out to be a pretty minor change all said. Rather than just returning early with sales that include a watched lot, we concatenate the watched lots array with the other lots, and if a user has bid on a lot that they've already watched, we filter out the watched lot and just return the one that they bid on so there aren't duplicates, reflecting the current state of the in progress sale (in the UI: user has bid, doesn't care about the fact that they've watched anymore). 

This pr is mostly just shuffling code around a bit, but the final outputted schema doesn't change at all, so nothing different with the tests. Gonna merge to get this to staging but happy to follow up.

@artsy/grow-devs

### DIFF
--- a/src/schema/v2/me/myBids.ts
+++ b/src/schema/v2/me/myBids.ts
@@ -147,6 +147,7 @@ export const MyBids: GraphQLFieldConfig<void, ResolverContext> = {
 
     // Clean invalid sales
     combinedSales = combinedSales.filter(Boolean)
+
     // Fetch all sale artworks from sale
     const saleSaleArtworks = await Promise.all(
       combinedSales.map((sale: any) => {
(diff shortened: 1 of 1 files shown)


src/schema/v2/me/myBids.ts
```diff
@@ -169,39 +170,47 @@ export const MyBids: GraphQLFieldConfig<void, ResolverContext> = {
...
+      const saleArtworks = watchedLotsFromSale
+        .filter((watchedLot) => {
+          // Check to see if a user has both watched AND bid on a lot, if so,
+          // only take the lot that user bid on and reject the watched one.
+          const foundWatchedAndBidOnLot = saleArtworksWithCausalityState.find(
+            (saleArtwork) => saleArtwork.lotState.internalID === watchedLot._id
+          )
+          if (!foundWatchedAndBidOnLot) {
+            return watchedLot
+          }
```
damassi, line 205: 200-205 is actually the only change

src/schema/v2/me/myBids.ts
```diff
@@ -169,39 +170,47 @@ export const MyBids: GraphQLFieldConfig<void, ResolverContext> = {
         return node.lot.saleId === sale._id
       })
 
-      // If lot isn't in causality it means we're just watching it and there's
-      // no bidding activity; return that status.
-      if (lots.length === 0) {
```
damassi, line 174: and this was the bug

PR #3089, opened 2021-04-12
Fix: add total count to partner artists connection.

kiryl-zubarau:
This PR adds the `totalCount` field to the artists connection.

image


sweir27:
Nice one!

PR #3091, opened 2021-04-12
chore(PX-3772): Remove image_count_less_than from MP

annacarey:
Remoes for `image_count_less_than` from the `partner` object in Metaphysics. The corresponding gravity work is [here](https://github.com/artsy/gravity/pull/13961). I also regenerated the schema so it would not include this attribute.

Let me know if I missed anything!

[Jira ticket](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=36&projectKey=PX&modal=detail&selectedIssue=PX-3772)

### DIFF
--- a/src/schema/v2/partner.ts
+++ b/src/schema/v2/partner.ts
@@ -48,10 +48,6 @@ const artworksArgs: GraphQLFieldConfigArgumentMap = {
   forSale: {
     type: GraphQLBoolean,
   },
-  imageCountLessThan: {
-    type: GraphQLInt,
-    description: "Return artworks with less than x additional_images.",
-  },
   missingPriorityMetadata: {
     type: GraphQLBoolean,
     description: "Return artworks that are missing priority metadata",
@@ -193,7 +189,6 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
             artwork_id?: string[]
             exclude_ids?: string[]
             for_sale: boolean
-            image_count_less_than?: number
             missing_priority_metadata?: boolean
             page: number
             published: boolean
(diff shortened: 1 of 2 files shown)


src/schema/v2/partner.ts
```diff
@@ -48,10 +48,6 @@ const artworksArgs: GraphQLFieldConfigArgumentMap = {
   forSale: {
     type: GraphQLBoolean,
   },
-  imageCountLessThan: {
-    type: GraphQLInt,
-    description: "Return artworks with less than x additional_images.",
-  },
```
damassi, line 54: One side note about removals: It's totally cool to refactor stuff out _as long as_ the mobile app doesn't intersect the code. As soon as Eigen starts using a slice of the schema, its basically permanent due to the fact that we have to support older versions of the app (which contain a copy of the schema / code from the point that the app was installed). 

It's an unfortunate and tricky aspect of mobile dev + graphql, but so it goes....

annacarey: OH interesting did not know this. Does that mean all of this needs to stay or just the schema part?

mdole: In this case, neither! We're OK to remove this because the mobile app isn't _using_ these parameters at all - even though they're in Eigen's copy of the schema, CMS is the only consumer. If there _was_ a call that used these parameters in Eigen, we would not want to remove them from either the schema or the...hm, I don't know what you call that. Partner schema, I guess. Because if we removed either part, attempts to send an `artworksConnection` field with an `imageCountLessThan` parameter would break. Happy to chat about this more if that's unclear!

mdole (approved):

src/schema/v2/partner.ts
annacarey: Yes thank you @mdole! @damassi and I chatted briefly about it during office hours and now I'm clear. Yes also checked Eigen and we are in the clear on this one!

damassi: Yup precisely @mdole

oxaudo: Cool! Yes - this case is good to remove because we def remember when we've added it and sure that we do not use it naywhere.

PR #3094, opened 2021-04-14
toolchain: Parallelize Jest tests on CI

damassi:
Addresses https://artsyproduct.atlassian.net/browse/FP-25 

This PR splits our Jest tests into two separate jobs -- v1 and v2 -- so that old code and new can run in parallel and speed up CI runs. 

Process:
- Took existing config and copied it to a jest.v2 config file
- Created jest.v1 file and scoped the test runner **only** to tests within `v1` folders

Before:
- ~22min 

After: 
- ~6min

🎉


.circleci/config.yml
```diff
@@ -68,24 +68,33 @@ workflows:
...
+
+          # TODO: Reenable code-cov uploads
+          # post-steps:
+          #   - run: mkdir -p ./coverage
+          #   - run:
+          #       name: Copy coverage artifacts
+          #       command: docker cp hokusai_metaphysics_1:/app/coverage ./
+          #       when: always
+          #   - codecov/upload:
+          #       file: ./coverage/lcov.info
```
damassi, line 89: Have to follow up on this, though TBH I'm not sure how much value code coverage really brings to a web service. It makes a lot of sense for 3rd party libs as a means of building trust, but internal to a company I suspect codecov is largely ignored.

package.json
```diff
@@ -27,6 +27,8 @@
     "schema-drift": "babel-node --no-warnings --extensions '.ts,.js' ./scripts/schema-drift.ts",
     "start": "yarn run dev",
     "test": "yarn type-check && yarn lint && jest",
+    "test:jest:v1": "yarn jest --config jest.config.v1.js --maxWorkers 3",
+    "test:jest:v2": "yarn jest --config jest.config.v2.js --maxWorkers 3",
```
zephraph, line 31: We should be careful with the extra workers here. Especially if you're parallelizing the execution of the tests. 

From [Circle's docs](https://circleci.com/docs/2.0/collect-test-data/#jest):
>Note: When running Jest tests, please use the --runInBand flag. Without this flag, Jest will try to allocate the CPU resources of the entire virtual machine in which your job is running. Using --runInBand will force Jest to use only the virtualized build environment within the virtual machine.

>For more details on --runInBand, refer to the Jest CLI documentation. For more information on these issues, see [Issue 1524](https://github.com/facebook/jest/issues/1524#issuecomment-262366820) and [Issue 5239](https://github.com/facebook/jest/issues/5239#issuecomment-355867359) of the official Jest repository

The challenge here is that are tests could fail in non-deterministic ways, depending on what other resources are running on the CI runner instance.

I'm good w/ trying it, but we'll need to observe and potentially roll back if we start seeing a lot of failures.

.circleci/config.yml
zephraph: You may have a point there, but we should probably have that discussion outside the scope of this change. Was it failing due to an overall lower coverage (given that the v1 tests are split out)?

package.json
```diff
@@ -27,6 +27,8 @@
     "schema-drift": "babel-node --no-warnings --extensions '.ts,.js' ./scripts/schema-drift.ts",
     "start": "yarn run dev",
     "test": "yarn type-check && yarn lint && jest",
```
zephraph, line 29: Should the `test` command be updated here? I guess it doesn't have to be given that the main config pulls in the sub configs.

zephraph (approved):
I've got some comments/feedback, but I don't think anythings particularly blocking. We should watch this, we might get spurious failures. There should be a conversation (on platform?) about code coverage. 

In the spirit of impact over perfection, I'm 👍

.circleci/config.yml
damassi: Its not that it was failing, its that I wasn't sure how to hop from the previous hokusai/test docker test context into a raw node context, eg this command up above: 
```
command: docker cp hokusai_metaphysics_1:/app/coverage ./
```

package.json
damassi: For sure 👍 This worker count of 3 is what we landed on in Force and its been going pretty good so far. Will keep an eye on it.

.circleci/config.yml
zephraph: Ahh, I see. You don't actually need that line. Jest has the coverage plugin so it'll automatically output the results to the `coverage` directory. All that was doing was copying it out of the container so that it could be uploaded. I think if you just deleted that run step it'd complete successfully.

PR #3104, opened 2021-04-21
Add shipping fees to artwork type

starsirius:
https://artsyproduct.atlassian.net/browse/PURCHASE-2573

This exposes `domesticShippingFee` and `internationalShippingFee` on `Artwork` type as `Money` type. CMS can use it to determine if the artwork of an inquiry order has sufficient metadata for partners to "counter" an offer that'd be transactable.

For artwork location, I think we can rely on the existing [`shippingOrigin`](https://github.com/artsy/gravity/blob/0a9e0da15053c7935d9f9051b9175ec34f34dbd6/app/models/domain/artwork.rb#L1223) field.

Screen Shot 2021-04-21 at 5 42 37 PM

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1292,6 +1292,9 @@ type Artwork implements Node & Searchable & Sellable {
   dimensions: dimensions
   displayLabel: String
   displayPriceRange: Boolean
+
+  # Domestic shipping fee.
+  domesticShippingFee: Money
   editionNumber: String
   editionOf: String
   editionSet(id: String!): EditionSet
@@ -1336,6 +1339,9 @@ type Artwork implements Node & Searchable & Sellable {
 
   # A type-specific ID likely used as a database ID.
   internalID: ID!
+
+  # International shipping fee.
+  internationalShippingFee: Money
 
   # Private text field for partner use
   inventoryId: String
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -1766,6 +1766,178 @@ describe("Artwork type", () => {
     })
   })
 
+  describe("#domesticShippingFee", () => {
+    const query = `
+      {
+        artwork(id: "richard-prince-untitled-portrait") {
+          domesticShippingFee {
+            major
+            minor
+            currencyCode
+          }
+        }
+      }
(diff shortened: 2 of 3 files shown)


_schemaV2.graphql
```diff
@@ -1292,6 +1292,9 @@ type Artwork implements Node & Searchable & Sellable {
   dimensions: dimensions
   displayLabel: String
   displayPriceRange: Boolean
+
+  # Domestic shipping fee.
+  domesticShippingFee: Money
```
oxaudo, line 1297: @starsirius - do you think we can rely on those to have real time artwork data that we can check on offer page?

oxaudo (approved):
This will work great if that is not cached.

_schemaV2.graphql
starsirius: Yes I think so. When accessing from CMS, we'll use the authenticated loader which is not cached.

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -1766,6 +1766,178 @@ describe("Artwork type", () => {
...
+              currencyCode: "JPY",
+              major: 10000,
+              minor: 10000,
+            },
+          },
+        })
+      })
+    })
+
+    it("returns null with no domestic_shipping_fee_cents", () => {
```
sepans, line 1830: 💯

_schemaV2.graphql
sepans: Yes, I had to do [some digging](https://artsy.slack.com/archives/CP9P4KR35/p1614706681075900) to remind myself how loaders would work:
When auth token is present the auth loaders are merged into unauth ones and in case of `artworkLoader` where [auth'ed](https://github.com/artsy/metaphysics/blob/a85149e85316ea9252ba916432adf4cec297de93/src/lib/loaders/loaders_with_authentication/gravity.ts#L20) and [unauth'ed](https://github.com/artsy/metaphysics/blob/89543d6347e22c4365cb0e82c15a6c5c27290b9e/src/lib/loaders/loaders_without_authentication/gravity.ts#L51) have the same name the auth loader replaces the auth.

sepans (approved):
🚀  Taking the liberty and merge this since it has 🍏 from all reviewers.

PR #3109, opened 2021-04-23
fix: make follow gene test more specific

olerichter00:
Resolves [FX-2876]

[FX-2876]: https://artsyproduct.atlassian.net/browse/FX-2876


olerichter00:
@squashongreen

PR #3112, opened 2021-04-26
feat(PX-3970): top banner for shows tab

kiryl-zubarau:
This PR:

- adds `featured` sort option
- does not filter by status if it's null
- adds `isFeatured` field to the `show` entity
- adds filtering by `isDisplayable` for the `showsConnection`

JIRA - https://artsyproduct.atlassian.net/browse/PX-3970

image


src/schema/v2/partner.ts
```diff
@@ -562,8 +567,9 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
             page,
             size,
             sort: args.sort,
-            status: args.status,
+            status: args.status || undefined,
```
dzucconi, line 570: Why can't this be `false`?

src/schema/v2/sorts/show_sorts.ts
```diff
@@ -5,6 +5,7 @@ const SHOW_SORTS = {
   END_AT_DESC: { value: "-end_at" },
   FEATURED_ASC: { value: "featured" },
   FEATURED_DESC: { value: "-featured" },
+  FEATURED: { value: "-featured,-end_at" },
```
dzucconi, line 8: Should this be `FEATURED_DESC_END_AT_DESC`?

src/schema/v2/partner.ts
kiryl-zubarau: if you mean something like `status: args.status || false` it throws an error
image

dzucconi: That answers my question yeah (I meant why the need to remove the arg vs let it hang as false). Weird though that a boolean flag can be true but not false :(

src/schema/v2/sorts/show_sorts.ts
kiryl-zubarau: Renamed!

PR #3118, opened 2021-04-28
feat: add all partner artists connection

kiryl-zubarau:
This PR adds `allArtistsConnection` to the `partner` type. Since we load all artists for the artist list on the client-side we decided with @sweir27  to move the loading logic to the server-side.

image

### DIFF
--- a/src/schema/v2/partner.ts
+++ b/src/schema/v2/partner.ts
@@ -36,6 +36,7 @@ import {
 import { deprecate } from "lib/deprecation"
 import { articleConnection } from "./article"
 import ArticleSorts, { ArticleSort } from "./sorts/article_sorts"
+import { allViaLoader } from "lib/all"
 
 import { truncate } from "lib/helpers"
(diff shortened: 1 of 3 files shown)


src/schema/v2/partner.ts
```diff
@@ -157,6 +158,72 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      allArtistsConnection: {
```
dzucconi, line 161: *Sort* of odd that this is even a connection if it *always* returns all the results, non-paginated.

dzucconi: Though I guess it makes sense if you want to access the `totalCount`, for instance

dzucconi (approved):
Feel free to merge. My only comment is that I could see this possibly being an argument on `artistsConnection` instead of it's own field. But I also trust that it's weird enough that this distinction doesn't matter much.

PR #3121, opened 2021-04-29
wip: fetch partner to determine artwork vat completeness

starsirius:
Open up PRs early and get feedback. I'll work on tests separately.

In https://github.com/artsy/gravity/pull/13995, we exposed a new `vat_requirement_complete` field on `Artwork` json response, and it relies on some fields on the `Partner` model such as `vat_status`. One issue we noticed was that the artwork json response is cached and is not updated when the partner gets updated.

We could invalidate artworks from the partners when, say, the `vat_status` on the partner was updated. But I wonder if it's better to not keep adding this caching complexity.

So this PR follows https://github.com/artsy/gravity/pull/14019 and updates `vatRequirementComplete` to fetch data individually closer to the source. UI can still use the same `vatRequirementComplete` flag and MP will do the heavy lifting.

### DIFF
--- a/src/lib/loaders/loaders_with_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_with_authentication/gravity.ts
@@ -220,6 +220,7 @@ export default (accessToken, userID, opts) => {
       { method: "DELETE" }
     ),
     notificationsFeedLoader: gravityLoader("me/notifications/feed"),
+    partnerAllLoader: gravityLoader((id) => `partner/${id}/all`),
     partnerArtworksLoader: gravityLoader(
       (id) => `partner/${id}/artworks`,
       {},
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -2309,10 +2309,64 @@ describe("Artwork type", () => {
       }
     `
 
-    it("returns artworks vat_requirement_complete", () => {
-      artwork.vat_requirement_complete = true
+    it("returns null when vat_required is not present", () => {
+      delete artwork.vat_required
+      return runQuery(query, context).then((data) => {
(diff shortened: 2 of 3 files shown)


src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -220,6 +220,7 @@ export default (accessToken, userID, opts) => {
       { method: "DELETE" }
     ),
     notificationsFeedLoader: gravityLoader("me/notifications/feed"),
+    partnerAllLoader: gravityLoader((id) => `partner/${id}/all`),
```
starsirius, line 223: Need to add a new authenticated loader to the `partner/:id/all` endpoint to get vat data.

sepans: (not related to this PR) haha I was looking if we have this `all` naming pattern anywhere else and found [this loader](https://github.com/artsy/metaphysics/blob/master/src/lib/loaders/loaders_with_authentication/gravity.ts#L239) this loaded that has `All` in the name but doesn't really call the `all` endpoint 😄

sepans (approved):
Makes sense. This is very interesting because I don't know if we have this pattern anywhere using MP to combine data coming from different loaders of the same microservice.

src/lib/loaders/loaders_with_authentication/gravity.ts
starsirius: Yeah potentially it was trying to differentiate from this [sale artworks loader](https://github.com/artsy/metaphysics/blob/e037268902b9a2da332dc211ae178f503e6a724c/src/lib/loaders/loaders_with_authentication/gravity.ts#L245).

We probably don't have many examples because Volt wasn't using MP at the time so there was no need, and now a lot of stuff are done through stitching.

PR #3125, opened 2021-05-04
chore: stitch updated gravity schema (PX-3983)

pepopowitz:
**~Blocked by [gravity#14027](https://github.com/artsy/gravity/pull/14027)~**

---

Associated with [PX-3983]

Stitches in an updated version of the gravity schema.

### Newly included in stitching

- an `agreement` field at the root level for querying a legal agreement (and associated type), from https://github.com/artsy/gravity/pull/14027.
- a `createSavedSearch` mutation, from https://github.com/artsy/gravity/pull/14004. As discussed [in slack](https://artsy.slack.com/archives/C1HH3KNJG/p1620162902020000), there's no reason to exclude this.

### Newly excluded from stitching

- GravQL's `sale`, `saleArtwork`, and `saleArtworks` root-level fields, from https://github.com/artsy/gravity/pull/13969 and https://github.com/artsy/gravity/pull/13976. They have name conflicts with existing MP fields/types, and [as discussed in slack](https://artsy.slack.com/archives/C1HH3KNJG/p1620162524018700) and [the original PR](https://github.com/artsy/gravity/pull/14007#issuecomment-829558613), they are not needed in MP. 

[PX-3983]: https://artsyproduct.atlassian.net/browse/PX-3983


bhoggard (approved):

ansor4:
hmm, I would've assumed we needed to extend the `Partner` type to stitch `Agreements` under that field, something like this: https://github.com/artsy/metaphysics/blob/04dd28b0c6aa2e969b98a3d424dc9eebd93a6ddb/src/lib/stitching/gravity/v2/stitching.ts#L112.

Then I think we would still need to direct the resolver to the gravql schema, like here: https://github.com/artsy/metaphysics/blob/04dd28b0c6aa2e969b98a3d424dc9eebd93a6ddb/src/lib/stitching/gravity/v2/stitching.ts#L533

Might be good to add a test in `stitching.test.ts` too, following the pattern: https://github.com/artsy/metaphysics/blob/a1d8173d7b788dd18ec8578524cb5c9a4118ef66/src/lib/stitching/gravity/__tests__/stitching.test.ts#L445

pepopowitz:
Once we got to the agreements page in CMS, I wasn't thinking of this agreement being in the context of a partner -- we have an agreement ID, we can get at it directly rather than through the current partner. 

I'm fine with either of these approaches, and I can't make a strong case for either of them over the other. Do you have strong feelings about your approach?

ansor4:
Oh yeah that's fair, I think the ticket we wrote had that field stitched under `Partner`, but this seems more straightforward! 👍  Makes me wonder if we want to gate the `Agreement` field at all on gravity's side (like, do we want to only let that field accessible by users with access to partners?), but that can wait.

pepopowitz:
> Makes me wonder if we want to gate the `Agreement` field at all on gravity's side

Yeah, I agree this is something for us to figure out -- I was uncertain about how it _should_ be gated, so I left it fully ungated in [that gravity PR](https://github.com/artsy/gravity/pull/14027), but we should definitely add a gate that makes sense.

bhoggard (approved):

brainbicycle (approved):

PR #3144, opened 2021-05-11
GRO-304: Use of Free Geo IP to find request Country Code

The-Beez-Kneez:
Updated MP to utilize freegeoip to return country code based on IP address


joeyAghion (commented):
Is this for Force, a mobile app, or something else? I ask because Force already uses a similar integration with a different service [here](https://github.com/artsy/force/blob/bd325f12d3f8296cabdfab6901f9edbdad9ab9d9/src/desktop/apps/geo/routes.coffee) and exposes it at `/geo/ip?ip=...`.

The data it returns includes country code and other details that seem like they would support this feature.

If Metaphysics needs its own, similar functionality, there'd be some value in employing the same service and library as Force does now (for consistency, and for simplicity when we upgrade or switch services, as has happened a few times in the past).

The-Beez-Kneez:
@joeyAghion I had a conversation a bit ago in Slack/#dev-help on this topic. 

https://artsy.slack.com/archives/CP9P4KR35/p1619455330109800

From my understanding it had not been used in quite some time (@damon even mentioned the possibility it was hacked). So I proposed a tech plan to use IP Stack. 

Are you saying your preference/recommendation is that we continue to utilize `/geo/ip?ip=...` as the go to service?

joeyAghion:
My goal is just that we not maintain multiple integrations for the same purpose. Force and Gravity both integrate with the freegeoip service today. Force's `/geo/*` endpoints get a few thousand requests per day based on logs, so I don't think it's unused. If IP Stack is significantly better, we should consider switching completely to it.

(The confusion around "hacking" may have been that Force [defaults](https://github.com/artsy/force/blob/8eb6efc43785c246a4b9e9d55f990b99e9701de5/src/desktop/config.coffee#L63) to a Heroku app name that we abandoned a while back and has since been taken over. Our deployed apps have been configured with specific geocoding endpoints for some time. I can send a PR to fix that default.)

joeyAghion:
Thanks yes--that context is helpful and Metaphysics makes sense.

The-Beez-Kneez:
So that said @joeyAghion is there a request or need for us to make a refactoring note for Gravity and the other services utilizing the /geo/*  endpoint? 

I had originally proposed a refactor of the systems utilizing it in Volt, but this could also extend to the other services using it. And I would argue that it definitely should if we are choosing to use IP Stack.

The-Beez-Kneez:
@joeyAghion updated to utilize the exiting `freegeoip` service

src/schema/v2/schema.ts
```diff
@@ -105,6 +105,9 @@ import { DeleteArtworkImageMutation } from "./deleteArtworkImageMutation"
 import { ArtworkOrEditionSetType } from "schema/v2/artworkOrEditionSet"
 import { AuctionResult } from "./auction_result"
 import ArtworkMediums from "./artworkMediums"
+import { RequestLocationField } from "./requestLocation"
+
+// import  { RequestLocation } from "./requestLocation"
```
jonallured, line 110:
```suggestion
```

The-Beez-Kneez: Removed

joeyAghion (commented):
Thanks for the update!

PR #3162, opened 2021-05-19
feat: add flag for works tab display

ivan-badyulya:
JIRA -> [Works tab displays when it is off on CMS](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=36&projectKey=PX&modal=detail&selectedIssue=PX-4098&quickFilter=228)

Added a flag connected with CMS which responsible for displaying the works tab on PPP


sweir27:
Looks good!

PR #3177, opened 2021-05-26
feat(CX-1464): add auction results count to the artist interface

MounirDhahri:
Jira ticket: https://artsyproduct.atlassian.net/browse/CX-1464

This PR adds the auction results count to the artist interface. This will be useful for us to display the links to the auction results tab for the new price insights feature. @edez1t  The query that we will need to run will look similar to the one in the screenshot below:

Screenshot 2021-05-26 at 15 30 56

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -988,6 +988,7 @@ type ArtistCounts {
     format: String
     label: String
   ): FormattedNumber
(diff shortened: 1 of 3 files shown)


olerichter00:
Looks great 🚀

edez1t (approved):
thanks :)

PR #3188, opened 2021-05-31
feat: Add articles connection query

olerichter00:
Resolves [1470](https://artsyproduct.atlassian.net/browse/CX-1470)

## Description

This PR adds an articles connection query to fetch published articles. An articles query already exists but it just fetches the last 10 articles without pagination capabilities.


src/schema/v2/schema.ts
```diff
@@ -1,111 +1,106 @@
 import {
```
olerichter00, line 1: Note: That's just my editor sorting the imports alphabetically...

MounirDhahri (approved):

MounirDhahri:
🎉

PR #3193, opened 2021-06-02
feat(PX-4058): partner transition part 3

kiryl-zubarau:
**Do not merge this PR until the first part will be merged https://github.com/artsy/force/pull/7682**

The 3rd step of the partner profile page transition. https://github.com/artsy/force/pull/7611
JIRA - https://artsyproduct.atlassian.net/browse/PX-4058

### DIFF
--- a/src/schema/v2/partner.ts
+++ b/src/schema/v2/partner.ts
@@ -526,7 +526,7 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
         resolve: ({ type, default_profile_id }) =>
           type === "Auction"
             ? `/auction/${default_profile_id}`
-            : `/${default_profile_id}`,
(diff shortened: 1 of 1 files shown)


src/schema/v2/profile.ts
```diff
@@ -35,7 +35,7 @@ export const ProfileType = new GraphQLObjectType<any, ResolverContext>({
     },
     href: {
       type: GraphQLString,
-      resolve: ({ id }) => `/${id}`,
+      resolve: ({ owner }) => `/partner/${owner.id}`,
```
sweir27, line 38: I think this change isn't correct (we can just remove it), since the `href` here should be for the profile, not the partner.

kiryl-zubarau: Removed

PR #3212, opened 2021-06-11
GRO-350: Updated to allow the partner.href property to be used for Artist Bio links

The-Beez-Kneez:
This is an update in MP to allow us to access the partner href that is used to create slugs. 

We are seeing issues with certain slugs not being updated to accurately reflect Partner name changes. This is because we are creating our associated partner hrefs using partnerID `const partnerHref = `${sd.APP_URL}/${partnerID}``

This update will allow us to pull in the correct href value for slug creation.


mzikherman (approved):

jonallured:
Wow I didn't realize it was this easy to get access to the full partner object - really nice work here. ✨

PR #3214, opened 2021-06-11
feat: hybrid conversation message+event connection

erikdstock:
This PR builds on the future friday work of #3073 to begin creating a combined connection of conversation events (union of messages + order updates). The same functionality currently exists in eigen- we are trying to move it into metaphysics so other clients can share this logic.

The generic logic around processing multiple sources with separate offsets is extracted into a submodule `hybridConnection` - see the new [README.md](https://github.com/artsy/metaphysics/blob/erikdstock/conversation-event-connection/src/schema/v2/fields/README.md) in that folder. We pass to a single function the pagination args (first/after) as well as a limit/offset fetching function for each source - in the case of exchange's order events they come as a single complete collection so logic is not as simple as a gravity query.

still to do:
- [x] Create a `fakeId` field in exchange's order history events so we don't have to make one up on the client. This way we can get a little closer to a guarantee they will be unique **(followup)**
- [x] Verify the order we want these returned in and whether we need a `sort` parameter on the connection
- [x] Make pagination work:
  - when using a cursor from a message
  - when using a cursor from an order update from exchange
  - with a selection of messages that spans both of these unioned types
- [x] Verify that my approach to adding a unique but meaningless id to our events (to fulfill the relay node interface) is ok 
- [x] Validate that this works with polling
- [x] Possibly add helpers like Money to the offer amount
- [x] Make sure that we don't have a problem with order updates dangling off the end of a list of messages [overfetched, because exchange gives us all at once while messages are paginated]
- [x] Tests

paired with @jo-rs on some bits, @lilyfromseattle months later on others
cc @artsy/purchase-devs


src/schema/v2/me/conversation/conversationEventConnection/conversationMessageEvents.ts
```diff
@@ -0,0 +1,67 @@
+import { convertConnectionArgsToGravityArgs } from "lib/helpers"
+import { get } from "lodash"
+
+const lastMessageId = (conversation) => {
+  return get(conversation, "_embedded.last_message.id")
```
damassi, line 5: Why use `get` here?

src/schema/v2/me/conversation/conversationEventConnection/conversationMessageEvents.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// Prepare the incoming args by adding a `before: ${lastMessageId}` if it is missing.
+export const prepareConversationArgs = (conversation, args) => {
+  const conversationArgs = { ...args }
```
damassi, line 10: No need to create intermediate object. Can just do `conversationArgs = args`, or even better just use `args` vs creating intermediate variables

src/schema/v2/me/conversation/conversationEventConnection/conversationOrderEvents.ts
```diff
@@ -0,0 +1,155 @@
...
+  // from args - presence of sellerId overrides buyer mode
+  sellerId?: string
+) => {
+  if (!exchangeLoader) return null
+  // this id changes for the buyer or seller
+  const viewerKey = sellerId
+    ? `sellerId: "${sellerId}"`
+    : `buyerId: "${userID}"`
+
+  const exchangeData = await exchangeLoader({
```
damassi, line 36: I wonder if it might be helpful to add `GraphQL` to graphql loaders to distinguish them from other more common rest loaders -- so like `exchangeGraphQLLoader` 🤔

src/schema/v2/me/conversation/conversationEventConnection/conversationOrderEvents.ts
```diff
@@ -0,0 +1,155 @@
...
+import gql from "lib/gql"
+
+/**
+ * fetch all order events for a conversation from exchange
+ * @param conversationId
+ * @param ctx
+ * @param { sellerId } - to authenticate as seller instead of buyer
+ * @returns
+ */
+export const fetchConversationOrderEvents = async (
```
damassi, line 23: Might be better to move this to its own file?

damassi (commented):
Not sure about the particulars, but code looks like a great start and is clear 👍

src/schema/v2/me/conversation/conversationEventConnection/conversationMessageEvents.ts
erikdstock: this is copied from our old `messagesConnection` so there are definite opportunities for refactoring.

erikdstock:
Checking in here, [this](https://www.notion.so/artsy/Building-a-relay-connection-from-multiple-sources-5581a75681b04f40b1a8efd923d4facd) is a private doc outlining the current approach, some stumbling blocks we've hit and our updated approach which will require encoding special cursors that contain an item's position in the total list as well as relative to others in the list. Still WIP, but other product work may take priority.

src/schema/v2/me/conversation/eventConnection/superCursor.ts
```diff
@@ -0,0 +1,62 @@
+import qs from "qs"
+import { URLSearchParams } from "url"
+
+type CursorState<T extends string> = { position: number } & Record<T, number>
+
+const PREFIX = "supercursor"
+const base64 = (str: string) => Buffer.from(str, "utf-8").toString("base64")
+const unBase64 = (str: string) => Buffer.from(str, "base64").toString("utf-8")
+
+export class SuperCursor<T extends string> {
```
damassi, line 10: Mind adding a description and `@example` of what this does?

erikdstock: Yep, but it is now covered by unit tests as well. In short it is an immutable object that contains a counter for each source you are querying.

Say you initialize a list of both bnmo artworks and auction lots from two different endpoints. You might call that with something like `fetchHybridConnection({ first: 10, after: 'asdfasdf' }, { lot: fetchSaleArtworks, bnmo: fetchBnmoArtworks })` and you will get back a connection of lots and sale artworks. _Internally_ this object is tracking 3 things: `position` which is the overall index of each element in total combined collection, plus `lot` and `bnmo` which is the **offset** within each individual connection.

If the cursor `asdfasdf` decodes to something like `offsets:position=9&lot=4&bnmo=6` (not sure about my index math there) it means that you are a) requesting page 2, after index 9 in the collection and b) requesting an offset of 4 and 6 respectively on each collection. In order to be certain we will have enough records we much get the 10 of each (the `first` arg) and then sort/slice.

Anyhow it's not an easy thing to explain on here, it's really something that requires the charlie day always sunny gif to fully communicate.

src/schema/v2/me/conversation/eventConnection/hybridConnection/README.md
```diff
@@ -0,0 +1,72 @@
+# Hybrid connection
```
damassi, line 1: Is this the right location for this util / readme?

src/schema/v2/me/conversation/eventConnection/hybridConnection/README.md
```diff
@@ -0,0 +1,72 @@
...
+It exports a single function, `fetchHybridConnection` which accepts a map of keys(arbitrary strings for each source) and values (fetchers which can respect `limit`/`offset` args). The function decodes the `after` cursor if present and fetches from each source using its own separately-tracked offset arguments. Then it attaches new offsets and cursors before returning the expected connection interface.
+
+Example:
+
+```ts
+const result = await fetchHybridConnection(
+  { first: 4 },
+  {
+    // a fetcher for messages
+    msg: fetchMessagesForPagination(
```
damassi, line 16: `msg` -> `messages`

src/schema/v2/me/conversation/eventConnection/hybridConnection/README.md
```diff
@@ -0,0 +1,72 @@
...
+      conversationID,
+      conversationMessagesLoader,
+      parent
+    ),
+    // a fetcher for order events
+    ord: fetchOrderEventsForPagination(
```
damassi, line 22: `ord` -> `orders` 

(We should strive to avoid abbreviation in code)

src/schema/v2/me/conversation/eventConnection/hybridConnection/fetchHybridConnection.ts
```diff
@@ -0,0 +1,139 @@
+// interface PaginatedSourceInput<T> {
```
damassi, line 1: Should this file be moved to `lib` or some such place?

src/schema/v2/me/conversation/eventConnection/hybridConnection/fetchHybridConnection.ts
```diff
@@ -0,0 +1,139 @@
...
+// }
+
+import { Connection, ConnectionArguments } from "graphql-relay"
+import { sortBy } from "lodash"
+import { hybridConnectionFromArraySlice } from "./hybridConnectionFromArraySlice"
+import { NodeWMeta } from "./hybridConnectionFromArraySlice"
+import { HybridOffsets } from "./hybridOffsets"
+
+// some typical relay pagination args (maybe there is a type for this)
+type SortDirection = "ASC" | "DESC"
```
damassi, line 13: Aren't we currently limited to only `DESC` atm?

src/schema/v2/me/conversation/eventConnection/hybridConnection/fetchHybridConnection.ts
```diff
@@ -0,0 +1,139 @@
...
+  /* the generic nodes themselves */
+  T extends Record<string, unknown>
+>(
+  { first, last, before, after }: ConnectionArguments,
+  fetchers: Record<K, FetcherForLimitAndOffset<T>>
+): Promise<Connection<T>> => {
+  // 1. check args
+  if (before || last) {
+    throw new Error(
+      "only descending pagination with first/after args are supported"
```
damassi, line 62:
```suggestion
      "Only descending pagination with first/after args are supported"
```

src/schema/v2/me/conversation/eventConnection/hybridConnection/fetchHybridConnection.ts
```diff
@@ -0,0 +1,139 @@
...
+    )
+  }
+  if (typeof first === "number") {
+    if (first <= 0) {
+      throw new Error('Argument "first" must be a non-negative integer')
+    }
+  } else {
+    throw new Error("'first' is required")
```
damassi, line 70:
```suggestion
    throw new Error("The argument `first` is invalid.")
```

If we're checking if something exists, then 
```
if (!first) {
  throw new Error("`first` is required")
}

File: src/schema/v2/me/conversation/eventConnection/hybridConnection/fetchHybridConnection.ts [hunk 13]
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  // assume latest messages are coming through first
```
damassi, line 73: This is super nit picky but noticing lots of inconsistent casing in comments; some are sentence case, some are lowercase, some are mixed. Best to stick with regular `Sentence case everywhere`.

src/schema/v2/me/conversation/eventConnection/hybridConnection/fetchHybridConnection.ts
```diff
@@ -0,0 +1,139 @@
...
+  // (don't support a sort arg yet - descending only)
+  const sort: "ASC" | "DESC" = "DESC"
```
damassi, line 75: If we only support DESC, then there shouldn't be any reference to ASC anywhere in the code

src/schema/v2/me/conversation/eventConnection/hybridConnection/fetchHybridConnection.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  // 2. determine offsets by deserializing cursor
+  const initialOffsets: HybridOffsets<K> = after
+    ? HybridOffsets.decode(after)
+    : HybridOffsets.empty(sources)
+
+  // 3. overfetch enough to fulfill request from either service
+  const requests = sources.map<ReturnType<FetcherForLimitAndOffset<T>>>(
+    (source) => {
+      return fetchers[source](limit, initialOffsets.state[source], sort)
```
damassi, line 87: `fetchers[source](limit, initialOffsets.state[source], sort)` Is a pretty hard to read invocation. Can we break this apart into self-documenting variables?

src/schema/v2/me/conversation/eventConnection/hybridConnection/fetchHybridConnection.ts
```diff
@@ -0,0 +1,139 @@
...
+    const nodes: any[] = resultArray.nodes
+    return nodes.map((node) => ({ ...node, _cursorMeta: { source } }))
+  })
+
+  // 6. sort the nodes (only supports sorting by createdAt/_at key)
+  const sorter = (node) => {
+    const nodeDate = Date.parse(node["createdAt"] || node["created_at"])
+    if (isNaN(nodeDate)) {
+      throw new Error(
+        "A node didn't have a createdAt or created_at, which is required"
```
damassi, line 112: This feels a bit rigid - do we need to have `createdAt` for this to work? That would immediately make it less reusable. Perhaps a sortBy argument could be passed.

src/schema/v2/me/conversation/eventConnection/index.ts
```diff
@@ -0,0 +1,62 @@
...
+        parent
+      ),
+      ord: fetchOrderEventsForPagination(
+        conversationID,
+        userID,
+        exchangeGraphQLLoader
+      ),
+    })
+
+    return result
```
damassi, line 60: 👌

src/schema/v2/me/myBids.ts
```diff
@@ -103,7 +103,7 @@ export const MyBids: GraphQLFieldConfig<void, ResolverContext> = {
   } | null> => {
     if (
       !(
-        causalityLoader &&
+        causalityGraphQLLoader &&
```
erikdstock, line 106: Loaders that make graphql queries were renamed on suggestion of @damassi.

erikdstock (commented):
Small PR Tour

damassi:
@erikdstock - i'm going to be out for the holidays the remainder of this week and next, but will be sure to give this a thorough review once I get back into town. Please hold on merge until then 🙏

src/schema/v2/me/conversation/eventConnection/hybridConnection/README.md
erikdstock: these keys get serialized in the cursors so I was shooting for brevity, but a different dsl could be possible.

src/schema/v2/me/conversation/eventConnection/hybridConnection/fetchHybridConnection.ts
erikdstock: i don't think so, removed this limitation.

src/lib/loaders/loaders_with_authentication/exchange.ts
```diff
@@ -13,13 +21,49 @@ export default (accessToken, opts) => {
...
+        Authorization: `Bearer ${token}`,
+        "Content-Type": "application/json",
+      },
+    })
+
+    const json = await response.json()
+    const { data: exchangeData, errors: exchangeErrors } = json
+    verbose(json)
+
+    // If the causality request failed for some reason, throw its errors.
```
sweir27, line 54: #superminor this comment should probably say `exchange` instead of `causality` 😄

damassi:
@erikdstock - where you able to test this implementation using two random connections? Like `artworksConnection` / `artistsConnection` and see how things behave?

src/lib/loaders/loaders_with_authentication/causality.ts
```diff
@@ -12,7 +12,7 @@ interface GraphQLArgs {
 }
 
 export const causalityLoaders = (_accessToken, _userID) => {
-  const causalityLoader = async ({
+  const causalityGraphQLLoader = async ({
```
damassi, line 15: 👍

src/lib/loaders/loaders_with_authentication/exchange.ts
```diff
@@ -1,7 +1,15 @@
 import factories from "../api"
 import config from "config"
+import fetch from "node-fetch"
+import urljoin from "url-join"
+import { GraphQLError } from "graphql"
+import { verbose } from "lib/loggers"
 
-const { EXCHANGE_APP_ID } = config
+const { EXCHANGE_APP_ID, EXCHANGE_API_BASE } = config
+interface GraphQLArgs {
```
damassi, line 9: I'm gonna be a bit nit-picky so please forgive 🙏  Line break after the const pls

src/lib/loaders/loaders_with_authentication/exchange.ts
```diff
@@ -13,13 +21,49 @@ export default (accessToken, opts) => {
...
+      body,
+      headers: {
+        Authorization: `Bearer ${token}`,
+        "Content-Type": "application/json",
+      },
+    })
+
+    const json = await response.json()
+    const { data: exchangeData, errors: exchangeErrors } = json
+    verbose(json)
```
damassi, line 52: Do we want to verbosely log json here? Perhaps only in `development`?

src/lib/loaders/loaders_with_authentication/exchange.ts
```diff
@@ -13,13 +21,49 @@ export default (accessToken, opts) => {
...
+
+    // If the causality request failed for some reason, throw its errors.
+    if (exchangeErrors) {
+      const errors = exchangeErrors.reduce((acc, error) => {
+        return acc + " " + error["message"]
+      }, "From exchange: ")
```
damassi, line 58: maybe `[exchange]: Error ${errors}`

src/lib/loaders/loaders_with_authentication/index.ts
```diff
@@ -3,15 +3,15 @@ import convectionLoaders from "./convection"
 import diffusionLoaders from "./diffusion"
 import impulseLoaders from "./impulse"
 import gravityLoaders from "./gravity"
-import exchangeLoaders from "./exchange"
+import exchangeGraphQLLoaders from "./exchange"
```
damassi, line 6: This should be exported as a named export, not a default export

src/schema/v2/fields/hybridConnection/README.md
```diff
@@ -0,0 +1,73 @@
...
+It exports a single function, `fetchHybridConnection` which accepts a map of keys(arbitrary strings for each source) and values (fetchers which can respect `limit`/`offset`/`sort` args). The function decodes the cursor if present and fetches from each source using its own separately-tracked offset arguments. A final function argument may be used to sort or transform the collection of nodes. Then it attaches new offsets and cursors before returning the expected connection interface.
+
+Example:
+
+```ts
+const result = await fetchHybridConnection({
+  args: { first: 4 },
+  fetchers: {
+    // a fetcher for messages
+    msg: ({limit, offset, sort}) => fetchMessages(limit, offset, sort),
```
damassi, line 16: we should avoid short-hand abbreviated variable names. `msg -> messages`

src/schema/v2/fields/hybridConnection/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+    // a fetcher for order events that requires a little
+    ord: async ({limit, offset}) => {
```
damassi, line 19: `ord -> orders`

src/schema/v2/fields/hybridConnection/README.md
```diff
@@ -0,0 +1,73 @@
...
+  transform: (args, nodes) => {
+    return nodes.sort((a, b) =>
+      args.sort === 'DESC'
+      ? Date.parse(b.created_at) - Date.parse(a.created_at)
+      : Date.parse(a.created_at) - Date.parse(b.created_at)
+    )
+  }
+)
+
+result ===
```
damassi, line 37: For examples, its a best practice for the entirety of the example / readme / etc to be valid js: 
```js
console.log(
 ... 
)

File: src/schema/v2/fields/hybridConnection/README.md [hunk 28]
```diff
@@ -0,0 +1,73 @@
...
+      },
+    ],
+    pageInfo: {
+      /* Page info */
+    },
+  }
+const { cursor } = result.edges[3]
+
+Buffer.from(cursor, "base64").toString("utf-8") ===
+  "offsets:_position=3&msg=3&ord=1"
```
damassi, line 64: its unclear what lines `const { cursor }` down to this `Buffer.from` is doing

src/schema/v2/fields/hybridConnection/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+**Caveats**
+
+- Only supports one-directional relay pagination args `first`/`after` and descending sort
```
damassi, line 69: We've discussed this offline, but is this `descending` sort still a requirement? It's understandable that `first / after` is required, not its less clear why descending sort is necessary. Is this tied to hard limits around the implementation? Or just a TODO? We should elaborate here on that.

src/schema/v2/fields/hybridConnection/README.md
```diff
@@ -0,0 +1,73 @@
...
+- Expects fetchers to accept `limit`/`offset`/`sort` args. If we want to support other pagination strategies that will be a new PR
+- If a new element is added to the front of the list (eg a new message in a chat) the offsets will all be off which will probably screw up querying. This seems like an inherent limitation of offset-based pagination.
```
damassi, line 71: What does this look like in practice? When would a new element be added? After a user has hit enter and a new message is added to the chat? Does this mean that this hybrid connection strategy only works on initial fetch, and can not be updated in real-time?

src/schema/v2/fields/hybridConnection/README.md
```diff
@@ -0,0 +1,73 @@
...
+  "offsets:_position=3&msg=3&ord=1"
+```
+
+**Caveats**
+
+- Only supports one-directional relay pagination args `first`/`after` and descending sort
+- Expects fetchers to accept `limit`/`offset` args (plus a descending sort)
+- If a new element is added to the front of the list (eg a new message in a chat) the offsets will all be off which will probably screw up querying. This seems like an inherent limitation of offset-based pagination.
+
+> (\*) Under the hood, the HybridOffsets tracks position (index in overall collection) and the source offsets differently: The former is the _0-based index_ in the entire collection, while the former is the _offset_ in that collection. Thus an 'empty' offsets object for the above example would look like `{ a: 0, b: 0 _position: null }`. A PR to clean up this confusing bit would be great.
```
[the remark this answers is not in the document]
damassi: Can we add an example of what this looks like in practice? As a developer distant from this work, its hard to understand what this means concretely, and how it might impact future work and / or use of this library.

src/schema/v2/fields/hybridConnection/fetchHybridConnection.ts
```diff
@@ -0,0 +1,116 @@
...
+    const nodeWithFullMeta = node as NodeWMeta<K, T>
+    nodeWithFullMeta._cursorMeta.offsets = offsets
+
+    nodesWithOffsets.push(nodeWithFullMeta)
+  })
+
+  // return requested number of nodes
+  const slicedNodes = nodesWithOffsets.slice(0, limit)
+
+  return hybridConnectionFromArraySlice({
```
damassi, line 112: small nit, since there's a lot going on in here and future debugging might be necessary, can we assign this line to a variable and then return the variable? That way we can add a breakpoint or console log inbetween and its easier to introspect.

src/schema/v2/fields/hybridConnection/hybridConnectionFromArraySlice.ts
```diff
@@ -0,0 +1,71 @@
+import { Connection } from "graphql-relay"
+import { HybridOffsets } from "./hybridOffsets"
+
+/**
+ * A node with an additional `_cursorMeta` property for storing intermediate data
+ * used to generate our hybrid cursors
+ */
+export type NodeWMeta<
```
damassi, line 8: Please avoid all abbreviations 🙏

src/schema/v2/fields/hybridConnection/hybridConnectionFromArraySlice.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+  })
+
+  const firstEdge = edges[0]
+  const lastEdge = edges[edges.length - 1]
+
+  const result = {
+    edges,
+    pageInfo: {
+      startCursor: firstEdge ? firstEdge.cursor : null,
```
damassi, line 60: could also do `firstEdge?.cursor ?? null` here and elsewhere

src/schema/v2/fields/hybridConnection/hybridOffsets.ts
```diff
@@ -0,0 +1,85 @@
...
+   * of the entire collection and thus starts at `null`, then 0, 1, etc.
+   */
+  static empty<U extends string>(keys: Array<U>): HybridOffsets<U> {
+    const offsets = keys.reduce<OffsetState<U>>(
+      (acc: OffsetState<U>, key: U) => {
+        return { ...acc, [key]: 0 }
+      },
+      { _position: null } as OffsetState<U>
+    )
+    return new HybridOffsets(offsets)
```
damassi, line 32: Here, as elsewhere, its a good habit to assign to a variable and return variable versus returning an operation

src/schema/v2/me/conversation/eventConnection/hybridConnection/README.md
damassi: lets avoid abbreviating. worrying about a descriptive string being serialized and therefore abbreviating is an over-optimization

damassi (approved):
This code and all of the cleanup you've done since last review is just fantastic 💯. Really nice work here. 

I left a few super-minor comments, but nothing blocking at this point. However, before merge I'd encourage a fuller elaboration of the `caveats` section in the README. Some of those points are unclear to outsiders, and the ideal is making known issues so well understood that another person could come in and address them independent of you. 

Really nice work @erikdstock (including types!), and double props for sticking this through over such a long period of time.

src/schema/v2/fields/hybridConnection/README.md
erikdstock: You're right - will update docs here.

src/schema/v2/fields/hybridConnection/README.md
erikdstock: Yeah this is not a helpful caveat. It's just describing the tradeoff between cursor- vs offset-based pagination. I'm just removing it.

src/schema/v2/fields/hybridConnection/fetchHybridConnection.ts
erikdstock: Done, will push changed code shortly.

src/schema/v2/fields/hybridConnection/hybridConnectionFromArraySlice.ts
erikdstock: Done, will push changed code shortly.

src/schema/v2/fields/hybridConnection/hybridConnectionFromArraySlice.ts
erikdstock: Done, will push changed code shortly.

src/schema/v2/fields/hybridConnection/README.md
erikdstock: Updated for clarity.

src/lib/loaders/loaders_with_authentication/exchange.ts
erikdstock: You're right, this is very early future friday leftovers 🔪

src/schema/v2/fields/hybridConnection/README.md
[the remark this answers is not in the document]
erikdstock: Updating for clarity:
> The HybridOffsets object (which becomes our cursor) tracks position in the combined collection and the source offsets differently: The former, `_position` (in the entire collection) is the _0-based index_, while the latter for each individual source is the _offset_ in that collection. Thus an 'empty' offsets object (before the first element has been added) would look like `{ sourceA: 0, sourceB: 0, _position: null }`. It's an implementation detail but a PR to clean up this confusing bit would be welcome.

src/schema/v2/me/conversation/eventConnection/hybridConnection/fetchHybridConnection.ts
erikdstock: Yeah  cleaned up some redundant code that would not make it through the graphql layer into a resolver as well.

src/schema/v2/me/conversation/eventConnection/hybridConnection/fetchHybridConnection.ts
erikdstock: yep, removed.

erikdstock:
#squashongreen 😬

PR #3217, opened 2021-06-14
forgot to include a schema update for my GRO-350 PR

The-Beez-Kneez:
This is to add the schema update from my GRO-350 PR

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -958,9 +958,13 @@ enum ArtistArtworksFilters {
 
 type ArtistBlurb {
   credit: String
+  partner: Partner
 
   # The partner id of the partner who submitted the featured bio.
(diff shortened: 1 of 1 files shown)


jonallured (approved):
LGTM but I'm surprised that CI didn't catch this omission - do we not have a check for this? @mzikherman @damassi @dzucconi

dzucconi:
This could conceivably happen if the pre-commit hooks didn't run?

jonallured:
Ah, the old git hook debate - I see this here:

https://github.com/artsy/metaphysics/blob/master/package.json#L173-L178

I guess I just want to call out again that hooks are a nice-to-have but I believe CI should be the judge about whether a PR is passing. Assuming hooks have successfully run is a mistake, imho.

PR #3218, opened 2021-06-14
feat: request only available for sale artist artworks

dimatretyak:
This PR resolves [FX-3002]
The type of this PR is: **Feature**
Description: "Other works by artist" rail should display available works only
[A little discussion in slack](https://artsy.slack.com/archives/CP9P4KR35/p1623674672071700)

#### Demo

https://user-images.githubusercontent.com/3513494/121940509-daffa280-cd56-11eb-8335-2164fcdb858a.mp4

[FX-3002]: https://artsyproduct.atlassian.net/browse/FX-3002

### DIFF
--- a/src/schema/v1/artwork/artworkContextGrids/ArtistArtworkGrid.ts
+++ b/src/schema/v1/artwork/artworkContextGrids/ArtistArtworkGrid.ts
@@ -46,12 +46,13 @@ export const ArtistArtworkGridType = new GraphQLObjectType<
           artwork,
         })
         gravityArgs.sort = "-published_at"
+        gravityArgs.filter = ["for_sale"]
 
         return artistArtworksLoader(artist.id, gravityArgs).then((artworks) => {
           if (!artworks) return null
 
           return connectionFromArraySlice(artworks, options, {
-            arrayLength: artistArtworkArrayLength(artist, options.filter),
+            arrayLength: artistArtworkArrayLength(artist, gravityArgs.filter),
             sliceStart: offset,
           })
         })
--- a/src/schema/v2/artwork/artworkContextGrids/__tests__/defaultContext.test.ts
+++ b/src/schema/v2/artwork/artworkContextGrids/__tests__/defaultContext.test.ts
@@ -34,7 +34,7 @@ describe("Default Context", () => {
       artist: {
         id: "andy-warhol",
         name: "Andy Warhol",
-        published_artworks_count: 123,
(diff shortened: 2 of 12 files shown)


src/schema/v2/artwork/artworkContextGrids/__tests__/defaultContext.test.ts
```diff
@@ -158,10 +159,7 @@ describe("Default Context", () => {
     expect(artistTitle).toEqual("Other works by Andy Warhol")
     expect(artistCtaTitle).toEqual("View all works by Andy Warhol")
     expect(artistctaHref).toEqual("/artist/andy-warhol")
-    expect(artistArtworks.edges.map(({ node }) => node.slug)).toEqual([
-      "artwork1",
-      "artwork2",
-    ])
+    expect(artistArtworks.edges.map(({ node }) => node.slug)).toEqual([])
```
dimatretyak, line 162: Maybe there are ideas/tips on how to make it so that when [the filter is active](https://github.com/artsy/metaphysics/blob/921ba0e68528cf2828567f10a2e10e2be0421200/src/schema/v1/artwork/artworkContextGrids/ArtistArtworkGrid.ts#L49), the elements are returned?

src/schema/v1/artwork/artworkContextGrids/ArtistArtworkGrid.ts
```diff
@@ -46,12 +46,13 @@ export const ArtistArtworkGridType = new GraphQLObjectType<
           artwork,
         })
         gravityArgs.sort = "-published_at"
+        gravityArgs.filter = ["for_sale"]
```
dblandin, line 49: Does the existing `for_sale` property on `gravityArgs` work in this case or do we need this new `filter` attribute?

```suggestion
        gravityArgs.for_sale = true
```

dimatretyak: We can pass only [filter param](https://github.com/artsy/gravity/blob/master/app/api/v1/artists_artworks_endpoint.rb#L14) for this ArtistArtworks endpoint

sweir27:
Seems straightforward to me! Nice fix-- I suspect this will have a huge impact.

dimatretyak:
> Seems straightforward to me! Nice fix-- I suspect this will have a huge impact.

That's right – this will fix the problem both in the mobile app and on the web at the same time

PR #3226, opened 2021-06-17
update: schema from exchange

kajatiger:
Resolves [PX-4251](https://artsyproduct.atlassian.net/browse/PX-4251)

Connected to the PR in Exchange where the ground work for the mutations was laid: 
https://github.com/artsy/exchange/pull/746

Add ARTA Shipment fields


kajatiger:
@dblandin I am not sure if we need to add the stitching in [here](https://github.com/artsy/metaphysics/blob/master/src/lib/stitching/exchange/v2/stitching.ts). Based on what criteria would I have to add it? How can I test this?

ansor4 (approved):
Nice!

PR #3242, opened 2021-06-23
feat: Add optionalField directive

olerichter00:
Addresses [CX-1132] and blocks https://github.com/artsy/eigen/pull/5023

[CX-1132]: https://artsyproduct.atlassian.net/browse/CX-1132

## Description

Adds `@optionalField` extension that includes all errors on fields that are annotated with the `@optionalField` directive.


damassi:
@olerichter00 - For folks passing by that might not have access to CX Jira (and other historical cross-service factors), its helpful to make PR descriptions [as complete as possible](https://github.com/artsy/README/blob/main/playbooks/engineer-workflow.md#pull-requests), versus just links to elsewhere.

src/extensions/optionalFieldsDirectiveExtension.ts
```diff
@@ -0,0 +1,32 @@
...
+import { flattenErrors, statusCodeForError } from "lib/graphqlErrorHandler"
+
+export const optionalFieldsDirectiveExtension = (documentAST, result) => {
+  const paths = getOptionalFieldsDirectivePaths(documentAST)
+  const extensions = {}
+  const optionalFields: any[] = []
+
+  if (paths.length && result.errors && result.errors.length) {
+    const optionalErrors = result.errors.filter(
+      (error) => paths.filter((path) => isEqual(error.path, path)).length
```
olerichter00, line 12: I wonder if errors that happen somewhere in a subtree of an optional field (annotated with @optionalField) should be included in the optionalField extension as well 🤔 

For example, fetching the `review` could fail in the following query. With our current implementation, the error wouldn't show up in the `optionalFields` extension because we check if the error has the exact same path as the `optionalField` directive.

```
query { 
  artwork(id: "an-id") @optionalField {
    title
    review {
       text
    }
  }
}
```

I guess we have two options:

1. Don't include errors from subtree: This could lead to some "Unable to load" screens because our Eigen error middleware would throw the error if it is not included in the `optionalFields` extension.

2. Include errors from subtree: Every component rendering a field annotated with `@optionalField` would need to handle null values everywhere in the subtree.

What do you think @brainbicycle?

olerichter00: This would be the implementation for Option 2.

```suggestion
      (error) => paths.filter((path) => isEqual(error.path.slice(0, path.length), path))
```

brainbicycle: I think option 2 makes the most sense to me

olerichter00: I agree, let's go with option 2 then!

olerichter00: Yeah, option 2 looks more intuitive.

I just want to mention that there could be cases where a component ends up in an "incomplete" state. For example, letting fetching the `title` fail in the following query could end up in a list of articles with missing titles.

```
query { 
  articles @optionalField {
    title
    ...
  }
}
```

simulator_screenshot_8F42235B-7221-4BBE-BA9F-A5E2E96B8F04

Although I guess that's fine because in practice the example above (fetching reviews fails) is more realistic and I can't find an example where just a single field with a basic type could fail.

brainbicycle (approved):
🙌 Awesome work!

src/index.ts
```diff
@@ -76,11 +77,17 @@ function createExtensions(document, result, requestID) {
...
+    document,
+    result
+  )
+
   const requestLoggerExtensions = enableRequestLogging
     ? fetchLoggerRequestDone(requestID)
     : {}
 
   const extensions = {
+    ...optionalFieldsExtensions,
```
MounirDhahri, line 90: **thought:** What should be executed first the `principalField` or the `optionalField` 🤔. I think it's safe the way you have it now since we won't be having both in the same code base ideally

MounirDhahri:
Great ✨

src/index.ts
olerichter00: This line just adds the `optionalFields` to the extensions in the reply:

```
"data": ...,
"extensions: {
  "optionalFields": [],
  "otherExtension: [],
}

Comment (olerichter00) [reply src/index.ts:90]: In **Eigen** we ignore every field that's annotated with `@optionalField` regardless if it is also a `@principalField`. And ideally, we don't use them in the same query :)

Comment (MounirDhahri) [reply src/index.ts:90]: great, thanks for explaining

**Review by MounirDhahri — APPROVED** [review]

File: src/index.ts [hunk 3]
```diff
@@ -32,6 +32,7 @@ import { ResolverContext } from "types/graphql"
 import { logQueryDetails } from "./lib/logQueryDetails"
 import { ErrorExtension } from "./extensions/errorExtension"
 import { LoggingExtension } from "./extensions/loggingExtension"
+import { optionalFieldsDirectiveExtension } from "./extensions/optionalFieldsDirectiveExtension"
```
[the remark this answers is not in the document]
olerichter00: It's a bit confusing but it's actually called `optionalFields` on purpose. The `principalField` extension contains a single field and the `optionalFields` extension contains an array of fields. We annotate fields in the query with the `optionalField` directive but the extension (contains all `optional fields` that have errors) is called `optionalFields`.

PR #3249, opened 2021-06-29
[PURCHASE-2747] - enable fetching a live artwork instead of the snapshot

xtina-starr:
Addresses: [PURCHASE-2747](https://artsyproduct.atlassian.net/browse/PURCHASE-2747)

**Issue:**
When we fetch items _(a collection of artworks or partner shows a user has inquired on)_, we are fetching a [snapshot version](https://github.com/artsy/impulse/blob/master/app/services/item_service.rb#L17-L46) of these items from impulse. This is particularly beneficial in the event a partner deletes or un-publishes an artwork or show. The conversation still maintains relevant context. However, when we need actual (not a snapshot) and up-to-date information from the artwork object (as is the case for the `isOfferableFromInquiry` property), this presents a problem.

Previously we got around this by creating a [new query renderer](https://github.com/artsy/eigen/pull/4637) and just fetching the artwork directly from gravity through metaphysics. This worked fine initially but introduces other [complexities](https://github.com/artsy/eigen/pull/4814#pullrequestreview-674613779). For example, when we refetch the conversation we also have to trigger a refetch to get the artwork, otherwise the `isOfferableFromInquiry` will be stale and show the "Make Offer" button when we shouldn't.

**Proposed solution:**
The proposed solution is to move the fetch of an up-to-date artwork to metaphysics as part of the conversation query. This is a WIP pull requests with two solutions:
1. Add an optional boolean argument that determines whether to get the snapshot or actual artwork.
2. A new `liveItem` field in the `ConversationItem`

**Screenshot of query:**
Screen Shot 2021-06-29 at 5 10 20 PM

Opening this WIP for purpose of discussion. More context about this issue in a [slack thread](https://artsy.slack.com/archives/C9YNS4X32/p1623442687199400).

Shout out to @erikdstock, our resident purchase graphql/metaphysics expert 😜 for helping me with some of the typing.

cc: @artsy/purchase-devs 

**Screenshot:**
Screen Shot 2021-07-07 at 5 13 56 PM

### DIFF
--- a/src/schema/v2/me/conversation/index.ts
+++ b/src/schema/v2/me/conversation/index.ts
@@ -134,16 +134,19 @@ const ConversationItem = new GraphQLObjectType<any, ResolverContext>({
     permalink: {
       type: GraphQLString,
     },
-    liveItem: {
+    liveArtworkItem: {
       type: ConversationItemType,
-      resolve: (parent, _args, { artworkLoader }) => {
-        if (parent.item_type === "Artwork") {
-          return artworkLoader(parent.properties.id).then((artwork) => ({
-            ...artwork,
-            __typename: "Artwork",
-          }))
+      resolve: (artworkItem, _args, { artworkLoader }) => {
+        if (artworkItem.item_type === "Artwork") {
+          return artworkLoader(artworkItem.properties.id).then((artwork) => {
+            return {
+              ...artwork,
+              __typename: "Artwork",
+            }
+          })
+        } else {
+          throw new Error("PartnerShow not supported.")
         }
-        throw new Error("Only artworks are supported.")
       },
     },
   },
@@ -364,17 +367,10 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
     },
 
     items: {
-      args: {
-        live: {
-          type: GraphQLBoolean,
-          description: "When true fetches an actual items instead of snapshots",
-          defaultValue: false,
-        },
-      },
       type: new GraphQLList(ConversationItem),
       description:
         "The artworks and/or partner shows discussed in the conversation.",
-      resolve: async (conversation, args, { artworkLoader }, _info) => {
+      resolve: async (conversation) => {
         const results = []
 
         for (const item of conversation.items) {
@@ -382,23 +378,9 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
             isExisty(item.properties) &&
             (item.item_type === "Artwork" || item.item_type === "PartnerShow")
           ) {
-            if (args.live && item.item_type === "Artwork") {
-              const artwork = await artworkLoader(item.properties.id)
-
-              const updatedArtwork = {
-                properties: artwork,
-                item_type: "Artwork",
-                title: artwork.title,
-                permalink: `https://www.artsy.net/artwork/${item.properties.id}`,
-              }
-
-              // @ts-ignore
-              results.push(updatedArtwork) // figure out how to type this as an artwork
-            } else {
-              // FIXME: Argument of type 'any' is not assignable to parameter of type 'never'.
-              // @ts-ignore
-              results.push(item)
-            }
+            // FIXME: Argument of type 'any' is not assignable to parameter of type 'never'.
+            // @ts-ignore
+            results.push(item)
           }
         }
         return results
--- a/src/schema/v2/me/conversation/index.ts
+++ b/src/schema/v2/me/conversation/index.ts
@@ -136,6 +136,7 @@ const ConversationItem = new GraphQLObjectType<any, ResolverContext>({
     },
     liveArtwork: {
       type: ConversationItemType,
+      description: "The actual, non-snapshotted artwork",
       resolve: async (conversationItem, _args, { artworkLoader }) => {
         if (conversationItem.item_type === "Artwork") {
           try {
--- a/src/schema/v2/me/conversation/index.ts
+++ b/src/schema/v2/me/conversation/index.ts
@@ -134,6 +134,18 @@ const ConversationItem = new GraphQLObjectType<any, ResolverContext>({
     permalink: {
       type: GraphQLString,
     },
+    liveItem: {
+      type: ConversationItemType,
+      resolve: (parent, _args, { artworkLoader }) => {
+        if (parent.item_type === "Artwork") {
+          return artworkLoader(parent.properties.id).then((artwork) => ({
+            ...artwork,
+            __typename: "Artwork",
+          }))
+        }
+        throw new Error("Only artworks are supported.")
+      },
+    },
   },
 })
 
@@ -352,10 +364,17 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
     },
 
     items: {
+      args: {
+        live: {
+          type: GraphQLBoolean,
+          description: "When true fetches an actual items instead of snapshots",
+          defaultValue: false,
+        },
+      },
       type: new GraphQLList(ConversationItem),
       description:
         "The artworks and/or partner shows discussed in the conversation.",
-      resolve: (conversation) => {
+      resolve: async (conversation, args, { artworkLoader }, _info) => {
         const results = []
 
         for (const item of conversation.items) {
@@ -363,9 +382,23 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
             isExisty(item.properties) &&
             (item.item_type === "Artwork" || item.item_type === "PartnerShow")
           ) {
-            // FIXME: Argument of type 'any' is not assignable to parameter of type 'never'.
-            // @ts-ignore
-            results.push(item)
+            if (args.live && item.item_type === "Artwork") {
+              const artwork = await artworkLoader(item.properties.id)
+
+              const updatedArtwork = {
+                properties: artwork,
+                item_type: "Artwork",
+                title: artwork.title,
+                permalink: `https://www.artsy.net/artwork/${item.properties.id}`,
+              }
+
+              // @ts-ignore
+              results.push(updatedArtwork) // figure out how to type this as an artwork
+            } else {
+              // FIXME: Argument of type 'any' is not assignable to parameter of type 'never'.
+              // @ts-ignore
+              results.push(item)
+            }
           }
         }
         return results
--- a/src/schema/v2/me/conversation/index.ts
+++ b/src/schema/v2/me/conversation/index.ts
@@ -152,7 +152,8 @@ const ConversationItem = new GraphQLObjectType<any, ResolverContext>({
             return null
           }
         } else if (conversationItem.item_type === "PartnerShow") {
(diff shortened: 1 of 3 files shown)


src/schema/v2/me/conversation/index.ts
```diff
@@ -352,20 +364,41 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
...
-            results.push(item)
+            if (args.live && item.item_type === "Artwork") {
+              const artwork = await artworkLoader(item.properties.id)
+
+              const updatedArtwork = {
+                properties: artwork,
+                item_type: "Artwork",
+                title: artwork.title,
+                permalink: `https://www.artsy.net/artwork/${item.properties.id}`,
+              }
```
xtina-starr, line 393: One thing about this solution is that we have to do some data massaging.

src/schema/v2/me/conversation/index.ts
```diff
@@ -134,6 +134,18 @@ const ConversationItem = new GraphQLObjectType<any, ResolverContext>({
...
+      resolve: (parent, _args, { artworkLoader }) => {
+        if (parent.item_type === "Artwork") {
+          return artworkLoader(parent.properties.id).then((artwork) => ({
+            ...artwork,
+            __typename: "Artwork",
+          }))
+        }
+        throw new Error("Only artworks are supported.")
+      },
+    },
```
xtina-starr, line 148: I'm in favor of this option because: 
1. We don't need do any data massaging
2. We are encapsulating this live artwork versus convoluting it with the other items (it feels a little cleaner).

erikdstock:
Really great explanation of the problem and other context. I'm in favor of this approach but interested in others thoughts!

src/schema/v2/me/conversation/index.ts
erikdstock: This is to me the code smell that makes me in favor of `liveItem` instead. Aside from not being up to date the `item` snapshot sent from impulse has its own shape and so when fetching a real artwork we have to do extra work to conform to the resolver's conversation-item shape. For this reason I said let's just make this a separate property and call it only when specifically requested. We could also add a `description`  that `item` is only a snapshot.

src/schema/v2/me/conversation/index.ts
starsirius: I agree on using a new field for this use case. While we are not going to support multiple conversation items in a conversation anytime soon, it might still be a good idea to keep it consistent with the current `items` and name this in plural form.

src/schema/v2/me/conversation/index.ts
```diff
@@ -134,6 +134,18 @@ const ConversationItem = new GraphQLObjectType<any, ResolverContext>({
     permalink: {
       type: GraphQLString,
     },
+    liveItem: {
+      type: ConversationItemType,
+      resolve: (parent, _args, { artworkLoader }) => {
+        if (parent.item_type === "Artwork") {
```
starsirius, line 140: I think `parent` here is a conversation, so maybe what we are looking for is the individual item from `parent.items`?

xtina-starr: @starsirius I thought the naming convention here was to call this parameter `parent`. Perhaps I should change it to `item` because it is of type `ConversationItemType`. Additionally, I `console.log` out the object to double check. Also the conversation object doesn't have the property `item_type`.

starsirius: Ohh I think you're right. It is a conversation item (now I expand the diff 😅 ), instead of a conversation.

joeyAghion:
How does this handle an artwork getting unpublished?

Speaking naïvely, it might be unexpected that the authorization rules and semantics around this data change based only on the argument.

joeyAghion:
Also, `live` is a little overloaded in the context of artworks. Would `published` be a suitable alternative (that might better communicate the authorization implications)?

xtina-starr:
@joeyAghion the `published` callout is a good catch. I added check for whether the artwork is published to handle this.

Regarding the naming, I'm afraid that `published` doesn't distinguish between the actual real artwork and the snapshotted artwork enough. How about something like `uncachedArtworkItem`?

src/schema/v2/me/conversation/index.ts
```diff
@@ -134,6 +134,24 @@ const ConversationItem = new GraphQLObjectType<any, ResolverContext>({
     permalink: {
       type: GraphQLString,
     },
+    liveArtworkItem: {
+      type: ConversationItemType,
+      resolve: (artworkItem, _args, { artworkLoader }) => {
```
xtina-starr, line 139: I changed the name of this from `parent` to `artworkItem` to be more explicit. Curious of your thoughts on this @starsirius?

src/schema/v2/me/conversation/index.ts
```diff
@@ -134,6 +134,24 @@ const ConversationItem = new GraphQLObjectType<any, ResolverContext>({
...
+        if (artworkItem.item_type === "Artwork") {
+          return artworkLoader(artworkItem.properties.id).then((artwork) => {
+            const updatedArtwork = {
+              ...artwork,
+              __typename: "Artwork",
+            }
+            return artwork.published ? updatedArtwork : null
```
xtina-starr, line 146: Added a check to make sure the artwork is published here per @joeyAghion [comment](https://github.com/artsy/metaphysics/pull/3249#issuecomment-873189728).

joeyAghion: This wasn't precisely what I was asking about (or maybe I just don't follow).

If the artwork isn't published, this resolver will _fail_ because it's not authorized to load the artwork's data. Is that expected? (And will clients tolerate it?) The intent of the snapshot was to represent the subset of artwork data that _could_ be exposed long after an artwork is updated or unpublished.

xtina-starr: Ah I understand what you're saying now. No I didn't realize that was the case but I think it's something we could maybe handle on the client.

src/schema/v2/me/conversation/index.ts
starsirius: Non-blocking but I think `conversationItem` might be better name, given that it could be a partner show (and we are actually checking the type below).

src/schema/v2/me/conversation/index.ts
```diff
@@ -355,7 +373,7 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
       type: new GraphQLList(ConversationItem),
       description:
         "The artworks and/or partner shows discussed in the conversation.",
-      resolve: (conversation) => {
+      resolve: async (conversation) => {
```
starsirius, line 376: I think this is not needed?

src/schema/v2/me/conversation/index.ts
```diff
@@ -134,6 +134,24 @@ const ConversationItem = new GraphQLObjectType<any, ResolverContext>({
...
+            const updatedArtwork = {
+              ...artwork,
+              __typename: "Artwork",
+            }
+            return artwork.published ? updatedArtwork : null
+          })
+        } else if (artworkItem.item_type === "PartnerShow") {
+          throw new Error("PartnerShow not supported.")
+        } else {
+          return null
```
starsirius, line 151: To be consistent, maybe we should also raise an error if the item type is not supported.

src/schema/v2/me/conversation/index.ts
starsirius: The goal is to fetch "live" artwork data as part of the same conversation query, instead of the snapshot, so we can render the "Make Offer" button properly. For example, an artwork can be updated to sold after a conversation is made.

If the live artwork becomes unavailable (e.g. unpublished or deleted), it makes sense that this field raises an exception (or returns null?) and clients need to handle it. I'm not entirely positive what the response looks like in this case, so we should test it and make sure clients handle it properly.

@xtina-starr let me know if you want to pair on this together!

src/schema/v2/me/conversation/index.ts
xtina-starr: Do you mean throw an error in the else block versus return null?

xtina-starr: I was more so following the [`ConversationItemType`](https://github.com/artsy/metaphysics/blob/2202f0c3491d566bde20351248092af97283b4e2/src/schema/v2/me/conversation/index.ts#L106-L119) pattern. I throw an error for the `PartnerShow` because one might expect for it to be supported due to the typing. If any other item where to come through I think I might expect it to return `null` as is the case with ConversationItemType. This keeps the implementation similar to how `item` works.

@starsirius let me know if you have strong opinion on this though.

xtina-starr:
I believe I've addressed most of the feedback on this PR. @joeyAghion do you have strong opinions on the name? @starsirius do you have any additional feedback?

joeyAghion:
No strong opinions--up to you.

src/schema/v2/me/conversation/index.ts
```diff
@@ -134,6 +134,26 @@ const ConversationItem = new GraphQLObjectType<any, ResolverContext>({
...
+            (artwork) => {
+              const updatedArtwork = {
+                ...artwork,
+                __typename: "Artwork",
+              }
+              return artwork.published ? updatedArtwork : null
+            }
+          )
+        } else if (conversationItem.item_type === "PartnerShow") {
+          throw new Error("PartnerShow not supported.")
```
erikdstock, line 151: Not sure if this makes sense here but since the field is called `liveArtworkItem` could this just return null? I suppose it depends if the client is sure that it is querying an artwork by this point. We probably don't want to throw an error if we are rendering something that may be an artwork (in which case we want the non-snapshotted version), or a partner show (in which case we don't care).

erikdstock (approved):
2 minor comments, nothing blocking!

src/schema/v2/me/conversation/index.ts
xtina-starr: @erikdstock this is a good point. So here is where we are [querying](https://github.com/artsy/force/blob/master/src/v2/Apps/Conversation/Components/Conversation.tsx#L243-L253) for `isOfferableFromInquiry` in force and here is where we could query for it in [eigen](https://github.com/artsy/eigen/blob/master/src/lib/Scenes/Inbox/Components/Conversations/ConversationCTA.tsx#L78-L80). They are both on an Artwork type. But do we feel we need to make the implementation less strict?

xtina-starr: If anyone were to query for this on PartnerShow it would cause the entire query to fail so that's worth keeping in mind.

erikdstock: I'd say if the name says artwork, return null if there is no artwork item. but now clients who know they are working with an artwork must understand that the item may be null (maybe it is anyways, like if the artwork has been unpublished)

xtina-starr: I think it's safe to return null here because of the naming as you pointed out.

PR #3256, opened 2021-07-02
Added pageCursors field to showsConnection of fair

anastasiapyzhik:
[FX-2992]

pageCursors field was added in order to replace lazy pagination with simple pagination of exhibitors on Force

Screen Shot 2021-07-02 at 11 30 50 AM


src/schema/v2/__tests__/fair.test.js
```diff
@@ -301,22 +275,42 @@ describe("Fair", () => {
...
+        shows: { pageCursors, pageInfo },
       },
+    } = data
+
+    const { around, first, previous } = pageCursors
+
+    expect(around.length).toBe(4)
+    expect(first).toBe(null)
+    expect(previous).toBe(null)
+    expect(pageInfo).toEqual({
```
mdole, line 307: **minor, non-blocking:** @anastasiapyzhik do you think we should add tests where we expect multiple pages? for example, we could try `first:1`, or perhaps`first:1`, get a result back, pull out a cursor, and then do `first: 1, after: cursor` to see more of these cursor fields being populated. Let me know what you think!

src/schema/v2/__tests__/fair.test.js
```diff
@@ -301,22 +275,42 @@ describe("Fair", () => {
...
+      endCursor: null,
```
mdole, line 308: **blocking:** This is a surprise to me - why is it `null`? In order to maintain backwards compatibility, I would expect this to equal `1234567890` since that is what it was in the previous test.

mdole (changes requested):
Mostly looking good! I did leave a "blocking" comment about the result of `endCursor` - we should avoid breaking changes, or we should be very certain they are not being used in any versions of the app that users still have on their phones. 

@anastasiapyzhik please feel free to reach out to me here or on Slack if you have questions or would like to discuss!

src/schema/v2/__tests__/fair.test.js
```diff
@@ -118,34 +118,6 @@ describe("Fair type", () => {
     })
   })
 
-  it("is_publically_visible returns false when profile is not published", () => {
```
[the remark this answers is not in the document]
anastasiapyzhik: There is a duplication here: [this test](https://github.com/artsy/metaphysics/blob/379da5785e4798d1b84b67fc1e5f22dd53735455/src/schema/v2/__tests__/fair.test.js#L121) fully duplicates [this one](https://github.com/artsy/metaphysics/blob/379da5785e4798d1b84b67fc1e5f22dd53735455/src/schema/v2/__tests__/fair.test.js#L93)
So I removed it as Eslint was blocking my commit
I guess it should've be done in a separate PR?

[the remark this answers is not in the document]
mdole: oh! good spot, then I think it makes sense to remove it. no need to create a separate PR, I was just confused about why it was here.

src/schema/v2/__tests__/fair.test.js
anastasiapyzhik: I was sure that [this method](https://github.com/artsy/metaphysics/blob/9fd312b2d27b2678d96aa7e5fbf9beedaf9244d2/src/schema/v2/fair.ts#L300) would return `pageInfo` object with proper `endCursor` field. [[connectionFromArraySlice method type]](https://github.com/graphql/graphql-relay-js/blob/b4cb206c13fc901b4e1e4f63299ac8f8658455c8/src/connection/arrayConnection.d.ts#L41)

I should've thought that my changes might be breaking 
And moreover it turned out not to be so easy and I had to dig a bit more into depth. And if I understand it right we invoke [this method](https://github.com/artsy/gravity/blob/fa4ed7b208f9f34129f2f9550c2d736cad3149b5/app/api/v1/fairs_shows_endpoint.rb#L41) on Gravity when paginating and it doesn't support jumping to a certain page, it only supports loading next bunch of results, so it seems like we have to make some changes on Gravity in order to change pagination logic of fair exhibitors on Force

So I was definitely in a hurry to merge this PR 😅  I expected that I would be able to adjust pagination once I have access to Force locally

Please correct me if I'm wrong

mdole: @anastasiapyzhik all good! We have to be extra careful with breaking changes to Metaphysics because it is also consumed by our iOS and Android apps.

That means that anyone using an older version of the app is making queries using an outdated version of our schema. So if we change the way a field works, it could cause their apps to break.

In this case, I believe it's not a problem because [we don't seem to be using that endCursor field in Eigen](https://github.com/artsy/eigen/blob/74722c329bb74f2851b10092c95df17c3e22557e/src/lib/Scenes/Fair/Components/FairExhibitors.tsx#L83-L113). There is only a "show more" button that fetches the next page, and no "last page" button that would require an end cursor.

It's possible that older versions of the app _did_ use `endCursor`, but hard to know with 100% confidence whether or not that is the case.

One other thing I realized while looking at this and the attached Jira ticket: do we actually need this change? I've started a [thread with Jonathan in Slack ](https://artsy.slack.com/archives/C9SATFLUU/p1625650842112600) to clarify the requirement.

If we do need the change, I think likely OK to merge. Let's see what Jonathan says!

anastasiapyzhik: @mdole I've already asked Jonathan about pagination [here](https://artsy.slack.com/archives/C9SATFLUU/p1625039727429100) so I was certain that these changes are relevant. Now I see that it should be separated from the current ticket, so I'll focus on sorting for now.

What do you think about pagination supporting by Gravity? I'm not good at ruby but as I see [this method](https://github.com/artsy/gravity/blob/fa4ed7b208f9f34129f2f9550c2d736cad3149b5/app/api/v2/util/cursor_helpers.rb#L13) doesn't process `page` value which will be surely needed

mdole: Oh awesome, sorry I missed that you had already checked in about pagination! I created a [separate ticket ](https://artsyproduct.atlassian.net/browse/FX-3071) for pagination in Jira.

And thanks for your patience while I looked into your question about pagination in Gravity. Short answer: yes, with the way the `fair_shows_endpoint` is currently set up, it would not allow us to easily paginate using Metaphysics. However, I think this is very fixable. I have included details in the above Jira ticket and assigned it to myself; I will hopefully complete it in the next 1 - 2 days.

Does that sound OK to you? Are you blocked by this pagination problem, or do you have sufficient other work?

anastasiapyzhik: @mdole that's totally OK, this issue doesn't block me. Thank you for diving into that!

src/schema/v2/__tests__/fair.test.js
anastasiapyzhik: @mdole updated tests so you can see it now 🙂

src/schema/v2/fair.ts
```diff
@@ -262,35 +266,48 @@ export const FairType = new GraphQLObjectType<any, ResolverContext>({
             type: GraphQLBoolean,
             defaultValue: false,
           },
+          page: {
+            type: GraphQLInt,
+            defaultValue: 1,
```
mdole, line 271: I noticed we often don't specify a default page number - I'm not really sure why (or if there are potential ramifications), so I started a thread about it in [#dev-help](https://artsy.slack.com/archives/CP9P4KR35/p1626703855113200). Let's see if there are any best practices we should be following!

src/schema/v2/fair.ts
```diff
@@ -262,35 +266,48 @@ export const FairType = new GraphQLObjectType<any, ResolverContext>({
...
-              connection.pageInfo.endCursor = next
-              connection.pageInfo.hasNextPage = !!next
-              return connection
+            ({ body, headers }) => {
+              const totalCount = parseInt(headers["x-total-count"] || "0", 10)
+
+              return {
+                totalCount,
+                pageCursors: createPageCursors(
+                  { ...options, page, size },
```
mdole, line 303: **minor:** I don't believe we need to pass `...options` in here, because we just [take this object and destructure it to access page and size](https://github.com/artsy/metaphysics/blob/3fcdbf0d4c9355802666d6184e6204573a8df01f/src/schema/v2/fields/pagination.ts#L97)

mdole (approved):
great! one small question and comment, but I think this is looking good. Assigning back to you @anastasiapyzhik , merge when you are ready!

src/schema/v2/fair.ts
anastasiapyzhik: Good catch!

src/schema/v2/fair.ts
anastasiapyzhik: [Seems that there is no significant difference](https://artsy.slack.com/archives/CP9P4KR35/p1626705184113800?thread_ts=1626703855.113200&cid=CP9P4KR35) so I merge

PR #3259, opened 2021-07-02
[PX-2975] Exposes schema necessary to support the galleries_institutions page on Artsy.net

sweir27:
Jira ticket: https://artsyproduct.atlassian.net/browse/PX-2975

When we created metaphysics v2, we intentionally commented out some fields at the root to avoid unnecessary fields.

It turns out, we do actually need a few of these for the /galleries and /institutions page on Artsy.net!


anandaroop (approved):
LGTM

dzucconi:
It appears that your pre-push hooks for updating the schema aren't running? https://github.com/artsy/metaphysics/pull/3266/files#diff-3a304b8e0fa043b030a2e5b3040547c6e4008ebebed2a9207ba67d27dc18192d

PR #3261, opened 2021-07-05
feat: show only for_sale artworks on "similar grid"

dimatretyak:
Type: **Feature**
Jira ticket: [FX-1834)](https://artsyproduct.atlassian.net/browse/FX-1834)
Changes on the Gravity side: [PR 14217](https://github.com/artsy/gravity/pull/14217)

### Description
Requests similar works that are available for sale only (_where it is supported_)


mdole (approved):
awesome! 🚀

PR #3266, opened 2021-07-07
feat(articles): adds featured option

dzucconi:
Very minor thing: simply adds a `featured` boolean option to the top-level `articles` field.

(Some unrelated schema changes showing up here.)


src/schema/v2/articles.ts
```diff
@@ -15,6 +15,10 @@ const Articles: GraphQLFieldConfig<void, ResolverContext> = {
     auctionID: {
       type: GraphQLString,
     },
+    featured: {
+      type: GraphQLBoolean,
+      defaultValue: true,
```
dzucconi, line 20: Woops; do not want a default. 😓

PR #3267, opened 2021-07-07
chore: dump MP schema

ansor4:
ran `yarn dump:local` and committed those changes, this should probably be automated!


anandaroop:
Thanks, I think @dzucconi noticed [over here](https://github.com/artsy/metaphysics/pull/3259#issuecomment-875677159) that for some reason this had not taken place over in that PR in the usual automated way 🤔

PR #3273, opened 2021-07-08
feat(shows): adds filter for event status

dzucconi:
To support the v3 shows page we need to add this argument to replicate the feed API functionality.


dzucconi:
... Peril is merging early without running specs I see ...

damassi:
(I've found that one can't add the flag too soon)

PR #3293, opened 2021-07-20
fix: fetch more than the first page

mdole:
In #3256, we accidentally introduced a bug where we would only ever fetch the first page of results. We introduced a default `page 1` argument, then [in Gravity](https://github.com/artsy/gravity/pull/14273), we introduced behavior that prioritized `page` over `cursor`. End result: we always fetch page 1 and nothing beyond.

This PR removes the default value for page so that we can once again fetch more than 1 page of results.

### DIFF
--- a/src/schema/v2/__tests__/fair.test.js
+++ b/src/schema/v2/__tests__/fair.test.js
@@ -321,7 +321,6 @@ describe("Fair", () => {
           shows: showsConnection(first: 1) {
             pageInfo {
               hasNextPage
-              endCursor
             }
             edges {
               node {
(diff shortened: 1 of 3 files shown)


src/schema/v2/__tests__/fair.test.js
```diff
@@ -352,41 +350,6 @@ describe("Fair", () => {
...
-          edges: [
-            {
-              node: {
-                name: "B",
-              },
-            },
-          ],
-        },
-      },
-    })
```
mdole, line 389: We spent some time looking at this spec, and frankly, we never quite understood a) how it was ever passing, and b) why we couldn't get it to pass. In its current form, the cursor passed to `after` is both the `endCursor` and the cursor for node `"A"`, which seems counterintuitive - why would an item that is _not_ the last item in the list have the same cursor as `endCursor`?

@anastasiapyzhik I think you mentioned that the `endCursor` might actually belong to the second-to-last item; maybe that's the behavior here and I'm just misunderstanding how `endCursor` works?

dblandin: One quote from the [relay spec](https://relay.dev/graphql/connections.htm):

> startCursor and endCursor must be the cursors corresponding to the first and last nodes in edges, respectively.

anastasiapyzhik: `endCursor` is the cursor of the **last** edge of the **page**

<details>
<summary>When we fetch the very first 3 edges we <b>don't</b> pass any cursor and you can see that <code>endCursor</code> is equal to the cursor of the 3d edge</summary>

Screenshot 2021-07-21 at 10 33 31

</details>

<details>
<summary>Then we use this <code>endCursor</code> as <code>after</code> (i.e. we fetch the next 3 edges <b>after</b> the 3d one)</summary>

Screenshot 2021-07-21 at 10 34 54

</details>

<details>
<summary>Big list of edges to ensure that pagination works correctly</summary>

Screenshot 2021-07-21 at 10 33 06

</details>

PR #3298, opened 2021-07-22
bugfix: Fix MyBids lot ordering; add CAUSALITY_TOKEN env var 

damassi:
Fixes an issue that was reported [here](https://artsy.slack.com/archives/C01ADJNCS5D/p1626791297147500?thread_ts=1626720243.123600&cid=C01ADJNCS5D).

#### Todo:
- [x] Add `CAUSALITY_TOKEN` to staging 
- [x] Add `CAUSALITY_TOKEN` to prod 
- [ ] Add `CAUSALITY_TOKEN` to citadel 

#### The gist of it is: 
1. Staging data has always been brittle when it comes to querying lots from causality and so we stuck to a few very specific jenkins-created gravity sales for testing things
1. This mostly worked ok because we create a few different kinds of sales, but in this case it revealed an edge case around data access where a causality-sync'd lot wasn't appearing correctly in the my bids component
1. When trying to fix, it revealed that we were returning causality lots based on a user generated JWT key via a `causalityLoader`. 
1. But for reasons that are unclear, user generated JWT keys were erroring out, and metaphysics was consuming the error.
1. Based on this [slack convo here](https://artsy.slack.com/archives/CP9P4KR35/p1626821863151300) I learned that we recently updated causality's general-purpose JWTs in may, so perhaps that led to a cascading failure around this. 
1. And so, I replaced the user-generated JWT with an ENV var `CAUSALITY_TOKEN` from 1p and now causality lot state began to appear
1. This then revealed that we were never accessing those kinds of sale lots correctly in a `filter` statement, leading to the bug 
1. Lastly, all of this was likely missed until now because we could never test production bidding, and seemed to only stick to specific Jenkins weekly sync sales during QA in staging. That's called... a failure 😢. 

> Thankfully all of these layers will be simplified shortly as Causality is being retired for a full Gravity bidding engine rebuild, unifying lot state within one system instead of two. 

_Anyways_ 

#### Repo steps:

1. Go to https://staging.artsy.net/auction/longhouse-reserve-benefit-auction-2021
2. Follow artist from first lot: https://staging.artsy.net/artwork/anni-albers-study-for-nylon-rug-3
3. Go to: https://staging.artsy.net/artwork/miya-ando-4-dot-14-dot-2020-dot-1-the-lyrids-meteor-shower-latitude-40-dot-760131-longitude-73-dot-980127
4. Watch lot, bid high
5. Go back to: https://staging.artsy.net/auctions
6. I can see image + artist info from lot 1 but bidding info from meteor shower lot

#### Before: 

Screen Shot 2021-07-21 at 10 15 08 PM

#### After: 

Screen Shot 2021-07-21 at 10 15 56 PM


src/lib/loaders/loaders_with_authentication/causality.ts
```diff
@@ -33,20 +25,23 @@ export const causalityLoaders = (_accessToken, userID) => {
...
     })
 
     const json = await response.json()
-    const { data: causalityData, errors: causalityErrors } = json
+    const { data: causalityData, error, errors: causalityErrors } = json
 
-    // If the causality request failed for some reason, throw its errors.
-    if (causalityErrors) {
+    if (error) {
+      throw new Error(`[loaders/causality.ts]: ${error.message}`)
```
damassi, line 37: The lack of this condition was consuming the error

src/lib/loaders/loaders_with_authentication/causality.ts
```diff
@@ -33,20 +25,23 @@ export const causalityLoaders = (_accessToken, userID) => {
       method: "POST",
       body,
       headers: {
-        Authorization: `Bearer ${token}`,
+        Authorization: `Bearer ${CAUSALITY_TOKEN}`,
```
damassi, line 28: @dzucconi - this should fix all of that weirdness you were seeing when trying to repo the bug. Lots should now all return (assuming you've updated your ENV with the value in 1pass)

dzucconi (approved):
I'm sure this was fun to get to the bottom of. 🙏

PR #3300, opened 2021-07-22
toolchain: Match setup script with force

damassi:
Related https://github.com/artsy/force/pull/8031

This updates the setup script to match force, with notable changes being:
- check for `yarn` before trying to install 
- check for `memcached` before trying to install
- add `setup` to `package.json` 
- add `sync-env` to `package.json`

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,13 +45,12 @@ It is currently used in production all over the place in
 
 To get yourself set up with all the project's dependencies:
 
-```
+```sh
 git clone https://github.com/artsy/metaphysics
 cd metaphysics
 
 # Run the setup script
-
-yarn setup
+./scripts/setup.sh
 ```
 
 This will pull the environment variables from aws into .env.shared.
@@ -161,20 +160,20 @@ this:
 - Install [Docker for Mac](https://github.com/artsy/hokusai#requirements) and
   [Hokusai](https://github.com/artsy/hokusai#setup)
 
-  ```
+  ```sh
   $ brew tap caskroom/cask && brew cask install docker
   $ pip install hokusai
   ```
 
   If you are using your system Python distribution, you may need to run this as:
 
-  ```
+  ```sh
   $ sudo pip install hokusai --ignore-installed
   ```
 
 - Configure Hokusai
 
-  ```
+  ```sh
   export AWS_ACCESS_KEY_ID={{ MY_AWS_ACCESS_KEY_ID }}
   export AWS_SECRET_ACCESS_KEY={{ MY_AWS_SECRET_ACCESS_KEY }}
   hokusai configure --kubectl-version {{ kubectl_version }} --s3-bucket {{ kubectl_config_s3_bucket }} --s3-key {{ kubectl_config_s3_key }}
@@ -184,11 +183,9 @@ this:
   Artsy staff should find follow the instructions in
   https://github.com/artsy/potential/blob/master/platform/Kubernetes.md#hokusai
 
-### Development
-
 - Start the server
 
-  ```
+  ```sh
   hokusai dev start
   ```
 
@@ -196,7 +193,7 @@ this:
 
 - Run tests in the Docker Compose test stack via Hokusai:
 
-  ```
+  ```sh
   hokusai test
   ```
 
--- a/README.md
+++ b/README.md
@@ -51,10 +51,10 @@ cd metaphysics
 
 # Run the setup script
 
-source ./scripts/setup.sh
+yarn setup
 ```
 
-This will pull the environment variables from aws into .env.shared. 
+This will pull the environment variables from aws into .env.shared.
 It will also overwrite .env with the values in .env.example. If you need to override any of these values
 or add new .env values place them in the .env file.
 
--- a/README.md
+++ b/README.md
@@ -63,16 +63,10 @@ or add new .env values place them in the .env file.
 With your dependencies set up, you can run Metaphysics by running:
 
 ```sh
-yarn dev
+yarn start
 ```
 
 Which will start the server on http://localhost:3000
-
-You can also use hokusai by running the below command.
-
-```sh
-hokusai dev start
-```
 
 Be sure that memcached is no longer running before starting hokusai by running
 
@@ -104,14 +98,18 @@ Introspection is available by default when developing.
 Introspection on staging and production are for internal use only, so artsy devs can use it to make development for MP clients (eigen, force, etc) easier, but it is and should not be used by any of the clients or anyone else.
 
 In order to set this up in your playground of choice (Postman, Insomnia, Altair, etc), you need to send the following header:
+
 ```
 Authorization: Bearer <secret>
 ```
+
 and replace `<secret>` with the value you get from hokusai using
+
 ```
 hokusai staging env get INTROSPECT_TOKEN
 hokusai production env get INTROSPECT_TOKEN
 ```
+
 or the contents of `Metaphysics INTROSPECT_TOKEN` in 1Password.
 
 ### Sample Queries
@@ -185,6 +183,14 @@ this:
 
   Artsy staff should find follow the instructions in
   https://github.com/artsy/potential/blob/master/platform/Kubernetes.md#hokusai
+
+### Development
+
+- Start the server
+
+  ```
+  hokusai dev start
+  ```
 
 ### Testing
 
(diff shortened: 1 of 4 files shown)


scripts/setup.sh
```diff
@@ -28,5 +39,5 @@ fi
 echo 'Updating .env.shared file (for shared configuration)...'
 aws s3 cp s3://artsy-citadel/dev/.env.metaphysics .env.shared || return
 
-echo 'Setup complete! To start the server on localhost:3000, run:
-  yarn dev'
+echo 'Setup complete! To start the server, run:
+  yarn start'
```
damassi, line 43: `start` points at the `dev` command behind the scenes, so this keeps things a little more vanilla

joeyAghion: Vanilla is delicious.

scripts/setup.sh
```diff
@@ -6,17 +6,28 @@
...
+  echo 'yarn is required for setup, installing...'
+  if ! brew --version &> /dev/null; then
+    echo 'brew is required to install yarn, see https://docs.brew.sh/Installation'
+    exit 0
+  fi
+  brew install yarn
+fi
+
+if ! memcached --version &> /dev/null; then
+  brew install memcached
```
joeyAghion, line 20: The `brew bundle` command is supposed to work similarly (only install what isn't already installed). Are you testing for each dependency in case developers have them installed other [non-brew] ways?

README.md
```diff
@@ -186,6 +184,14 @@ this:
   Artsy staff should find follow the instructions in
   https://github.com/artsy/potential/blob/master/platform/Kubernetes.md#hokusai
 
+### Development
```
joeyAghion, line 187: I think this needs to be indented further, so it's in the "Docker and Kubernetes setup" section. Otherwise it reads as if we're recommending doing development on docker, and I don't think we're there yet. 😅

README.md
```diff
@@ -51,10 +51,10 @@ cd metaphysics
 
 # Run the setup script
 
-source ./scripts/setup.sh
+yarn setup
```
joeyAghion, line 54: The setup script is responsible for installing `yarn`, so I don't think we should recommend running it this way. (I realize practically everyone already has `yarn` installed, but the setup script aims to work for a brand new developer and environment rather than make those assumptions.)

damassi: Ah, super good point 👍 Will revert

package.json
```diff
@@ -25,7 +25,9 @@
     "prepare": "patch-package",
     "prettier-project": "prettier --write 'src/**/*.{js,ts,tsx,md,graphql}'",
     "schema-drift": "babel-node --no-warnings --extensions '.ts,.js' ./scripts/schema-drift.ts",
+    "setup": " ./scripts/setup.sh",
```
ovasdi, line 28: I like this!
Will add to force tomorrow along with `sync-env` and check to test is shared config is out of date, unless you beat me to it :-)

scripts/setup.sh
damassi: > developers have them installed other [non-brew] ways
Yup, exactly. I've run into a lot of trouble in the past with multiple installs so its nice to check ahead of time.

package.json
damassi: Joey brought up a good point about missing `yarn` 🤦

ovasdi:
> Joey brought up a good point about missing yarn 🤦

Yea i saw it. So removed my comment.

scripts/setup.sh
```diff
@@ -6,17 +6,28 @@
 #
 # Commands that may fail have "|| return" to avoid continuing or interfering with terminal.
 
-if [ ! -z $NVM_DIR ]; then # skip nvm steps if not available
-  echo 'Configuring node 12...'
-  source ~/.nvm/nvm.sh
-  nvm install || return
+# Install yarn if it does not exist.
+if ! yarn versions &> /dev/null; then
```
ovasdi, line 10: What do you think of using this for checking if a tool exists? I like the conciseness and the fact that its not actually calling it. @pvinis shared this approach in [this slack thread](https://artsy.slack.com/archives/CP9P4KR35/p1626998393239400?thread_ts=1626988667.231100&cid=CP9P4KR35)
```suggestion
if ! which yarn > /dev/null; then
```

scripts/setup.sh
```diff
@@ -6,17 +6,28 @@
...
+  echo 'yarn is required for setup, installing...'
+  if ! brew --version &> /dev/null; then
```
ovasdi, line 12:
```suggestion
  if ! which brew > /dev/null; then
```

scripts/setup.sh
```diff
@@ -6,17 +6,28 @@
...
+    echo 'brew is required to install yarn, see https://docs.brew.sh/Installation'
+    exit 0
+  fi
+  brew install yarn
+fi
+
+if ! memcached --version &> /dev/null; then
```
ovasdi, line 19:
```suggestion
if ! which memcached > /dev/null; then
```

scripts/setup.sh
damassi: Much better 👌

scripts/setup.sh
```diff
@@ -6,17 +6,28 @@
 #
 # Commands that may fail have "|| return" to avoid continuing or interfering with terminal.
```
ovasdi, line 7: Because most of the `|| return` have been removed, perhaps we should remove the comment [L7](https://github.com/artsy/metaphysics/blob/3c79914b378e75b040a8074919070c6148835a03/scripts/setup.sh#L7) and update the comment on [L4](https://github.com/artsy/metaphysics/blob/3c79914b378e75b040a8074919070c6148835a03/scripts/setup.sh#L4) to match the README change:

```
Run like:
  ./scripts/setup.sh
```

scripts/setup.sh
```diff
@@ -6,17 +6,28 @@
...
 
-if [ ! -z $NVM_DIR ]; then # skip nvm steps if not available
-  echo 'Configuring node 12...'
-  source ~/.nvm/nvm.sh
-  nvm install || return
+# Install yarn if it does not exist.
+if ! which yarn > /dev/null; then
+  echo 'yarn is required for setup, installing...'
```
ovasdi, line 11:
```suggestion
  echo 'yarn is required for setup, installing with brew...'
```

scripts/setup.sh
```diff
@@ -6,17 +6,28 @@
...
+  if ! which brew > /dev/null; then
```
ovasdi, line 12: On second thought, I am proposing removing this entirely. The script [assumes](https://github.com/artsy/metaphysics/blob/3c79914b378e75b040a8074919070c6148835a03/scripts/setup.sh#L1) you have pre-reqs from [potential setup](https://github.com/artsy/potential/blob/main/scripts/setup) (which installs brew) and some contributors may want to install `yarn` by other means. So I think stopping the script here is redundant considering if brew is missing L16 will blow up and let the using know that `brew: command not found`. This will be more consistent with `memcached` install step below.

scripts/setup.sh
```diff
@@ -28,5 +39,5 @@ fi
 echo 'Updating .env.shared file (for shared configuration)...'
 aws s3 cp s3://artsy-citadel/dev/.env.metaphysics .env.shared || return
```
ovasdi, line 40: This is the last `|| return` so we may as well remove this too?
```suggestion
aws s3 cp s3://artsy-citadel/dev/.env.metaphysics .env.shared || 'Unable to download shared configuration, ensure you have S3 access!'
```

ovasdi:
@damassi after looking this over once more, I wanted to make a few more suggestions.

Last thing, consider adding `set -e` at the top so the script stops if it encounters an error in a couple of places its not guarded against with `echo`.

scripts/setup.sh
```diff
@@ -6,17 +6,28 @@
...
+if [[ ! -z $NVM_DIR ]]; then # skip if nvm is not available
+  echo "Installing Node..."
+  source ~/.nvm/nvm.sh
+  nvm install
+fi
 
-echo 'Installing node packages...'
-yarn install || return
+echo "Installing dependencies..."
+yarn install || echo 'Unable to install dependencies using yarn!'
```
ovasdi, line 30: Let's let this fail organically instead of echo and continue. If yarn does not install deps, the project wouldn't run anyway...
```suggestion
yarn install
```

damassi:
Feedback incorporated 👍

PR #3304, opened 2021-07-26
chore: refining isHangable logic to enable View in Room for more works with depth

lilyfromseattle:
# PURCHASE-2719
User can still see view 2D works on wall that now have Depth

While looking into this story, I realized we were already enabling VIR for works with a depth of >1. Based on [this discussion](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=41&projectKey=PURCHASE&modal=detail&selectedIssue=PURCHASE-2719&quickFilter=61) with stakeholders, I refined the logic to make `isHangable` always true and thus enable VIR for all works in certain categories (like "painting"), always false for all works in other categories (like "sound"), and for all other works, it is true if the depth of the work is less than 3.

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -471,12 +471,36 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
       isHangable: {
         type: GraphQLBoolean,
         resolve: (artwork) => {
-          const is3D =
-            _.includes(artwork.category, "sculpture") ||
-            _.includes(artwork.category, "installation") ||
-            _.includes(artwork.category, "design")
+          const categories3D = [
+            "installation",
+            "design",
+            "performance",
+            "sound",
+            "fashion",
+            "architecture",
+            "books_and_portfolios",
+            "jewelry",
+            "other",
+          ]
+          const categories2D = [
+            "print",
+            "drawing_collage_other_work_on_paper",
+            "painting",
+            "photography",
+            "posters",
+            "work_on_paper",
+          ]
 
-          return !is3D && isTwoDimensional(artwork) && !isTooBig(artwork)
+          const areIncluded = (category) =>
+            _.includes(artwork.category, category)
+
+          const is3DCategory = categories3D.some(areIncluded)
+          const is2DCategory = categories2D.some(areIncluded)
+
+          return (
+            (is2DCategory || (!is3DCategory && isTwoDimensional(artwork))) &&
+            !isTooBig(artwork)
+          )
         },
       },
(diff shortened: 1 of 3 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -471,12 +471,33 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+          const is2DCategory =
+            _.includes(artwork.category, "print") ||
+            _.includes(
+              artwork.category,
+              "drawing_collage_other_work_on_paper"
+            ) ||
+            _.includes(artwork.category, "painting") ||
+            _.includes(artwork.category, "photography") ||
+            _.includes(artwork.category, "posters") ||
+            _.includes(artwork.category, "work_on_paper")
```
jo-rs, line 495: There's a duplicate `sound`, but also maybe this can be less repetitive? Something like:
```
const categories3D = ["installation", "design", "performance", "sound", "fashion", "architecture", "books_and_portfolios", "jewelry", "other"]
const categories 2D = ["print", "drawing_collage_other_work_on_paper", "painting", "photography", "posters", "work_on_paper"]

const areIncluded = (category) => _.includes(artwork.category, category) 

const is3DCategory = categories3D.some(areIncluded)
const is2DCategory = categories2D.some(areIncluded)
```
I'm sure there are a million ways of simplifying this, but definitely non-blocking and up to you.

src/schema/v2/artwork/utilities.ts
```diff
@@ -1,7 +1,7 @@
 import { parse } from "url"
 
 export const isDimensional = (value) => parseFloat(value) > 0
-export const isTinyDimensional = (value) => parseFloat(value) > 1
+export const isTinyDimensional = (value) => parseFloat(value) > 3
```
jo-rs, line 4: Is the `3` coming from somewhere in particular or arbitrary? 🤔

jo-rs (approved):
Looks great! 💯 
Just a small non-blocking comment and a question.

jo-rs:
Also, don't forget the semantic PR title 😬

src/schema/v2/artwork/index.ts
lilyfromseattle: Totally forgot about this PR, my bad! Making changes now

src/schema/v2/artwork/utilities.ts
lilyfromseattle: Chosen by stakeholders

jo-rs (approved):
Nice nice 💯

PR #3308, opened 2021-07-29
feat: add root-level fairOrganizer field

mdole:
In our quest to update Force's fair organizer app, we realized that we don't have a good way of fetching a fair organizer from Metaphysics when it's not nested inside of something else.

In the "old" [implementation of the fair organizer app in Force](https://github.com/artsy/force/blob/bb0564b88c7aa06f407fd8e9a75577c3ad0f8f0f/src/desktop/apps/fair_organizer/index.coffee#L22), we hit Gravity's `/fair_organizer/:id` route, e.g. `/fair_organizer/the-armory-show`, which gives us back a very limited amount of information about the fair organizer - including the `profile_id`. We then use that ID as a param and pass it to Gravity's `/fairs` endpoint: `/fairs?fair_organizer_id=the-armory-show`. This gets us a paginated list of fairs associated with the fair organizer, which we can then display.

In Metaphysics right now, we [support](https://github.com/artsy/metaphysics/blob/629ff9048de8f497bafc168da1f434199c757bc9/src/schema/v2/fair.ts#L317-L319) fetching a fair organizer as a nested field inside a fair edition, but as far as I can tell, there's no way to just fetch a fair organizer based on a fair organizer ID or slug. There _is_ [a bit of code](https://github.com/artsy/metaphysics/blob/629ff9048de8f497bafc168da1f434199c757bc9/src/schema/v2/partner.ts#L45) in the `partner` endpoint that seems to indicate that you can fetch a `partner` with type `FairOrganizer`, but I believe this is actually incorrect - I have been unable to find anything in Gravity to indicate that a `Partner` can be a fair organizer. Let me know if I missed something though!

So: this PR adds a root-level `fairOrganizer` field and adds a nested `fairsConnection`, so that we can do it all in one go.

Last note: currently, our `get /fair_organizer` endpoint in Gravity [enforces](https://github.com/artsy/gravity/blob/889a590fe3d4109038c269b39acc4fa858487269/app/api/v1/fair_organizers_endpoint.rb#L62) an `admin true` block. I'm frankly not sure how Force deals with this and whether or not it would cause problems with this new field in production, but it was definitely stopping me from fetching results while testing. I got around it by removing the block and running MP against local grav, which worked fine. I'm opening a [separate PR in Gravity](https://github.com/artsy/gravity/pull/14326) to remove that requirement, because it seems strange/unnecessary for `get` requests. Leave a comment there if I'm wrong/it's unnecessary :)

After this PR, you can run something like this:
```graphql
{
  fairOrganizer(id:"the-armory-show") {
    name
    profileID
    website
    profile {
      id
      href
      isPubliclyVisible
    }
    fairsConnection(first: 5) {
      totalCount
      edges {
        cursor
        node {
         name
        }
      }
    }
  }
}
```

...and get back something like this:
```graphql
{
  "data": {
    "fairOrganizer": {
      "name": "The Armory Show",
      "profileID": "the-armory-show",
      "website": "https://www.thearmoryshow.com/",
      "profile": {
        "id": "UHJvZmlsZTo1NTFiMWE2OTcyNjE2OTJiNTJlMTBlMDA=",
        "href": "/the-armory-show",
        "isPubliclyVisible": true
      },
      "fairsConnection": {
        "totalCount": 8,
        "edges": [
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjA=",
            "node": {
              "name": "The Armory Show 2020"
            }
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjE=",
            "node": {
              "name": "The Armory Show 2019"
            }
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjI=",
            "node": {
              "name": "The Armory Show 2018"
            }
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjM=",
            "node": {
              "name": "The Armory Show 2017"
            }
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjQ=",
            "node": {
              "name": "The Armory Show 2016"
            }
          }
        ]
      }
    }
  },
  "extensions": {
    "requests": {
      "gravity": {
        "requests": {
          "fair_organizer/the-armory-show": {
            "time": "0s 1.923ms",
            "cache": true,
            "length": "N/A"
          },
          "profile/the-armory-show": {
            "time": "0s 1.129ms",
            "cache": true,
            "length": "N/A"
          },
          "fairs?fair_organizer_id=the-armory-show&page=1&size=5&total_count=true": {
            "time": "0s 1.136ms",
            "cache": true,
            "length": "N/A"
          }
        }
      }
    },
    "requestID": "d5492be0-f07e-11eb-ac59-1bb61fb25790"
  }
}
```

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -6232,7 +6232,7 @@ type Fair implements EntityWithFilterArtworksConnectionInterface & Node {
   marketingCollections(size: Int): [MarketingCollection]!
   mobileImage: Image
   name: String
-  organizer: organizer
+  organizer: FairOrganizer
   profile: Profile
 
   # This connection only supports forward pagination. We're replacing Relay's default cursor with one from Gravity.
@@ -6334,6 +6334,23 @@ type FairExhibitorsGroup {
 
   # Letter exhibitors group belongs to
   letter: String
+}
+
+type FairOrganizer {
+  about: String
+
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
+  name: String
+  profile: Profile
+  profileID: ID
+
+  # A slug ID.
+  slug: ID!
+  website: String
 }
 
 enum FairSorts {
@@ -8760,20 +8777,6 @@ enum OrderedSetSorts {
 
 union OrderParty = Partner | User
 
-type organizer {
-  # A globally unique ID.
-  id: ID!
-
-  # A type-specific ID likely used as a database ID.
-  internalID: ID!
-  profile: Profile
-  profileID: ID
-
-  # A slug ID.
-  slug: ID!
-  website: String
-}
-
 type PageCursor {
   # first cursor on the page
   cursor: String!
@@ -9996,6 +9999,12 @@ type Query {
     # The slug or ID of the Fair
     id: String!
   ): Fair
+
+  # A fair organizer, e.g. The Armory Show
+  fairOrganizer(
+    # The slug or ID of the Fair organizer
+    id: String!
+  ): FairOrganizer
 
   # A list of Fairs
   fairs(
@@ -12551,6 +12560,12 @@ type Viewer {
     # The slug or ID of the Fair
     id: String!
   ): Fair
+
+  # A fair organizer, e.g. The Armory Show
+  fairOrganizer(
+    # The slug or ID of the Fair organizer
+    id: String!
+  ): FairOrganizer
 
   # A list of Fairs
   fairs(
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -6338,6 +6338,12 @@ type FairExhibitorsGroup {
 
 type FairOrganizer {
   about: String
+  fairsConnection(
+    after: String
+    before: String
+    first: Int
+    last: Int
+  ): FairConnection
 
   # A globally unique ID.
   id: ID!
--- a/src/lib/loaders/loaders_without_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_without_authentication/gravity.ts
@@ -72,6 +72,7 @@ export default (opts) => {
       { headers: true }
     ),
     fairLoader: gravityLoader((id) => `fair/${id}`),
+    fairOrganizerLoader: gravityLoader((id) => `fair_organizer/${id}`),
     fairsLoader: gravityLoader("fairs", {}, { headers: true }),
     filterArtworksLoader: gravityLoader(
       "filter/artworks",
--- a/src/schema/v2/__tests__/fair_organizer.test.ts
+++ b/src/schema/v2/__tests__/fair_organizer.test.ts
@@ -1,0 +1,59 @@
+import gql from "lib/gql"
+import { runQuery } from "schema/v2/test/utils"
+
+describe("fairOrganizer", () => {
+  const query = gql`
+    {
+      fairOrganizer(id: "the-armory-show") {
+        name
+        website
+        profile {
+          href
+        }
+      }
+    }
+  `
+  const fairOrganizerLoader = jest.fn().mockResolvedValue(
+    Promise.resolve({
+      id: "the-armory-show",
+      name: "The Armory Show",
+      website: "https://www.thearmoryshow.com/",
+      profile_id: "the-armory-show",
+    })
+  )
+  const profileLoader = jest.fn().mockResolvedValue(
+    Promise.resolve({
+      id: "the-armory-show",
+      href: "/the-armory-show",
+    })
+  )
+
+  const context = {
+    fairOrganizerLoader: fairOrganizerLoader,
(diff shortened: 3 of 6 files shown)


_schemaV2.graphql
```diff
@@ -6336,6 +6336,29 @@ type FairExhibitorsGroup {
...
+  internalID: ID!
+  name: String
+  profile: Profile
+  profileID: ID
+
+  # A slug ID.
+  slug: ID!
+  website: String
+}
+
```
anastasiapyzhik, line 6361: Looks great! But according to [this figma spec](https://www.figma.com/file/aM3b5sodn04vMTigbQXACi/Fairs-2021?node-id=3139%3A5588) we want to get cover image, icon image, startAt/endAt dates, and social links that are missing here.

Does that make sense? I'm just a bit unclear on what exactly do we show on fair organizer page. Seems like there's info about some closest/main event of the organizer's in overview block cause it has start/end dates, _Art Paris: Art Fair_ writing on icon image. Or fair organizer is an event itself?

So my question is should we get above mentioned data from `fairOrganizer` endpoint or from `fairsConnection`?
In the first case it seems that the schema is not complete

mdole: @anastasiapyzhik those are really good questions, thanks for flagging!

I did some digging, and here's what I found:
- Cover image & icon image: these can be accessed via the `profile` field included on `fairOrganizer`:
```graphql
{
  fairOrganizer(id:"art-paris") {
    profile {
      icon {
        imageVersions
        url(version:"square") 
        # note that we store several different image types, listed with imageVersions; when we
        # implement these pages we should make sure we're using the right size for the situation!
      }
      image {
        imageVersions
        url(version:"large")
      }
    }
  }
}
```
- startAt/endAt: our [current implementation](https://github.com/artsy/force/blob/ac64ba6786e85b8231bfb5e5f2e419fd3bd3b5be/src/desktop/apps/fair_organizer/client/index.coffee#L46-L54) gets these dates from an upcoming fair if one exists (e.g. for Art Paris, those dates are from `art-paris-2021`). The fair organizer model itself does not store any information about editions of the fair. So we should use `fairsConnection` to fetch these dates going forward.
- Social links: OK, this is a really weird one 😅 we don't currently make these available via the API as individual links - they're simply part of a fair organizer's `about` text. For example, here's the "about" text that is returned from our API for Art Paris:
```
"Art Paris, the leading spring event for modern and contemporary art, supports and celebrates the French art scene and invites visitors to discover the Spanish and Portuguese art from the 1950s to the present day, from modern masters to contemporary artists.\r\n\r\nIn the 20 years since its first edition, Art Paris has become Paris’s must-see spring fair for modern and contemporary art, once again bringing together more than 150 galleries from over 20 different countries with a total of almost 1,000 artists.\r\n\r\nIn 2021, Art Paris will become the first art event to move into the ephemeral Grand Palais which will be installed on the Champ de Mars in Paris, while renovation work is carried out on the historic Parisian monument.\r\n\r\n**Follow the Fair**\r\n[Facebook](https://www.facebook.com/ArtParisArtFair/) | [Instagram](https://www.instagram.com/artparisartfair/) | [Twitter](https://twitter.com/ArtParisArtFair)"
```
So in this case, it seems like we will want to add new fields to our `FairOrganizer` model in Gravity to support storing these links in a structured way. I've ticketed that work [here](https://artsyproduct.atlassian.net/browse/FX-3141).

Also, pasting these links so you can quickly see what API responses for different fields look like in case it's useful:
Fair profile: https://stagingapi.artsy.net/api/v1/profiles?id[]=the-armory-show
Fairs filtered by fair organizer: https://stagingapi.artsy.net/api/v1/fairs?fair_organizer_id=the-armory-show
Fair organizer: https://stagingapi.artsy.net/api/v1/fair_organizer/the-armory-show

I hope that helps! Let me know if you have other questions :)

mdole:
> The V2 schema in this PR has breaking changes with Force. Remember to update the Force schema if necessary.
> Type 'organizer' was removed
> Field 'Fair.organizer' changed type from 'organizer' to 'FairOrganizer'

I'm not 100% sure whether or not this would cause problems for Eigen. Does renaming the type for a field cause queries from an older version of the schema to break, even if the same properties are available on the field? E.g. let's say we already have this query in the app:

```graphql
{
  fair(id: "the-armory-show-2017") {
    name
    organizer {
      slug
    }
  }
}
```
This would still be a valid query with the newer schema, but under the hood, the type of `organizer` has been renamed. I _think_ that should work fine, right? Would it cause us to get warnings or errors sent to Sentry even if it executes correctly though? 🤔

dblandin:
> Does renaming the type for a field cause queries from an older version of the schema to break, even if the same properties are available on the field?

I think this change should be pretty safe unless we have some very specific checks on `__typename` for the current type. Definitely worth QAing with the app build.

dblandin (approved):
Looks good to me 🎉

PR #3310, opened 2021-07-30
feat: expose profile and add owner type

mdole:
For comparison with #3308!


mdole:
Closing since we adopted https://github.com/artsy/metaphysics/pull/3308

PR #3314, opened 2021-08-03
Upgrading node to v14

chr-tatu:
## Description
Upgrading node to v14. Jira ticket: [PLATFORM-3459](https://artsyproduct.atlassian.net/browse/PLATFORM-3459)

## Additional info
I was trying to upgrade to node 16 initially, but one of the dependencies `appmetrics` doesn’t support Node 16. See https://github.com/RuntimeTools/appmetrics/blob/master/.travis.yml.

⚠️  After the upgrade one of the optional dependencies is broken: fsevents.
🥺  Please squash and merge.
🧪  The artsy/orbs#129 PR has to be merged and then the test rerun. I already tested with the canary yarn orb for 6.0.0. All tests passed.

## Related PRs
- artsy/orbs#128
- artsy/orbs#129


chr-tatu:
@icirellik : I rerun the pipeline after yarn@6.0.0 was deployed, but apparently because I committed the version change before the version was available, there is this step that looks failed in the pipeline, but actually everything is alright. If you check here, the last run of the pipeline was  green: https://app.circleci.com/pipelines/github/artsy/metaphysics?branch=chr-tatu%2FPLATFORM-3459-node-14

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "repository": "https://github.com/artsy/metaphysics",
   "engines": {
-    "node": "12",
+    "node": "14",
```
icirellik, line 7: What do you think about changing this to `>=12`? I feel like this would make the transition easier until all the node projects are on v14.

icirellik:
```suggestion
    "node": ">=12",
```

PR #3318, opened 2021-08-04
fix: add savedSearchesConnection arguments first/last/before

gkartalis:
# Description

savedSearchesConnection was unable to take args for connection and we needed to explicitly declare them on the gravity/v2/stiching.ts file like that:

` after: String
    before: String
    first: Int
    last: Int`


dimatretyak (approved):
🚀

dblandin (approved):

PR #3337, opened 2021-08-11
feature: Add include variable to artworkModules field

olerichter00:
Contributes towards [CX-1796] by adding an `include` variable to `artworkModules` to specify exactly which modules will be returned from the query. So far we can only exclude modules.

[CX-1796]: https://artsyproduct.atlassian.net/browse/CX-1796


leamotta (approved):
LGTM

PR #3338, opened 2021-08-11
fix: Fix Python Installation Path

icirellik:
In the latest CircleCI Node image python is installed as python 3, this
sets an alias so that yarn post install hooks execute successfully.


ovasdi (approved):
:shipit:

PR #3345, opened 2021-08-13
feat: expose algolia app id under system>algolia query field

dblandin:
Clients need both an app ID and API key to initialize an API client.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -76,6 +76,7 @@ type Agreement {
 
 type Algolia {
   apiKey: String
+  appID: String
 }
 
 type AnalyticsArtist {
--- a/src/config.ts
+++ b/src/config.ts
@@ -6,6 +6,7 @@ require("dotenv").config({
 })
 
 const {
+  ALGOLIA_APP_ID,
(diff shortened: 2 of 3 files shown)


_schemaV2.graphql
```diff
@@ -76,6 +76,7 @@ type Agreement {
 
 type Algolia {
   apiKey: String
+  appID: String
```
gkartalis, line 79: **observation:** noticed that inside this file in most of the cases we use camelCase when having id as the second part of the string i.e. `partnerId` vs `partnerID` although both looks good to me

gkartalis (approved):

gkartalis:
Looks good! 🚀

_schemaV2.graphql
dblandin: I think most of our V2 schema leans toward `somethingID` naming. These schema design [docs](https://github.com/artsy/README/blob/main/playbooks/graphql-schema-design.md#id-fields) in artsy/README mention IDs of objects in our database but kind of applies here as well.

gkartalis: Sounds good! feel free to merge whenever

PR #3352, opened 2021-08-17
Fix partner category query by properly passing category ID to partners endpoint

sweir27:
This PR fixes the bug reported in #product-find-explore, where all of the category-based carousels on the /galleries page were returning the same set of galleries.

I tracked this down to a bug in [this PR](https://github.com/artsy/metaphysics/pull/3259). When exposing the `partnerCategories` field in the v2 schema, I realized that we weren't properly converting camelCase to snake_case for all of the arguments under the sub-`partners` field. I saw an opportunity to simplify this resolver, but in doing so missed [a crucial piece of logic](https://github.com/artsy/metaphysics/pull/3259/files#diff-8e9b07cef48b76d0559107831dc93046d531c35b68dc8c9154052ab3d9b9e2a3L34)! (Where we explicitly pass the parent partner category as an argument).

This PR does the more verbose, but correct, thing and pulls in the full logic from the `partners` resolver (slightly modified). If you see a way to simplify, let me know!

### DIFF
--- a/src/schema/v2/partner_category.ts
+++ b/src/schema/v2/partner_category.ts
@@ -9,6 +9,7 @@ import {
   GraphQLBoolean,
 } from "graphql"
 import { ResolverContext } from "types/graphql"
+import { clone } from "lodash"
 
 export const PartnerCategoryType = new GraphQLObjectType<any, ResolverContext>({
   name: "PartnerCategory",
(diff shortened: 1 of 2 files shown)


src/schema/v2/partner_category.ts
```diff
@@ -27,7 +28,37 @@ export const PartnerCategoryType = new GraphQLObjectType<any, ResolverContext>({
...
+          { partnersLoader }
+        ) => {
+          const options: any = {
+            default_profile_public: defaultProfilePublic,
+            eligible_for_carousel: eligibleForCarousel,
+            eligible_for_listing: eligibleForListing,
+            eligible_for_primary_bucket: eligibleForPrimaryBucket,
+            eligible_for_secondary_bucket: eligibleForSecondaryBucket,
+            has_full_profile: hasFullProfile,
+            partner_categories: [id],
```
anandaroop, line 51: Ahh, good catch!

anandaroop (approved):
Thanks so much for this fix!

> If you see a way to simplify, let me know!

Since the new resolver differs only by one lil arg from the one its based on I think something like this could work:

```typescript
resolve: (parent, args, context) => {
  return Partners.resolve(
    parent,
    { ...args, partner_categories: [parent.id] },
    context
  )
},
```

I tried locally, and that _appears_ to give me the same results as the more explicit/verbose version in this PR. 

I leave it up to you if you want to vet this alternative — else I'm fine if we merge this as-is in order to get this 🐛 squashed!

sweir27:
AH thank you @anandaroop -- will update, that's much better. 😄

anandaroop:
As discussed in Slack I pushed up that change.

And we are finally 🟢 after some Circle memory issues 😅 — gonna merge!

PR #3353, opened 2021-08-17
feat(PX-4286): Users can see viewing rooms on the gallery profile page

Serge0n:
Jira Ticket: https://artsyproduct.atlassian.net/browse/PX-4286

This PR expands `ViewingRoom` with the `exhibitionPeriod` field
Related PR for Force: https://github.com/artsy/force/pull/8224

image

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -13014,6 +13014,7 @@ type ViewingRoom {
 
   # Datetime after which the viewing room is no longer viewable
   endAt: ISO8601DateTime
+  exhibitionPeriod: String
   heroImageURL: String @deprecated(reason: "Use image field instead")
   href: String
   image: ARImage
--- a/src/lib/stitching/gravity/__tests__/stitching.test.ts
+++ b/src/lib/stitching/gravity/__tests__/stitching.test.ts
@@ -470,6 +470,62 @@ describe("gravity/stitching", () => {
     })
   })
 
+  describe("#exhibitionPeriod", () => {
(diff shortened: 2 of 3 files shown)


ansor4 (changes requested):
I think this is fine, it seems like overkill to add this logic, but might be easier when adding Viewing Rooms view to Partner view in Eigen. I would just add tests for this new calculated field, I think this might be the right file:

https://github.com/artsy/metaphysics/blob/b076ddd353252783e72567ebb95b165ff4f5ca28/src/lib/stitching/gravity/__tests__/stitching.test.ts

Serge0n:
> I think this is fine, it seems like overkill to add this logic, but might be easier when adding Viewing Rooms view to Partner view in Eigen. I would just add tests for this new calculated field, I think this might be the right file:
> 
> https://github.com/artsy/metaphysics/blob/b076ddd353252783e72567ebb95b165ff4f5ca28/src/lib/stitching/gravity/__tests__/stitching.test.ts

I added couple tests. Do I understand correctly that I don't need to test all possible `exhibitionPeriod` values? Because it has already been tested in: https://github.com/artsy/metaphysics/blob/49e4b10a0f0f0e6589bd436511e6e1a4664b6a98/src/lib/date.ts#L157

ansor4 (approved):
Yeah I think this is fine for now, ideally we wouldn't want to have "Invalid date" shown, but I think that's out of scope for this PR. thanks!

PR #3360, opened 2021-08-24
feature: Slim Docker

icirellik:
This PR reduces the scope of our production image thus decreasing its
size and increasing its security.

The old size is 661 MiB
The new size is 271 MiB


artsyjian (approved):
nice! lgtm.

mc-jones (approved):

PR #3363, opened 2021-08-25
feat: supports additional fields for inquiries

dzucconi:
Adds additional fields to the mutations for me and collector profile; as well as exposing a few more on both types.


jonallured (approved):
LGTM!! 🎉

PR #3374, opened 2021-08-30
feat: stitch Gravity-backed algolia type into system root query field

dblandin:
Note: In the example below, we're using `__typename` to trigger the full query resolution. Without it, `system` comes back `null`. I've seen this behavior for other queries that only rely on stitched fields. It seems like you need some part of your query that is not relying on stitched resolvers for the full data resolution to trigger.

**Request**

```graphql
{
  system {
    __typename
    algolia {
      apiKey
      appID
      indices {
        name
        displayName
      }
    }
  }
}
```

**Response**

```json
{
  "data": {
    "system": {
      "__typename": "System",
      "algolia": {
        "apiKey": "<api key>",
        "appID": "<app id>",
        "indices": [
          {
            "name": "Artist_staging",
            "displayName": "Artist"
          }
        ]
      }
    }
  }
}
```


src/data/gravity.graphql
```diff
@@ -1231,6 +1247,19 @@ A partner agreement or errors object
 """
 union PartnerAgreementOrErrorsUnion = Errors | PartnerAgreement
 
+"""
+A location
+"""
+type PartnerLocation {
```
anandaroop, line 1253: I gather this [is a new type](https://github.com/artsy/gravity/pull/14388/files) that came along for the ride with this PR

anandaroop (approved):
LGTM!

PR #3379, opened 2021-08-31
feat(external): exposes galaxy entity fields

dzucconi:
This PR exposes auction houses, fairs, and galleries from Galaxy under a field named `external`.

I didn't make these connections for simplicities sake as they don't need to be paginated — they are just used for auto-complete.


src/schema/v2/External/External.ts
```diff
@@ -0,0 +1,92 @@
...
+  name: "External",
+  fields: {
+    auctionHouses: {
+      args: {
+        term: { type: GraphQLString },
+        size: { type: GraphQLInt },
+      },
+      type: new GraphQLNonNull(
+        new GraphQLList(new GraphQLNonNull(extenralAuctionHouseType))
+      ),
```
dzucconi, line 33: Shout out not having nullable types everywhere 🙄

src/schema/v2/External/ExternalAuctionHouse.ts
```diff
@@ -0,0 +1,50 @@
+import {
+  GraphQLInt,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLString,
+} from "graphql"
+import { ResolverContext } from "types/graphql"
+
+export interface Response {
```
dzucconi, line 9: Types here we just generated with QuickType then validated against the models in Galaxy.

jonallured (approved):
LGTM!! 🎉

PR #3393, opened 2021-09-03
Change saleMessage price formatting

copperkraft:
solves [https://artsyproduct.atlassian.net/browse/PURCHASE-2824]
might solve [https://artsyproduct.atlassian.net/browse/PURCHASE-2404]

Change saleMessage field to display price formatted at metaphysics instead of gravity 
Results: (as is => to be)
AUD $420 => A$420
HKD $420 => HK$420
$420 => US$420
AUD $400-420 => AU$400-AU$420

It affected only saleMessage field, but in other places similar approach could be taken. 
It could affect some tests in other projects as well as integrity tests.

### DIFF
--- a/src/lib/moneyHelpers.ts
+++ b/src/lib/moneyHelpers.ts
@@ -1,7 +1,7 @@
 import currencyCodes from "lib/currency_codes.json"
 import numeral from "numeral"
 
-const symbolOnly = ["USD", "GBP", "EUR", "MYR"]
+const symbolOnly = ["GBP", "EUR", "MYR"]
 
 /**
  * Adds the currency to a price and returns the display text.
@@ -42,11 +42,14 @@ export const isCurrencySupported = (currency: string): boolean => {
  * Builds price display text (e.g. "$100").
  */
 export const priceDisplayText = (
-  priceCents: number,
+  priceCents: number | [number, number],
   currency: string,
   format: string
 ): string => {
-  return currencyPrefix(currency) + priceAmount(priceCents, currency, format)
+  if (typeof priceCents === "number") {
+    return currencyPrefix(currency) + priceAmount(priceCents, currency, format)
+  }
+  return priceRangeDisplayText(priceCents[0], priceCents[1], currency, format)
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/lib/moneyHelpers.ts
```diff
@@ -42,11 +42,14 @@ export const isCurrencySupported = (currency: string): boolean => {
...
-  priceCents: number,
+  priceCents: number | [number, number],
   currency: string,
   format: string
 ): string => {
-  return currencyPrefix(currency) + priceAmount(priceCents, currency, format)
+  if (typeof priceCents === "number") {
+    return currencyPrefix(currency) + priceAmount(priceCents, currency, format)
+  }
+  return priceRangeDisplayText(priceCents[0], priceCents[1], currency, format)
```
erikdstock, line 52: Minor: This might be a little more readable as an `if/else` or even a ternary like `return (typeof PriceCents === 'number') ?  currencyPrefix(currency) + priceAmount(priceCents, currency, format) : priceRangeDisplayText(priceCents[0], priceCents[1], currency, format)` (or maybe not for that one)

src/schema/v2/artwork/index.ts
```diff
@@ -57,9 +57,10 @@ import { ArtworkContextGrids } from "./artworkContextGrids"
 import { PageInfoType } from "graphql-relay"
 import { getMicrofunnelDataByArtworkInternalID } from "../artist/targetSupply/utils/getMicrofunnelData"
 import { InquiryQuestionType } from "../inquiry_question"
+import { priceDisplayText } from "lib/moneyHelpers"
 
 const has_price_range = (price) => {
-  return new RegExp(/\-/).test(price)
+  return new RegExp(/-/).test(price)
```
erikdstock, line 63: what is this change doing?

erikdstock (commented):
Looking good here! Some minor comments.

src/schema/v2/artwork/index.ts
copperkraft: That's unnesesary escaping, it was displayed as a warning in IDE, so i removed it.

src/lib/moneyHelpers.ts
copperkraft: As priceRangeDisplayText calls priceRangeDisplay I look at number input as at recursion base case instead of simple OR statement.

ivan-badyulya (approved):

PR #3394, opened 2021-09-06
feat(PX-4369): add location to artwork schema

leamotta:
Part of https://artsyproduct.atlassian.net/browse/PX-4369
Add the possibility to query for artwork location, using the REST endpoint in gravity.

Screenshot 2021-09-06 at 16 16 17

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1461,6 +1461,7 @@ type Artwork implements Node & Searchable & Sellable {
   layers: [ArtworkLayer]
   listPrice: ListPrice
   literature(format: Format): String
+  location: Location
   manufacturer(format: Format): String
 
   # Represents the **materials** used in this work, such as _oil and acrylic on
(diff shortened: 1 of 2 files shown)


src/schema/v1/artwork/index.ts
```diff
@@ -501,6 +502,10 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
       literature: markdown(({ literature }) =>
         literature.replace(/^literature:\s+/i, "")
       ),
+      location: {
+        type: LocationType,
+        resolve: ({ location }) => location,
+      },
```
dblandin, line 508: **question**: Are there clients using Metaphysics V1 which will use this new field? If not, we could consider making this a V2-only addition.

dblandin (approved):
LGTM! - Left one question about V1 usage.

src/schema/v1/artwork/index.ts
leamotta: Not really, it will be used by Pulse, which queries the V2. I was unsure about this, seems fair to make it only a V2 addition

PR #3407, opened 2021-09-10
FX-3277 Send fair organizer id to gravity instead of it's pofile id

anastasiapyzhik:
[FX-3277]

In the current implementation when requesting fairsConnection of fairOrganizer we send to gravity fairOrganizer's profile id instead of it's own id and thus getting an error in cases those values are not equal 

**Before**

Screenshot 2021-09-10 at 15 57 04

**After**

Screenshot 2021-09-10 at 15 55 41

[FX-3277]: https://artsyproduct.atlassian.net/browse/FX-3277


mdole (approved):
Sweet! Thanks for the fix @anastasiapyzhik !

PR #3408, opened 2021-09-10
fix(FX-3294): Group fair exhibitors by numeric or special character within `#` group

dimatretyak:
solves [FX-3294] 

### Acceptance criteria
- Special characters will go first, then numbers

### Response
```javascript
{
  "data": {
    "fair": {
      "exhibitorsGroupedByName": [
        {
          "letter": "#",
          "exhibitors": [
            {
              "partner": {
                "name": "{Suit}"
              }
            },
            {
              "partner": {
                "name": "193 Gallery"
              }
            },
            {
              "partner": {
                "name": "313 Art Project"
              }
            },
          ]
        },
        {
          "letter": "A",
          "exhibitors": [
            {
              "partner": {
                "name": "A2Z Art Gallery"
              }
            },
          ]
        },
      ]
    }
  }
}
```

[FX-3294]: https://artsyproduct.atlassian.net/browse/FX-3294

### DIFF
--- a/src/schema/v2/fair.ts
+++ b/src/schema/v2/fair.ts
@@ -356,17 +356,16 @@ export const FairType = new GraphQLObjectType<any, ResolverContext>({
           if (!root._id) {
             return []
           }
+          interface Exhibitor {
+            name: string
+            profile_id: string
+            id: string
+            partner_id: string
+          }
           const exhibitor_groups: {
             [letter: string]: {
               letter: string
-              exhibitors: [
-                {
-                  name: string
-                  profile_id: string
-                  id: string
-                  partner_id: string
-                }
-              ]
+              exhibitors: Exhibitor[]
             }
           } = {}
           const fetch = allViaLoader(fairPartnersLoader, { path: root._id })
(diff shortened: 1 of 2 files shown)


src/schema/v2/fair.ts
```diff
@@ -381,7 +381,13 @@ export const FairType = new GraphQLObjectType<any, ResolverContext>({
             for (const fairExhibitor of fairExhibitors) {
               const names = fairExhibitor.name.split(" ")
               const firstName = names[0]
-              const letter = firstName.charAt(0).toUpperCase()
+              let letter = firstName.charAt(0).toUpperCase()
+
+              // Numeric or special symbol
+              if (/[^A-Z]/i.test(letter)) {
```
iskounen, line 387: **nitpick (non-blocking):** This regular expression will also catch exhibitors whose names start with accented characters like `Á` or `Ö`. I don't know if there are any galleries in our system that start with accented characters, so I am marking this as non-blocking. (I couldn't find any in the ongoing fairs.)

iskounen (commented):
LGTM!

src/schema/v2/fair.ts
dimatretyak: Yes, I also thought about this case, but I also didn't find such names in the system. To solve this problem, we can use `deburr` from [lodash](https://lodash.com/docs/4.17.15#deburr) and move exhibitors whose names start with accented characters (like, `Á`) into a group with the letter `A` ([on the frontend](https://github.com/artsy/force/blob/179755b196356b5b93681f288fe9273cd3dff8e4/src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx#L11), we can only choose from the values `A-Z`, `#`)

iskounen: That's a wonderful solution! It makes sense that any exhibitor starting with `Á`, `Å`, etc. would be grouped with `A`.

mdole: I think using `deburr` is a good idea; I also don't immediately see any partners whose names start with special characters but since it doesn't add a ton of complexity I think it's worth doing

mdole (approved):
Looks good to me - @dimatretyak I'll assign to you to merge after the `deburr` change is in!

PR #3410, opened 2021-09-10
fix: [HomePageArtworkModules] Respect exclude argument

damassi:
Noticed that the `exclude` key wasn't being fully respected. This fixes that and makes the logic a little more linear.


dzucconi (approved):
🙏

PR #3412, opened 2021-09-13
fix(FX-3300): return only running fairs

dimatretyak:
The type of this PR is: **Fix**
Solves: [FX-3300]

### Description
* Art Fair rail is appearing on the homepage, even though the fair has not yet launched. Clicking into the fair leads to a 404 error.
* Fairs should not be appearing on homepage rail ahead of start date.

### Demo
It's now **September 13**
```javascript
{
  "data": {
    "homePage": {
      "artworkModules": [
        {
          "key": "current_fairs",
          "context": {
            "name": "The Armory Show 2021",
            "exhibitionPeriod": "Sep 9 – 30"
          }
        }
      ]
    }
  }
}
```

GraphiQL 2021-09-13 15-30-05

[FX-3300]: https://artsyproduct.atlassian.net/browse/FX-3300

### DIFF
--- a/src/schema/v2/home/__tests__/home_page_artwork_modules.test.js
+++ b/src/schema/v2/home/__tests__/home_page_artwork_modules.test.js
@@ -1,5 +1,6 @@
 /* eslint-disable promise/always-return */
 import { map, find } from "lodash"
+import moment from "moment"
 import { runAuthenticatedQuery } from "schema/v2/test/utils"
 
 describe("HomePageArtworkModules", () => {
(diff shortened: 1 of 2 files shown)


src/schema/v2/home/fetch.ts
```diff
@@ -12,18 +12,26 @@ import {
...
 ) => {
   return fairsLoader({
     size: 5,
     active: true,
+    sort: "-start_at",
     has_homepage_section: true,
   }).then(({ body: fairs }) => {
     if (fairs.length) {
+      const now = moment.utc()
+      const runningFairs = filter(fairs, (fair) => {
```
iskounen, line 28: **question:** Should this method return `undefined` if there are no running fairs (like it does on line 42)?

src/schema/v2/home/__tests__/home_page_artwork_modules.test.js
```diff
@@ -355,4 +356,51 @@ describe("HomePageArtworkModules", () => {
       expect(keys).toEqual(["followed_galleries"])
     })
   })
+
+  it("doesn't return fairs that haven't opened yet", async () => {
```
iskounen, line 360: **suggestion:** It may help to have a test for the case where there are no running fairs.

iskounen (commented):
Thanks for looking into this problem!

src/schema/v2/home/fetch.ts
dimatretyak: That's a good point. I think what should still be returned in this case is `undefined`. For example, because of https://github.com/artsy/metaphysics/blob/1b0fe1d9f8554406e9139bfe04c1fd976077b10d/src/schema/v2/home/logged_out_modules.ts#L11 (if an empty array is returned, the check will pass)

src/schema/v2/home/fetch.ts
```diff
@@ -12,23 +12,33 @@ import {
...
 import { LoadersWithoutAuthentication } from "lib/loaders/loaders_without_authentication"
+import moment from "moment"
 
 export const featuredFair = (
   fairsLoader: LoadersWithoutAuthentication["fairsLoader"]
 ) => {
   return fairsLoader({
     size: 5,
     active: true,
+    sort: "-start_at",
```
dblandin, line 23: FYI, I believe the fairs API endpoint also supports a `status` param which you might be able to use in order to filter down to running fairs without the explicit datetime comparison here.

dimatretyak: You're right. The fair endpoint supports a status param. Thanks!

PR #3416, opened 2021-09-13
feature: Babel as a Development Dependency

icirellik:
Description:

Since babel is not needed at runtime this moves babel to be a development dependency.


damassi:
LGTM 👍

PR #3420, opened 2021-09-14
feat: Affinity score

olerichter00:
Resolves [CX-1816]

Depends on https://github.com/artsy/vortex/pull/323

[Tech Plan](https://www.notion.so/artsy/App-Homefeed-Make-Implicit-Collector-Interest-Available-44010887c7074d3eac6a6a9bdd9f8c68)

## Description

Adds a new query to fetch new works by artists the user interacted with.

**The Query:**

```
query HomeQuery {
  me {
    newWorksByInterestingArtists(first: 10) {
      pageInfo {
        hasNextPage
      }
      totalCount
      edges {
        node {
          internalID
          slug
        }
      }
    }
  }
}

```

[CX-1816]: https://artsyproduct.atlassian.net/browse/CX-1816


src/lib/apis/vortex.ts
```diff
@@ -0,0 +1,18 @@
+import urljoin from "url-join"
+import { assign } from "lodash"
+import fetch from "./fetch"
+import config from "config"
+
+const { VORTEX_API_BASE, VORTEX_TOKEN } = config
+
+export default (path, accessToken, fetchOptions = {}) => {
```
damassi, line 8: We should avoid using `default` exports when we can

src/lib/loaders/loaders_without_authentication/index.ts
```diff
@@ -5,6 +5,7 @@ import geminiLoaders from "./gemini"
 import geodataLoaders from "./geodata"
 import gravityLoaders from "./gravity"
 import positronLoaders from "./positron"
+import vortexLoaders from "./vortex"
```
damassi, line 8: Ditto about default export

src/lib/loaders/loaders_without_authentication/vortex.ts
```diff
@@ -0,0 +1,14 @@
+import factories from "../api"
+
+export default (opts) => {
```
damassi, line 3: Ditto about default export. (We'll also want to be sure to use a full function name so that we get better stack traces)

src/schema/v2/me/index.ts
```diff
@@ -38,6 +38,7 @@ import LotStanding from "./lot_standing"
 import LotStandings from "./lot_standings"
 import { MyBids } from "./myBids"
 import { MyCollection } from "./myCollection"
+import newWorksByInterestingArtists from "./new_works_by_interesting_artists"
```
damassi, line 41: Ditto

damassi (approved):
Looks really good @olerichter00. Just a couple super minor convention comments. 

Also, since this is a totally new field in the schema, can we update the description with a runable query [similar to here](https://github.com/artsy/metaphysics/pull/3407) and make reference to the field name in the PR title?

olerichter00:
#squashongreen

PR #3425, opened 2021-09-15
feat: [Profile] Add `location` field

damassi:
Adds a new `location` field to the `Profile` type and a little bit of alphabetization cleanup.

### DIFF
--- a/src/schema/v2/profile.ts
+++ b/src/schema/v2/profile.ts
@@ -20,10 +20,6 @@ export const ProfileType = new GraphQLObjectType<any, ResolverContext>({
     bio: {
       type: GraphQLString,
     },
-    fullBio: {
-      type: GraphQLString,
-      resolve: ({ full_bio }) => full_bio,
-    },
     counts: {
       resolve: (profile) => profile,
       type: new GraphQLObjectType<any, ResolverContext>({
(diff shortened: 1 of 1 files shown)


src/schema/v2/profile.ts
```diff
@@ -63,17 +58,26 @@ export const ProfileType = new GraphQLObjectType<any, ResolverContext>({
...
     },
+    isPubliclyVisible: {
+      type: GraphQLBoolean,
+      resolve: (profile) => profile && profile.published && !profile.private,
+    },
     isPublished: {
       type: GraphQLBoolean,
       resolve: ({ published }) => published,
     },
+    location: {
```
damassi, line 69: Only addition -- the rest is just alphabetization for better readability

src/schema/v2/profile.ts
```diff
@@ -63,17 +58,26 @@ export const ProfileType = new GraphQLObjectType<any, ResolverContext>({
...
+      type: GraphQLString,
+      resolve: ({ location }) => location,
```
dzucconi, line 71: This isn't a proper structured location?

damassi: Not sure I follow

dzucconi: Ah OK. Both user profiles and shows have locations that are more like this: https://github.com/artsy/metaphysics/blob/master/src/schema/v2/me/myLocation.ts

dzucconi (approved):

PR #3426, opened 2021-09-16
fix(PX-4576): use "Shipping: Сalculated in checkout" for artworks with arta shipping

kiryl-zubarau:
This PR fixes shipping label for artworks with Arta shipping.

Discussed here - https://artsy.slack.com/archives/C9ZJYFDNF/p1629972053009400
JIRA - https://artsyproduct.atlassian.net/browse/PX-4576

image

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -2118,7 +2118,7 @@ describe("Artwork type", () => {
       return runQuery(query, context).then((data) => {
         expect(data).toEqual({
           artwork: {
-            shippingInfo: "Shipping: calculated at checkout",
+            shippingInfo: "Shipping: Calculated in checkout",
           },
         })
       })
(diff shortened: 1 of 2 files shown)


src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -2112,6 +2112,18 @@ describe("Artwork type", () => {
       })
     })
 
+    it("is set to calculated at checkout when Arta shipping", () => {
+      artwork.arta_enabled = true
+
+      return runQuery(query, context).then((data) => {
+        expect(data).toEqual({
+          artwork: {
+            shippingInfo: "Shipping: Calculated in checkout",
```
oxaudo, line 2121: I assume that was approved by Joh. Not an expert here but `IN` seems like a strange thing to use here. But what do I know:)

oxaudo (approved):
Looks good to me!

src/schema/v2/artwork/__tests__/artwork.test.js
kiryl-zubarau: I'm also not an expert 🙂. I took latest variant from JIRA.

PR #3432, opened 2021-09-17
feat: support adding country code to address field

ashleyjelks:
To support phone number validation, we need to add a  phone number country code field to the Address model in [Gravity](https://github.com/artsy/gravity/pull/14447) and enable storing values in the db via MP mutations. This PR introduces the necessary schema updates to support this ch.

https://artsyproduct.atlassian.net/browse/PURCHASE-2938

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -15087,6 +15087,9 @@ type UserAddress {
   # Phone number
   phoneNumber: String
 
+  # Phone number country code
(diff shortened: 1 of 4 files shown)


erikdstock (approved):

starsirius (approved):
LGTM! We might want to merge and deploy this PR _after_ the upstream Gravity change just to be safe.

PR #3433, opened 2021-09-18
fix: Remove duplicated "/api" from vortexGraphqlLoader path

olerichter00:
Addresses [CX-1891]

## Description

Removes duplicated `/api` from `vortexGraphqlLoader` path.

[CX-1891]: https://artsyproduct.atlassian.net/browse/CX-1891


olerichter00:
#squashongreen

PR #3435, opened 2021-09-20
fix: Filter out sold artworks for newWorksByInterestingArtists

olerichter00:
Addresses [CX-1891]

Changes the `newWorksByInterestingArtists` resolver in order to filter out sold artworks for.

[CX-1891]: https://artsyproduct.atlassian.net/browse/CX-1891


araujobarret (approved):
All good 🚀

PR #3438, opened 2021-09-21
docs: Copy schema design doc over from playbook

damassi:
Copying this over from [our playbook](https://github.com/artsy/README/blob/43c400d81ff9fee7276c3dd934de26b985da362f/playbooks/graphql-schema-design.md) (which feels a bit buried there).


joeyAghion (commented):
Can we just link to the shared playbook? I hate the thought of this being duplicated, and I think the doc _is_ applicable to more systems than just Metaphysics.

damassi:
Ah, a link is ideal; will do 👍

PR #3441, opened 2021-09-21
feat: [Home] Use HomePageArtworkModuleTypes enum

damassi:
Noticed that we weren't getting proper types / intellisense for requesting home modules, looked, and realized that the enum wasn't being assigned: 

Screen Shot 2021-09-21 at 4 36 07 PM

Have looked through Force and Eigen and it appears this field isn't yet being used, though it might be accessed in some old version of the mobile app. Thoughts?

### DIFF
--- a/src/schema/v2/home/home_page_artwork_module.ts
+++ b/src/schema/v2/home/home_page_artwork_module.ts
@@ -15,6 +15,7 @@ import {
   GraphQLFieldConfig,
 } from "graphql"
 import { ResolverContext } from "types/graphql"
+import { HomePageArtworkModuleTypes } from "./types"
 
 let possibleArgs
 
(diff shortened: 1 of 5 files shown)


src/schema/v2/home/home_page_artwork_module.ts
```diff
@@ -61,7 +62,7 @@ const HomePageArtworkModule: GraphQLFieldConfig<void, ResolverContext> = {
       description: "ID of generic gene rail to target",
     },
     key: {
-      type: GraphQLString,
```
damassi, line 64: The only real change

dzucconi:
This will be a breaking schema change though? We'd have to have a safe migration path with like a duplicate argument, deprecate the old one, migrate force/eigen, remove deprecated argument?

dzucconi:
However yeah, I've seen us doing this in a few places (using strings instead of enums) and it's very frustrating 😓

damassi:
> deprecate the old one, migrate force/eigen, remove deprecated argument?

Looking through force and eigen this field isn't used anywhere, so there's nothing to migrate. Is that what you mean?

PR #3453, opened 2021-10-01
Fix price range display when price is in the lowest or highest bucket (PURCHASE-2970)

starsirius:
Fixes https://artsyproduct.atlassian.net/browse/PURCHASE-2970

When an artwork is set as displaying price range and is in the lowest (`[nil, 1000]`) or highest (`[10_000_000, nil]`) bucket, the price is displayed incorrectly with an exact price, rather than the "Under [price]" or "[price] and up" language.

This fixes it by updating the `priceRangeDisplayText` logic and the `saleMessage` resolver. More details inline.

It's not ideal there if duplicate and inconsistent formatting logic in MP and Gravity that deserves a bigger discussion. This only fixes the immediate issue.

#### Price in the lowest bucket
Screen Shot 2021-09-30 at 11 46 45 PM
Screen Shot 2021-09-30 at 11 47 05 PM

#### Price in the highest bucket
Screen Shot 2021-10-01 at 12 36 03 AM
Screen Shot 2021-10-01 at 12 36 22 AM

### DIFF
--- a/src/lib/moneyHelpers.ts
+++ b/src/lib/moneyHelpers.ts
@@ -39,21 +39,18 @@ export const isCurrencySupported = (currency: string): boolean => {
 }
 
 /**
- * Builds price display text (e.g. "$100").
+ * Builds price display text (e.g. "US$100").
  */
 export const priceDisplayText = (
-  priceCents: number | [number, number],
+  priceCents: number,
   currency: string,
   format: string
 ): string => {
-  if (typeof priceCents === "number") {
-    return currencyPrefix(currency) + priceAmount(priceCents, currency, format)
-  }
-  return priceRangeDisplayText(priceCents[0], priceCents[1], currency, format)
+  return currencyPrefix(currency) + priceAmount(priceCents, currency, format)
 }
 
 /**
- * Builds price range display text (e.g. "$100–$200" or "VUV Vt100–Vt200")..
+ * Builds price range display text (e.g. "US$100–US$200" or "VUV Vt100–Vt200")..
  */
 export const priceRangeDisplayText = (
   lowerPriceCents: number | null,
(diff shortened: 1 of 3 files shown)


src/lib/moneyHelpers.ts
```diff
@@ -39,49 +39,46 @@ export const isCurrencySupported = (currency: string): boolean => {
 }
 
 /**
- * Builds price display text (e.g. "$100").
+ * Builds price display text (e.g. "US$100").
  */
 export const priceDisplayText = (
-  priceCents: number | [number, number],
```
starsirius, line 45: Reverting this change from https://github.com/artsy/metaphysics/pull/3393 to make the responsibility clear: formatting a single number.

src/lib/moneyHelpers.ts
```diff
@@ -39,49 +39,46 @@ export const isCurrencySupported = (currency: string): boolean => {
...
 
 /**
- * Builds price range display text (e.g. "$100–$200" or "VUV Vt100–Vt200")..
+ * Builds price range display text (e.g. "US$100–US$200" or "VUV Vt100–Vt200")..
  */
 export const priceRangeDisplayText = (
-  lowerPriceCents: number,
-  higherPriceCents: number,
+  lowerPriceCents: number | null,
+  higherPriceCents: number | null,
```
starsirius, line 57: [`price_cents`](https://github.com/artsy/gravity/blob/8b1259a48a655aee0d9aa929577c5d08ac8e7e42/app/models/util/for_sale.rb#L208-L220) array from Gravity could return unbounded on either end, for example [`[nil, 100000]`](https://github.com/artsy/gravity/blob/eaf860818d504442f5d666f3cf43667cc5116571/spec/models/domain/artwork_spec.rb#L2037-L2043). This updates the logic of price range display to behave in a similar way as the [money_helper](https://github.com/artsy/money_helper/blob/8fb2110afa122c4ae32ba557000d2c0521fb2bf8/spec/money_helper_spec.rb#L199-L205) gem.

olerichter00 (approved):
Looks good to me 👍 Feel free to merge whenever you're ready.

starsirius:
Merging and giving this a try! Thanks for the review, @olerichter00 !

PR #3458, opened 2021-10-04
fix(FX-3421): pass fair id instead of profile id

dimatretyak:
Jira ticket: [FX-3421]

### Description
We passed `profileID` instead of `fair ID`. As a result, we received an error message `Fair Organizer Not Found`

GraphiQL 2021-10-04 17-33-50


mdole (approved):
LGTM - going to go ahead and merge this so that we can get the fix on production. Ty @dimatretyak !

PR #3459, opened 2021-10-04
fix(city): throws 404 to support principal field

dzucconi:
Allows us to support 404s for city pages in Force.

### DIFF
--- a/src/schema/v2/city/index.ts
+++ b/src/schema/v2/city/index.ts
@@ -27,6 +27,7 @@ import { StaticPathLoader } from "lib/loaders/api/loader_interface"
 import { BodyAndHeaders } from "lib/loaders"
 import { sponsoredContentForCity } from "lib/sponsoredContent"
 import { createPageCursors } from "../fields/pagination"
+import { HTTPError } from "lib/HTTPError"
 
 export interface TCity {
   slug: string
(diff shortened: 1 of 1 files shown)


src/schema/v2/city/index.ts
```diff
@@ -234,7 +235,7 @@ const lookupCity = (slug: string, cities: TCity[]) => {
   const city = cities.find((c) => c.slug === slug)
 
   if (!city) {
-    throw new Error(`City "${slug}" not found`)
+    throw new HTTPError(`City "${slug}" not found`, 404)
```
damassi, line 238: Didn't know we needed to use this to support `@principleField` 👍 

I'll add a todo about adding a note around here https://github.com/artsy/metaphysics/blob/master/src/lib/getPrincipalFieldDirectivePath.ts

damassi: I'd also be helpful to co-locate all of this principle field stuff under a `directives/principleField` directory rather than spreading it throughout the codebase: 

Screen Shot 2021-10-04 at 4 22 26 PM
:

mzikherman:
Did this wind up working? If so, sweet!

src/schema/v2/city/index.ts
dzucconi: This isn't explicitly a principal field thing, I suppose. We only need to raise this type of error here because we're not getting a status code from the server since it's just being thrown in a hash lookup.

dzucconi:
It does indeed @mzikherman

PR #3483, opened 2021-10-15
fix: select causality state using lot id instead of index

erikdstock:
This PR corrects a bug in myBids where causality lot state is paired with the wrong lot. It also adds types, improves our testing setup and assertions, and includes a light refactor. There are some clarifications in here that we should check against existing behavior:
- watched lot without a lot standing (bid on lot): `isHighestBidder` will be false and `lotState` will be `null`
- closed lots will have watched-only lots removed

- testing asserts on ordering of lots as well as presence of watched-only lots in active vs closed
- testing increases number of expected lots so that incorrectly-paired lots will break the test (i think).

If we aren't comfortable with a change this size, the first commit (plus some of the testing diff) should do the trick.

paired with @laurabeth with 👀 from @jo-rs


erikdstock:
#squashongreen

PR #3486, opened 2021-10-17
Revert "fix: select causality state using lot id instead of index"

erikdstock:
Reverts artsy/metaphysics#3483

This looked okay in testing but broke my bids in staging. Reverting for now.


erikdstock:
#mergeongreen

PR #3491, opened 2021-10-19
fix: isWatching should be false if a lot has a bid

erikdstock:
https://github.com/artsy/metaphysics/pull/3488 Is in staging and a major bugfix, but it has a bug - `SaleArtwork.isWatching` should only be true if the user is watching _but_ has not bid on a lot. This PR fixes that and makes the logic a little simpler as well.

Verified in insomnia and updated tests.


erikdstock:
#squashongreen

PR #3518, opened 2021-11-01
feat: Add flag to exclude purchased artworks from My Collection query

olerichter00:
Addresses [CX-1868]

## Description

Adds a flag to exclude purchased artworks from "My Collection" query. This is necessary to only show artworks that have been automatically added to "My Collection" to specific users until we release the feature for all users.

[CX-1868]: https://artsyproduct.atlassian.net/browse/CX-1868?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/me/myCollection.ts
+++ b/src/schema/v2/me/myCollection.ts
@@ -69,6 +69,7 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
     },
     excludePurchasedArtworks: {
       type: GraphQLBoolean,
+      defaultValue: false,
       description:
         "Exclude artworks that have been purchased on Artsy and automatically added to the collection.",
     },
--- a/src/schema/v2/me/myCollection.ts
+++ b/src/schema/v2/me/myCollection.ts
@@ -67,13 +67,23 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
         },
       }),
     },
+    excludePurchasedArtworks: {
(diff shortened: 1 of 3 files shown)


src/schema/v2/me/myCollection.ts
```diff
@@ -67,13 +67,23 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
         },
       }),
     },
+    excludePurchasedArtworks: {
+      type: GraphQLBoolean,
```
olerichter00, line 71: @MounirDhahri feel free to commit and add a default value here:

```suggestion
      type: GraphQLBoolean,
      defaultValue: false,
```

MounirDhahri: 🙏

MounirDhahri (approved):
👍👍👍

olerichter00:
#squashongreen

src/schema/v2/me/myCollection.ts
```diff
@@ -67,13 +67,24 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
...
+      defaultValue: false,
```
brainbicycle, line 72: should this be true instead for backwards compatibility?

MounirDhahri:
😬 good catch

PR #3525, opened 2021-11-04
fix: default to excluding purchased artworks from my collection 

brainbicycle:
This changes the default behavior of metaphysics to ask gravity to exclude users imported purchases from their collection. This is to maintain backwards compatibility with old clients until the feature is ready. 

Conversation here on justification: https://github.com/artsy/gravity/pull/14610


jonallured (approved):
LGTM!!

lordkiz (approved):

PR #3531, opened 2021-11-08
feat(schema): exposes partner category

dzucconi:
Closes [GRO-602](https://artsyproduct.atlassian.net/browse/GRO-602)

So this apparently existed but was disabled with [this commit](https://github.com/artsy/metaphysics/commit/dde8efc4b76dcfbeb3bf2d68bbb09ce42bab04ad). This PR re-enables the field `partnerCategory(id)` field.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10720,6 +10720,12 @@ type Query {
     size: Int
   ): [PartnerCategory]
 
+  # A PartnerCategory
+  partnerCategory(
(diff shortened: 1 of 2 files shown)


damassi:
Some backstory is that when we moved to MPv2, we disabled anything that wasn't being actively used (or something like that), only to then opt back in when the need arose.

dzucconi:
Thanks @damassi — figured you'd have the story.

PR #3546, opened 2021-11-17
feat(SWA-60): phone number validation

kiryl-zubarau:
This PR adds phone number validation that uses `google-libphonenumber` package

JIRA - https://artsyproduct.atlassian.net/browse/SWA-60

My proposal consists of: 

- Prevalidate phone numbers before usage
    1. If the phone number is valid - get region code(for country select) and phone number in the national format(for input, but not sure that is a good idea)
    2. If the phone number is not valid - use the default region(USA) and empty input.
- Save phone numbers in international format

**Phone number parsing**
Before validation phone number should be parsed by `google-libphonenumber`. For now, if `regionCode`(optional argument) is empty I try to parse a phone number as a phone in international format if it doesn't work I try to parse using the most expectable regions(US and GB for now). The problem here is that GB phone number in national format may be a valid US phone number if I left `regionCode` empty. Example:
image

**National format**
Some countries 😄   add not only zero symbols before the number:
image

So I'm not sure that it will look good at the input:
 | +375 v | 8 029 516-69-75  (imagine phone input)

Some other examples:
image


kiryl-zubarau:
With valid phone number: 

https://user-images.githubusercontent.com/79979820/143856533-70cfa27a-ff4f-4b1e-911b-d51319686a54.mov

With invalid phone number: 

https://user-images.githubusercontent.com/79979820/143856979-af9994b5-599b-49df-b888-a215b3c4b18c.mov

Validation: 

https://user-images.githubusercontent.com/79979820/143857450-74cf328b-20bc-4f9d-8800-0151decd2371.mov

src/schema/v2/phoneNumber.ts
```diff
@@ -0,0 +1,126 @@
...
+
+            case PhoneNumberUtil.ValidationResult.TOO_LONG:
+              return "Too long"
+
+            case PhoneNumberUtil.ValidationResult.INVALID_COUNTRY_CODE:
+              return "Invalid country code"
+          }
+
+          if (!phoneUtil.isValidNumber(parsedPhone)) {
+            return "Invalid number"
```
iskounen, line 69: **suggestion:** Instead of returning a string, I think that it may be more useful for clients if Metaphysics returns an enum value from `{ TOO_SHORT, TOO_LONG, INVALID_COUNTRY_CODE, INVALID_NUMBER }`. What do you think?

iskounen (commented):
Thank you for explaining how we decided to pre-validate existing phone numbers, and for providing videos.

iskounen:
@damassi, I've requested your review on this PR because I think that the new `phoneNumber` query can be used across our system (account settings, BNMO checkout, Sell with Artsy, shipping addresses) and would like to see if you have any thoughts about it.

_schemaV2.graphql
```diff
@@ -9922,6 +9923,29 @@ enum PartnersSortType {
...
+  E164
+  INTERNATIONAL
+  NATIONAL
+  RFC3966
+}
+
+type PhoneNumberType {
+  countryCode: String
+  display(format: PhoneNumberFormats): String
+  error: PhoneNumberErrors
```
iskounen, line 9943: **question (non-blocking):** Is this the pattern that other queries use to report errors in Metaphysics?

iskounen (approved):

_schemaV2.graphql
kiryl-zubarau: honestly i don't know about patterns that metaphysics uses to report errors. I just added an enum with possible errors and change field name from `errorMessage` to `error`.

iskounen: Got it. I think we can launch anonymous submission with this implementation, and follow-up with another ticket if there is a pattern that we should follow.

@mzikherman is there a preferred pattern that Metaphysics queries should follow for reporting errors to clients?

PR #3550, opened 2021-11-18
feat: short exhibition dates for shows + fairs

brainbicycle:
Add a `shortExhibitionPeriod` property that shows exhibition period display text in the old style with shortened months. In the app in some places we are displaying these dates we are pretty space constrained so in order to prevent wrapping to next line or cutting off text as much as possible we should use this instead. 

Instead of: `"February 25 – May 24, 2015"` => `"Feb 25 - May 24, 2015"`

Way around my hacky way of fixing this in Eigen: https://github.com/artsy/eigen/pull/5802

This logic was updated here to show longer months and it does look nicer so we should use it still where we can: https://github.com/artsy/metaphysics/pull/3497

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -6603,6 +6603,9 @@ type Fair implements EntityWithFilterArtworksConnectionInterface & Node {
   organizer: FairOrganizer
   profile: Profile
 
+  # A formatted description of the start to end dates with abbreviated months
+  shortExhibitionPeriod: String
+
   # This connection only supports forward pagination. We're replacing Relay's default cursor with one from Gravity.
   showsConnection(
     after: String
@@ -12118,6 +12121,9 @@ type Show implements EntityWithFilterArtworksConnectionInterface & Node {
   # Link to the press release for this show
   pressReleaseUrl: String
 
+  # A formatted description of the start to end dates with abbreviated months
+  shortExhibitionPeriod: String
+
   # A slug ID.
   slug: ID!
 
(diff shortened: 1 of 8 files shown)


pvinis:
I wonder if there is a field (or if it makes sense to add one) that returns two dates that the clients can display as needed 🤔.

brainbicycle:
> I wonder if there is a field (or if it makes sense to add one) that returns two dates that the clients can display as needed 🤔.

there is already but the display logic is kind of complicated and think if we leave up to clients it will drift as we have seen in past with stuff like auction time display

mdole (approved):
good thinking @brainbicycle! we didn't consider the space constraints of mobile devices when shipping the original change 😅

dblandin:
I'm sorry for the post-merge feedback, but one thought I had was if this could be supported by an argument on the existing `exhibitionPeriod` rather than introducing a new query field.

```graphql
{
  fair(id: "example") {
    exhibitionPeriod # "February 25 – May 24, 2015"
  }
}
```

```graphql
{
  fair(id: "example") {
    exhibitionPeriod(format: SHORT) # "Feb 25 - May 24, 2015"
  }
}
```

brainbicycle:
> I'm sorry for the post-merge feedback, but one thought I had was if this could be supported by an argument on the existing `exhibitionPeriod` rather than introducing a new query field.
> 
> ```graphql
> {
>   fair(id: "example") {
>     exhibitionPeriod # "February 25 – May 24, 2015"
>   }
> }
> ```
> 
>  ```graphql
> {
>   fair(id: "example") {
>     exhibitionPeriod(format: SHORT) # "Feb 25 - May 24, 2015"
>   }
> }
> ```

yeah this is a good suggestion thanks, I may ask for a pair on updating the field to a function if I don't get too far

PR #3555, opened 2021-11-22
feat(CX-2131): Add artist recommendations

olerichter00:
Addresses [CX-2131]

## Description

Adds a new `artistRecommendations` connection to `me`. The connection fetches artist recommendations for the current user from **Vortex** and the artist details from **Gravity**.

### Query

```graphql
query ArtistRecommendationsQuery {
  me {
    artistRecommendations(first: 10) {
      totalCount
      edges {
        node {
          artistId
          score
        }
      }
    }
 }
}
```

[CX-2120]: https://artsyproduct.atlassian.net/browse/CX-2120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CX-2131]: https://artsyproduct.atlassian.net/browse/CX-2131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/me/__tests__/artistRecommendations.test.ts
+++ b/src/schema/v2/me/__tests__/artistRecommendations.test.ts
@@ -128,31 +128,43 @@ const mockVortexResponse = {
   },
 }
 
-const mockArtistsResponse = [
-  {
-    _id: "608a7417bdfbd1a789ba092a",
-    artists_count: 3,
-    birthday: "",
-    blurb: "",
-    consignable: false,
-    deathday: "",
-    forsale_artists_count: 2,
-    group_indicator: "individual",
-    id: "yayoi-kusama",
-    image_url: null,
-    image_urls: {},
-    image_versions: [],
-    medium_known_for: null,
-    name: "Yayoi Kusama",
-    nationality: "",
-    original_height: null,
-    original_width: null,
-    public: true,
-    published_artists_count: 3,
-    sortable_id: "kusama-yayoi",
-    target_supply: false,
-    target_supply_priority: null,
-    target_supply_type: null,
-    years: "",
-  },
-]
+const mockArtistsResponse = {
+  body: [
+    {
+      _id: "5d6f9f7e3a583e000e51956d",
+      artworks_count: 1,
+      birthday: "",
+      blurb: "",
+      consignable: false,
+      deathday: "",
+      forsale_artworks_count: 0,
+      group_indicator: "individual",
+      id: "1-plus-1-plus-1",
+      image_url:
+        "https://d32dm0rphc51dk.cloudfront.net/jS7hjyXq3OKxNqWZPJeLPg/:version.jpg",
+      image_urls: {
+        square:
+          "https://d32dm0rphc51dk.cloudfront.net/jS7hjyXq3OKxNqWZPJeLPg/square.jpg",
+        four_thirds:
+          "https://d32dm0rphc51dk.cloudfront.net/jS7hjyXq3OKxNqWZPJeLPg/four_thirds.jpg",
+        large:
+          "https://d32dm0rphc51dk.cloudfront.net/jS7hjyXq3OKxNqWZPJeLPg/large.jpg",
+        tall:
+          "https://d32dm0rphc51dk.cloudfront.net/jS7hjyXq3OKxNqWZPJeLPg/tall.jpg",
+      },
+      image_versions: ["square", "four_thirds", "large", "tall"],
+      medium_known_for: null,
+      name: "1+1+1",
+      nationality: "Swedish Icelandic Finnish",
+      original_height: 5237,
+      original_width: 3491,
+      public: true,
+      published_artworks_count: 0,
+      sortable_id: "1-plus-1-plus-1",
+      target_supply: false,
+      target_supply_priority: null,
+      target_supply_type: null,
+      years: "",
+    },
+  ],
+}
--- a/src/schema/v2/me/__tests__/artistRecommendations.test.ts
+++ b/src/schema/v2/me/__tests__/artistRecommendations.test.ts
@@ -40,9 +40,9 @@ describe("artistRecommendations", () => {
         "edges": Array [
           Object {
             "node": Object {
-              "id": "QXJ0aXN0OjYwOGE3NDE3YmRmYmQxYTc4OWJhMDkyYQ==",
-              "internalID": "608a7417bdfbd1a789ba092a",
-              "slug": "yayoi-kusama",
+              "id": "QXJ0aXN0OjVkNmY5ZjdlM2E1ODNlMDAwZTUxOTU2ZA==",
+              "internalID": "5d6f9f7e3a583e000e51956d",
+              "slug": "1-plus-1-plus-1",
             },
           },
         ],
--- a/src/schema/v2/me/__tests__/artistRecommendations.test.ts
+++ b/src/schema/v2/me/__tests__/artistRecommendations.test.ts
@@ -12,7 +12,6 @@ describe("artistRecommendations", () => {
             node {
               internalID
               slug
-              id
             }
           }
         }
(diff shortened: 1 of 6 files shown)


shaq31415926:
I would change the query to artistRecommendations(first: 20), otherwise based on my limited knowledge, looks good to me and we reviewed one output together.

shaq31415926 (changes requested):
I would change the query to artistRecommendations(first: 20), otherwise based on my limited knowledge, looks good to me and we reviewed one output together. Thank you.

src/schema/v2/me/artistRecommendations.ts
```diff
@@ -0,0 +1,84 @@
...
+    if (!vortexGraphqlLoader || !artistsLoader) return
+
+    const { page, size, offset } = convertConnectionArgsToGravityArgs(args)
+
+    // Fetch artist IDs from Vortex
+
+    const vortexResult = await vortexGraphqlLoader({
+      query: gql`
+        query artistRecommendationsQuery {
+          artistRecommendations(first: ${MAX_ARTISTS}) {
```
MounirDhahri, line 36: **thought/issue:** would it be possible to set an `offset` here and have `first` to be equal to `size` as we get it from the client. That way we will avoid making the same query when the user is trying to get to the second page and it will also fix the issue that we will only be having a maximum of 50 `MAX_ARTISTS` recommended artists

src/schema/v2/me/artistRecommendations.ts
```diff
@@ -0,0 +1,84 @@
...
+              }
+            }
+          }
+        }
+      `,
+    })()
+
+    const artistIds: string[] = vortexResult.data?.artistRecommendations?.edges?.map(
+      (edge) => edge?.node?.artistId
+    )
```
MounirDhahri, line 51: **nitpick:** I saw this pattern being used a few times in metaphysics, maybe it's time to add `extractNodes` as a helper like we do in Eigen https://github.com/artsy/eigen/blob/5b6d470e1177ced6dbc997cc3401d9876cfcbed0/src/lib/utils/extractNodes.ts

src/schema/v2/me/artistRecommendations.ts
MounirDhahri: the query offset would be then based on the `recommendedArtists` and not the artists themselves

src/schema/v2/me/artistRecommendations.ts
```diff
@@ -0,0 +1,84 @@
...
+        await artistsLoader({
+          ids: artistIds,
+          size,
+          offset,
+        })
+      ).body
+
+      // Apply order from Vortex result (score ASC)
+
+      artists = artistIds.map((artistId) => {
```
MounirDhahri, line 68:
```suggestion
      artists = artistIds.filter((artistId) => {
```
**issue:** Shouldn't this be `filter` 🤔. In the case where `artistIds` will initially have 50 recommended artists and we have queried gravity for only `size = 10` artists, artists will be an array containing `undefined` values coming from `find`

MounirDhahri:
Nice PR 👏 I left few comments, I would be happy to provide more context if something wasn't clear

olerichter00:
> I would change the query to artistRecommendations(first: 20), otherwise based on my limited knowledge, looks good to me and we reviewed one output together. Thank you.

Good point. We can adjust the query in the Client (**Eigen** in this case) to query 20 artists instead of 10.

src/schema/v2/me/artistRecommendations.ts
olerichter00: Oh yes, we can actually just forward the pagination arguments from the query to **Vortex**. I'll change that.

src/schema/v2/me/artistRecommendations.ts
olerichter00: Good catch!

We actually don't have to pass `offset` and `size` to **Gravity**. The number of artist IDs coming from **Vortex** will be enough. 

I added `compact` to remove all `undefined` values in the case an artist can't be found in **Gravity**.
```
artists = compact(
        artistIds.map((artistId) => {
          return find(artists, { _id: artistId })
        })
      )
```

MounirDhahri: aha. thanks 🙌

src/schema/v2/me/artistRecommendations.ts
olerichter00: Good idea! I added an `extractNodes` helper 👍

src/schema/v2/me/__tests__/artistRecommendations.test.ts
```diff
@@ -0,0 +1,208 @@
...
+          }
+        }
+      `,
+    })
+    expect(artistsLoader).toHaveBeenCalledWith({
+      ids: ["608a7417bdfbd1a789ba092a", "608a7416bdfbd1a789ba0911"],
+    })
+  })
+
+  it("doesn't return artists if no artist recommendations are present", async () => {
```
shaq31415926, line 77: nice!

src/schema/v2/me/artistRecommendations.ts
```diff
@@ -0,0 +1,79 @@
...
+import { connectionFromArray } from "graphql-relay"
+import { convertConnectionArgsToGravityArgs, extractNodes } from "lib/helpers"
+import { CursorPageable, pageable } from "relay-cursor-paging"
+import { createPageCursors } from "schema/v1/fields/pagination"
+import { ResolverContext } from "types/graphql"
+import gql from "lib/gql"
+import { artistConnection } from "schema/v2/artist"
+import { find } from "lodash"
+
+const MAX_ARTISTS = 50
```
shaq31415926, line 11: just for your info, we are only recommending twenty artists per user right now. however, we may change that so this looks good.

shaq31415926 (commented):
looks good to mr, thanks very much.

src/schema/v2/me/artistRecommendations.ts
MounirDhahri: I can't find the compact in your last commit 🤔

src/schema/v2/me/artistRecommendations.ts
olerichter00: Good catch and this is confusing indeed. It's an implementation detail and related to how we implement the pagination and not directly related to the amount of artists we query. I will add a comment to make it easier to understand.

src/schema/v2/me/artistRecommendations.ts
```diff
@@ -0,0 +1,79 @@
+import { GraphQLFieldConfig, GraphQLInt } from "graphql"
+import { connectionFromArray } from "graphql-relay"
+import { convertConnectionArgsToGravityArgs, extractNodes } from "lib/helpers"
+import { CursorPageable, pageable } from "relay-cursor-paging"
+import { createPageCursors } from "schema/v1/fields/pagination"
+import { ResolverContext } from "types/graphql"
+import gql from "lib/gql"
+import { artistConnection } from "schema/v2/artist"
+import { find } from "lodash"
+
```
olerichter00, line 10:
```suggestion

// This limits the maximum number of artists we receive from Vortex and is related to how we implement the Connection in this resolver.
```

MounirDhahri (dismissed):

MounirDhahri (approved):
Thanks for the changes 🚀 this is clear to go

MounirDhahri:
#squashongreen

src/schema/v2/me/artistRecommendations.ts
olerichter00: Forgot to push and added it now 🙇

PR #3558, opened 2021-11-22
fix: short show exhibition period as a format arg

brainbicycle:
some PR review follow up from https://github.com/artsy/metaphysics/pull/3550

Instead of having a separate field add a format arg to the existing exhibitionPeriod field, right now just `LONG` and `SHORT` supported.

### DIFF
--- a/src/schema/v2/fair.ts
+++ b/src/schema/v2/fair.ts
@@ -139,14 +139,18 @@ export const FairType = new GraphQLObjectType<any, ResolverContext>({
       exhibitionPeriod: {
         type: GraphQLString,
         description: "A formatted description of the start to end dates",
-        resolve: ({ start_at, end_at }) => dateRange(start_at, end_at, "UTC"),
-      },
-      shortExhibitionPeriod: {
-        type: GraphQLString,
-        description:
-          "A formatted description of the start to end dates with abbreviated months",
-        resolve: ({ start_at, end_at }) =>
-          dateRange(start_at, end_at, "UTC", true),
+        args: {
+          format: {
+            type: GraphQLString,
+            description:
+              "SHORT or LONG, SHORT returns a shorter exhbition period description, LONG returns a longer description, defaults to LONG",
+          },
+        },
+        resolve: ({ start_at, end_at }, args) => {
+          const { format } = args
+          const useShortFormat = format === "SHORT"
+          return dateRange(start_at, end_at, "UTC", useShortFormat)
+        },
       },
       formattedOpeningHours: {
         type: GraphQLString,
--- a/src/schema/v2/fair.ts
+++ b/src/schema/v2/fair.ts
@@ -12,7 +12,7 @@ import Profile from "./profile"
 import Image from "./image"
 import Artist from "./artist"
 import Partner, { PartnerType } from "./partner"
-import { ShowsConnection } from "./show"
+import { ExhibitionPeriodFormatType, ShowsConnection } from "./show"
 import { LocationType } from "./location"
 import {
   SlugAndInternalIDFields,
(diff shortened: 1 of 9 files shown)


src/schema/v2/show.ts
```diff
@@ -56,6 +57,18 @@ const FollowArtistType = new GraphQLObjectType<any, ResolverContext>({
   }),
 })
 
+export const ExhibitionPeriodFormatType = new GraphQLEnumType({
+  name: "ExhibitionPeriodFormat",
+  values: {
+    SHORT: {
+      value: "SHORT",
```
dblandin, line 64: **suggestion**: Might be worth adding descriptions with an example formatted periods?

```typescript
{
  description: "Long formatted period e.g. February 25 – May 24, 2015"
}
```

```typescript
{
  description: "Short formatted period e.g. Feb 25 - May 24, 2015"
}
```

src/schema/v2/fair.ts
```diff
@@ -139,14 +139,18 @@ export const FairType = new GraphQLObjectType<any, ResolverContext>({
...
-        type: GraphQLString,
-        description:
-          "A formatted description of the start to end dates with abbreviated months",
-        resolve: ({ start_at, end_at }) =>
-          dateRange(start_at, end_at, "UTC", true),
+        args: {
+          format: {
+            type: ExhibitionPeriodFormatType,
+            description:
+              "SHORT or LONG, SHORT returns a shorter exhbition period description with months abbreviated, defaults to LONG",
```
dblandin, line 146: **suggestion**: Instead of implicitly including default behavior, the GraphQL SDL supports an explicit `defaultValue` option. With an explicit default value in place, you might also be able to make the argument `non null`:

```typescript
{
  exhibitionPeriod: {
    type: GraphQLString,
    description: "A formatted description of the start to end dates",
    args: {
      format: {
        type: new GraphQLNonNull(ExhibitionPeriodFormatType),
        description:  "Formatting option to apply to exhibition period",
        defaultValue: ExhibitionPeriodFormatType.LONG.value
      }
    }
  }
}
```

src/schema/v2/show.ts
```diff
@@ -338,14 +351,18 @@ export const ShowType = new GraphQLObjectType<any, ResolverContext>({
...
+          format: {
+            type: ExhibitionPeriodFormatType,
+            description:
+              "SHORT or LONG, SHORT returns a shorter exhbition period description with months abbreviated, defaults to LONG",
+          },
+        },
+        resolve: ({ start_at, end_at }, args) => {
+          const { format } = args
+          const useShortFormat = format === "SHORT"
+          return dateRange(start_at, end_at, "UTC", useShortFormat)
```
dblandin, line 364: **suggestion**: Related to the comment above, if `format` is always either `LONG` or `SHORT` and never null, we could pass that variable directly through to `dateRange` and let that function decide what to do.

src/schema/v2/show.ts
```diff
@@ -56,6 +57,18 @@ const FollowArtistType = new GraphQLObjectType<any, ResolverContext>({
...
 })
 
+export const ExhibitionPeriodFormatType = new GraphQLEnumType({
+  name: "ExhibitionPeriodFormat",
+  values: {
+    SHORT: {
+      value: "SHORT",
+    },
+    LONG: {
+      value: "LONG",
```
dblandin, line 67: **suggestion**: We're a bit inconsistent here but unless the backend needs the value to the uppercase, which is sometimes need to support sorting via Gravity, I think most other enums have lowercase values.

```typescript
{
  SHORT: { value: "short" },
  LONG: { value: "long" }
}
```

dblandin (commented):
Looking good! Left a few comments 👍🏼

dblandin (approved):
Looks great!! 👍

PR #3564, opened 2021-11-24
fix: make format arg not required

brainbicycle:
former PR made the format arg required which was incompatible with Eigen's existing schema. This makes the arg not required but we still have a default.


dblandin (approved):
LGTM - good to know that you can't rely a specified default value to satisfy a NonNull type.

PR #3568, opened 2021-11-29
[COORDINATE-DEPLOY] feat: relax VAT requirements for artworks (TX-135)

starsirius:
https://artsyproduct.atlassian.net/browse/TX-135

We want to enable BNMO (including Make Offer in Inquiries) for artworks located in relevant countries as long as partners have provided a VAT status (and necessary exemption documentation for us to review), regardless of the selected status.

With this, clients (i.e. Volt) will not block taking actions on MOI orders as long as there is _a_ VAT status set. The recording below shows the flow of selecting exempt and the blocking banner going away. Note that the actions are still disabled because of metadata that's also required by Make Offer in Inquiries.

vat-moi

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -2424,11 +2424,11 @@ describe("Artwork type", () => {
       })
     })
 
-    it("returns true when artwork requires vat and partner is vat registered", () => {
+    it("returns true when artwork requires VAT and partner has any VAT status", () => {
       artwork.vat_required = true
       artwork.partner = { id: "123" }
       context.partnerAllLoader = jest.fn(() =>
-        Promise.resolve({ vat_registered: true })
+        Promise.resolve({ vat_status: "anything" })
       )
 
       return runQuery(query, context).then((data) => {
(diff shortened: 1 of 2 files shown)


starsirius:
Self-assign this to myself to coordinate deployment. Any feedback would be appreciated! 🙏

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -2437,11 +2437,24 @@ describe("Artwork type", () => {
       })
     })
 
-    it("returns false when artwork requires vat and partner is not vat registered", () => {
+    it("returns false when artwork requires VAT and partner does not have a VAT status", () => {
       artwork.vat_required = true
       artwork.partner = { id: "123" }
       context.partnerAllLoader = jest.fn(() =>
-        Promise.resolve({ vat_registered: false })
+        Promise.resolve({ vat_status: null })
```
erikdstock, line 2444: so `vatRequirementComplete` now requires that a vat status be _present_, but doesn't make demands beyond that? If so I wonder if we could add another test for the `true` case in addition to this one:
https://github.com/artsy/metaphysics/pull/3568/files#diff-1bca24dc978ef5f6e034b1bf83eb9f55a14bcb7181e8bc18df4e6790ed5e520dR2416-R2425

erikdstock (commented):
Minor question!

src/schema/v2/artwork/__tests__/artwork.test.js
starsirius: > so vatRequirementComplete now requires that a vat status be present, but doesn't make demands beyond that?

Yes, we only require some status to be set for the new relaxed requirements.

Is [this test](https://github.com/artsy/metaphysics/pull/3568/files#diff-1bca24dc978ef5f6e034b1bf83eb9f55a14bcb7181e8bc18df4e6790ed5e520dR2427) you're referring to, or you think we can add more coverage?
```javascript
it("returns true when artwork requires VAT and partner has any VAT status")
```

starsirius:
Going to merge and get ready for the release tomorrow!

PR #3569, opened 2021-11-30
[feat]: update convection schema

AlicanAkyuz:
### Description 

Ticket: [SWA-134](https://artsyproduct.atlassian.net/browse/SWA-134)

As part of the above ticket's implementation [in Convection](https://github.com/artsy/convection/pull/1184), our DB schema has slightly changed with the addition of a new column called `created_by` on Submissions, which is also reflected Convection's graphQL schema since this new field is automatically added for sorting only. 

This PR updates Convection's schema in MP to reflect the above change.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -6506,6 +6506,12 @@ enum ConsignmentSubmissionSort {
   # sort by created_at in descending order
   CREATED_AT_DESC
 
+  # sort by created_by in ascending order
+  CREATED_BY_ASC
+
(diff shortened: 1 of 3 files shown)


Gon4ar5 (approved):

iskounen (approved):
These changes look good to me. It may be helpful to add a bit more context about why you're making this change in the PR description.

AlicanAkyuz:
@iskounen apologies for being inexplicit, I've added some context in the description.

PR #3577, opened 2021-12-02
feat: adds update password mutation

dzucconi:
Supports https://github.com/artsy/force/pull/8997

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -9051,6 +9051,11 @@ type Mutation {
     input: UpdateConversationMutationInput!
   ): UpdateConversationMutationPayload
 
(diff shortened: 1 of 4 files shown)


MounirDhahri:
Hey Damon, I know we are updating the password right now using a REST call. I don't remember who mentioned why this was the case before, but I think it's something related to not exposing user passwords in metaphysics in sentry or so 🤷‍♂️. Is it no longer the case, if so I will update the logic on Eigen as well to use this mutation. cc @brainbicycle

MounirDhahri:
Thanks for looking into this, I guess we should then update Eigen to use this mutation soon

PR #3579, opened 2021-12-02
[DO NOT MERGE] Example hybrid connection implementation

lilyfromseattle:
Paired with @erikdstock on this example implementation of a hybrid connection.

It uses 2 standard gravity endpoints, but the artworkLoader fails/times out due to a quirk in its implementation (can explain in slack). Swapping it out for another generic gravity loader (fairs) does work for a basic query without accessing any of the fields. In order to get a fully working implementation we need 2 objects that:
- Can be queried with their total count
- Can be sorted according to the same criteria for consistent collection ordering (ie created_at)

Our original connection PR: #3214

cc @damassi

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1733,6 +1733,28 @@ union ArtworkMutationType =
     ArtworkMutationDeleteSuccess
   | ArtworkMutationFailure
 
+union ArtworkOrArtist = Artist | Artwork
+
+# A connection to a list of items.
+type ArtworkOrArtistConnection {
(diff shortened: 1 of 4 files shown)


damassi:
This example is looking 💯 . Really really nice!!

erikdstock:
I was trying to think of two objects that could be queried and sorted separately and together in a logical way - one thought is `fairs` and `sales`, which are both sortable by `start_at`.

damassi:
@erikdstock - why close this example?

PR #3580, opened 2021-12-03
feat(FX-3624): Add "sizes" attribute to SearchCriteria and SearchCriteriaAttributes

anastasiapyzhik:
[FX-3624]

At the moment, the principle of operation of the Size filter on Force (we can choose several size values) and Eigen (we can choose only one size value) is different and we decided to make the same behavior on the two platforms the same (the user can select multiple filter values). For this reason, we need to add support for the sizes attribute in SearchCriteria, SearchCriteriaAttributes.

[FX-3624]: https://artsyproduct.atlassian.net/browse/FX-3624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -2834,6 +2834,12 @@ type ArtworksAggregationResults {
   slice: ArtworkAggregation
 }
 
+enum ArtworkSizes {
+  LARGE
+  MEDIUM
+  SMALL
+}
+
 enum ArtworkSorts {
   availability_desc
     @deprecated(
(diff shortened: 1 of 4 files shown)


dimatretyak (commented):
Tested locally. Looks great 💯 

#### Mutation
mutation

#### Query
query

dblandin (approved):
Nice! Looks great 💯

PR #3594, opened 2021-12-10
chore: Update target branch on update repo for Eigen [RFC Follow-up]

MounirDhahri:
**Description:**

This PR updates the target branch for Eigen to be the new default branch "main" instead of "master".

PS: I already updated the default branch Eigen to be **main**

This comes as a follow-up from this [RFC](https://github.com/artsy/README/issues/386)


rquartararo (approved):

MounirDhahri:
🛳🛳🛳

PR #3595, opened 2021-12-10
feat: adopt inclusive language - main default branch

dblandin:
Favor inclusive language in the Metaphysics codebase per this [RFC][].

[RFC]: artsy/README#427


MounirDhahri:
niiiice

PR #3598, opened 2021-12-13
feat: Add artwork_location to artwork

olerichter00:
Addresses [CX-2218]

## Description

Add artwork_location to artwork

It needs [gravity PR](https://github.com/artsy/gravity/pull/14759) to be deployed first. 

[CX-2218]: https://artsyproduct.atlassian.net/browse/CX-2218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -133,6 +133,7 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      artworkLocation: { type: GraphQLString },
       availability: { type: GraphQLString },
       category: {
         type: GraphQLString,
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -133,7 +133,10 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
-      artworkLocation: { type: GraphQLString },
+      artworkLocation: {
+        type: GraphQLString,
+        resolve: (artwork) => artwork.artwork_location,
+      },
       availability: { type: GraphQLString },
       category: {
         type: GraphQLString,
--- a/src/schema/v2/me/__tests__/myCollectionCreateArtworkMutation.test.ts
+++ b/src/schema/v2/me/__tests__/myCollectionCreateArtworkMutation.test.ts
@@ -9,6 +9,7 @@ const artworkDetails = {
   medium: "Painting",
   price_paid_cents: 10000,
   price_paid_currency: "USD",
+  artwork_location: "Berlin",
 }
 const artworkLoader = jest.fn().mockResolvedValue(artworkDetails)
 
@@ -54,6 +55,7 @@ const computeMutationInput = ({
           ... on MyCollectionArtworkMutationSuccess {
             artwork {
               medium
+              artworkLocation
               pricePaid {
                 display
               }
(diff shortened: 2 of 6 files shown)


src/schema/v2/me/myCollectionCreateArtworkMutation.ts
```diff
@@ -71,6 +71,9 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
     height: {
       type: GraphQLString,
     },
+    artwork_location: {
+      type: GraphQLString,
+    },
```
olerichter00, line 76:
```suggestion
    artworkLocation: {
      type: GraphQLString,
    },
```

src/schema/v2/me/myCollectionUpdateArtworkMutation.ts
```diff
@@ -136,6 +141,7 @@ export const myCollectionUpdateArtworkMutation = mutationWithClientMutationId<
         artists: artistIds,
         cost_currency_code: costCurrencyCode,
         cost_minor: costMinor,
+        artwork_artwork_location: artworkLocation,
```
MounirDhahri, line 144: **thought:** is this intentionally called `artwork_artwork_location`

MounirDhahri (commented):
should we also add a test for this maybe?

src/schema/v2/me/myCollectionUpdateArtworkMutation.ts
olerichter00: This should be 

```suggestion
        artwork_location: artworkLocation,
```

olerichter00 (commented):
Looks good to me 🚀

src/schema/v2/me/__tests__/myCollectionCreateArtworkMutation.test.ts
```diff
@@ -40,6 +40,7 @@ const computeMutationInput = ({
           editionSize: ${JSON.stringify(editionSize)}
           externalImageUrls: ${JSON.stringify(externalImageUrls)}
           height: "20"
+          artworkLocation: "Berlin"
```
MounirDhahri, line 43: **issue(non-blocker)**: I think we should also add the `artworkLocation` to the query to make sure it's coming back

MounirDhahri (commented):
Nice job, I left one minor comment and we should be clear to go

MounirDhahri (approved):
awesome! thanks

src/schema/v2/artwork/index.ts
```diff
@@ -133,6 +133,10 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      artworkLocation: {
```
oxaudo, line 136: Sorry for late comment here. Do I read this correctly - that we've added `artwork_location` here that is different from `location` and is only applicable to myCollection artworks? Maybe worth adding a `description` to this field here. 

Suspect it will be very confusing at some point if somebody looks at artwork and `location` is different than `artworkLocation` and nothing matches `shippingOrigin`.

Apologies if it has been discussed and resolved already.

rajsam003: Good point. I will add a description. Thanks

PR #3608, opened 2021-12-15
docs: Artwork's location description

rajsam003:
This PR helps in adding descriptions for My Collection artwork's location which is discussed [here](https://github.com/artsy/metaphysics/pull/3598#discussion_r769124885).

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -136,6 +136,8 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
       artworkLocation: {
         type: GraphQLString,
         resolve: (artwork) => artwork.artwork_location,
+        description:
+          'Represents the location of the artwork for "**My Collection**" artworks',
       },
       availability: { type: GraphQLString },
       category: {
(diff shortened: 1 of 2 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -136,6 +136,8 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
       artworkLocation: {
         type: GraphQLString,
         resolve: (artwork) => artwork.artwork_location,
+        description:
+          'Represents the location of the artwork for "**My Collection**" artworks',
```
olerichter00, line 140: I wonder if `**My Collection**` is necessary here or if `"My Collection"` would be easier to read 🤔

olerichter00 (approved):
Thanks for adding the missing description 🙇

src/schema/v2/artwork/index.ts
rajsam003: Haha. I just wanted to give more prominence. I will change it now.

olerichter00 (approved):

PR #3609, opened 2021-12-15
chore: use Squash on Green label when pushing schema updates to client repos

pvinis:
Follow up after https://artsy.slack.com/archives/C02BAQ5K7/p1639509813109500.

We tend to prefer a squash-merge to avoid the merge commit these days.


dblandin (approved):
LGTM!

PR #3621, opened 2021-12-17
feat(schema): exposes a field for getting authentication status

dzucconi:
Picking up from https://github.com/artsy/force/pull/9093 — this gives us a single field for determining auth status.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2321,6 +2321,12 @@ enum AuthenticationProvider {
   GOOGLE
 }
 
+enum AuthenticationStatus {
+  INVALID
+  LOGGED_IN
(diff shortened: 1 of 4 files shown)


damassi (approved):
Beautiful 😘 

This file is simple enough, but i'll leave it to you to add a test or not. Nothing blocking.

dzucconi:
Yeah; nothing really happening to test! (A single conditional)

damassi:
There's actually a lot of behavior in this. No loader? logged out. Loader throws? Check the (specific) error being returned. Does it return the correct enum? Etc. Its low enough level that test would be good, but you're right, it is simple enough code.

PR #3623, opened 2021-12-20
GRO-686: Settings delete account module

The-Beez-Kneez:
This is a MP update to create a mutation to allow a user to delete their account. It will be paired with the ongoing work in force to update and migrate the `edit-setting` modules.


src/schema/v2/me/delete_account_mutation.ts
```diff
@@ -0,0 +1,80 @@
...
+  any,
+  any,
+  ResolverContext
+>({
+  name: "DeleteAccount",
+  description: "Delete User Artsy Account",
+  inputFields: {
+    id: {
+      type: new GraphQLNonNull(GraphQLString),
+    },
```
dzucconi, line 54: I know we accept an `explanation` which we'll want to also support here.

dzucconi: `url` as well

https://github.com/artsy/gravity/blob/d53fd1a51711b98103feabdc64503be9d893b585/app/api/v1/me_endpoint.rb#L100-L101

dzucconi: And actually: we don't want to accept an ID at all. This will be hitting the `me` endpoint; you can only delete the logged in user.

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -72,6 +72,11 @@ export default (accessToken, userID, opts) => {
       {},
       { method: "DELETE" }
     ),
+    deleteUserAccountLoader: gravityLoader(
+      (id) => `user/${id}`,
```
dzucconi, line 76: So this is actually `me` not the admin URL

dzucconi: https://github.com/artsy/gravity/blob/d53fd1a51711b98103feabdc64503be9d893b585/app/api/v1/me_endpoint.rb#L96-L102

_schemaV2.graphql
```diff
@@ -6127,6 +6137,15 @@ type DeepZoomImageSize {
   Width: Int
 }
 
+input DeleteAccountInput {
+  clientMutationId: String
```
jonallured, line 6141: Why does this appear here? 🤔

src/schema/v2/schema.ts
```diff
@@ -225,6 +226,7 @@ export default new GraphQLSchema({
       createGeminiEntryForAsset: CreateGeminiEntryForAsset,
       createUserInterest: createUserInterestMutation,
       deleteCreditCard: deleteCreditCardMutation,
+      deleteUserAccount: deleteUserAccountMutation,
```
dzucconi, line 229: I'd name this something like `deleteMyAccountMutation`

The-Beez-Kneez:
@dzucconi : @jonallured and I are close on this but we are having issues, the deleteUserAccountLoaders are not working for us. We thought that passing an access token would remedy this but it is not happening. We have been trying to delete a sake account in staging.

src/schema/v2/schema.ts
jonallured: We sorta just followed the example by the mutation directly above this one but maybe that one isn't named real well?

dzucconi:
So long as you're using the `me` endpoint you should be fine. All the headers required are passed in so long as you're properly logged in in GraphQL.

src/schema/v2/me/delete_account_mutation.ts
```diff
@@ -0,0 +1,73 @@
...
+  description: "Delete User Artsy Account",
+  outputFields: {
+    userAccountOrError: {
+      type: AccountMutationType,
+      resolve: (result) => result,
+    },
+  },
+  mutateAndGetPayload: ({ deleteUserAccountLoader }) => {
+    if (!deleteUserAccountLoader) {
+      throw new Error("You need to be signed in to perform this action")
```
jonallured, line 58: @dzucconi we're hitting this throw line when passing an `x-access-token` header - is that what you'd expect? 🤔

src/schema/v2/me/delete_account_mutation.ts
The-Beez-Kneez: Updated.

The-Beez-Kneez:
> So you'll have to do more than this
> 
> ```
> explanation: { type: GraphQLString, description: 'Some string explaining what this is' }
> ```
> 
> And you'll also have to pull out the arg and pass it into the loader below

Not really sure what you mean by this set of instructions. More how? And what is being passed to a loader?

The-Beez-Kneez:
Updated!

src/schema/v2/schema.ts
The-Beez-Kneez: Done

src/schema/v2/me/delete_account_mutation.ts
```diff
@@ -0,0 +1,82 @@
...
+  any,
+  any,
+  ResolverContext
+>({
+  name: "DeleteAccount",
+  description: "Delete User Artsy Account",
+  inputFields: {
+    explanation: {
+      description: "Reason for deleting the account.",
+      type: new GraphQLNonNull(GraphQLString)
```
dzucconi, line 53: Both these arguments shouldn't be non null as they are optional.

dzucconi, line 53:
```suggestion
      type: GraphQLString
```

src/schema/v2/me/delete_account_mutation.ts
```diff
@@ -0,0 +1,82 @@
...
+    },
+    url: {
+      description: "Referrer location",
+      type: new GraphQLNonNull(GraphQLString)
```
dzucconi, line 57:
```suggestion
      type: GraphQLString
```

dzucconi (approved):
Should be fine to merge after accepting my suggestions to make the args optional.

src/schema/v2/me/delete_account_mutation.ts
The-Beez-Kneez: @dzucconi What should they be if not non null?

src/schema/v2/me/delete_account_mutation.ts
The-Beez-Kneez: @dzucconi Using only the `GraphQLString` causes us to run into an error of: `This expression is not constructable. Type 'GraphQLScalarType' has no construct signatures.`

dzucconi: That shouldn't be the case but I'm AFK until later. There should be plenty of examples in the codebase if you take a look around.

The-Beez-Kneez: Changed.

jonallured:
Ok I think we got it thanks so much for the persistence @The-Beez-Kneez!! 🏆

PR #3624, opened 2021-12-20
feat(artwork): exposes downloadableImageUrl

dzucconi:
So: 

Downloadable images for admins are currently broken in Force.

Previously we were constructing this URL with some methods on the old Backbone models and then proxying the download. I don't really see a reason though why we can't just conditionally display this URL directly on the client-side — this will redirect. Let me know if I'm missing something obvious. This endpoint will redirect to the S3 asset with a `X-Amz-Expires=43200` query param.

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -59,6 +59,7 @@ import { getMicrofunnelDataByArtworkInternalID } from "../artist/targetSupply/ut
 import { InquiryQuestionType } from "../inquiry_question"
 import { priceDisplayText, priceRangeDisplayText } from "lib/moneyHelpers"
 import { LocationType } from "schema/v2/location"
+import config from "config"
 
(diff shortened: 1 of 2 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -448,6 +449,19 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+        type: GraphQLString,
+        resolve: ({ id, images }) => {
+          if (!images || !images.length) return null
+
+          const defaultImage =
+            images.find((image) => image.is_default) || images[0]
+
+          if (!defaultImage) return null
+
+          return `${config.GRAVITY_API_BASE}/artwork/${id}/image/${defaultImage.id}/original.jpg`
```
joeyAghion, line 462: Note that you can avoid specifying the image id if you only need the default:

`/artwork/:id/default_image/original.jpg`

PR #3625, opened 2021-12-21
feat: add attributionClass to MyCollectionCreateArtworkInput

fladson:
This PR resolves to add attribution Class in MyCollection artwork - [CX-2224](https://artsyproduct.atlassian.net/browse/CX-2224)

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1607,6 +1607,13 @@ enum ArtworkAggregation {
   TOTAL
 }
 
+enum ArtworkAttributionClassType {
+  LIMITED_EDITION
+  OPEN_EDITION
+  UNIQUE
+  UNKNOWN_EDITION
+}
+
 # A connection to a list of items.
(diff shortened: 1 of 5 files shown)


olerichter00 (changes requested):
Looking good! 🙌 

Could you please add a description to the PR (so we don't rely on third-party services like Jira to find out what this change is about) and have a look if we can update the corresponding test files as well?

_schemaV2.graphql
```diff
@@ -1607,6 +1607,13 @@ enum ArtworkAggregation {
   TOTAL
 }
 
+enum ArtworkAttributionClassType {
```
fladson, line 1610: Is `ConsignmentAttributionClass` referring to the same thing? Just wondering if we could unify it into only one enum.

olerichter00: That would be great but I guess `ConsignmentAttributionClass` is not an input type, right?

olerichter00 (approved):
🚀 🚀 🚀

PR #3633, opened 2021-12-22
remove unused graphiql dep

mc-jones:
This PR removes the unused `graphiql` dependency and its peer dependencies `react` and `react-dom`. This also addresses [this depenabot alert](https://github.com/artsy/metaphysics/security/dependabot/yarn.lock/graphiql/open), noting that this wasn't an actual vulnerability, since it wasn't implemented and we aren't using dynamically set Schema URLs.

### Considerations
`Graphiql` is built into `express-graphql` and is already enabled [here](https://github.com/artsy/metaphysics/blob/6439278795f6a15a1217c9baf42791c17eff76ec/src/index.ts#L243) in non-production mode.


icirellik (approved):
Looks good to me, I'm adding a note that `graphiql` is loaded statically from npm in case anyone else is as confused as I was about how it's loaded given we are no longer installing it.

https://github.com/graphql/express-graphql/blob/a7c3805d4eb6eb4eade6c48e2f086573f4488976/src/renderGraphiQL.ts#L158

PR #3636, opened 2021-12-23
Remove unused apollo libs and settings

joeyAghion:
This removes `apollo-server-express`, which was never successfully used in production. A few other files and dependencies appear to only have ever supported that optional mode, so I removed them as well.

I don't understand exactly _why_ it was never successful, but the [discussion](https://artsy.slack.com/archives/CA8SANW3W/p1576170343344900) indicates that all the endpoints started 404-ing when Apollo was briefly enabled.

In the meantime, security patches like https://github.com/artsy/metaphysics/pull/3528 have been open and failing.

I've already unset the associated config in staging, which should be safe because the default is to _not_ enable Apollo:

```
hokusai staging env unset ENABLE_APOLLO
```

It's already unset in production.


tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "allowSyntheticDefaultImports": true,
     "baseUrl": "./src",
     "checkJs": false,
-    "lib": ["es2017"],
+    "lib": ["es2019"],
```
joeyAghion, line 7: Not sure why my change led to [these type failures](https://app.circleci.com/pipelines/github/artsy/metaphysics/10519/workflows/2c97577b-e08d-434d-b17d-b1573efba310/jobs/21520), but bumping this target as instructed by the warnings worked.

src/index.ts
```diff
@@ -136,134 +133,82 @@ function startApp(appSchema, path: string) {
...
-        validationRules,
-        extensions: ({ document, result }) =>
-          createExtensions(document, result, requestID),
-        formatError: (error) => ({
-          // better errors formatting for clients
-          // See errorMiddleware in  https://github.com/relay-tools/react-relay-network-modern/blob/master/README.md#built-in-middlewares
-          message: error.message,
-          stack: !PRODUCTION_ENV ? error.stack.split("\n") : null,
-        }),
-      }
```
damassi, line 261: 😢 👋  We could always revive this

icirellik (approved):

PR #3640, opened 2022-01-03
feat: expose collector profile privacy field

fladson:
Relates to https://github.com/artsy/gravity/pull/14815

### DIFF
--- a/src/schema/v2/me/collector_profile.ts
+++ b/src/schema/v2/me/collector_profile.ts
@@ -30,10 +30,7 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
   intents: { type: new GraphQLList(GraphQLString) },
   loyaltyApplicantAt: date,
   name: { type: GraphQLString },
-  privacy: {
-    type: GraphQLInt,
-    resolve: ({ privacy }) => privacy,
-  },
+  privacy: { type: GraphQLInt },
   professionalBuyerAppliedAt: date,
   professionalBuyerAt: date,
   selfReportedPurchases: {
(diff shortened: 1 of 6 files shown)


fladson:
All I did here was to search for `collector_level` and `collectorLevel` and replicate the changes for `privacy`, let me know if we need to make more changes or remove unintended code.

src/schema/v2/me/collector_profile.ts
```diff
@@ -30,6 +30,10 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
   intents: { type: new GraphQLList(GraphQLString) },
   loyaltyApplicantAt: date,
   name: { type: GraphQLString },
+  privacy: {
+    type: GraphQLInt,
+    resolve: ({ privacy }) => privacy,
+  },
```
MounirDhahri, line 36: nice 🙌
btw, you don't need `line 35` here since the field name and the property you're destructing are the the same :)

src/schema/v2/me/index.ts
```diff
@@ -215,6 +215,12 @@ export const meType = new GraphQLObjectType<any, ResolverContext>({
       type: GraphQLFloat,
       resolve: ({ price_range }) => price_range?.split(":")[1],
     },
+    privacy: {
+      type: GraphQLInt,
+      resolve: ({ privacy }) => {
+        return privacy
+      },
```
MounirDhahri, line 222: same not as above here

MounirDhahri (approved):
Nice 🙌 Thanks!
LGTM. It would be great though if you add a test for the new privacy property inside `collector_profile.test.js`

MounirDhahri (approved):
Thank you!
LGTM 🚢🚢🚢

PR #3657, opened 2022-01-11
feat(articles): fleshes out article schema

dzucconi:
Extended schema to support rebuilt article pages.

### DIFF
--- a/src/schema/v2/article/index.ts
+++ b/src/schema/v2/article/index.ts
@@ -1,4 +1,6 @@
 import {
+  GraphQLBoolean,
+  GraphQLEnumType,
   GraphQLFieldConfig,
   GraphQLInt,
   GraphQLList,
@@ -21,6 +23,8 @@ import { ArticleSectionCallout } from "./sections/ArticleSectionCallout"
 import { ArticleSectionEmbed } from "./sections/ArticleSectionEmbed"
 import { ArticleSectionImageSet } from "./sections/ArticleSectionImageSet"
 import { ArticleSectionSocialEmbed } from "./sections/ArticleSectionSocialEmbed"
+import { take } from "lodash"
+import { ArticleHero } from "./models"
 
 export const ArticleType = new GraphQLObjectType<any, ResolverContext>({
   name: "Article",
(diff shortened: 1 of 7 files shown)


dzucconi:
These will be 'technically' breaking changes but won't matter since nothing is using it yet.

src/schema/v2/article/models.ts
```diff
@@ -0,0 +1,85 @@
...
+  isTypeOf: ({ type }) => {
+    return type === "image"
+  },
+  fields: {
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
+      resolve: ({ url }) => {
+        return uuid(url, uuid.URL)
+      },
+    },
```
dzucconi, line 24: This is the only slightly odd thing about this PR. I wanted all the `figures` to have IDs because this dramatically simplifies how we can implement the zoom gallery view since we know which ID we click on. Here I'm just constructing an ID based on the URL of the image.

src/schema/v2/article/index.ts
```diff
@@ -53,26 +56,91 @@ export const ArticleType = new GraphQLObjectType<any, ResolverContext>({
...
+    layout: {
+      type: new GraphQLNonNull(
+        new GraphQLEnumType({
+          name: "ArticleLayout",
+          values: {
+            STANDARD: { value: "standard" },
+            FEATURE: { value: "feature" },
+          },
+        })
+      ),
```
damassi, line 79: Ah, thats how you do this. I didn't quite understand what was happening over in positrons schema in this regard

src/schema/v2/article/index.ts
```diff
@@ -53,26 +56,91 @@ export const ArticleType = new GraphQLObjectType<any, ResolverContext>({
...
-            ArticleSectionCallout,
-            ArticleSectionEmbed,
-            ArticleSectionImageCollection,
-            ArticleSectionImageSet,
-            ArticleSectionSocialEmbed,
-            ArticleSectionText,
-            ArticleSectionVideo,
-          ],
-        })
+      type: new GraphQLNonNull(
```
damassi, line 125: For the future: should we basically always be using new GraphQLNonNull unless its obvious not to?

damassi (approved):

src/schema/v2/article/index.ts
dzucconi: For anything we can assume either won't be null or will have a default, yeah. There's a lot in our schema that could be non-nullable. In particular this crops up annoyingly with arrays of things where we can assume there's always going to be an empty array (or we can default to one).

But yeah this is why everything in our codebase is guarded with `?` even when not necessary.

PR #3666, opened 2022-01-17
refactor: make CollectorProfile#privacy String

fladson:
Relates to #3640

After some discussion, we decided to have the privacy field as String in Gravity, so we needed to reflect the changes here as well.

See https://github.com/artsy/gravity/pull/14815 for more context.


olerichter00 (approved):
Thanks a lot for the fix 👍

PR #3668, opened 2022-01-17
fix: Set excludePurchasedArtworks to false

olerichter00:
Addresses [CX-2269]

## Description

There is no need anymore to exclude purchased artworks from "My Collection". Therefore we can default `excludePurchasedArtworks` to false so clients don't have to send the param anymore.

Relates to https://github.com/artsy/eigen/pull/6036

[CX-2269]: https://artsyproduct.atlassian.net/browse/CX-2269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MounirDhahri:
Thanks!

PR #3673, opened 2022-01-19
chore: update Gravity schema

mdole:
_bump!_


mdole:
confirmed that we don't need to worry about `lotEvents` being removed here: https://artsy.slack.com/archives/C01PBEBRB0V/p1642602364044700

PR #3672, opened 2022-01-19
feat: Add otherRelevantPosition to collector profile

olerichter00:
Addresses [CX-2240]

🔴  Depends on https://github.com/artsy/gravity/pull/14812

## Description

Add otherRelevantPosition to collector profile.

[CX-2240]: https://artsyproduct.atlassian.net/browse/CX-2240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3674, opened 2022-01-19
fix(article): filters out unpublished artworks

dzucconi:
So articles may have unpublished works and since we are loading the actual artwork now we need to catch and filter them out.


_schemaV2.graphql
```diff
@@ -13189,6 +13192,9 @@ input UpdateMyProfileInput {
   # Additional personal notes.
   notes: String
 
+  # Collector's position with relevant institutions
+  otherRelevantPosition: String
+
```
dzucconi, line 13197: This comes from https://github.com/artsy/metaphysics/pull/3672 — not this PR.

PR #3675, opened 2022-01-19
fix: Update GraphQL schema

olerichter00:
This PR updates the schema to reflect the changes of https://github.com/artsy/metaphysics/pull/3672.


dzucconi:
This got updated in my PR just now https://github.com/artsy/metaphysics/pull/3674/files so we can close this but thanks for following up!

PR #181, opened 2016-04-06
@craigspaeth: Money fields on SaleArtwork

dzucconi:
Included accounting, which should make it easy to throw in currency conversions.

I think this includes some breaking changes so I'll get that sorted in Force tomorrow and rebased with master here. Let me know if there's something I need to do to coordinate with Prediction, however.

PR #3692, opened 2022-01-25
fix: disambiguate currency symbols

ashleyjelks:
This PR introduces an optional parameter on the amount field that when passed will return the `disambiguate_symbol` subfield if present. 

This should allow us to display `US$XXX` instead of `$XXX` to distinguish US from other dollar-based currencies (e.g. AUD,   HKD) optionally, and without introducing breaking changes in all of the places the money helper is used.

This is a requirement from [NX-2405](https://artsyproduct.atlassian.net/browse/NX-2405) and it fixes this [bug](https://github.com/artsy/force/pull/8963).

[NX-3031](https://artsyproduct.atlassian.net/browse/NX-3031)

Thank you to @starsirius for their help on this!

cc @artsy/negotiate-devs

### DIFF
--- a/src/schema/v2/fields/__tests__/money.test.ts
+++ b/src/schema/v2/fields/__tests__/money.test.ts
@@ -62,7 +62,7 @@ describe("amount", () => {
         obj: { symbol: "$", currencyCode: "GBP" },
         args: {},
       })
-    ).toMatchInlineSnapshot(`"$12.34"`)
+    ).toMatchInlineSnapshot(`"£12.34"`)
   })
 
   it("prefers argument symbol over currencyCode", () => {
@@ -70,13 +70,30 @@ describe("amount", () => {
       getResult({
         obj: { symbol: "£", currencyCode: "GBP" },
       })
-    ).toMatchInlineSnapshot(`"$12.34"`)
+    ).toMatchInlineSnapshot(`"£12.34"`)
   })
 
   it("doesn't break when it can't find a currencyCode", () => {
     expect(getResult({ obj: { currencyCode: "BLAH" } })).toMatchInlineSnapshot(
       `"$12.34"`
     )
+  })
+
+  it("returns disambiguate symbol over symbol when optional param passed", () => {
+    expect(
+      getResult({
+        obj: { currencyCode: "USD", symbol: "$" },
+        args: { disambiguate: true },
+      })
+    ).toMatchInlineSnapshot(`"US$12.34"`)
+  })
+
+  it("returns symbol over disambiguate symbol by default", () => {
+    expect(
+      getResult({
+        obj: { currencyCode: "USD", symbol: "$" },
+      })
+    ).toMatchInlineSnapshot(`"$12.34"`)
   })
 })
 
--- a/src/schema/v2/fields/money.ts
+++ b/src/schema/v2/fields/money.ts
@@ -7,6 +7,7 @@ import {
   GraphQLFloat,
   GraphQLInputObjectType,
   GraphQLNonNull,
+  GraphQLBoolean,
 } from "graphql"
 import { ResolverContext } from "types/graphql"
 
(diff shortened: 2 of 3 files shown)


src/schema/v2/fields/money.ts
```diff
@@ -50,11 +51,17 @@ export const amount = (centsResolver) => ({
       type: GraphQLString,
       defaultValue: ",",
     },
+    disambiguate: {
+      type: GraphQLBoolean,
+      defaultValue: false,
+    },
```
starsirius, line 57: I like that we're aiming for the real goal here, which is to disambiguate ambiguous symbols. This aligns with the [ruby money gem](https://github.com/RubyMoney/money/blob/2d9153b02257783d75a5f3fe9b80bf8250bc67be/lib/money/money/formatter.rb#L175-L182) that MP's [source of truth](https://github.com/artsy/metaphysics/blob/c4763ac50a5b25143723f618a489f5c1ad40eed5/src/lib/currency_codes.json) is based on. This also helps push us towards more consistent display across surface areas, vs. supporting every arbitrary formatting.

starsirius: Unrelated to this PR, I'd be interested in consolidating the source of truth currency_code.json (which was [copied over from the money gem](https://github.com/artsy/metaphysics/pull/883) and has been diverged from that) with the gem. Potentially we can try similar approach in [address_helper](https://github.com/artsy/address_helper/pull/1), and move the file to the [money_helper](https://github.com/artsy/money_helper) gem as the single place. I might try to spike on that on a #FF.

starsirius (approved):
I think we need to update the test; otherwise, looks good to me! 🎉

src/schema/v2/fields/money.ts
```diff
@@ -50,11 +51,18 @@ export const amount = (centsResolver) => ({
...
   },
   resolve: (obj, options) => {
     const cents = centsResolver(obj)
+
     const symbol =
-      options.symbol || obj.symbol || symbolFromCurrencyCode(obj.currencyCode)
+      options.symbol ||
+      obj.symbol ||
+      symbolFromCurrencyCode(obj.currencyCode, options.disambiguate)
```
starsirius, line 65: I did some digging to learn why we'd need `obj.symbol` since it's redundant with `obj.currencyCode` and also curious what objects expose a symbol in addition to the currency code. It turned out it's introduced in https://github.com/artsy/metaphysics/pull/181 for [sale artworks](https://github.com/artsy/gravity/blob/ef9b996155ccb9f6adeffddaff03213f8e96c3e8/app/models/domain/sale_artwork.rb#L142), and we just used the symbol, not the currency code, for formatting at that point.

_I think_ eventually we want to use the currency code on the object instead of the symbol (since it can be ambiguous). For now I'd suggest we keep the behavior and update the description of the `disambiguate` argument to make it clear that it'll only have effect when there is no symbol explicitly passed in as an argument or symbol from the object itself. We can separately investigate if it's ok to ignore the symbol from the object.

In our use cases, Exchange just exposes the [currency code](https://github.com/artsy/exchange/blob/2b581ebbd9d4d2d1af3327ae546af89f1ea8ce08/app/graphql/types/order_interface.rb#L18) and not the symbol, so I think the `disambiguate` here would work.

src/schema/v2/fields/__tests__/money.test.ts
```diff
@@ -78,6 +78,23 @@ describe("amount", () => {
...
+    ).toMatchInlineSnapshot(`"US$12.34"`)
+  })
+
+  it("returns symbol over disambiguate symbol by default", () => {
+    expect(
+      getResult({
+        obj: { currencyCode: "USD", symbol: "$" },
+      })
+    ).toMatchInlineSnapshot(`"$12.34"`)
+  })
```
starsirius, line 97: So based on https://github.com/artsy/metaphysics/pull/3692/files#r794203952, I think we can have more realistic specs to cover our use cases, which is without `symbol` on the object.

```typescript
  it("returns disambiguate symbol when optional param passed", () => {
    expect(
      getResult({
        obj: { currencyCode: "USD" },
        args: { disambiguate: true },
      })
    ).toMatchInlineSnapshot(`"US$12.34"`)
  })

  it("returns currency symbol by default", () => {
    expect(
      getResult({
        obj: { currencyCode: "USD" },
      })
    ).toMatchInlineSnapshot(`"$12.34"`)
  })
```

src/schema/v2/fields/money.ts
```diff
@@ -74,11 +82,22 @@ export const amount = (centsResolver) => ({
...
-    ? currencyCodes[currencyCode.toLowerCase()] &&
-        currencyCodes[currencyCode.toLowerCase()].symbol
-    : null
+export const symbolFromCurrencyCode = (currencyCode, disambiguate = false) => {
+  if (!currencyCode) {
+    return null
+  }
+
+  const disambiguateSymbol =
+    // currencyCode === "USD" &&
```
starsirius, line 91: 🙈

PR #3700, opened 2022-01-28
fix(CX-2262): price_paid_cents check for zero value

dariakoko:
Changed condition that handles the price paid for the artwork in a user's 'my collection. With this fixes collectors can enter the price paid for the artwork equal to 0 and it will be displayed in the artwork details. - daria

Before: 
Simulator Screen Shot - iPhone 12 Pro - 2022-01-28 at 14 09 21

After:
Simulator Screen Shot - iPhone 12 Pro - 2022-01-28 at 14 10 19


olerichter00 (approved):
Thanks for the fix 🙇

PR #3708, opened 2022-02-01
Adding in the query for Notification Preferences: GRO-704

The-Beez-Kneez:
This PR adds a query/mutation duo for the notification preferences.

https://artsyproduct.atlassian.net/browse/GRO-704

### DIFF
--- a/src/schema/v2/notification_preferences.ts
+++ b/src/schema/v2/notification_preferences.ts
@@ -3,9 +3,11 @@ import {
   GraphQLNonNull,
   GraphQLFieldConfig,
   GraphQLList,
+  GraphQLInputObjectType,
   GraphQLObjectType,
   GraphQLString,
 } from "graphql"
+import { mutationWithClientMutationId } from "graphql-relay"
 import { ResolverContext } from "types/graphql"
 
 const subGroupfields = {
@@ -36,6 +38,11 @@ const NotificationPreferenceType = new GraphQLObjectType({
   fields: subGroupfields,
 })
 
+const NotificationPreferenceInputType = new GraphQLInputObjectType({
+  name: "NotificationPreferenceInput",
+  fields: subGroupfields,
+})
+
 export const notificationPreferences: GraphQLFieldConfig<
   void,
   ResolverContext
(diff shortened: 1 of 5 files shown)


src/schema/v2/notification_preferences.ts
```diff
@@ -0,0 +1,115 @@
...
+  GraphQLInputObjectType,
+  GraphQLObjectType,
+  GraphQLString,
+} from "graphql"
+import { mutationWithClientMutationId } from "graphql-relay"
+import { ResolverContext } from "types/graphql"
+
+const subGroupfields = {
+  id: {
+    type: GraphQLString,
```
dzucconi, line 15: So I'm not sure what the difference between input and output in terms of requirement but — should any of these other fields also be non-nullable? Presumably at least ID

dzucconi (approved):
LGTM — feel free to merge unless there's a non-nullable update to be made.

damassi:
There's some stuff going on in that mutation and we should be testing behavior. If there's another mutation that has some setup we can copy from there.

jonallured:
Does [02e3a01](https://github.com/artsy/metaphysics/pull/3708/commits/02e3a013a1e5d1f2eb9b9ba481adeb89380bd86b) address your concerns @damassi?

src/schema/v2/notification_preferences.ts
jonallured: Yeah all these things will be non-null, updated with [132d848](https://github.com/artsy/metaphysics/pull/3708/commits/132d848b6539e4f4e14a9d9b25fa200cd00dee5f). Note technically all Gravity cares about in the POST endpoint is the status/name duo and the rest are ignored but providing the rest doesn't hurt anything.

damassi:
Yup, thank you @jonallured! If there's logic, there's behavior; if there's behavior, it needs to be tested. Sometimes forward momentum can feel more important than tests, but I'm starting to worry that this is becoming a excuse and we need to reign it in a bit, especially because other teams (and artsy, historically) has written test coverage that is close to 100%. 

For those reading, a reminder about **why** we write tests: it's an insurance policy. Future developers very rarely understand the implications of why code was written and how it impacts larger systems. This is more than just a philosophical stance. Using types alone is insufficient, because developers who are not experts at typescript will just change the types to make them more flexible or to accommodate changes in code. However, said developers **_will not just arbitrarily change tests_**. It is the one and only tool that engineers have to ensure changes are properly thought out.

damassi (approved):

jonallured:
Ok I'm going to merge this sucker thanks all!

PR #3710, opened 2022-02-02
Feat [CX-2257]: Expose collector profile from partner

MrSltun:
This PR resolves [CX-2257](https://artsyproduct.atlassian.net/browse/CX-2257)

Exposes the collector profile from an inquiry requests

This PR requires artsy/gravity#14950 to be merged to work


src/schema/v2/me/userInterests.ts
```diff
@@ -34,7 +34,7 @@ export const userInterestCategoryEnum = new GraphQLEnumType({
 
 export const userInterestInterestUnion = new GraphQLUnionType({
   name: "UserInterestInterest",
-  types: [ArtistType, GeneType],
+  types: () => [ArtistType, GeneType],
```
MrSltun, line 37: this change is to resolve `ArtistType` being `undefined`, see [this slack thread 🔒](https://artsy.slack.com/archives/CP9P4KR35/p1643790512633099) for more context

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -267,6 +267,13 @@ export default (accessToken, userID, opts) => {
       {},
       { headers: true }
     ),
+    partnerInquirerProfileLoader: gravityLoader<
```
MounirDhahri, line 270:
```suggestion
    partnerInquirerCollectorProfileLoader: gravityLoader<
```

src/schema/v2/partnerInquirerProfile.ts
```diff
@@ -0,0 +1,99 @@
...
+  bio: {
+    type: GraphQLString,
+  },
+  profession: { type: GraphQLString },
+  otherRelevantPositions: {
+    type: GraphQLString,
+    description: "Collector's position with relevant institutions",
+    resolve: ({ other_relevant_positions }) => other_relevant_positions,
+  },
+  emailVerified: {
```
MounirDhahri, line 51: Let's maybe rename this to `emailConfirmed` to match the me route. See https://github.com/artsy/metaphysics/blob/ad9fc1d6aaffc1ee4c625c616e4fe46db2a16e23/_schemaV2.graphql#L8797

src/schema/v2/__tests__/partnerInquirerProfile.test.ts
```diff
@@ -0,0 +1,60 @@
...
+        partner(id: "catty-partner") {
+          inquiryRequest(inquiryId: "inquiry-id") {
+            collectorProfile {
+              name
+              location {
+                city
+                country
+              }
+              profession
+              bio
```
MounirDhahri, line 38: would you mind also adding some of the new fields like `isActiveBidder` `isActiveInquirer` to the test query

MounirDhahri (commented):
I left some small comments, the rest looks good to me to go. Thanks for this big effort

src/schema/v2/partnerInquirerCollectorProfile.ts
```diff
@@ -0,0 +1,99 @@
...
+  },
+  isActiveBidder: {
+    type: GraphQLBoolean,
+    resolve: ({ previously_registered_for_auction }) =>
+      previously_registered_for_auction ?? false,
+  },
+  collectorProfileArtists: {
+    type: CollectorProfileArtists,
+    resolve: ({ collector_artists }) => collector_artists,
+  },
```
olerichter00, line 72: I wonder if we can give this a more descriptive name. Are these actually artists the user collects or artists the user is interested in?

olerichter00 (changes requested):
This is great work 🌟 
I've just added a comment about the name of the `collectorProfileArtists` field.

src/schema/v2/partnerInquirerCollectorProfile.ts
MrSltun: for now, this is the list of artists the user is interested in, but later it will also include the list of artists collected by the user

olerichter00: Ok, very difficult to find a good name for it. Maybe a good description works as well :)

src/schema/v2/partnerInquirerCollectorProfile.ts
```diff
@@ -0,0 +1,101 @@
...
+      artwork_inquiry_requests_count >= 25,
+  },
+  isActiveBidder: {
+    type: GraphQLBoolean,
+    resolve: ({ previously_registered_for_auction }) =>
+      previously_registered_for_auction ?? false,
+  },
+  collectorProfileArtists: {
+    type: new GraphQLList(CollectorProfileArtists),
+    description: "List of artists names the Collector is interested in",
```
olerichter00, line 72:
```suggestion
    description: "List of artists the Collector is interested in.",
```

olerichter00 (approved):
🚀 🚀 🚀

dzucconi:
How did we wind up with a separate collector profile type for partner inquiries when we already have a CollectorProfile. This means they are completely incompatible on the front-end. Are they really distinct things? Can we fix this?

MrSltun:
@dzucconi My reasoning for it was to expose these fields only within the context of an inquiry request, which probably was not the best decision
We can fix it as a follow-up on #4628 and only use the `CollectorProfile` type

dzucconi:
Ah, I see. Yeah we want to think about our types as modeling our domain. It'd be good to just deprecate and fix this up rather than let these things grow apart and splinter further.

araujobarret:
@MrSltun I can open a PR tomorrow to deprecate it will assign it to CX folks as soon as I have something

PR #3714, opened 2022-02-02
feat: Update MyBids endpoint to include sales loaded from gravity that are not synced to causality

annacarey:
[BID-180](https://artsyproduct.atlassian.net/browse/BID-180)

Putting up a draft for the MP work to allow us to no longer rely on syncing data to Causality. This PR loads in lot standings from a gravity endpoint for sales where events are not synced to causality. The lot standings from gravity are combined with the Causality lot standings (and formatted to match).


src/schema/v2/me/myBids.ts
```diff
@@ -379,3 +396,32 @@ function withSaleInfo(sale): SaleSortingInfo {
...
+
+function getSoldStatus(lotStanding): string {
+  const { sale_artwork } = lotStanding
+  const { artwork } = sale_artwork
+  if (artwork.forSale) {
+    return "ForSale"
+  } else if (artwork.sold) {
+    return "Sold"
+  } else {
+    return "Passed" // this is incorrect
```
annacarey, line 425: What does "passed" mean? Any idea where to find that from the lot standing?

src/schema/v2/me/myBids.ts
```diff
@@ -379,3 +396,32 @@ function withSaleInfo(sale): SaleSortingInfo {
...
+        internalID: lotStanding.id,
+        saleId: lotStanding.bidder.sale.id,
+        soldStatus: getSoldStatus(lotStanding),
+        sellingPriceCents: lotStanding.highest_bid_amount_cents,
+      },
+    }
+  })
+}
+
+function getSoldStatus(lotStanding): string {
```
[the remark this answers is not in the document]
damassi: "verbose" and "ugly" is almost always better than "clever" and "concise"! (this is the hardest thing for so many to understand)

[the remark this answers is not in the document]
damassi: If there were many (many) more options I often prefer a switch 
```tsx

switch (true) {
  case artwork.forSale: {
    return "ForSale"
  }
  case artwork.sold: {
    return "Sold"
  }
  default: {
    return "Passed"
  }
}
```
But again, for this to be clearer than a conventional if/else, there probably needs to be more cases. 

Or if you want the above to be assigned to a variable, can use an immediate function invocation:

```tsx 
const saleStatus = (() => {
  switch (true) {
  case artwork.forSale: {
    return "ForSale"
  }
  case artwork.sold: {
    return "Sold"
  }
  default: {
    return "Passed"
  }
})()
```

src/schema/v2/me/myBids.ts
mzikherman: i _think_ that's something for live auctions so can probably skip/not worry about it here

src/schema/v2/me/myBids.ts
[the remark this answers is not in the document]
mzikherman: +1 to verbose over cleverly concise. No need to outsmart our future selves like that! 😆

src/schema/v2/me/myBids.ts
```diff
@@ -379,3 +396,33 @@ function withSaleInfo(sale): SaleSortingInfo {
...
+  gravityLotStandings
+): LotStandingResponse {
+  return gravityLotStandings.map((lotStanding) => {
+    const { sale_artwork, id, bidder, highest_bid_amount_cents } = lotStanding
+    return {
+      lot: {
+        bidCount: sale_artwork.bidder_positions_count,
+        reserveStatus: sale_artwork.reserve_status,
+        internalID: id,
+        saleId: bidder.sale.id,
```
mzikherman, line 410: This should be `bidder.sale._id` (since id is a slug)

src/schema/v2/me/myBids.ts
```diff
@@ -379,3 +396,33 @@ function withSaleInfo(sale): SaleSortingInfo {
     liveBiddingStarted,
   }
 }
+
+function formatGravityLotStandingsToMatchCausalityLotStandings(
+  gravityLotStandings
+): LotStandingResponse {
+  return gravityLotStandings.map((lotStanding) => {
+    const { sale_artwork, id, bidder, highest_bid_amount_cents } = lotStanding
+    return {
```
mzikherman, line 405: I think there's a `isHighestBidder` needed too.

src/schema/v2/me/myBids.ts
```diff
@@ -379,3 +396,33 @@ function withSaleInfo(sale): SaleSortingInfo {
     liveBiddingStarted,
   }
 }
+
+function formatGravityLotStandingsToMatchCausalityLotStandings(
+  gravityLotStandings
+): LotStandingResponse {
+  return gravityLotStandings.map((lotStanding) => {
+    const { sale_artwork, id, bidder, highest_bid_amount_cents } = lotStanding
```
mzikherman, line 404: since the lot standing is a 'synthetic' object, it has the following properties: https://github.com/artsy/gravity/blob/1ffe27dc7d5a55b36757ed1cf8998ca64b963f49/app/services/lot_standing_service.rb#L37-L40

(so no `id` or `highest_bid_amount_cents`). In general the `lotState` object prob needs a few more things added to it!

src/schema/v2/me/myBids.ts
```diff
@@ -379,3 +396,33 @@ function withSaleInfo(sale): SaleSortingInfo {
...
+        sellingPriceCents: highest_bid_amount_cents,
+      },
+    }
+  })
+}
+
+function getSoldStatus(lotStanding): string {
+  const { sale_artwork } = lotStanding
+  const { artwork } = sale_artwork
+  if (artwork.forSale) {
```
mzikherman, line 421: This is probably `artwork` from the JSON response for a lot standing, and not a GraphQL type, so would be using underscored fields from the Gravity API (and not any GraphQL fields).

mzikherman:
I pushed a small update that lines up the data so the expected sale artwork data is being populated. Check out my commit, the issue was an id vs _id (id vs. slug) being used as the key in an object and so data wasn't matching up.

nspinazz89 (commented):
Looking good so far!

src/schema/v2/me/myBids.ts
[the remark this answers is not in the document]
annacarey: Thank you both. I still get caught up with this stuff but my first more-basic-syntax instinct is probably not just easier for me to understand, but also easier for others

src/schema/v2/me/myBids.ts
```diff
@@ -172,25 +179,35 @@ export const MyBids: GraphQLFieldConfig<void, ResolverContext> = {
     // Fetch everything in parallel
     const [
       causalityResponse,
+      newBiddingEngineLotStandingsResponse,
```
annacarey, line 182: Rename variable to something about not synced to causality

src/schema/v2/me/myBids.ts
bhoggard: Passed is from live auction integrations, where the lot didn't sell.

mzikherman:
Amazing!

damassi:
Yah, props on digging into this insane code

PR #3723, opened 2022-02-03
feat: Add submitOfferOrderWithConversation mutation

araujobarret:
[NX-3043]

@artsy/negotiate-devs 

This PR adds a new mutation that stitches to `Exchange` on `commerceSubmitOrderWithOffer` and then creates the inquiry passing the `orderId` to the `InquiryRequest` endpoint in Gravity that is going to handle the conversation creation and the link of the conversation and the order in exchange.

[NX-3043]: https://artsyproduct.atlassian.net/browse/NX-3043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -546,6 +554,76 @@ export const exchangeStitchingEnvironment = ({
...
+
+            if (orderOrError.error || !orderOrError.order) {
+              return submitOrderWithOffer
+            }
+
+            try {
+              const {
+                order: { internalID: orderId },
+              } = orderOrError
+              await submitArtworkInquiryRequestLoader({
```
araujobarret, line 613: @erikdstock I didn't refactor the previous mutation, but maybe we could reuse this loader in the previous mutation, idk the details of how we use it but it looks like they have something in common
Could be a good thing to keep Gravity as the source of truth in terms of linking the conversation with the order there

src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -546,6 +554,76 @@ export const exchangeStitchingEnvironment = ({
             return offerResult
           },
         },
+        submitOrderOffer: {
```
lean8086, line 557: Super minimal question: shouldn't it be **offer order** instead of **order offer**? (`submitOfferOrder`) I mostly see it expressed that way in our code and not the other way around. Plus, I think it's semantically an offer order, right? 😛

araujobarret: I used Exchange's as the base, the name of the mutation there is `submitOrderWithOffer` since Offer is a sub thing under Order I thought that made sense but I can change it

lean8086: nvm, I see it's because we use commerceSubmitOrderWithOffer

src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -254,10 +254,18 @@ export const exchangeStitchingEnvironment = ({
       orders(first: Int, last: Int, after: String, before: String, mode: CommerceOrderModeEnum, sellerId: String, sort: CommerceOrderConnectionSortEnum, states: [CommerceOrderStateEnum!]): CommerceOrderConnectionWithTotalCount
     }
 
+    extend input CommerceSubmitOrderWithOfferInput {
+      note: String
+      artworkId: ID
+    }
```
starsirius, line 260: Since `submitOrderOffer` is a separate mutation from `commerceSubmitOrderWithOffer`, I wonder if it's better to have its own input type, rather than expanding `CommerceSubmitOrderWithOfferInput`.

And actually, since we are creating an Exchange order first. Maybe the `note` and `artworkId` are already available from the created order, and clients don't need to pass them explicitly as arguments?

src/lib/stitching/exchange/v2/stitching.ts
starsirius, line 557: I'd plus one saying `offerOrder` (rather than `orderOffer`) since it reads better to me order is the parent model and the actual thing we're submitting.

I'd also suggest a way to make it clear that this mutation will submit the offer order _and_ create an associated inquiry/conversation. Don't have a good suggestion about naming (maybe `submitConversationalOfferOrder` or just `submitOfferOrderAndCreateConversation`?) but at least we can include a description or comment.

src/lib/stitching/exchange/v2/stitching.ts
araujobarret: Thanks for taking a look @starsirius 🙌 

1 - I was asking myself the same question when extending, tried to keep it like this if we change something on exchange we would be automatically extending it here, but I guess I need to spread the other exchange props and use them in the args to make it work as I expect, but even though you think that we need to create a different one?

2 - good point, we can extend the WrapQuery to grab those inputs and use them as well, going to do it tomorrow morning!

src/lib/stitching/exchange/v2/stitching.ts
araujobarret: Makes sense to me, going to rename it better to avoid confusion and provide a description, I missed that one

src/lib/stitching/exchange/v2/stitching.ts
araujobarret: So we don't need to extend any input if those fields are already in the Exchange response 🙏

araujobarret: @starsirius all the requested changes in [91a98eb](https://github.com/artsy/metaphysics/pull/3723/commits/91a98ebbe8651fd6269a4cd840bd88851398b95f)

src/lib/stitching/exchange/v2/stitching.ts
araujobarret: I confess I preferred `submitOfferOrderWithConversation` with a description simple like `Submits an OfferOrder and creates a conversation for it`. But if stills confusing I can change it again

lidimayra (approved):

PR #3726, opened 2022-02-04
fix: Load otherRelevantPositions from collector profile

olerichter00:
Addresses [CX-2179]

## Description

This PR fixes the issue of `otherRelevantPositions` returning `null` by fetching the field from the Gravity collector profile endpoint instead of Gravity's me endpoint.

[CX-2179]: https://artsyproduct.atlassian.net/browse/CX-2179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/me/index.ts
+++ b/src/schema/v2/me/index.ts
@@ -9,11 +9,12 @@ import {
   GraphQLString,
 } from "graphql"
 import { includesFieldsOtherThanSelectionSet } from "lib/hasFieldSelection"
+import { StaticPathLoader } from "lib/loaders/api/loader_interface"
 import date from "schema/v2/fields/date"
 import initials from "schema/v2/fields/initials"
 import { IDFields, NodeInterface } from "schema/v2/object_identification"
 import { ResolverContext } from "types/graphql"
-import Image, { normalizeImageData } from "../image"
+import Image from "../image"
 import { PhoneNumber } from "../phoneNumber"
 import { SaleArtworksConnectionField } from "../sale_artworks"
 import { ArtistRecommendations } from "./artistRecommendations"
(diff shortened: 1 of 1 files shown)


src/schema/v2/me/index.ts
```diff
@@ -239,7 +244,7 @@ export const meType = new GraphQLObjectType<any, ResolverContext>({
     otherRelevantPositions: {
       type: GraphQLString,
       description: "Collector's position with relevant institutions",
-      resolve: ({ other_relevant_positions }) => other_relevant_positions,
+      resolve: collectorProfileResolver("other_relevant_positions"),
```
olerichter00, line 247: `other_relevant_positions` is not exposed from the `meLoader` but from the `collectorProfileLoader`.

src/schema/v2/me/index.ts
```diff
@@ -53,6 +54,22 @@ import { SavedArtworks } from "./savedArtworks"
 import { ShowsByFollowedArtists } from "./showsByFollowedArtists"
 import { WatchedLotConnection } from "./watchedLotConnection"
 
+/**
+ * @deprecated: Please use the CollectorProfile type instead of adding fields to me directly.
+ */
```
olerichter00, line 59: For some reason, we started adding collector profile attributes to `me` instead of `me.collectorProfile`. I've added this helper function with a deprecation warning to make sure we add these fields to `me.collectorProfile` in the future.

MounirDhahri (approved):
Thanks for fixing this Ole

PR #3743, opened 2022-02-09
Fix[SWA-150]: Add loaders without authentification

Gon4ar5:
Fix for [SWA-150](https://artsyproduct.atlassian.net/browse/SWA-150)

We need to add some loaders for unauthenticated requests(cause we use XAPP token instead of X-ACCESS)


kiryl-zubarau (dismissed):

iskounen:
We decided to close this PR because we will continue using access tokens (instead of app tokens) to authenticate requests from Convection to Gravity (via Metaphysics).

PR #3747, opened 2022-02-09
fix [CX-2257]: update  partner inquirer collector profile fields

MrSltun:
This PR resolves [CX-2257](https://artsyproduct.atlassian.net/browse/CX-2257)

This is a follow up on #3710 

## Description
This PR adds `userIntroduction` field into the Partner inquirer collector profile endpoint to be used by the old design of the the collector profile
It also adds `affiliatedGalleryIds` and `affiliatedGalleryIds` fields to be used in the collector profile controller in volt

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -8152,6 +8152,11 @@ type ImageURLs {
 }
 
 type InquirerCollectorProfile {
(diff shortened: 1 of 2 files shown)


olerichter00 (approved):
🚀

PR #3751, opened 2022-02-10
feat: Omit draft submissions from My Collection artwork

olerichter00:
Addresses [CX-2295]

## Description

Omit draft submissions from My Collection artwork.

### DIFF
--- a/src/schema/v2/me/myCollection.ts
+++ b/src/schema/v2/me/myCollection.ts
@@ -152,13 +152,21 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
   },
 }
 
-const enrichArtworks = (artworks: any, submissions: any) => {
+const enrichArtworks = (
+  artworks: any[],
+  submissions: any[],
+  filterDraftSubmissions = true
+) => {
   if (submissions.length === 0) {
     return artworks
   }
 
+  const filteredSubmissions = filterDraftSubmissions
+    ? submissions.filter((submission) => submission.state !== "draft")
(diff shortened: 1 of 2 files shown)


src/schema/v2/me/myCollection.ts
```diff
@@ -152,13 +152,21 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
   },
 }
 
-const enrichArtworks = (artworks: any, submissions: any) => {
+const enrichArtworks = (
+  artworks: any[],
+  submissions: any[],
+  filterDraftSubmissions = true
```
nickskalkin, line 158: Is there a need for `filterDraftSubmissions`? We always filter out draft submissions, so I don't see the need for this argument

nickskalkin (changes requested):

src/schema/v2/me/myCollection.ts
olerichter00: I wanted to make it explicit that we filter them out when we enrich the artworks. Just adding `filter(a => a.state !== 'draft')` hides the fact that we filter them out. But I'm open to changing it.

nickskalkin: I would add a documentation comment explaining the behaviour then. It doesn't seems right for me to add an optional attribute just to highlight the behaviour.

olerichter00: 👍

nickskalkin (approved):

PR #3752, opened 2022-02-10
[CX-2328]: adds inProgress to artworkConsignmentSubmission

nickskalkin:
The goal is to hide Artwork Edit Button when the submission is in progress. This is a BE part of the story: adds inProgress field to artworkConsignmentSubmission type

Jira - https://artsyproduct.atlassian.net/browse/CX-2328

### DIFF
--- a/src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
+++ b/src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
@@ -1,4 +1,4 @@
-import { GraphQLObjectType, GraphQLString } from "graphql"
+import { GraphQLObjectType, GraphQLString, GraphQLBoolean } from "graphql"
 import { ResolverContext } from "types/graphql"
 
 const ArtworkConsignmentSubmissionType = new GraphQLObjectType<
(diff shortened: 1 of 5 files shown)


src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
```diff
@@ -23,6 +23,24 @@ const ArtworkConsignmentSubmissionType = new GraphQLObjectType<
...
+            "approved",
+            "hold",
+            "open",
+          ]
+
+          return inProgressSubmissionStates.includes(
+            consignmentSubmission.state
+          )
+            ? true
+            : false
```
olerichter00, line 41:
```suggestion
          return inProgressSubmissionStates.includes(
            consignmentSubmission.state
          )
```

olerichter00 (changes requested):
Great work 🌟 

We probably need to fetch the submission inside of the [artwork resolver](https://github.com/artsy/metaphysics/blob/46e659da1f7d9d98717ff7f011cb56768a83ed25/src/schema/v2/artworks.ts#L28) as well to make the field accessible on the My Collection artwork detail screen ([query](https://github.com/artsy/eigen/blob/5486de3052e95b1b341bf88f0362e5491776e6af/src/lib/Scenes/MyCollection/Screens/Artwork/MyCollectionArtwork.tsx#L19)).

src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
nickskalkin: ahah, I'm idiot :) thanks for catching this

olerichter00 (approved):
LGTM

PR #3753, opened 2022-02-10
Remove unneeded conditionals from inquireable

lidimayra:
[NX-3081] We don't need to check here if an artwork is not `ecommerce` in order to determine if it's `inquireable`. This is Gravity responsibility.

See: https://github.com/artsy/gravity/pull/14971

[NX-3081]: https://artsyproduct.atlassian.net/browse/NX-3081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -531,7 +531,7 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
       isInquireable: {
         type: GraphQLBoolean,
         description: "Do we want to encourage inquiries on this work?",
-        resolve: ({ ecommerce, inquireable }) => !ecommerce && inquireable,
+        resolve: ({ inquireable }) => inquireable,
       },
       isInAuction: {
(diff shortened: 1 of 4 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -531,7 +531,7 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
       isInquireable: {
         type: GraphQLBoolean,
         description: "Do we want to encourage inquiries on this work?",
-        resolve: ({ ecommerce, inquireable }) => !ecommerce && inquireable,
+        resolve: ({ inquireable }) => inquireable,
```
araujobarret, line 534: makes total sense to me ✔️

araujobarret (approved):
Looks good! Idk why we had this e-commerce check before 🤔

PR #3762, opened 2022-02-11
feat(alerts): render human-friendly criteria labels

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-3216

Supersedes:  ~https://github.com/artsy/gravity/pull/14926~

This gives Metaphysics the ability to add human-readable search criteria into the `SearchCriteria` response coming from upstream in Gravity.

So, whereas Gravity's GraphQL API might produce a response like this…

grav

Metaphysics can now additionally fulfill a new `labels` field on `SearchCriteria`…

mp

See the first commit (https://github.com/artsy/metaphysics/pull/3762/commits/7fec9bda1e4f078903c1566c95fd61580c596147) where the basic plumbing is laid down. Subsequent commits add the various facets, one at a time.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -12388,6 +12388,9 @@ type SearchCriteria {
   href: String!
   inquireableOnly: Boolean
   internalID: ID!
+
+  # Human-friendly labels that are added by Metaphysics to the upstream SearchCriteria type coming from Gravity
+  labels: [SearchCriteriaLabel]
   locationCities: [String!]!
   majorPeriods: [String!]!
   materialsTerms: [String!]!
@@ -12445,6 +12448,18 @@ type SearchCriteriaEdge {
 
   # The item at the end of the edge.
   node: SearchCriteria
+}
+
+# Human-friendly representation of a single SearchCriteria filter
+type SearchCriteriaLabel {
+  # The GraphQL field name of the filter facet
+  field: String
+
+  # The human-friendly name of the filter facet
+  name: String
+
+  # The human-friendly value of the filter facet
+  value: String
 }
 
 # A search criteria or errors object
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -12390,7 +12390,7 @@ type SearchCriteria {
   internalID: ID!
 
   # Human-friendly labels that are added by Metaphysics to the upstream SearchCriteria type coming from Gravity
-  labels: [SearchCriteriaLabel]
+  labels: [SearchCriteriaLabel!]!
   locationCities: [String!]!
   majorPeriods: [String!]!
   materialsTerms: [String!]!
@@ -12453,13 +12453,13 @@ type SearchCriteriaEdge {
 # Human-friendly representation of a single SearchCriteria filter
 type SearchCriteriaLabel {
   # The GraphQL field name of the filter facet
-  field: String
+  field: String!
 
   # The human-friendly name of the filter facet
-  name: String
+  name: String!
 
   # The human-friendly value of the filter facet
-  value: String
+  value: String!
 }
 
 # A search criteria or errors object
--- a/src/lib/stitching/gravity/v2/stitching.ts
+++ b/src/lib/stitching/gravity/v2/stitching.ts
@@ -796,6 +796,7 @@ export const gravityStitchingEnvironment = (
               offerable
               materialsTerms
               locationCities
+              majorPeriods
             }
             `,
           resolve: resolveSearchCriteriaLabels,
--- a/src/lib/stitching/gravity/v2/stitching.ts
+++ b/src/lib/stitching/gravity/v2/stitching.ts
@@ -798,6 +798,7 @@ export const gravityStitchingEnvironment = (
               locationCities
               majorPeriods
               colors
+              partnerIDs
             }
             `,
           resolve: resolveSearchCriteriaLabels,
--- a/src/lib/stitching/gravity/v2/stitching.ts
+++ b/src/lib/stitching/gravity/v2/stitching.ts
@@ -82,7 +82,7 @@ export const gravityStitchingEnvironment = (
         ): UserAddressConnection
       }
       extend type SearchCriteria {
-        labels: [SearchCriteriaLabel]
+        labels: [SearchCriteriaLabel!]!
       }
       extend type UserAddress {
         id: ID!
--- a/src/lib/stitching/gravity/v2/stitching.ts
+++ b/src/lib/stitching/gravity/v2/stitching.ts
@@ -785,6 +785,7 @@ export const gravityStitchingEnvironment = (
             ... on SearchCriteria {
               artistIDs
               attributionClass
+              additionalGeneIDs
             }
             `,
           resolve: resolveSearchCriteriaLabels,
--- a/src/lib/stitching/gravity/v2/stitching.ts
+++ b/src/lib/stitching/gravity/v2/stitching.ts
@@ -797,6 +797,7 @@ export const gravityStitchingEnvironment = (
               materialsTerms
               locationCities
               majorPeriods
+              colors
             }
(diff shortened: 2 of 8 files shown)


_schemaV2.graphql
```diff
@@ -12447,6 +12450,18 @@ type SearchCriteriaEdge {
...
+  # The GraphQL field name of the filter facet
+  field: String
+
+  # The human-friendly name of the filter facet
+  name: String
+
+  # The human-friendly value of the filter facet
+  value: String
+}
+
```
anandaroop, line 12464: This is a new type defined here in MP, and is used to define the new `labels` field which is inserted into the `SearchCriteria` type:

label

src/schema/v2/__tests__/searchCriteriaLabel.test.ts
```diff
@@ -0,0 +1,375 @@
...
+    const labels = await resolveSearchCriteriaLabels(parent, _, _, _)
+
+    expect(labels).toIncludeAllMembers([
+      {
+        name: "Price",
+        value: expect.anything(), // TODO: fix this placeholder formatting
+        field: "priceRange",
+      },
+    ])
+  })
```
anandaroop, line 180: In the interest of momentum I'm wiring up the basics here, but deferring the persnickety formatting for size and price to future PRs, since they likely involve some user and/or locale wrangling. (We already have some subtasks [here](https://artsyproduct.atlassian.net/browse/FX-3216) for that, too)

src/schema/v2/searchCriteriaLabel.ts
```diff
@@ -0,0 +1,346 @@
...
+    offerable,
+    materialsTerms,
+    locationCities,
+    majorPeriods,
+    colors,
+    partnerIDs,
+  } = parent
+
+  const { artistLoader, partnerLoader } = context
+
```
anandaroop, line 77: The artist and partner labels are actually fulfilled by going back to Gravity to request the names belonging to the corresponding IDs that are stored in the SearchCriteria.

src/schema/v2/searchCriteriaLabel.ts
```diff
@@ -0,0 +1,346 @@
...
+    nft: "NFT",
+    design: "Design",
+    drawing: "Drawing",
+    installation: "Installation",
+    "film-slash-video": "Film/Video",
+    jewelry: "Jewelry",
+    "performance-art": "Performance Art",
+    reproduction: "Reproduction",
+    "ephemera-or-merchandise": "Ephemera or Merchandise",
+  }
```
anandaroop, line 152: During an earlier knowledge share, I showed a version where gene names (corresponding to the Medium filter) were _also_ being fetched from Gravity.

But it makes sense to do those with a static lookup here, since there are only a dozen or so possibilities (unlike artists and partners which are unbounded collections).

src/schema/v2/searchCriteriaLabel.ts
```diff
@@ -0,0 +1,346 @@
...
+        value: partner.name,
+        field: "partnerIDs",
+      }
+    })
+  )
+}
+
+function capitalizeWords(str: string) {
+  return str.replace(/\b([a-z])/g, (_match, p1) => p1.toUpperCase())
+}
```
anandaroop, line 346: This was a quick local solution to the capitalization needs here — since I couldn't seem to find something like this in MP.

But I wonder if it makes sense to actually just import https://github.com/artsy/to-title-case into MP?

src/schema/v2/searchCriteriaLabel.ts
```diff
@@ -0,0 +1,346 @@
...
+    value: city,
+    field: "locationCities",
+  }))
+}
+
+function getPeriodLabels(majorPeriods: string[]) {
+  if (!majorPeriods?.length) return []
+
+  const DISPLAY_TEXT: Record<string, string> = {
+    "2020": "2020–Today",
```
mdole, line 290: time to set a slack reminder `/remind me to update this label on jan 1 2030` 😀

src/schema/v2/searchCriteriaLabel.ts
```diff
@@ -0,0 +1,346 @@
...
+    materialsTerms,
+    locationCities,
+    majorPeriods,
+    colors,
+    partnerIDs,
+  } = parent
+
+  const { artistLoader, partnerLoader } = context
+
+  const labels: any[] = []
```
mdole, line 78: Just confirming my understanding: is it correct to say that we can't type this from the get-go because we're not sure what type `getArtistLabels` and `getPartnerLabels` will return?

src/schema/v2/searchCriteriaLabel.ts
mdole: Yeah, that sounds reasonable to me. This did make me think about whether we could further consolidate our filter logic in MP in the future - basically doing what's in this PR for our filter menus to avoid clients like [Force doing a similar conversion](https://github.com/artsy/force/blob/880f2132d53d993e5f1ac9afe5507c24f1226247/src/v2/Components/ArtworkFilter/ArtworkFilters/MediumFilter.tsx#L89). Definitely not something for right now, but maybe down the road.

mdole (approved):
Awesome work @anandaroop 🎉  this is looking great to me, and big +1 to your decision to split some of the fiddly term wrangling (dimensions, price, etc.) into separate tasks so that we can do that work in parallel and this PR doesn't get stuck in limbo.

Not much else to add!

src/schema/v2/searchCriteriaLabel.ts
anandaroop: Good q — it sure seems like `labels` should be typed as `const labels: SearchCriteriaLabel[]`

But that ends up being incorrect because, as you suspect, some of those `get*` methods return a simple `{name, value, field}` object and some return an array of such objects. (Some even may return `undefined`)

It's only after that resulting `labels` array gets compacted and flattened on L100 that we can really assert that it is `SearchCriteriaLabel[]`

Prior to that it's technically `SearchCriteriaLabel | SearchCriteriaLabel[] | undefined)[]` which seemed ugly and not helpful — but idk maybe that is actually more informative than `any`?

src/lib/stitching/gravity/v2/stitching.ts
```diff
@@ -80,6 +81,9 @@ export const gravityStitchingEnvironment = (
           before: String
         ): UserAddressConnection
       }
+      extend type SearchCriteria {
+        labels: [SearchCriteriaLabel]
```
dblandin, line 85: It's harder to make these fields non-nullable later so unless we anticipate `labels: null` or `labels: [null]`, we might want to opt for the stricter return type now.

```suggestion
        labels: [SearchCriteriaLabel!]!
```

src/schema/v2/searchCriteriaLabel.ts
```diff
@@ -0,0 +1,346 @@
...
+export const SearchCriteriaLabel = new GraphQLObjectType<
+  SearchCriteriaLabel,
+  ResolverContext
+>({
+  name: "SearchCriteriaLabel",
+  description:
+    "Human-friendly representation of a single SearchCriteria filter",
+  fields: {
+    field: {
+      type: GraphQLString,
```
dblandin, line 33: Similarly here - unless we anticipate null values, would be best to make the return types non-nullable from the beginning:

```suggestion
      type: new GraphQLNonNull(GraphQLString),
```

dblandin (approved):
This looks great!! So excited to have this field in the schema 🥳 

Left a couple comments regarding nullability in the schema. LGTM otherwise!

src/schema/v2/searchCriteriaLabel.ts
anandaroop: Replaced in ac2bff95257c98543b4006ee7a652e89819ad7c7

PR #3765, opened 2022-02-14
[CX-2343]: Adds consignment submission statuses to artworks

nickskalkin:
Jira: https://artsyproduct.atlassian.net/browse/CX-2343

The task is to add the rest of the statuses from this document: https://www.notion.so/artsy/SWA-My-Collection-integration-8d02370063b444e19036dc19f8e73915#b8d078fc5c064e0592c67da44fd824e8

Another change - now it is possible to query consignmentSubmission directly by artwork:

```
{
        artwork(id: ...) {
		title
		internalID
		consignmentSubmission {
			inProgress
		}
	}
}
```

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -154,6 +154,20 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
       },
       consignmentSubmission: {
         type: ArtworkConsignmentSubmissionType,
+        resolve: async ({ id }, _options, { submissionsLoader }) => {
+          if (!submissionsLoader) {
+            return
+          }
+
+          const submissions = await submissionsLoader({ size: 1000 })
+          const filteredSubmissions = submissions.filter(
+            (submission) => submission.state !== "draft"
+          )
+
+          return filteredSubmissions.find((submission) => {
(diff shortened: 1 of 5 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -154,6 +154,20 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+
+          const submissions = await submissionsLoader({ size: 1000 })
+          const filteredSubmissions = submissions.filter(
+            (submission) => submission.state !== "draft"
+          )
+
+          return filteredSubmissions.find((submission) => {
+            return submission.my_collection_artwork_id === id
+          })
+        },
```
olerichter00, line 170: Have you checked how many requests to Convection we will fire when we load many artworks? If it just fires one request, this solution is pretty nice 👍

olerichter00 (commented):
Is this ready for review?

src/schema/v2/artwork/index.ts
nickskalkin: I've checked local metaphysics logs and I see only 1 requests to convection. If you know a better way how to check if the code triggers >1 request - please let me know

olerichter00 (approved):

PR #3771, opened 2022-02-15
[NX-3084] Handle missing scenario in shippingInfo logic

lidimayra:
[NX-3084]

We weren't handling one possible scenario when building `shippingInfo`: The possibility of domestic shipping fee being null while international shipping fee is present. This was leading to the following string being returned for this case:

`Shipping: null domestic, $1000 rest of world`

Changes in this PR changes this to:
`Shipping: $1000 worldwide`


src/schema/v2/artwork/index.ts
```diff
@@ -809,6 +809,12 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           )
             return `Shipping: ${domesticShipping} ${shippingRegion} only`
 
+          if (
+            artwork.domestic_shipping_fee_cents == null &&
+            artwork.international_shipping_fee_cents !== null
+          )
```
oxaudo, line 815: Hi:) Happy to discuss it in person but I think this is not a true statement here. Don't think we would ever apply international shipping cents to domestic shipping so the phrase `worldwide` is misleading here. 

This state should never be possible for bnmo works (ui constraints supposed to prevent it in CMS) and for non bnmo works we were never using/displaying this data. As we are starting to do it now for inquiry works - I think we need to reconsider validations on artwork and decide what exactly we want to show on inquiry works.

As far as this pr concerned - I think if we don't want to display `Shipping calculated at checkout here` the more proper solution here is to return empty string in case domestic is not present but international is present.

lidimayra: oooh, got it, many thanks for all the context, and sorry about the confusion!! 🙇‍♀️ 
I think `Shipping calculated at checkout` makes sense at this point, so we can keep consistency with the other scenarios. Just made this change and squashed it.

src/schema/v2/artwork/index.ts
```diff
@@ -809,6 +809,12 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           )
             return `Shipping: ${domesticShipping} ${shippingRegion} only`
 
+          if (
```
araujobarret, line 812:
```suggestion
          if (!artwork.domestic_shipping_fee_cents && !artwork.international_shipping_fee_cents) {
```
would cover undefined and null cases here

despite the suggestion what about the case where the `international_shipping_fee_cents` is undefined shouldn't we also cover it with || instead of &&?

araujobarret: nevermind my suggestion, I see we have handling when the value is 0(false) this suggestion would mess with the current logic there

src/schema/v2/artwork/index.ts
oxaudo: This change is better but I'm wondering if the more proper way to handle it would be updating the first couple of line of this method and using this newly added `process_with_arta_shipping`.

So something like:
```
resolve: (artwork) => {
          if (artwork.process_with_arta_shipping) {
            return "Shipping: Calculated in checkout"
          }

          if (artwork.domestic_shipping_fee_cents == null)
            return "Shipping, tax, and additional fees quoted by seller"
      
          //and the test is without change from master....
```

I *think* it would most accurately describe what will happen to shipping. If the work is arta or we'll silently opt it into Arta - we can say that collector can see shipping at checkout.
The artwork is not complete - i.e. missing domestic shipping - we'll let partner define what is the correct shipping will be and the incomplete offer will be generated. 

Sorry for all the back and force @lidimayra . I'm just trying to figure out what is actually true for collector here.

lidimayra: No apologies needed, I'm actually thankful for your patience, I'm learning tons in the process!! ❤️ 

Just to confirm I got it: we were already displaying`Shipping, tax, and additional fees quoted by seller` message in case `artwork.domestic_shipping_fee_cents` is null and `artwork.international_shipping_fee_cents` is also null. Apparently, we don't actually care about `artwork.international_shipping_fee` in order to control this message, right? In case `artwork.international_shipping_fee` is null, this condition must be enough to display this, so I changed the existing example to remove the unwanted part. And this won't ever be reached in case artwork will be processed with Arta, because in this scenario, it won't be quoted by seller, it will be quoted according to Arta shipping option.

Please, let me know in case this doesn't sound good.

araujobarret (approved):

oxaudo (approved):
Looks great to me now except for this redundant now `if`. I'm out next week but please feel free to merge when ready :)

src/schema/v2/artwork/index.ts
```diff
@@ -809,6 +806,12 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
```
oxaudo, line 809: So this whole `if` is redundant now as the cae would be covered above on line `767` and the code here would never be executed. So I think we should kill 809 - 814 lines here @lidimayra

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -2165,7 +2164,7 @@ describe("Artwork type", () => {
     })
 
     it("is set to calculated at checkout when Arta shipping", () => {
```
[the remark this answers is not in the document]
lidimayra: Thanks for the suggestion, just performed this change and squashed it!

src/schema/v2/artwork/index.ts
lidimayra: Absolutely, I was supposed to remove these, and I totally forgot about it! Just squashed this change

lidimayra:
Thank you so much @oxaudo for putting so much effort into reviewing this PR ensuring not only that we have messages that tell the truth but also that we, as a team, have a deep understanding of this flow, by communicating so clearly the reasoning behind every suggestion!!

Your comments here were a masterclass and I really appreciate it! 🙌 ❤️

PR #3772, opened 2022-02-15
feat(article): corrects author

dzucconi:
So I'm deprecating two fields which are the denormalized author data; these have IDs that don't actually match up with what's in the database... they appear to just be some stub sub-objects. The `author_ids` has the correct author data in it. We can still use the denormalized data for the `byline` so this PR just adds the `authors` field and loads those (to get the bio + image) in addition to deprecating the old author fields.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -684,7 +684,8 @@ input ARImageInput {
 }
 
 type Article implements Node {
-  author: Author
+  author: Author @deprecated(reason: "Use `byline` or `authors` instead")
+  authors: [Author!]
 
   # The byline for the article. Defaults to "Artsy Editorial" if no authors are present.
   byline: String
@@ -695,6 +696,7 @@ type Article implements Node {
   ): [Article!]!
   channelID: String
   contributingAuthors: [Author]
+    @deprecated(reason: "Use `byline` or `authors` instead")
   description: String
   hero: ArticleHero
   href: String
@@ -2567,6 +2569,7 @@ type Author {
   # A globally unique ID.
   id: ID!
   image: Image
+  initials(length: Int = 3): String
 
   # A type-specific ID.
   internalID: ID!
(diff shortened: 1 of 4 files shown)


_schemaV2.graphql
```diff
@@ -2562,13 +2564,17 @@ type AuthenticationType {
...
+
   # A globally unique ID.
   id: ID!
+  image: Image
+  initials(length: Int = 3): String
 
   # A type-specific ID.
   internalID: ID!
   name: String
-  profileHandle: String
```
dzucconi, line 2571: `profileHandle` doesn't show up anywhere in Positron. It's not used in Force either. `twitterHandle` is an actual field, however.

dzucconi:
Well, I'm apparently wrong about this. I really wish someone was around who understood this but for instance: I make a new article and don't change anything about authorship and this is the list of authors associated with it:



Which is clearly non-sense.

dzucconi:
OK... here's the grim story:

Articles have this admin panel where I can set authorship.



`Primary Author` is... anything? I can just type any arbitrary string here and it saves. Apparently these things have IDs too. I guess this is just some orphaned sub-object on the Article itself.

`Authors` is autocompleted from the list of authors in Positron. These are real objects that I can query through the Authors endpoint.

`Contributing Authors` are just Artsy users, autocompleted from the entire database of Artsy users.

Now the really weird part is that, if I don't set any author I wind up getting that list of users above. I have no idea what that is and how that came to be. But `Staging Delete Me` is a sample author that I just created to see and apparently it got added to this default list... so that's... non-ideal. So basically: the `author_ids` list is effectively useless and we can't actually get at the bio/avatar without fixing this bug?

dzucconi:
Oh and: `author_ids` does get populated as the `Authors` area there; which would be fine if this default bug didn't exist. Back to the drawing board WRT how to fix this mess.

PR #3774, opened 2022-02-15
fix(article): correctly handles authors and bylines

dzucconi:
OK; so after some digging: https://artsy.slack.com/archives/C01ADJNCS5D/p1644957604225169

"Primary Author" = `author` is only used to control the article slug; and is (probably) broken anyway.
"Authors" = `author_ids` are what we really want for attribution.
"Contributing Authors" = `contributing_authors` we should just ignore this. This maybe had a real function at some point but is now currently used to hack around a front-end bug where we weren't correctly attributing articles in some contexts. Seems pretty wild that such an elaborate work around existed for so long but there you go.

Now `byline` also loads the correct authors from `author_ids` and I fixed the empty array bug on `authors`.


src/schema/v2/article/index.ts
```diff
@@ -33,28 +33,36 @@ export const ArticleType = new GraphQLObjectType<any, ResolverContext>({
     ...IDFields,
     cached,
     authors: {
-      type: new GraphQLList(new GraphQLNonNull(AuthorType)),
+      type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(AuthorType))),
       resolve: async ({ author_ids }, _args, { authorsLoader }) => {
+        console.log("HERE", author_ids)
```
dzucconi, line 38: Woops

PR #3776, opened 2022-02-16
feat: add artsy total and tax type to commerce order type (TX-96)

starsirius:
https://artsyproduct.atlassian.net/browse/TX-96

Depends on https://github.com/artsy/exchange/pull/1071. This update the Metaphysics schema with updated `artsyTotalCents` and `taxTypes` fields for commerce orders.

cc @artsy/transact-devs


_schemaV2.graphql
```diff
@@ -9423,6 +9481,12 @@ type Mutation {
     # Parameters for SubmitPendingOffer
     input: CommerceSubmitPendingOfferInput!
   ): CommerceSubmitPendingOfferPayload
+
+  # Update value of impulse_conversation_id on order
+  commerceUpdateImpulseConversationId(
```
starsirius, line 9486: Looks like this is the result of https://github.com/artsy/exchange/pull/1054. @lidimayra do you have context if this mutation will be or is being used? I assume without it in the schema, clients can't actually use it.

starsirius:
Merging and going to give this a try in https://github.com/artsy/volt/pull/5719!

PR #3759, opened 2022-02-11
feat: Add metric preferences to me

olerichter00:
Addresses [CX-2318]

## Description

Add metric preferences to me.

[CX-2318]: https://artsyproduct.atlassian.net/browse/CX-2318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Co-authored-by: dariakoko

PR #3779, opened 2022-02-17
fix: Add user metric preference default values

olerichter00:
Follow-up for https://github.com/artsy/metaphysics/pull/3759

## Description

This PR adds default values to user metric preferences. These default values are just needed because we cannot guarantee that Gravity always sends values due to caching issues.


dariakoko (approved):
Thanks for the changes and explanation Ole! 👍

PR #3787, opened 2022-02-21
Fix [CX-2257]: CMS Collector Profile 

MrSltun:
This PR is a follow up to #3747

This PR updates the CMS Collector Profile fields to remove unused fields and fixes the `emailConfirmed` resolver


olerichter00 (approved):
👏

PR #3789, opened 2022-02-21
feat[FX-3451]: human-readable size labels for SearchCriteria

mdole:
~~WIP: planning to move some of the parsing and formatting of size values to a util package so that they can be shared between Force, Eigen, and Metaphysics, per [this conversation](https://artsy.slack.com/archives/C9SATFLUU/p1645445717561499) on Slack.~~

~~We should also consider if there's any logic for other filter formatting that could be moved to this util! Prices perhaps? Planning on creating that package tomorrow~~

We ended up having more conversations about this and deciding that we want to store in/cm units on SearchCriteria in Gravity, which means we'll need to update this parsing anyway. For now, we'll go with the copy/paste approach to unblock displaying these labels.

Jira ticket: https://artsyproduct.atlassian.net/browse/FX-3451

### DIFF
--- a/src/schema/v2/searchCriteriaLabel.ts
+++ b/src/schema/v2/searchCriteriaLabel.ts
@@ -12,6 +12,8 @@ export const SIZES = {
   LARGE: "Large (over 100cm)",
 }
 
+const ONE_IN_TO_CM = 2.54
+
 type SearchCriteriaLabel = {
   /** The GraphQL field name of the filter facet */
   field: string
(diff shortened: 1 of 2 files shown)


src/schema/v2/searchCriteriaLabel.ts
```diff
@@ -174,11 +183,37 @@ function getSizeLabels(sizes: string[]) {
...
   return sizes.map((size) => ({
     name: "Size",
-    value: toTitleCase(size.toLowerCase()), // TODO: this a placeholder, we need to format these properly
+    value: SIZES[`${size}`],
     field: "sizes",
   }))
 }
 
+const convertToCentimeters = (element: number) => {
+  return Math.round(element * ONE_IN_TO_CM)
```
nickskalkin, line 192: Q: do we use the same rounding logic in other parts of the system?

nickskalkin (approved):
nice 👍

src/schema/v2/searchCriteriaLabel.ts
mdole: Good question - yeah, we do. For example, if you visit [this Force page](https://www.artsy.net/artist/zhao-zhao/works-for-sale?width=3.937007874015748-39.37007874015748) where a custom size filter has been applied, you can see that the URL has un-rounded values, while the label above the grid has rounded them off to the nearest centimeter.

This function is copied basically verbatim from [Force's SizeFilter file](https://github.com/artsy/force/blob/d85e7e8bc7928489c707000a92e66d726450a194/src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx#L58)

PR #3791, opened 2022-02-22
feat: [FX-3452] human-readable values for Price filters

nickskalkin:
Jira - https://artsyproduct.atlassian.net/browse/FX-3452

The task is to make price ranges in SearchCriteriaLabel human-readable

Screenshot 2022-02-22 at 14 39 49


mdole (approved):
🚢 let's do it!

PR #3797, opened 2022-02-23
feat: Artwork recommendations

olerichter00:
Addresses [CX-2371]

## Description

This PR adds the new `artworkRecommendations` connection to `me`. The connection fetches artwork recommendations for the current user from **Vortex** and the artwork details from **Gravity**.

### Query

```graphql
query ArtworkRecommendationsQuery {
  me {
    artworkRecommendations(first: 10) {
      totalCount
      edges {
        node {
          artistId
          score
        }
      }
    }
 }
}
```

[CX-2371]: https://artsyproduct.atlassian.net/browse/CX-2371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/me/artworkRecommendations.ts
+++ b/src/schema/v2/me/artworkRecommendations.ts
@@ -1,0 +1,85 @@
+import { GraphQLFieldConfig, GraphQLInt } from "graphql"
+import { connectionFromArray } from "graphql-relay"
+import gql from "lib/gql"
+import { convertConnectionArgsToGravityArgs, extractNodes } from "lib/helpers"
+import { compact, find } from "lodash"
+import { CursorPageable, pageable } from "relay-cursor-paging"
+import { createPageCursors } from "schema/v1/fields/pagination"
+import { artworkConnection } from "schema/v2/artwork"
+import { ResolverContext } from "types/graphql"
+
+// This limits the maximum number of artworks we receive from Vortex and is related to how we implement the Connection in this resolver.
+const MAX_ARTWORKS = 50
+
+export const ArtworkRecommendations: GraphQLFieldConfig<
+  void,
+  ResolverContext
+> = {
+  description: "A connection of artwork recommendations for the current user.",
+  type: artworkConnection.connectionType,
+  args: pageable({
+    page: { type: GraphQLInt },
+  }),
+  resolve: async (
+    _root,
+    args: CursorPageable,
+    { vortexGraphqlLoader, artworksLoader }
+  ) => {
+    if (!vortexGraphqlLoader || !artworksLoader) return
+
+    const { page, size } = convertConnectionArgsToGravityArgs(args)
+
+    // Fetch artwork IDs from Vortex
+
+    const vortexResult = await vortexGraphqlLoader({
+      query: gql`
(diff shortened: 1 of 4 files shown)


src/schema/v2/me/artworkRecommendations.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (!vortexGraphqlLoader || !artworksLoader) return
+
+    const { page, size } = convertConnectionArgsToGravityArgs(args)
+
+    // Fetch artwork IDs from Vortex
+
+    const vortexResult = await vortexGraphqlLoader({
+      query: gql`
+        query artworkRecommendationsQuery {
+          artworkRecommendations(first: ${MAX_ARTWORKS}) {
```
olerichter00, line 37: Fetching all artwork recommendations because Vortex pagination would need a cursor which we don't have. This works anyway because we have a limited number of recommendations per user.

src/schema/v2/me/artworkRecommendations.ts
```diff
@@ -0,0 +1,82 @@
...
+    // Fetch artwork details from Gravity
+
+    let artworks: any[] = []
+
+    if (artworkIds?.length) {
+      artworks = await artworksLoader({
+        ids: artworkIds,
+      })
+
+      // Apply order from Vortex result (score ASC)
```
leamotta, line 66: what do you think about adding a test for this specific behaviour?

src/schema/v2/me/artworkRecommendations.ts
```diff
@@ -0,0 +1,82 @@
...
+import gql from "lib/gql"
+import { convertConnectionArgsToGravityArgs, extractNodes } from "lib/helpers"
+import { compact, find } from "lodash"
+import { CursorPageable, pageable } from "relay-cursor-paging"
+import { createPageCursors } from "schema/v1/fields/pagination"
+import { artworkConnection } from "schema/v2/artwork"
+import { ResolverContext } from "types/graphql"
+
+// This limits the maximum number of artworks we receive from Vortex and is related to how we implement the Connection in this resolver.
+const MAX_ARTWORKS = 50
```
leamotta, line 12: We are already accepting a `first` parameter, can we use that? or what's the benefit of using the constant?

olerichter00: Vortex doesn't allow a `after` parameter and we would have to work with cursors (see this comment: https://github.com/artsy/metaphysics/pull/3797#discussion_r815860540).

Let me know if you have an idea how to make this work.

src/schema/v2/me/artworkRecommendations.ts
olerichter00: Good catch. The test case implicitly tested the order. I've added it to the description of the test to make it more clear.

src/schema/v2/me/artworkRecommendations.ts
leamotta: I guess it's ok then, we can reevaluate later

leamotta (approved):

PR #3808, opened 2022-02-25
fix(bidder-position): Fix error status handling

damassi:
This fixes some error handling issues we were seeing with the bidder position. I'm not sure how it worked before, unless we changed the structure of our error response in Gravity? In any case this works:

From grav: 
Screen Shot 2022-02-25 at 11 27 09 AM

In MP: 
Screen Shot 2022-02-25 at 12 10 54 PM

Screen Shot 2022-02-25 at 12 11 26 PM

Additionally, added a new field, `rawError`, which returns the error as it originates from gravity. I have no clue why we aren't using this in the UI? It's quite informative (as one would expect). 

cc @artsy/grow-devs

### DIFF
--- a/src/schema/v2/me/__tests__/bidder_position_mutation.test.js
+++ b/src/schema/v2/me/__tests__/bidder_position_mutation.test.js
@@ -17,14 +17,11 @@ const query = `
         }
         messageHeader
         messageDescriptionMD
+        rawError
       }
     }
   }
   `
-
-const errorMessageTemplate = `{ Error: https://stagingapi.artsy.net/api/v1/me/bidder_position?
-        artwork_id=mohammed-qasim-ashfaq-surrealism&max_bid_amount_cents=3700000&
-        sale_id=bidding-test - `
 
 describe("Bidder position mutation", () => {
   const createBidderPosition = {
@@ -50,10 +47,12 @@ describe("Bidder position mutation", () => {
 
   describe("error", () => {
     it("creates correct message when bid is not high enough", async () => {
-      const errorObjectString = `{"type": "param_error", "message":"Please enter a bid higher than $37,000.","detail"\
-:{"base":["Please enter a bid higher than $37,000"]}}`
       const errorMessage = {
-        message: errorMessageTemplate + errorObjectString,
+        body: {
+          type: "param_error",
+          message: "Please enter a bid higher than $37,000.",
+          detail: { base: ["Please enter a bid higher than $37,000"] },
+        },
       }
       const context = {
         createBidderPositionLoader: sinon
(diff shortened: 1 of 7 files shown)


src/schema/v2/me/__tests__/bidder_position_mutation.test.js
```diff
@@ -65,17 +64,21 @@ describe("Bidder position mutation", () => {
 
       expect(data.createBidderPosition.result.position).toBeNull()
       expect(data.createBidderPosition.result.messageHeader).toEqual(
-        "Your bid wasn’t high enough"
+        "Your bid wasn't high enough"
```
dzucconi, line 67: Should leave the correct apostrophe in

dzucconi: Oh, these are returned from Gravity

src/schema/v2/me/bidder_position_messages.ts
```diff
@@ -1,59 +1,59 @@
...
   {
-    id: "SALE_CLOSED",
-    gravity_key: "Sale Closed to Bids",
+    status: "SALE_CLOSED",
+    message: "Sale Closed to Bids",
     header: "Lot closed",
-    description_md: () =>
-      "Sorry, your bid wasn’t received\nbefore the lot closed.",
+    getDescription: () =>
+      "Sorry, your bid wasn't received before the lot closed.",
```
dzucconi, line 36: Oh — I guess not. Yeah; just a quick run through to use correct punctuation would be nice

dzucconi (approved):
I think this is fine but generally characterizing these statuses as "errors" doesn't seem quite accurate?

src/schema/v2/types/bidder_position_result.ts
```diff
@@ -13,14 +13,15 @@ export const BidderPositionResultType = new GraphQLObjectType<
...
     messageDescriptionMD: {
       type: GraphQLString,
-      resolve: ({ message_description_md }) => message_description_md,
     },
     position: {
       type: BidderPosition.type,
     },
+    rawError: {
+      type: GraphQLString,
+    },
```
dzucconi, line 25: Like; if these were errors I'd expect one of the following: this information is in the errors key not the normal returned result. Or we do the `FooOrError` type thing.

damassi:
> think this is fine but generally characterizing these statuses as "errors" doesn't seem quite accurate?

Right, yeah this is mostly just a bugfix PR. Why the decisions were made that were made is another story, but we already have Eigen relying on it, so not going to change things up too much atm. I think this cleanup bit is a good starting point for doing it right in the future, however.

PR #3815, opened 2022-03-01
feat: Comparable auction results for artworks

olerichter00:
Addresses [CX-2288]

## Description

Comparable auction results for artworks.


MounirDhahri (approved):

src/schema/v2/artwork/__tests__/comparableAuctionResults.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+describe("ComparableAuctionResults", () => {
+  let context = {}
+
+  beforeEach(() => {
+    context = {
+      artworkLoader: sinon
+        .stub()
+        .withArgs(mockArtwork.id)
+        .returns(Promise.resolve(mockArtwork)),
```
damassi, line 12: (non-blocking heads up!) It's hard to avoid older copy pasta stuff in MP, but we shouldn't need to use sinon. [Jest idioms + simple functions work just fine and are simpler](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/me/__tests__/artistRecommendations.test.ts#L23-L35).

PR #3824, opened 2022-03-03
feat(alerts): add support for new colors

gkartalis:
# Description 

Follow-up on https://github.com/artsy/metaphysics/pull/3762

- Adds new colors and keeps old ones as `OLD_COLORS`

more context on this [slack thread](https://artsy.slack.com/archives/C9SATFLUU/p1646316037871739)


anandaroop (approved):
Nice catch!

If you don't mind I'll push up a small commit to update specs to use the new color names only, for the benefit of future-us.

PR #3843, opened 2022-03-09
chore: Remove express-force-ssl

icirellik:
Removes the no longer required package `express-force-ssl`


joeyAghion (approved):

dblandin (approved):
Thanks for this follow-up! 💯

PR #3844, opened 2022-03-09
feat: Expose artist affinity based artwork recs GRO-792

jonallured:
This PR is a lot! It does a few things:

* refactors the stitching code that mutates the Vortex schema
* copies the latest Vortex schema into MP
* erases much of the updates via that stitching code
* adds in the artist affinity based artwork recs query
* updates the link to Vortex so that it'll pass through an app token

My goal is to make a query like this:

```graphql
query {
  analyticsArtworkRecommendations(userId:"BLAH") {
    edges {
      node {
        artworkId
        score
      }
    }
  }
}
```

With a header like this:

```
x-xapp-token: BLAH
```

And have that sent to Vortex like so:

```graphql
query {
  artworkRecommendations(userId:"BLAH") {
    edges {
      node {
        artworkId
        score
      }
    }
  }
}
```

With a header like this:

```
Authorization: Bearer: BLAH
```

So this is my solution to this puzzle - what do you think? Is there another, better way to do this? Open to feedback for sure!!

https://artsyproduct.atlassian.net/browse/GRO-792

/cc @artsy/grow-devs


src/lib/stitching/vortex/link.ts
```diff
@@ -35,10 +34,14 @@ export const createVortexLink = () => {
...
       }
+
+      const token = graphqlContext.appToken || VORTEX_TOKEN
+      const bearer = `Bearer ${token}`
+
       return {
         headers: {
           ...headers,
-          Authorization: `Bearer ${VORTEX_TOKEN}`,
+          Authorization: bearer,
```
damassi, line 44: This approach looks good! the graphql context is exactly how one passes this kind of data around.

src/lib/stitching/vortex/schema.ts
```diff
@@ -20,41 +20,45 @@ export const executableVortexSchema = ({
...
+    }
+  })
+
+  const transforms = [
+    ...(removeRootFields ? [filterTransform] : []),
+    renameTypesTransform,
+    renameRootFieldTransform,
+  ]
+
+  return transformSchema(schema, transforms)
```
damassi, line 63: Appreciate the refactor! but I would suggest keeping this as it was:
- the left is the pattern we use all through our stitched services and the code all looks roughly the same everywhere, and also matches the docs
- from a more personal pref perspective the left is fundamentally simpler and more straightforward. One can read straight down and it says quite literally exactly whats going on in a linear fashion that requires very little thought. The right now has lots of new variables to parse, regexes, stuff being defined and then later executed out of order, requiring visual jumping back and forth to grok and so on. 

If this bit had more inherent complexity I'd say 👍 but for what it is, and for the sake of MP-wide consistency we should just focus on the auth token aspects for now. 

Of course, this is non-blocking! Just my 2c at a glance.

damassi (approved):
This looks good! One suggestion to not refactor the rename types stuff, but not gonna block it you disagree 😄

olerichter00 (approved):

src/lib/stitching/vortex/schema.ts
jonallured: I pondered this overnight and decided to take your advice - did it like so:

```
$ git checkout main -- src/lib/stitching/vortex/schema.ts
```

Then I adjusted the `removeRootFieldList` to include `artistAffinities` and `artistRecommendations` to match the change introduced in my refactor.

I'd like to rebase this PR then so that git history will include the refactoring if we ever change our minds and want it back. Thanks @damassi!!

PR #3854, opened 2022-03-11
fix(article): early null on media

dzucconi:
🙄


src/schema/v2/article/index.ts
```diff
@@ -158,6 +158,14 @@ export const ArticleType = new GraphQLObjectType<any, ResolverContext>({
           url: { type: GraphQLString },
         },
       }),
+      resolve: ({ media }) => {
+        if (!media) return null
+
+        // Positron returns an object with null properties rather than an early null
```
dzucconi, line 164: I guess this is pervasive. _So_ annoying.

PR #3855, opened 2022-03-11
Exposes new field for artsy international shipping

jpotts244:
Part of https://artsyproduct.atlassian.net/browse/PX-4898

Exposes new boolean field on artwork model to denote international arta shipping
https://github.com/artsy/gravity/pull/15110/files

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -704,6 +704,11 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLBoolean,
         resolve: ({ arta_enabled }) => arta_enabled,
       },
+      artsyShippingInternational: {
+        type: GraphQLBoolean,
+        resolve: ({ artsy_shipping_international }) =>
+          artsy_shipping_international,
+      },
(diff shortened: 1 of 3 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -704,6 +704,11 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLBoolean,
         resolve: ({ arta_enabled }) => arta_enabled,
       },
+      artsyShippingInternational: {
```
leamotta, line 707: I'm wondering if we also need to change the [shipping info string ](https://github.com/artsy/metaphysics/pull/3855/files#diff-54b84e0c164089fc202e3439d8ded89ab4925442fb613f4e36a9c22acdb66cb8L760)
We could go with "Calculated in checkout" when it's either `process_with_arta_shipping` or this, what do you think?

jpotts244: Ooo good catch, didnt realize that was happening in MP. Will adjust

leamotta (approved):
Nice!

PR #3857, opened 2022-03-14
fix: Remove comparable works date param when empty string

olerichter00:
## Description

Remove comparable works date param when empty string.


MrSltun (approved):
LGTM! 🌟

MounirDhahri:
Thanks for making this fix, I would recommend adding a test to avoid this regression in the future.

PR #3871, opened 2022-03-16
chore(price filter): support new histogram aggregation

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-3801

Follow-up to https://github.com/artsy/gravity/pull/15118

Simply exposes the `simple_price_histogram` aggregation that's now available in Gravity's `api/v1/filter/artworks` endpoint.

This new aggregation is currently only supported only on _artist_ artwork grids, a constraint that is not modeled here but enforced by Gravity instead.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1936,6 +1936,7 @@ enum ArtworkAggregation {
   PARTNER_CITY
   PERIOD
   PRICE_RANGE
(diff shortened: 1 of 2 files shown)


anandaroop:
Pretty trivial PR, I'll go ahead and merge it so that we can't start to wire up a review app, cc @dimatretyak.

PR #3880, opened 2022-03-18
feat: add the ability to return value and displayValue fields in SearchCriteriaLabel

dimatretyak:
### Description
This PR adds the ability to query the initial value of search criteria, not just the human-readable label. This changes will allow us to replace the current implementation of label extraction for pills on Force/Eigen, where aggregations are now needed.

### Response before
```
[
  {
    "field": "attributionClass",
    "name": "Rarity",
    "value": "Unique"
  },
  {
    "field": "additionalGeneIDs",
    "name": "Medium",
    "value": "Prints"
  },
  {
    "field": "sizes",
    "name": "Size",
    "value": "Small (under 40cm)"
  },
  {
    "field": "acquireable",
    "name": "Ways to Buy",
    "value": "Buy Now"
  },
  {
    "field": "partnerIDs",
    "name": "Galleries and Institutions",
    "value": "3 White Dots"
  }
]
```

### Response now
```
[
  {
    "field": "attributionClass",
    "name": "Rarity",
    "value": "unique", // search criteria value
    "displayValue": "Unique" // human-readable label (it used to be a `value` field)
  },
  {
    "field": "additionalGeneIDs",
    "name": "Medium",
    "value": "prints",
    "displayValue": "Prints"
  },
  {
    "field": "sizes",
    "name": "Size",
    "value": "SMALL",
    "displayValue": "Small (under 40cm)"
  },
  {
    "field": "acquireable",
    "name": "Ways to Buy",
    "value": "true",
    "displayValue": "Buy Now"
  },
  {
    "field": "partnerIDs",
    "name": "Galleries and Institutions",
    "value": "3-white-dots",
    "displayValue": "3 White Dots"
  }
]
```

### DIFF
--- a/src/schema/v2/searchCriteriaLabel.ts
+++ b/src/schema/v2/searchCriteriaLabel.ts
@@ -21,11 +21,11 @@ type SearchCriteriaLabel = {
   /** The human-friendly name of the filter facet */
   name: string
 
-  /** The human-friendly label of the filter facet */
-  label: string
-
   /** The human-friendly value of the filter facet */
   value: string
+
+  /** The human-friendly label of the filter facet */
+  displayValue: string
 }
 
 /**
@@ -49,13 +49,13 @@ export const SearchCriteriaLabel = new GraphQLObjectType<
       type: GraphQLNonNull(GraphQLString),
       description: "The human-friendly name of the filter facet",
     },
-    label: {
-      type: GraphQLNonNull(GraphQLString),
-      description: "The human-friendly label of the filter facet",
-    },
     value: {
       type: GraphQLNonNull(GraphQLString),
       description: "The human-friendly value of the filter facet",
+    },
+    displayValue: {
+      type: GraphQLNonNull(GraphQLString),
+      description: "The human-friendly label of the filter facet",
     },
   },
 })
@@ -124,7 +124,7 @@ async function getArtistLabels(artistIDs: string[], artistLoader) {
       const artist = await artistLoader(id)
       return {
         name: "Artist",
-        label: artist.name,
+        displayValue: artist.name,
         value: id,
         field: "artistIDs",
       }
@@ -137,7 +137,7 @@ function getRarityLabels(attributionClasses: string[]) {
 
   return attributionClasses.map((attributionClass) => ({
     name: "Rarity",
-    label: allAttributionClasses[attributionClass].name,
+    displayValue: allAttributionClasses[attributionClass].name,
     value: attributionClass,
     field: "attributionClass",
   }))
@@ -171,7 +171,7 @@ function getMediumLabels(additionalGeneIDs: string[]) {
 
   return additionalGeneIDs.map((geneID) => ({
     name: "Medium",
-    label: MEDIUM_GENES[geneID],
+    displayValue: MEDIUM_GENES[geneID],
     value: geneID,
     field: "additionalGeneIDs",
   }))
@@ -196,7 +196,7 @@ function getPriceLabel(priceRange: string): SearchCriteriaLabel | undefined {
 
   return {
     name: "Price",
-    label,
+    displayValue: label,
     value: priceRange,
     field: "priceRange",
   }
@@ -207,7 +207,7 @@ function getSizeLabels(sizes: string[]) {
 
   return sizes.map((size) => ({
     name: "Size",
-    label: SIZES[`${size}`],
+    displayValue: SIZES[`${size}`],
     value: size,
     field: "sizes",
   }))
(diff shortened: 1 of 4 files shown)


src/schema/v2/searchCriteriaLabel.ts
```diff
@@ -267,28 +281,32 @@ function getWaysToBuyLabels(waysToBuy: {
...
+      value: "acquireable",
       field: "acquireable",
     })
 
   if (atAuction)
     labels.push({
       name: "Ways to Buy",
-      value: "Bid",
+      label: "Bid",
+      value: "atAuction",
```
dimatretyak, line 293: **question**: is it normal that a string is returned as `value`, and not a boolean value?

dblandin: For boolean filters, a boolean might make more sense.

src/schema/v2/searchCriteriaLabel.ts
```diff
@@ -117,7 +124,8 @@ async function getArtistLabels(artistIDs: string[], artistLoader) {
       const artist = await artistLoader(id)
       return {
         name: "Artist",
-        value: artist.name,
```
dblandin, line 120: **question**: This might be a disruptive change as we already have frontend clients (including mobile apps) using this `value` attribute for the pill label. Would it be less disruptive to add a new `rawValue` attribute and leave the existing `value` attribute as is?

I like the naming in this PR but we might be able to jump to it right away.

dimatretyak: Eigen doesn't use these changes. We will need to make changes only for Force. And I wanted to consult with you about this later and clarify how to do better for Force:
* temporarily comment out [changes](https://github.com/artsy/force/pull/9526) –> push commented changes -> wait for staging/production build - make the necessary changes according to the updated scheme and push it
* make new changes according to the updated scheme and push it, but in such a situation, when opening on `user/alerts`, there will be a 500 error (temporarily, until new changes are merged)
* use a different naming `rawValue` (as suggested by you)

dblandin: > Eigen doesn't use these changes

Ah, I didn't realize we were only using these labels on Force. That simplifies things a bit.

Although, looking at the full query, would it be odd to have label as an attribute on the SearchCriteriaLabel type? labels  > label? Maybe `rawValue` is a bit more intention revealing?

cc/ @anandaroop if you have thoughts on the naming

```graphql
{
  me {
    savedSearchesConnection(first: 5) {
      edges {
	node {
	  labels {
	    name
	    field
	    value
	  }
	}
      }
    }
  }
}

```

anandaroop: Interesting, I agree labels > label fields a little weird.

To summarize, for my own benefit...

```gql
# 1) currently live

labels {
  name      # sizes
  field     # Size
  value     # Small (under 40cm)"
  <missing> # SMALL
 }

# 2) this PR

labels {
  name   # sizes
  field  # Size
  value  # SMALL
  label  # Small (under 40cm)"
 }

# 3) devon suggestion

labels {
  name     # sizes
  field    # Size
  rawValue # SMALL
  value    # Small (under 40cm)"
 }

# 4) another option?

labels {
  name         # sizes
  field        # Size
  value        # SMALL
  displayValue # Small (under 40cm)"
 }
```

I'd be fine with 3 or 4 — and if backwards compatibility is a concern, that seems to point to 3.

Works for me!

dblandin: I know we use `displayValue` elsewhere so that's a good suggestion.

@dimatretyak Up to you! We can go with #4 if it's not too much trouble to temporary disable the feature in Force. Otherwise, #3 would also work but I don't love the `rawValue` attribute name.

dimatretyak: I agree with 4 👍

dimatretyak: @dblandin @anandaroop thanks for the discussion and help!

src/schema/v2/searchCriteriaLabel.ts
dimatretyak: Hmm 🤔  Can we specify several GraphQL types, which should be returned in `type` [here](https://github.com/artsy/metaphysics/blob/4e59d5bfa35bdb5052216f4094540269908622d2/src/schema/v2/searchCriteriaLabel.ts#L53) (in our case it should be `GraphQLString` and `GraphQLBoolean`)?

dimatretyak: > I wonder if a union type might be useful here?

I tried using `GraphQLUnionType` and the result is the following query structure. IMHO such a request will not be particularly convenient to use and it is better to use the solution you suggested (when true is returned in string format)

```
labels {
  displayValue
  field
  value {
    ... on BooleanValue {
      stringValue: value
    }
    
    ... on StringValue {
      booleanValue: value
    }
  }
}
```

anandaroop (approved):
Nice! And thanks for thinking through the sequencing with regards to Force.

PR #3896, opened 2022-03-24
feat[FX-3805]: enable switching to gravity for marketing collections

mdole:
This PR enables us to toggle between Gravity and KAWS as the source of MarketingCollections.

To actually make this change in staging/prod, we would run the following:
```
hokusai staging env set ENABLE_GRAVITY_MARKETING_COLLECTIONS=true && hokusai staging refresh
```
But we should wait to test it out until we've at least implemented more of the MarketingCollection fields (and we should definitely test locally first 😀 )

Example of a query to local MP that hits Gravity staging for a collection:
Screen Shot 2022-03-24 at 12 00 36 PM

Jira ticket: https://artsyproduct.atlassian.net/browse/FX-3805


Mitch-Henson (approved):
Looking good!

Should we link this PR or ticket to the other tickets that are implementing the gravity changes?

Also, should we also add in an environment check that @nickskalkin recommended to make it even more defensive? I forget how that conversation resolved.

src/config.ts
```diff
@@ -157,6 +158,8 @@ export default {
   ENABLE_ASYNC_STACK_TRACES,
   ENABLE_METRICS,
   ENABLE_GEOLOCATION_LOGGING: ENABLE_GEOLOCATION_LOGGING === "true",
+  ENABLE_GRAVITY_MARKETING_COLLECTIONS:
+    ENABLE_GRAVITY_MARKETING_COLLECTIONS === "true",
```
nickskalkin, line 162: does it makes sense to make it false by default instead?

nickskalkin (approved):
👍

src/config.ts
mdole: ah so this actually isn't a default, but an explicit boolean check: if the value of the env variable `ENABLE_GRAVITY_MARKETING_COLLECTIONS` is anything other than `true`, then the output from our config would be `false`.

If we removed the `ENABLE_GRAVITY_MARKETING_COLLECTIONS === "true",` we might run into something like `ENABLE_GRAVITY_MARKETING_COLLECTIONS=asdkfjhkjaldh`, which JavaScript would regard as truthy. Unlikely to happen, but seemed better to make it explicit :)

mdole:
Thanks for the feedback, both! @Mitch-Henson re your comments:

> Should we link this PR or ticket to the other tickets that are implementing the gravity changes?

Good thought! I actually think it's better to just let it be linked to its own Jira ticket (which is in turn linked to the epic). 

When I have a PR that is explicitly related to another PR, I'll link them so that it's easy to see the connection, but this is a somewhat weird case where this is related to several in-flight PRs and several PRs/tickets that don't exist yet. I feel like it's better for future-us to navigate back to the Jira epic if we need to review the other work that was completed here. Does that sound reasonable to you?

> Also, should we also add in an environment check that @nickskalkin recommended to make it even more defensive? I forget how that conversation resolved.

Oh yeah! I spaced that. Thinking on it more, I feel like we actually shouldn't introduce that. I think part of the point in using an env flag is that we can toggle to release any time we want - which stops being the case if we also have a hardcoded environment check. We'd have to also PR a change in that case.

I'm also just not that worried about this env var accidentally getting set in production 🤷 I'm happy to bear the blame if something bad comes about as a result!

src/config.ts
nickskalkin: makes sense, thank you for explanation! @mdole

anandaroop (approved):
Nice, this looks more straightforward than I feared!

> But we should wait to test it out until we've at least implemented more of the MarketingCollection fields (and we should definitely test locally first 😀 )

Not sure how much precedent we have for spinning up Metaphysics review apps, but that could also be a good way to really vet this toggle.

mdole:
> Not sure how much precedent we have for spinning up Metaphysics review apps, but that could also be a good way to really vet this toggle.

Ooo very interesting idea! I'm down to explore that when we think we might be ready for QA

PR #3903, opened 2022-03-25
chore(GRO-844): Remove input values that aren't used by Gravity

jacobherrington:
The `channel` and `id` fields aren't used by [Gravity](https://github.com/artsy/gravity/blob/main/app/api/v1/notification_preferences_endpoint.rb), so we can remove them and simplify the component using this data!


jonallured (approved):
Love it thanks!!

PR #3638, opened 2022-01-03
fix: Medium as optional on My Collection form

rajsam003:
This PR resolves [CX-2232](https://artsyproduct.atlassian.net/browse/CX-2232) changes the medium field as optional in add/update artwork form on my collection.

PR #3905, opened 2022-03-28
fix: Make title unrequired again in myCollectionUpdateArtworkMutation

olerichter00:
## Description

`title` doesn't need to be required when updating an artwork. It's just required when creating an artwork to prevent artworks without a title in My Collection. This change is needed to update the submission_id when a submission gets created for a My Collection artwork.

Context can be found here: https://github.com/artsy/metaphysics/pull/3638

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10054,7 +10054,7 @@ input MyCollectionUpdateArtworkInput {
   pricePaidCurrency: String
   provenance: String
   submissionId: String
-  title: String!
+  title: String
   width: String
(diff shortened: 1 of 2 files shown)


olerichter00:
> @olerichter00 I dont understand why `title` doesn't need to be required while updating. Is it ok if the user removes the title while updating details?

This is to be able to update other fields without updating `title` as well. If `title` is not provided, it just won't be updated at all.

rajsam003 (approved):
👍🏽

PR #3917, opened 2022-03-29
chore: add submission id to artwork interface

MounirDhahri:
- add submission id to artwork interface

Screenshot 2022-03-29 at 13 02 38


dariakoko (approved):
🚀  thanks a lot!

PR #3927, opened 2022-03-31
feat: Add data transfer bool for user

jonallured:
Small one - turns out there was only one missing field for user data. I wrongly assumed some other fields were on user but they are actually on the SaleProfile model in Gravity. That means all I had to do here was add this flag to the query and mutation in order for Forque to be able to edit it.

While I was working through this I realized that I wrongly typed the phone number is NOT optional - it can def be `nil` in Gravity.


damassi:
@jonallured - for these gravity field pass through PRs feel free to self-merge and just mark a reviewer!

PR #3928, opened 2022-03-31
feat(users): make users connection pageable

damassi:
This updates our `users` connection to be properly pageable: 

Screen Shot 2022-03-31 at 11 17 43 AM


src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -387,7 +387,7 @@ export default (accessToken, userID, opts) => {
       {},
       { method: "PUT" }
     ),
-    usersLoader: gravityLoader("users"),
+    usersLoader: gravityLoader("users", {}, { headers: true }),
```
dzucconi, line 390: Annoying but: did you double check in the v1 schema that nothing uses this loader? It won't show up on project-wide find.

damassi: Yup, I did

PR #3929, opened 2022-04-01
fix: parse error for sizes in `previewSavedSearch`

dimatretyak:
Type: **Fix**

### Description
[ArtworkSizes](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/artwork/artworkSizes.ts#L3-L16) values are stored in lowercase and for this reason we received an error when [parsing values](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/searchCriteriaLabel.ts#L210) in [getSizeLabels](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/searchCriteriaLabel.ts#L205-L214)

### Before
error

### After
fixed-error

### DIFF
--- a/src/schema/v2/__tests__/previewSavedSearch.test.ts
+++ b/src/schema/v2/__tests__/previewSavedSearch.test.ts
@@ -27,4 +27,41 @@ describe("previewSavedSearch", () => {
       },
     ])
   })
+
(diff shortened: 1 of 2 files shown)


dblandin (approved):
LGTM 👍🏼

PR #3937, opened 2022-04-06
feat(alerts): expose a medium type's corresponding filter gene

anandaroop:
_Leaving this as a DRAFT until we can answer the question below_

https://artsyproduct.atlassian.net/browse/FX-3820

This allows artworks' medium types to be associated with their corresponding genes — which is what we actually use for filtering on the frontend in our artwork grids.

## Context

- Every artwork has a medium type field, which is exposed in Metaphysics as `{ artwork { mediumType { name … } } }`

- This ultimately derives from `Artwork#category` in the corresponding Gravity record; this value is originally supplied by the partner in CMS

- But that medium type is not sufficient for our desired Alerts functionality, because alerts are based on our artworks grids' filter facets — and the medium facet actually filters on the artworks' _genes_, through [a bit of indirection that is described more fully in this Notion document](https://www.notion.so/artsy/Medium-types-vs-medium-filters-39310c0278d64675b755e6db46fec363).

- We therefore expose that corresponding `filterGene` now, which can be accessed from the root…

  ```graphql
  {
    artworkMediums {
      name
      filterGene { # NEW
        slug
        name
      }
    }
  }
  ```

  … Or from an artwork…

  ```graphql
  {
    artwork(id: "pablo-picasso-guernica") {
      slug
      mediumType {
        name
        filterGene { # NEW
          slug
          name
        }
      }
    }
  }
  ```

It's that last query that will be useful in the Artwork Alerts context.

## Questions

- As of now, three of these associated Genes — **Ephemera or Merchandise**; **Reproduction**; **NFT** — are actually _unpublished_, which means that they don't return a proper name, slug etc. Can we publish these, and who's the DRI who could tell us so?

  Screen Shot 2022-04-06 at 6 23 20 PM

### DIFF
--- a/src/lib/artworkMediums.ts
+++ b/src/lib/artworkMediums.ts
@@ -4,12 +4,14 @@ export default {
     long_description:
       "Includes architectural models; buildings (e.g., house, temple).",
     name: "Architecture",
+    mediumFilterGeneSlug: "architecture-1",
   },
   "Books and Portfolios": {
     id: "Books and Portfolios",
     long_description:
       "Includes albums; artist books; illustrated manuscripts; portfolios (i.e., a group of works such as photographs or prints).",
     name: "Books and Portfolios",
+    mediumFilterGeneSlug: "books-and-portfolios",
   },
   "Design/Decorative Art": {
     id: "Design/Decorative Art",
@@ -18,6 +20,7 @@ export default {
       "Includes containers (e.g., basket, bowl, box, vase); furniture (e.g., cabinet, chair, table); lighting; mirrors; tableware (e.g., flatware, plates).",
     ].join(" "),
     name: "Design/Decorative Art",
+    mediumFilterGeneSlug: "design",
   },
   "Drawing, Collage or other Work on Paper": {
     id: "Drawing, Collage or other Work on Paper",
@@ -26,6 +29,7 @@ export default {
       "This also includes paintings where paper is the support.",
     ].join(" "),
     name: "Drawing, Collage or other Work on Paper",
+    mediumFilterGeneSlug: "work-on-paper",
   },
   "Ephemera or Merchandise": {
     id: "Ephemera or Merchandise",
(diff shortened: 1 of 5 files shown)


src/lib/artworkMediums.ts
```diff
@@ -4,12 +4,14 @@ export default {
     long_description:
       "Includes architectural models; buildings (e.g., house, temple).",
     name: "Architecture",
+    mediumFilterGeneSlug: "architecture-1",
```
anandaroop, line 7: Strange but true. There are two genes that were called "Architecture" at one point or another: one for artistic discipline, and one for subject matter. We're interested in the first one, but it was created later and ended up with this janky slug.

src/schema/v2/__tests__/artworkMediums.test.ts
```diff
@@ -19,4 +20,37 @@ describe("ArtworkMediums type", () => {
...
+
+    const result = await runQuery(query, context)
+
+    expect(context.geneLoader).toHaveBeenCalledTimes(19)
+
+    expect(result.artworkMediums[0].filterGene).toEqual({
+      slug: "architecture-1",
+      name: "Architecture",
+    })
+  })
```
dimatretyak, line 55: 👏 👏 👏

anandaroop:
Based on the discussion at https://artsy.slack.com/archives/C03T8CNFQ/p1649346613862799?thread_ts=1644356509.834049&cid=C03T8CNFQ …

… It seems we can publish the three genes in question. So this PR now requires a migration:

```rb
Gene.find("ephemera-or-merchandise").update(published: true)
Gene.find("nft").update(published: true)
Gene.find("reproduction").update(published: true)
```

dimatretyak (approved):

anandaroop:
I've published the three genes in prod now as well, going to go ahead and merge this.

anandaroop:
Confirmed that with those genes being published we get valid data for them.

(Note that the "Other" medium type still returns a null gene, as expected

Screen Shot 2022-04-08 at 12 36 30 PM
)

PR #3882, opened 2022-03-18
feat [CX-2427]: Adding artist insights into my collection artworks

MrSltun:
Resolves [CX-2427]

This is a follow up on artsy/vortex#426 to expose the new vortex endpoint

## Description
This PR introduces Market Price Insights field for My Collection Artworks and returns the data in a `marketPriceInsights` field for each artwork inside my collection if the data is available

## Example
An example query would be like this:
```graphql
query {
  me {
    myCollectionConnection(first: X) {
      edges {
        node {
          ...
          marketPriceInsights {
            demandRank
          }
          ...
        }
      }
    }
  }
}
```

and if the data is available, the query results should be similar something like this:
```json
{
  "data": {
    "me": {
      "myCollectionConnection": {
        "edges": [
          {
            "node": {
              ...
              "marketPriceInsights": {
                "demandRank": 0.914
              },
              ...
            }
          },
          ...
        ]
      }
    }
  }
}
```

In case the data is not available, the `marketPriceInsights` field would be `null`

[CX-2427]: https://artsyproduct.atlassian.net/browse/CX-2427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3946, opened 2022-04-08
Fix: Update price insights type

MrSltun:
This is a follow up on #3882

## Description
This PR changes the type of the Market Price Insights inside My Collection Artwork type, since the current one uses the `MarketPriceInsights` from Vortex, which has the ID, a non-nullable field, that is returned as `null` in Metaphysics

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1840,7 +1840,7 @@ type Artwork implements Node & Searchable & Sellable {
   # Represents partner's location
   location: Location
   manufacturer(format: Format): String
-  marketPriceInsights: MarketPriceInsights
+  marketPriceInsights: PriceInsights
 
   # Represents the **materials** used in this work, such as _oil and acrylic on
   # canvas_. (This should not be confused with the artwork attribute called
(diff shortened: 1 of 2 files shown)


_schemaV2.graphql
```diff
@@ -10961,6 +10961,12 @@ type PriceInsightEdge {
   node: MarketPriceInsights
 }
 
+type PriceInsights {
```
olerichter00, line 10964: **(suggestion):** What do you think about naming it `ArtworkPriceInsights` because it's used for a sub-field of `Artwork`?

olerichter00 (commented):
👏

_schemaV2.graphql
```diff
@@ -2120,6 +2120,12 @@ union ArtworkMutationType =
 
 union ArtworkOrEditionSetType = Artwork | EditionSet
 
+type ArtworkPriceInsights {
+  artistId: String
```
olerichter00, line 2124: **(question):** Why do we return the artist ID?

olerichter00 (approved):
Just added a question about the artist ID in the type. Feel free to merge.

_schemaV2.graphql
MrSltun: Good question actually, I only added it because it is used in the query, would it be okay to remove it?
because I don't think we need this information inside the ArtworkPriceInsights

PR #3952, opened 2022-04-11
feat: Admin can manage sale profile data for a user

jonallured:
This PR is setting up more fields for user management in Forque. It adds Gravity resolvers around the `SaleProfile` records in Mongo and then uses them in a nested field in the `user` query and a top-level `updateUserSaleProfile` mutation. I _think_ everything is pretty standard here but would def love any feedback!


src/schema/v2/user.ts
```diff
@@ -11,6 +11,61 @@ import { LocationType } from "schema/v2/location"
...
+      )
+    }
+
+    return userSaleProfileLoader(sale_profile_id)
+      .then((result) => {
+        return result
+      })
+      .catch((err) => {
+        if (err.statusCode === 404) {
+          return false
```
dzucconi, line 63: How does this error out in practice? Would be helpful to throw a descriptive error.

src/schema/v2/user.ts
```diff
@@ -11,6 +11,61 @@ import { LocationType } from "schema/v2/location"
...
+    if (!userSaleProfileLoader) {
+      throw new Error(
+        "You need to be signed in as an admin to perform this action"
+      )
+    }
+
+    return userSaleProfileLoader(sale_profile_id)
+      .then((result) => {
+        return result
+      })
```
dzucconi, line 60: This then block isn't needed

src/schema/v2/user.ts
dzucconi: Oh this is just the fetch. I think you should just return null here?

src/schema/v2/user.ts
```diff
@@ -29,6 +84,12 @@ export const UserType = new GraphQLObjectType<any, ResolverContext>({
       description: "The given phone number of the user.",
       type: GraphQLString,
     },
+    saleProfileId: {
+      description: "The id of the user's sale profile.",
+      type: GraphQLString,
+      resolve: ({ sale_profile_id }) => sale_profile_id,
+    },
```
dzucconi, line 91: This field doesn't need to be exposed

jonallured: I think it does because the Forque code needs it in order to send the mutation!

src/schema/v2/users/updateUserSaleProfileMutation.ts
```diff
@@ -0,0 +1,79 @@
...
+      .filter((key) => key !== "id")
+      .reduce((acc, key) => {
+        if (key === "addressLine1") {
+          return { ...acc, address_1: args[key] }
+        } else if (key === "addressLine2") {
+          return { ...acc, address_2: args[key] }
+        } else {
+          return { ...acc, [snakeCase(key)]: args[key] }
+        }
+      }, initialValue)
```
dzucconi, line 66: `non-blocking` This looks rather complicated for what could just be manually written out

```tsx
const gravityOptions = {
  address_1: args.addressLine1,
  // ...etc
}
```

dzucconi (approved):
Looks good to me. I'd remove the `saleProfileId` however

src/schema/v2/user.ts
dzucconi: You'd just access it through the `saleProfle`

src/schema/v2/users/updateUserSaleProfileMutation.ts
damassi: +1. dumb, redundant and literal wins the game in graphql

src/schema/v2/users/updateUserSaleProfileMutation.ts
```diff
@@ -0,0 +1,79 @@
...
+    try {
+      return await updateUserSaleProfileLoader?.(args.id, gravityOptions)
+    } catch (err) {
+      if ("body" in (err as any)) {
+        const e = err as GravityError
+        throw new Error(e.body.text ?? e.body.error ?? e.body.message)
+      }
+
+      throw err
+    }
```
damassi, line 77: @jonallured / @dzucconi - since this error handling approach is now being copied all around, can we put it inside a reusable helper (and where would that helper be located)?

src/schema/v2/user.ts
jonallured: I see now - removed with [85cf29f](https://github.com/artsy/metaphysics/pull/3952/commits/85cf29fc45edf19c076a38ad0199cd057b5c5441) thanks for the nudge!!

src/schema/v2/users/updateUserSaleProfileMutation.ts
jonallured: Dumb coming right up! [bda9fe5](https://github.com/artsy/metaphysics/pull/3952/commits/bda9fe5408afbf2ae24299f15f570e1e070dac1a)

src/schema/v2/users/updateUserSaleProfileMutation.ts
jonallured: You might be onto something @damassi 😝 

```
jon@juggernaut:~/code/metaphysics(jonallured/feat-admin-can-manage-sale-profile-data-for-a-user)% ag "interface GravityError"
src/schema/v2/me/unlinkAuthenticationMutation.ts
11:interface GravityError {

src/schema/v2/me/deleteUserInterestMutation.ts
75:interface GravityError {

src/schema/v2/me/linkAuthenticationMutation.ts
18:interface GravityError {

src/schema/v2/me/createUserInterestMutation.ts
87:interface GravityError {

src/schema/v2/me/updateMyPasswordMutation.ts
20:interface GravityError {

src/schema/v2/users/updateUserMutation.ts
22:interface GravityError {

src/schema/v2/users/updateUserSaleProfileMutation.ts
25:interface GravityError {
```

But I'd like to break that off into another PR so I can land this work if that's ok!

damassi: sounds good!

jonallured:
Oops I def meant to assign you @damassi my bad 😝 - seems like this is good to merge so I'm going to do that and then I can open the Forque PR to take advantage of this stuff. Thanks all!

PR #3954, opened 2022-04-11
Revert "fix:fallback to email if user.name is null"

jacobherrington:
This reverts commit a1cd5718b5cb69595f796f421dcb2f4e912687c7.

We have backfilled the null names to the string `collector`, so this workaround is not necessary at the moment.


bhoggard:
Awesome!

PR #3960, opened 2022-04-12
chore: Push schema changes to forque

jonallured:
I mentioned that I was missing the automated schema updates over in Forque and @damassi helped me see that this is pretty trivial to add so that's what this PR does! I did a little refactoring to make it even easier to add things in the future.

### DIFF
--- a/scripts/push-schema-changes.js
+++ b/scripts/push-schema-changes.js
@@ -64,19 +64,18 @@ async function main() {
 
     execSync("yarn dump:staging")
 
-    await updateSchemaFile({
-      repo: "eigen",
-      body: `${defaultBody} #nochangelog`,
-    })
-    await updateSchemaFile({ repo: "force" })
-    await updateSchemaFile({
-      repo: "volt",
-      dest: "vendor/graphql/schema/metaphysics.json",
-    })
-    await updateSchemaFile({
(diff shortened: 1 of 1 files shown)


scripts/push-schema-changes.js
```diff
@@ -61,20 +62,19 @@ async function main() {
...
-    await updateSchemaFile({ repo: "force" })
-    await updateSchemaFile({
-      repo: "volt",
-      dest: "vendor/graphql/schema/metaphysics.json",
-    })
-    await updateSchemaFile({
-      repo: "pulse",
-      dest: "vendor/graphql/schema/metaphysics.json",
-    })
+    const updateOptionses = [
```
damassi, line 65: Is `Optionses` a real word? Maybe `reposToUpdate = [` might be a bit clearer?

damassi (approved):
One tiny naming q, otherwise LGTM 👍 

(Also, nice lil performance improvement with the `Promise.all` 👌 )

scripts/push-schema-changes.js
jonallured: That's for @pepopowitz he loves it when I name things like that! JK your suggestion is better - will update.

PR #3963, opened 2022-04-13
feat(FX-3850): Add ArtworksConnection under MarketingCollection

nickskalkin:
Jira: https://artsyproduct.atlassian.net/browse/FX-3850

This PR is a part of [Kaws deprecation](https://www.notion.so/artsy/Deprecating-Kaws-7be8b8b764534ea7838224f06252b99a) story

Adds `artworksConnection` under gravity-backed `marketingCollection` type.

Screenshot 2022-04-12 at 19 51 23

### DIFF
--- a/src/lib/stitching/gravity/v2/stitching.ts
+++ b/src/lib/stitching/gravity/v2/stitching.ts
@@ -87,9 +87,14 @@ const gravityMarketingCollectionsExtensions = gql`
   extend type HomePage {
     marketingCollectionsModule: HomePageMarketingCollectionsModule
   }
+  extend type MarketingCollection {
+    artworksConnection(${argsToSDL(pageableFilterArtworksArgsWithInput).join(
+      "\n"
+    )}): FilterArtworksConnection
+  }
 `
 
-const gravityMarketingCollectionsResolvers = (gravitySchema) => {
+const gravityMarketingCollectionsResolvers = (gravitySchema, localSchema) => {
   return {
     Artist: {
       marketingCollections: {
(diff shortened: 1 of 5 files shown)


src/schema/v2/__tests__/filterArtworksConnection.test.ts
```diff
@@ -619,4 +620,63 @@ describe("artworksConnection", () => {
...
+      config.ENABLE_GRAVITY_MARKETING_COLLECTIONS = true
+
+      const query = gql`
+        {
+          artworksConnection(
+            input: {
+              marketingCollectionID: "kaws-toys"
+              first: 1
+              after: ""
+              aggregations: [TOTAL]
```
mdole, line 663: #minor: do we need these? I think artworksConnection should work fine without them, but let me know if I'm wrong!

src/lib/stitching/gravity/v2/stitching.ts
```diff
@@ -205,17 +210,51 @@ const gravityMarketingCollectionsResolvers = (gravitySchema) => {
...
+        ) => {
+          return info.mergeInfo.delegateToSchema({
+            schema: localSchema,
+            operation: "query",
+            fieldName: "artworksConnection",
+            args: {
+              marketingCollectionID,
+              ...(!!context.currentArtworkID && {
+                excludeArtworkIDs: [context.currentArtworkID],
+              }),
```
mdole, line 234: @nickskalkin was this added so that in the future if we implement an "other works in this collection" rail, we don't show the work the user is already viewing (similar to what already exists for artist series)? 

If so, I would consider leaving it out for now because it's hard to know if/when we would actually implement that rail. Not something I feel extremely strong about, but I'm generally in favor of avoiding the addition of code that we're not 100% sure we need 🙂

mdole (approved):
Sweet! Two small questions, but nothing blocking. Merge when ready!

src/schema/v2/__tests__/filterArtworksConnection.test.ts
nickskalkin: removed, we don't need it, thanks!

src/lib/stitching/gravity/v2/stitching.ts
nickskalkin: good catch, thanks! Removed

PR #3965, opened 2022-04-13
feature(artwork page) Add data and models to back embedded video on the artwork page

laurabeth:
1/2 to address [GRO-884]

Adds fixture data for the POC, as well as preliminary types and fields. 

[GRO-884]: https://artsyproduct.atlassian.net/browse/GRO-884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/types/runtime/gravity/Video.ts
```diff
@@ -0,0 +1,31 @@
...
+    },
+    height: {
+      description: "The height of the video",
+      type: GraphQLInt,
+      resolve: ({ height }) => height,
+    },
+    width: {
+      description: "The width of the video",
+      type: GraphQLInt,
+      resolve: ({ width }) => width,
```
dzucconi, line 22: I believe these resolvers are redundant because by default graphql.js will use the field name for object access.

src/types/runtime/gravity/Video.ts
```diff
@@ -0,0 +1,31 @@
...
+      resolve: ({ src }) => src,
+    },
+    height: {
+      description: "The height of the video",
+      type: GraphQLInt,
+      resolve: ({ height }) => height,
+    },
+    width: {
+      description: "The width of the video",
+      type: GraphQLInt,
```
dzucconi, line 21: I'd make all these fields non-null since we're—currently—guaranteed to have this data. `new GraphQLNonNull(GraphQLInt)`

src/types/runtime/gravity/Video.ts
```diff
@@ -0,0 +1,31 @@
...
+      resolve: ({ width }) => width,
+    },
+  },
+})
+
+export const Video = Record({
+  src: String,
+  height: Number,
+  width: Number,
+})
```
dzucconi, line 31: You can remove this runtype

src/schema/v2/artwork/artworkVideos.ts
```diff
@@ -0,0 +1,62 @@
+export const videoList = {
```
dzucconi, line 1:
```suggestion
export const VIDEOS = {
```

dzucconi:
Looking at the Force PR — it seems like it'd be better to expose a union type of `figures` rather than a single `video`

src/types/runtime/gravity/Video.ts
```diff
@@ -0,0 +1,31 @@
+import { GraphQLInt, GraphQLObjectType, GraphQLString } from "graphql"
```
damassi, line 1: This file should live in [src/schema/v2](https://github.com/artsy/metaphysics/tree/main/src/schema/v2) alongside the rest of the schema

src/types/runtime/gravity/Video.ts
damassi: Yup, the only reason we ever need to resolve fields like this is if we're accessing `snake_case` props from Gravity that we' by convention convert to `camelCase`. 

So like: 
```tsx
someField: {
  type: ... 
  resolve: ({ some_field }) => some_field
}
```

src/types/runtime/gravity/Video.ts
laurabeth: cool, thanks for the heads up. hoping this solves another problem I ran into. 🤞🏽

src/schema/v2/artwork/index.ts
```diff
@@ -1286,6 +1288,13 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLString,
         resolve: ({ size_bucket }) => size_bucket,
       },
+      video: {
+        description: "returns a Video object for the associated artwork",
+        type: VideoType,
+        resolve: ({ id }) => {
```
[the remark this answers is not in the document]
laurabeth: Concerning.

src/schema/v2/types/Video.ts
```diff
@@ -0,0 +1,29 @@
...
+    },
+    height: {
+      description: "The height of the video",
+      type: GraphQLNonNull(GraphQLInt),
+      resolve: ({ height }) => height,
+    },
+    width: {
+      description: "The width of the video",
+      type: GraphQLNonNull(GraphQLInt),
+      resolve: ({ width }) => width,
```
damassi, line 26: No need to resolve `height` and `width` up above

dzucconi (approved):

PR #3966, opened 2022-04-13
chore:add placeholder for cascadingEndTimeIntervalMinutes

bhoggard:
re: https://artsyproduct.atlassian.net/browse/BID-348

This is a placeholder with a `/ 60` calculation until Gravity has been updated.

### DIFF
--- a/src/schema/v2/sale/__tests__/index.test.ts
+++ b/src/schema/v2/sale/__tests__/index.test.ts
@@ -847,6 +847,7 @@ describe("Sale type", () => {
         start_at: "2022-03-09 09+07:00",
         end_at: "2022-03-12 09+07:00",
         cascading_end_time_interval: 120,
+        cascading_end_time_interval_minutes: 2,
       })
 
       expect(response.sale.formattedStartDateTime).toEqual("March 9 – 12, 2022")
--- a/src/schema/v2/sale/__tests__/index.test.ts
+++ b/src/schema/v2/sale/__tests__/index.test.ts
@@ -767,6 +767,7 @@ describe("Sale type", () => {
       {
         sale(id: "foo-foo") {
           formattedStartDateTime
+          cascadingEndTimeIntervalMinutes
         }
       }
     `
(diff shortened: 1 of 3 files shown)


src/schema/v2/sale/index.ts
```diff
@@ -178,6 +178,12 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
         resolve: ({ cascading_end_time_interval }) =>
           cascading_end_time_interval,
       },
+      cascadingcascadingEndTimeIntervalMinutesEndTimeInterval: {
```
annacarey, line 181: I think there's a typo? Can this be `cascadingEndTimeIntervalMinutes`?

annacarey (changes requested):
Confused about the namings but maybe that was unintentional. Shouldn't the field just be `cascadingEndTimeIntervalMinutes`?

src/schema/v2/sale/index.ts
dzucconi: 😮

src/schema/v2/sale/index.ts
```diff
@@ -178,6 +178,12 @@ export const SaleType = new GraphQLObjectType<any, ResolverContext>({
...
+        type: GraphQLInt,
+        description: "Amount of minutes in between each lot closing.",
+        resolve: ({ cascading_end_time_interval }) =>
+          cascading_end_time_interval / 60,
```
mzikherman, line 185: This might be odd for sales that don't have this field enabled.

Since this field is nullable and we're using the presence to determine if a sale has this enabled, we should leave the type as-is (nullable `GraphQLInt`), and probably `return null` if `cascading_end_time_interval` is not set, otherwise we do the ` / 60` thing.

src/schema/v2/sale/__tests__/index.test.ts
```diff
@@ -847,6 +847,7 @@ describe("Sale type", () => {
         start_at: "2022-03-09 09+07:00",
         end_at: "2022-03-12 09+07:00",
         cascading_end_time_interval: 120,
+        cascading_end_time_interval_minutes: 2,
```
mzikherman, line 850: Spec that requests the new field (for a sale w/ and w/o cascading) might be nice!

src/schema/v2/sale/index.ts
mzikherman: 6chm67

mzikherman (approved):

annacarey (approved):

src/schema/v2/sale/index.ts
annacarey: Omg @mzikherman!

dzucconi: 🥇

PR #3968, opened 2022-04-14
feat: stitch in create bank debit setup for order mutation from Exchange (TX-270)

starsirius:
https://artsyproduct.atlassian.net/browse/TX-270

Depends on https://github.com/artsy/exchange/pull/1129. This stitches in the `CreateBankDebitSetupForOrder` mutation from Exchange that allows clients to create a SetupIntent for ACH.

### Success

Screen Shot 2022-04-15 at 10 58 42 PM

### Error

Screen Shot 2022-04-15 at 10 57 17 PM


_schema.graphql
```diff
@@ -4175,6 +4175,11 @@ type CollectorProfileType {
   self_reported_purchases: String
 }
 
+# Represents either an action or a potential failure
+union CommerceActionOrFailureUnion =
```
narikazu, line 4179: Question: Do you happen to know how this prefix `Commerce` routes to `Exchange` mutation?

narikazu (dismissed):
LGTM

_schema.graphql
starsirius: The [rename](https://github.com/artsy/metaphysics/blob/c46da5f9c93f7360f8c3ab1c3a9aa97ec03c0112/src/lib/stitching/exchange/schema.ts#L25-L37) happens here I believe.

PR #3969, opened 2022-04-14
chore: update grav schema to include MarketingCollectionGroup

mdole:
Just a schema bump now that we include these groups in Gravity's schema!


nickskalkin (approved):
👍

PR #3989, opened 2022-04-20
Kajatiger/px 4965 add field in order graphql type to differentiate domestic international shipping

kajatiger:
Update after PR on exchange: https://github.com/artsy/exchange/pull/1130

### DIFF
--- a/src/data/exchange.graphql
+++ b/src/data/exchange.graphql
@@ -133,6 +133,7 @@ type BuyOrder implements Order {
   requestedFulfillment: RequestedFulfillmentUnion
   seller: OrderPartyUnion!
   sellerTotalCents: Int
+  shippingRadius: String
   shippingTotalCents: Int
   source: OrderSourceEnum!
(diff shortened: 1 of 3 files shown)


src/data/exchange.graphql
```diff
@@ -1,8 +1,3 @@
-"""
-Represents either an action or a potential failure
-"""
-union ActionOrFailureUnion = OrderRequiresAction | OrderWithMutationFailure
-
```
kajatiger, line 5: @starsirius I am feeling a bit uneasy. I copied my newest schema from exchange to metaphysics after adding the `shippingRadius` field on the order. But somehow this seems to remove some other things that you had just added a week ago. Why is this not on exchange main?

leamotta (approved):

src/data/exchange.graphql
starsirius: Sorry for the delay. Looks like this has been resolved. Let me know if I can still help!

PR #3991, opened 2022-04-20
feat(FX-3873): dev can pass an image as an argument to `doNotUseImageSearch` query

dimatretyak:
Jira ticket: [FX-3873]

### Useful links
* [Spike](https://www.notion.so/artsy/Spike-Developer-can-pass-an-image-as-an-argument-to-a-query-in-Metaphysics-29536efe82fe4743836b0bbe29ccaab4)
* [Technical Plan](https://www.notion.so/artsy/Reverse-Image-Search-for-Artworks-aka-Shazam-for-Art-33c9ece7293641b697788bad921c58bd)
* [How to test query in Insomnia](https://www.notion.so/artsy/Spike-Developer-can-pass-an-image-as-an-argument-to-a-query-in-Metaphysics-29536efe82fe4743836b0bbe29ccaab4#7388622c8b19427081e5dcc71e436a72)

### Acceptance Criteria
- [x] Can execute GraphQL query against Metaphysics staging using Insomnia (or another GraphQL client) to verify an image can be passed as an argument.
- [x] Returns an error if queried against production
- [x] Echos back simple file metadata captured by the `graphql-upload` package
- [x] Only available for logged in users

### Response example
```graphql
{
	"data": {
		"doNotUseImageSearch": {
			"__typename": "ImageSearch",
			"encoding": "7bit",
			"filename": "image.jpeg",
			"mimetype": "image/jpeg"
		}
	},
	"extensions": {
		"requests": {},
		"requestID": "8d1f7ff0-c1a1-11ec-831e-dd079d21ba20"
	}
}
```

### Demo
https://user-images.githubusercontent.com/3513494/164680294-301ce00b-cd51-48a6-ab1f-945a9ab6262f.mp4

[FX-3873]: https://artsyproduct.atlassian.net/browse/FX-3873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mzikherman:
Fascinating! My initial thought is to be hesitant about allowing uploading files directly to a webserver, generally.

Instead, an alternate setup (I'm not sure the exact use case, this looks to be a staging-only experiment), might be to upload a file directly to S3, into our `artsy-media-uploads` (or whatever) bucket/folder, (which can use private ACL's as well). Then, the key to that file on S3 can be passed in as a string argument where needed. There should be examples out there of uploads directly to S3 from Eigen, as well as other apps.

Whatever backend is responsible for accessing the uploaded image and doing _something_ with it, could accept that S3 key and (with proper AWS credentials of its own) access it directly from S3.

dimatretyak:
> I'm not sure the exact use case, this looks to be a staging-only experiment

Yes, that's right. At the moment, this query can only be used on staging env


> Instead, an alternate setup (I'm not sure the exact use case, this looks to be a staging-only experiment), might be to upload a file directly to S3, into our artsy-media-uploads (or whatever) bucket/folder, (which can use private ACL's as well). Then, the key to that file on S3 can be passed in as a string argument where needed. There should be examples out there of uploads directly to S3 from Eigen, as well as other apps.

> Whatever backend is responsible for accessing the uploaded image and doing something with it, could accept that S3 key and (with proper AWS credentials of its own) access it directly from S3.

The ьфшт idea is to make Integration with TinEye to execute search requests. You can find more info and context [here](https://artsyproduct.atlassian.net/browse/FX-3751) and in [technical plan](https://www.notion.so/artsy/Reverse-Image-Search-for-Artworks-aka-Shazam-for-Art-33c9ece7293641b697788bad921c58bd#05527d855fef4acaa3e1937a866ebefc)

src/index.ts
```diff
@@ -207,13 +211,33 @@ function startApp(appSchema, path: string) {
...
+
+    const status = error?.status ?? 500
+    const message = error?.message ?? "Something went wrong"
+
+    return res.status(status).json({
+      message,
+      stack: !PRODUCTION_ENV ? error.stack.split("\n") : null,
+    })
+  })
+
```
dimatretyak, line 240: **question**: Should global error handler be used in Express to return error message and stack trace (see screenshot #1, #2)? Otherwise, we will always get 404 response code and the corresponding message text for this error (see attached screenshot #3).

On the one hand, it can help developers to quickly understand what is working wrong and find error. For example, as it was in my cases when I started working with `graphql-upload` and always got `404 Cannot POST /v2` in response to requests where I sent the file. It turned out that I used the wrong query structure and an error was thrown in the `graphql-upload` middleware for this reason. But due to the fact that we did not have a global error handler, the default `404 error` was always returned.

But on the other hand, won't I break something in another place or in another application if I add these changes? What do you think about this? Or maybe add a global error handler only for staging/development environments?

### Screenshot 1
error-1

### Screenshot 2
error-2


### Screenshot 3
error-4

dimatretyak, line 240: **question**: Should global error handler be used in Express to return error message and stack trace (_see screenshot 1, 2_)? Otherwise, we will always get 404 status code and the corresponding message text for this error (_see attached screenshot 3_).

On the one hand, it can help developers to quickly understand what is working wrong and find error. For example, as it was in my case when I started working with `graphql-upload` and always got `404 Cannot POST /v2` in response to requests where I sent the file. It turned out that I used the wrong query structure and an error was thrown in the `graphql-upload` middleware for this reason. But due to the fact that we did not have a global error handler, the default `404 error` was always returned.

But on the other hand, won't I break something in **another place or in another application** if I add these changes?

What do you think about this? Or maybe add a global error handler only for staging/development environments?

#### Screenshot 1
error-1


#### Screenshot 2
error-2

#### Screenshot 3 (without global error handler)
error-3

src/lib/graphqlUpload.ts
```diff
@@ -0,0 +1,28 @@
...
+  GRAPHQL_UPLOAD_MAX_FILES,
+  GRAPHQL_UPLOAD_MAX_FILE_SIZE_IN_BYTES,
+} = config
+
+export const graphqlUploadMiddleware = (req, res, next) => {
+  const accessToken = req.headers["x-access-token"] as string | undefined
+
+  if (PRODUCTION_ENV && req.is("multipart/form-data")) {
+    throw new Error("You cannot upload file for production")
+  }
```
dimatretyak, line 15: Acceptance Criteria has an item "Returns an error if queried against production". Maybe it's better not to throw an error (I'm afraid that `multipart/form-data` can be used somewhere else and such a check will break the logic there)?

dimatretyak:
### Response for small size file
https://user-images.githubusercontent.com/3513494/164745012-8a37d955-5cfa-44d6-9f56-fc28275ead63.mp4

### Response for large file size
https://user-images.githubusercontent.com/3513494/164745062-d388bfa4-2727-4cc1-b5a9-f3fc371172fc.mp4

### Response for unauthenticated user
https://user-images.githubusercontent.com/3513494/164745238-52267beb-f0b9-44a5-885e-0dd2af4a6e07.mp4

### Response for production env
https://user-images.githubusercontent.com/3513494/164745131-6b04a46c-ebef-43c2-89e2-f580796049fb.mp4

src/lib/graphqlUpload.ts
```diff
@@ -0,0 +1,28 @@
+import { graphqlUploadExpress } from "graphql-upload"
+import config from "../config"
+
+const {
+  PRODUCTION_ENV,
+  GRAPHQL_UPLOAD_MAX_FILES,
+  GRAPHQL_UPLOAD_MAX_FILE_SIZE_IN_BYTES,
+} = config
+
+export const graphqlUploadMiddleware = (req, res, next) => {
```
anandaroop, line 10: Typing could be improved here:

```ts
import { RequestHandler } from "express"

// ...

export const graphqlUploadMiddleware: RequestHandler = (req, res, next) => {
```

anandaroop (approved):
Nice spike @dimatretyak 👏🏽 

@mzikherman the idea here is that MP acts as a pass-through for an image snapped on a user device, and then sent onward to TinEye for reverse image matching. Because the image data is transient, we never planned to store it anywhere (though of course it does get temporarily buffered to a tmp dir or something).

src/lib/graphqlUpload.ts
dimatretyak: O, good point. Fixed

src/lib/graphqlUpload.ts
```diff
@@ -0,0 +1,28 @@
...
+  const accessToken = req.headers["x-access-token"] as string | undefined
+
+  if (PRODUCTION_ENV && req.is("multipart/form-data")) {
+    throw new Error("You cannot upload file for production")
+  }
+
+  // Only available for logged in users and non-production env
+  if (!PRODUCTION_ENV && accessToken) {
```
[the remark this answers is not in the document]
dimatretyak: Yes, that's right. This check will end silently. We can check something like this. What do others think about this?

```javascript
if (!PRODUCTION_ENV && isMultipartForm && !accessToken) {
  throw new Error("Only logged in user can upload file")
}
```

src/index.ts
dblandin: Interesting that 404 is the default error response. I guess that might be because we don't currently respond to `multipart/form-data` requests?

This sounds like a nice change but I'd suggest splitting it out into its own PR where we can discuss the implementation separate from this spike for image uploading. Some early feedback though: I might extract the handler into a separate file and perhaps have a specific environment variable to explicitly enable the middleware and verbose error responses.

src/config.ts
```diff
@@ -209,4 +209,6 @@ export default {
   VORTEX_API_BASE,
   VORTEX_APP_ID,
   VORTEX_TOKEN,
+  GRAPHQL_UPLOAD_MAX_FILE_SIZE_IN_BYTES: 5 * 1000 * 1000, // 5 MB
+  GRAPHQL_UPLOAD_MAX_FILES: 1,
```
dblandin, line 213: **suggestion**: Looks like we're not always consistent with alphabetizing these lines but we could move these two lines right before `GRAVITY_API_BASE`.

src/config.ts
```diff
@@ -209,4 +209,6 @@ export default {
   VORTEX_API_BASE,
   VORTEX_APP_ID,
   VORTEX_TOKEN,
+  GRAPHQL_UPLOAD_MAX_FILE_SIZE_IN_BYTES: 5 * 1000 * 1000, // 5 MB
```
dblandin, line 212: **suggestion**: It looks like the recommended average size when searching for images against the TinEye API is 100kb ([source](https://services.tineye.com/developers/mobileengine/api_reference/)). What to you think of reducing this default to 300kb to start? 5MB sounds pretty big.

src/index.ts
```diff
@@ -208,12 +209,27 @@ function startApp(appSchema, path: string) {
   })
 
   app.use("/batch", bodyParser.json(), graphqlBatchHTTPWrapper(graphqlServer))
-  app.use(graphqlServer)
+  app.use(graphqlUploadMiddleware, graphqlServer)
```
dblandin, line 212: **question** Is there a difference between having these in the same vs different `app.use` invocations? Also, would we consider having an explicit feature flag for this middleware instead of relying on "not production"?

```typescript
const { ENABLE_GRAPHQL_UPLOAD } = config

if (ENABLE_GRAPHQL_UPLOAD) {
  app.use(graphqlUploadExpress({
    maxFileSize: GRAPHQL_UPLOAD_MAX_FILE_SIZE_IN_BYTES,
    maxFiles: GRAPHQL_UPLOAD_MAX_FILES,
  }))
}

app.use(graphqlServer)
```

src/schema/v2/imageSearch.ts
```diff
@@ -0,0 +1,71 @@
...
+  description: "Get an image info",
+  args: {
+    image: {
+      description: "Image file",
+      // TODO: Remove `as unknown` when "graphql" is updated to version 16.x.x
+      // Wrong TS declaration for GraphQLScalarType in node_modules/graphql/type/definition.d.ts
+      type: new GraphQLNonNull((GraphQLUpload as unknown) as GraphQLScalarType),
+    },
+  },
+  resolve: async (_root, { image }) => {
```
dblandin, line 41: **question**: Would this be a reasonable spot to check for user authentication instead of at the middleware layer?

We could either do something simple that's best off of the request header values...
```typescript
if (!meLoader) throw new Error("You need to be signed in to perform this action")
```

... or something more robust that actually ensures that the token is still valid as is done [here](https://github.com/artsy/metaphysics/blob/1d0ecaf0813f78f5bd0c602b2b3c80eca6dcf426/src/schema/v2/authenticationStatus.ts):

```typescript
if (!meLoader) throw new Error("You need to be signed in to perform this action")

try {
  await meLoader()
} catch (err) {
}
```

dblandin (commented):
Looking good! Left a few comments. 💯

src/config.ts
dimatretyak: > What to you think of reducing this default to 300kb to start?

I have nothing against it 😅

src/schema/v2/imageSearch.ts
dimatretyak: I [did this](https://github.com/artsy/metaphysics/pull/3991/commits/da8d441d7f3096799ffbf48d0093b6ed438eb72d#diff-8ca357c42a580b97146120dc9b6348eb12c0f59514e8dd4de609734e2042d247R42-R48) initially. But the disadvantage of this approach is that even an unauthorized user could upload a file (you can read a little more about it [here](https://www.notion.so/artsy/Spike-Developer-can-pass-an-image-as-an-argument-to-a-query-in-Metaphysics-29536efe82fe4743836b0bbe29ccaab4#90e03d1c3b644d5fa3de2d3421763d1c

https://www.notion.so/artsy/Spike-Developer-can-pass-an-image-as-an-argument-to-a-query-in-Metaphysics-29536efe82fe4743836b0bbe29ccaab4#a37eae0dbcf64e33ad4b677f775f9e6a

https://www.notion.so/artsy/Spike-Developer-can-pass-an-image-as-an-argument-to-a-query-in-Metaphysics-29536efe82fe4743836b0bbe29ccaab4#7324467ebc124d55abf23b257b1250a9

https://www.notion.so/artsy/Spike-Developer-can-pass-an-image-as-an-argument-to-a-query-in-Metaphysics-29536efe82fe4743836b0bbe29ccaab4#a001580801a14a6db59c40cc61ce0c7d))

dblandin: Right, that makes sense. Thanks for writing those extra details!

I think before we make a change at this middleware layer, maybe we can stress test the resolver-level auth checking and see if that might be sufficient for now? There might be other solutions like IP request throttling to explore. As is, the middleware check doesn't seem very robust. An abuser could easily satisfy the check by providing a fake value in the `x-access-token` header, right? We don't actually ensure that the token is valid at this level.

What do you think? We're still early on in this project so I think we can start with a simpler implementation, enabled in staging only, as we iterate.

dimatretyak: > An abuser could easily satisfy the check by providing a fake value in the x-access-token header, right?

Yes, you are absolutely right about this and for this reason I also don't quite like the approach with custom middleware

> What do you think? We're still early on in this project so I think we can start with a simpler implementation, enabled in staging only, as we iterate.

If you don't mind placing the logic in graphql resolver, then I will be only glad of it 👍

src/index.ts
dimatretyak: > would we consider having an explicit feature flag for this middleware instead of relying on "not production"?

Do you want to read this value from `process.env`?

dimatretyak: > Is there a difference between having these in the same vs different app.use invocations?

If I understand you correctly, there is no difference. We needed to perform some additional checks ([like this](https://github.com/artsy/metaphysics/blob/530d65e7dcbd44aa8087c2419451419d400a2233/src/lib/graphqlUpload.ts#L14-L16)) and for this reason I created `graphqlUploadMiddleware` for this

dblandin: Yep, our configuration setup relies on `process.env` so we can follow that pattern for a new `ENABLE_GRAPHQL_UPLOAD` flag ([example](https://github.com/artsy/force/blob/6724e12b80e75655d29a46f541ad364c5efb47c7/src/config.ts#L50-L57)).

src/schema/v2/imageSearch.ts
dblandin: > If you don't mind placing this logic in GraphQL resolver

Sounds good! I'd feel comfortable with that + an explicit feature flag to enable the new middleware 👍🏼

src/index.ts
dimatretyak: Added

src/index.ts
dimatretyak: > This sounds like a nice change but I'd suggest splitting it out into its own PR where we can discuss the implementation separate from this spike for image uploading

I agree. I'll do it

src/index.ts
```diff
@@ -208,7 +209,7 @@ function startApp(appSchema, path: string) {
...
```
dblandin, line 212: **suggestion**: I still think we should hoist up the package middleware here, instead of defining a custom middleware, and only call `app.use` if the feature flag is enabled. Makes the change more explicit and ensures that we're not modifying the middleware chain unless that flag is enabled:

```typescript
app.use("/batch", bodyParser.json(), graphqlBatchHTTPWrapper(graphqlServer))

if (ENABLE_GRAPHQL_UPLOAD) {
  app.use(graphqlUploadExpress({
    maxFileSize: GRAPHQL_UPLOAD_MAX_FILE_SIZE_IN_BYTES,
    maxFiles: GRAPHQL_UPLOAD_MAX_FILES,
  }))
}

app.use(graphqlServer)
```

The `graphqlUploadExpress` should call next() unless a request has a `multipart/form-data` content type. I think the default error response of 404 if a client does execute a request with this content-type and the feature flag is disabled is totally fine for this iteration.

What do you think?

dblandin (dismissed):
One last suggestion to simplify the middleware chain build up a bit more but otherwise LGTM! Nice work!! 💯

src/index.ts
dimatretyak: > I think the default error response of 404 if a client does execute a request with this content-type and the feature flag is disabled is totally fine for this iteration

small clarification: in this scenario, we will not get a 404 error (see screenshot), since `express-graphql` is next in the middleware chain

image

dblandin: Gotcha. Thanks for the clarification! 👍🏼

I think that's a fine error message for now.

dblandin (approved):
💯

src/schema/v2/imageSearch.ts
```diff
@@ -0,0 +1,82 @@
...
+      // Wrong TS declaration for GraphQLScalarType in node_modules/graphql/type/definition.d.ts
+      type: new GraphQLNonNull((GraphQLUpload as unknown) as GraphQLScalarType),
+    },
+  },
+  resolve: async (_root, { image }, { meLoader }) => {
+    if (!meLoader) {
+      throw new Error("You need to be signed in to perform this action")
+    }
+
+    // Verifying that the token is still valid
```
mzikherman, line 46: Nice

src/index.ts
```diff
@@ -208,6 +212,16 @@ function startApp(appSchema, path: string) {
   })
 
   app.use("/batch", bodyParser.json(), graphqlBatchHTTPWrapper(graphqlServer))
+
+  if (ENABLE_GRAPHQL_UPLOAD) {
```
mzikherman, line 216: Is this feature flag intending to be used in staging (or prod too)? Either way, nice to have it toggle-able.

dimatretyak: Yes, that's right. For first iteration this flag will be enable for staging env only

mzikherman:
Feel free to merge! We can always disable it via the toggle if there's any unexpected issues.

Even if meant to be more transient, I still think an eventual production-ized version of this would likely be better to have uploads directly to S3 (bypassing MP for file uploads entirely). We could probably setup some S3 policy or lambda to scrub these regularly, or it's probably fine to keep them.

anandaroop (approved):

PR #4005, opened 2022-04-26
feat(SWA-333): add recently sold query

kiryl-zubarau:
This PR adds new query `recentlySoldArtworks` for the SWA landing page. The query based on static information provided by Merchandising and Auctions team.

JIRA - https://artsyproduct.atlassian.net/browse/SWA-333

Screenshot from 2022-04-26 09-47-37


_schemaV2.graphql
```diff
@@ -15196,6 +15227,12 @@ type Viewer {
   previewSavedSearch(
     attributes: PreviewSavedSearchAttributes
   ): PreviewSavedSearch
+  recentlySoldArtworks(
```
iskounen, line 15230: **nitpick:** Adding a description to this connection would improve the formatting of this file.

**question:** Why does this appear twice in the schema?

src/schema/v2/__tests__/recentlySoldArtworks.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      `
+
+    const context = {
+      artworksLoader,
+    }
+
+    const response = await runQuery(query, context)
+    const recentlySoldArtworks = response.recentlySoldArtworks
+
+    expect(recentlySoldArtworks.edges).toHaveLength(20)
```
iskounen, line 39: **suggestion:** If you change `20` to `recentlySoldArtworks.length`, this test assertion will make more sense to future developers.

src/schema/v2/recentlySoldArtworks.ts
```diff
@@ -0,0 +1,173 @@
...
+export const RecentlySoldArtworks: GraphQLFieldConfig<void, ResolverContext> = {
+  args: pageable({}),
+  type: connectionDefinitions({ nodeType: RecentlySoldArtworkType })
+    .connectionType,
+  resolve: async (_, options, { artworksLoader }) => {
+    const { page, size } = convertConnectionArgsToGravityArgs(options)
+    const totalCount = recentlySoldArtworks.length
+    const ids = recentlySoldArtworks.map(({ artworkId }) => artworkId)
+
+    const response = await artworksLoader({ ids })
```
iskounen, line 36: **suggestion (non-blocking):** Since we're already hard-coding the list of recently sold artworks, why not also hard-code the artwork data that the rails need and skip this step?

kiryl-zubarau: I would prefer to use fresh artwork information just in case since looks like the query correctly works with cache.

fladson:
LGTM, I would just be more explicit about the not dynamic part, it may be misleading. Can we rename the endpoint or type to include `static` somewhere?

src/schema/v2/recentlySoldArtworks.ts
```diff
@@ -0,0 +1,173 @@
...
+import { pageable } from "relay-cursor-paging"
+import { createPageCursors } from "schema/v1/fields/pagination"
+import { ResolverContext } from "types/graphql"
+import { ArtworkType } from "./artwork"
+
+const RecentlySoldArtworkType = new GraphQLObjectType<any, ResolverContext>({
+  name: "RecentlySoldArtworkType",
+  fields: {
+    lowEstimateUSD: {
+      type: GraphQLString,
```
starsirius, line 13: For pricing fields, rather than returning a pre-formatted string, I'd encourage us to move towards the [`Money` type](https://github.com/artsy/metaphysics/blob/eb885d10dd0f8ac101e7c8bc4bc20e58055b38dd/src/schema/v2/fields/money.ts#L158-L189) which implements good practices around currencies, supports formatting by clients, etc.

src/schema/v2/recentlySoldArtworks.ts
kiryl-zubarau: [Gravity doesn't return deleted/unpublished artworks](https://github.com/artsy/gravity/blob/acf176e92bb053b11a9947db22554e33074238a6/app/api/v1/artworks_endpoint.rb#L195-L198), so I'm not sending such artworks to the client. this helps to prevent redirects to the artwork's deleted pages on artwork card click.

_schemaV2.graphql
kiryl-zubarau: > Why does this appear twice in the schema?

First one it's a field in the `Query` type the second one in the `Viewer` type.
Here is a bit described the `Viewer` type https://artsy.slack.com/archives/CP9P4KR35/p1599075665072400

src/schema/v2/recentlySoldArtworks.ts
iskounen: Understood 👍 Thanks for adding "static" to the description to make this clear for future developers.

iskounen (approved):

PR #4007, opened 2022-04-26
feat: Allow creating artists within myCollectionCreateArtworkMutation

olerichter00:
Addresses [CX-2527]

## Description

This change will allow clients to add an array of artists to the mutation `myCollectionCreateArtworkMutation`. Within the resolver, the artists will be created and then associated with the artwork. This is a follow up for https://github.com/artsy/gravity/pull/15238


src/schema/v2/artwork/index.ts
```diff
@@ -141,10 +147,17 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
         },
-        resolve: ({ artists }, { shallow }, { artistLoader }) => {
+        resolve: (
+          { artists },
+          { shallow },
+          { authenticatedArtistLoader, artistLoader }
+        ) => {
           if (shallow) return artists
+
+          const loader = authenticatedArtistLoader || artistLoader
```
olerichter00, line 157: We need an authenticated request to Gravity in case we fetch a personal artist (artists created by the user). In the case the request to MP is not authenticated, `authenticatedArtistLoader` is undefined and we fall back to the unauthenticated `artistLoader`.

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
```diff
@@ -154,16 +153,29 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
...
 
+    if (!artistIds?.length && !artists?.length) {
+      return new Error("You need to provide either artist IDs or artists")
+    }
+
+    // Create artists if `artist` is provided in the input fields
+    if (artists?.length) {
+      const newArtistIDs = await createArtists(artists, createArtistLoader)
+
+      artistIds = [...(artistIds || []), ...newArtistIDs]
```
olerichter00, line 176: It's not really necessary at the moment, but I thought it would be useful because in the future we might want to link several artists to one artwork and mix Artsy's artists and personal/custom artists.

damassi (approved):
Looks good!

PR #4016, opened 2022-04-28
feat: Introduce myCollectionAuctionResults

olerichter00:
Addresses [CX-2512]

## Description

Introduces `myCollectionAuctionResults` endpoint under `me` which returns auction results for artists from a user's "My Collection".

**Query:**

```graphql
{
  me {
    myCollectionAuctionResults(first: 10) {
      edges {
        node {
          id
        }
      }
    }
  }
}

```

[CX-2512]: https://artsyproduct.atlassian.net/browse/CX-2512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/me/myCollectionAuctionResults.ts
+++ b/src/schema/v2/me/myCollectionAuctionResults.ts
@@ -114,18 +114,8 @@ const MyCollectionAuctionResults: GraphQLFieldConfig<any, ResolverContext> = {
       )
 
       return merge(
-        {
-          pageCursors: createPageCursors(
-            {
-              page,
-              size,
-            },
-            total_count
-          ),
-        },
-        {
-          totalCount: total_count,
-        },
+        { pageCursors: createPageCursors({ page, size }, total_count) },
+        { totalCount: total_count },
         connectionFromArraySlice(enrichedAuctionResults, options, {
           arrayLength: total_count,
           sliceStart: offset,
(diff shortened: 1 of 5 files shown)


MounirDhahri (approved):
Niice 🙌 thanks Ole

src/schema/v2/me/myCollectionAuctionResults.ts
```diff
@@ -0,0 +1,142 @@
...
+          )
+
+          auctionResult.artist = artist
+          return auctionResult
+        }
+      )
+
+      return merge(
+        { pageCursors: createPageCursors({ page, size }, total_count) },
+        { totalCount: total_count },
```
MounirDhahri, line 120: **thought:** shouldn't this be `filteredAuctionResults.length` since total_count might return items without an artist id (as per your comment in line 101)

olerichter00: `total_count` here is the total count of auction results for the given query from `Diffusion`. When we filter out auction results in line 101 it just removes auction results without matching artists from the current page. I can't remember why we needed to filter them out, but it should not brake pagination.

MounirDhahri: ah cool, I got you now. thank you

PR #4029, opened 2022-05-04
feat(FX-3878): Developer can search for matching artworks by image via Metaphysics

dimatretyak:
Jira ticket: https://artsyproduct.atlassian.net/browse/FX-3878

### Acceptance Criteria
* Given an image is indexed to TinEye
* When I query for the image using what should be a match
* Then I get back some results

### Description
Define a new root-level query field which returns limited results following a search against TinEye

```graphql
{
  reverseImageSearch(image: $file) {
    results {
      score
      filepath
      ...
    }
  }
}
```

#### Success
https://user-images.githubusercontent.com/3513494/167187549-3ae3dbf7-c6f2-4a23-9091-23480fbe3d73.mp4

### Fail
https://user-images.githubusercontent.com/3513494/167187674-d6086254-15db-46bf-9e9d-20c642db1f82.mp4


src/schema/v2/artworkImageSearch.ts
```diff
@@ -0,0 +1,153 @@
...
+      }
+    }
+
+    return {
+      errors: response.error.map((error) => ({
+        message: error,
+        code: "invalid",
+        path: [info.fieldName],
+      })),
+    }
```
dimatretyak, line 151: According to the conditions of the jira ticket, we should be able to get errors in this way
```graphql
{
  artworkImageSearch(image: $file) {
    ... on Errors {
      errors {
        message
      }
    }
  }
}
```

But usually such pattern is used in resolvers in Gravity (for example, as [here](https://github.com/artsy/gravity/blob/main/app/graphql/mutations/create_user_address.rb#L13-L22)) and there are no such use cases in Metaphysics. 

**Question**: Maybe it will be enough if we just throw error in this way?

```javascript
throw new Error(response.error.join("\n"))
```

src/schema/v2/artworkImageSearch.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  return {
+    errors: response.error.map((error) => ({
+      message: error,
+      code: "invalid",
+      path: [info.fieldName],
+    })),
+  }
+}
```
dimatretyak, line 144: According to the conditions of the jira ticket, we should be able to get errors in this way
```graphql
{
  artworkImageSearch(image: $file) {
    ... on Errors {
      errors {
        message
      }
    }
  }
}
```

But usually such pattern is used in resolvers in Gravity (for example, as [here](https://github.com/artsy/gravity/blob/main/app/graphql/mutations/create_user_address.rb#L13-L22)) and there are no such use cases in Metaphysics. 

**Question**: Maybe it will be enough if we just throw error in this way?

```javascript
throw new Error(response.error.join("\n"))
```

dblandin: Yep, agreed. I think we can rely on root-level errors for this first iteration and not introduce a union type 👍🏼

dimatretyak: Updated

_schemaV2.graphql
```diff
@@ -14916,6 +14956,12 @@ type Viewer {
   # List of all artwork attribution classes
   artworkAttributionClasses: [AttributionClass]
 
+  # Search for matching artworks by image
+  artworkImageSearch(
```
dblandin, line 14960: **suggestion**: Uff, naming is hard. I wonder if this is the right name to use here. What do you think of renaming this new root query field to `reverseImageSearch`? I feel like that might capture the intent more clearly.

As is, the name here might imply that this query field is to search for artwork images.

src/lib/loaders/loaders_without_authentication/tineye.ts
```diff
@@ -0,0 +1,35 @@
...
+          },
+        },
+      },
+    })
+
+    return response.body
+  }
+
+  return {
+    searchArtworkByImageLoader,
```
dblandin, line 33: **suggestion**: Should we call this a `tineyeSearchLoader` or something similar? TinEye itself doesn't have any awareness of artworks so I wonder if we should avoid the term `artwork` here.

We'll likely need to add that artwork awareness in Metaphysics by adding an `artwork` field in `ReverseImageSearchResult`, backed by a gravity API loader.

dblandin (commented):
Looking good!! 💯

_schemaV2.graphql
dimatretyak: > Uff, naming is hard

Oh, I understand you perfectly 😅

image

dimatretyak: > What do you think of renaming this new root query field to reverseImageSearch?

I like this name 👍

dimatretyak: Done

src/lib/loaders/loaders_without_authentication/tineye.ts
dimatretyak: > Should we call this a tineyeSearchLoader or something similar?

Renamed

dblandin (dismissed):
Awesome! 💯 LGTM

dblandin:
One random thought - it's interesting that including a check for a valid user session via the `/api/v1/me` API endpoint adds an additional second of latency. I wonder if Metaphysics could start verifying user tokens or if there's a different API endpoint we could use that would offer a quicker turnaround. Even if `/api/v1/me` accepted a `compact` / `shallow` query param to include less data.

Or if `/api/v1/me` continues to be the preferred way to check, can we improve its performance?

Although in a browser, it seems like that endpoint returns in around ~200ms. What else is involved in the GraphQL extension timing to get the overall time closer to ~900ms?

src/lib/loaders/loaders_without_authentication/tineye.ts
```diff
@@ -0,0 +1,35 @@
...
+          value: image,
+          options: {
+            filename,
+            contentType,
+          },
+        },
+      },
+    })
+
+    return response.body
```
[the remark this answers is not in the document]
mzikherman: I might actually skip this whole loader/abstraction if you'd like. The use of `DataLoader` isn't compelling for this use case. Instead, you can put this method (and that `apis/tiney.ts` file) into a more conventional helper and import/refer to it in your resolver (vs. trying to use this loader/factory thing and having that show up in `context`, etc.)

src/schema/v2/reverseImageSearch.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export const reverseImageSearchResolver = async (_root, args, context) => {
+  const { image } = args
+  const { meLoader, tineyeSearchLoader } = context
+
+  if (!meLoader) {
+    throw new Error("You need to be signed in to perform this action")
+  }
+
+  // Verifying that the token is still valid
```
mzikherman, line 91: Wondering if/why the verification is needed?

src/schema/v2/reverseImageSearch.ts
```diff
@@ -0,0 +1,133 @@
...
+  GraphQLList,
+} from "graphql"
+import { GraphQLUpload } from "graphql-upload"
+import { ResolverContext } from "types/graphql"
+
+export const ReverseImageSearchResultMatchRect = new GraphQLObjectType({
+  name: "ReverseImageSearchResultMatchRect",
+  fields: () => ({
+    left: {
+      type: new GraphQLNonNull(GraphQLFloat),
```
mzikherman, line 18: liking the typing in this file!

src/schema/v2/reverseImageSearch.ts
```diff
@@ -0,0 +1,133 @@
...
+      resolve: ({ query_match_rect }) => query_match_rect,
+    },
+  }),
+})
+
+export const ReverseImageSearchResults = new GraphQLObjectType({
+  name: "ReverseImageSearchResults",
+  fields: {
+    results: {
+      type: new GraphQLList(ReverseImageSearchResult),
```
mzikherman, line 78: #minor should this be nullable (i'd imagine an empty result set would be an empty list vs a `null`)?

src/schema/v2/__tests__/reverseImageSearch.test.ts
```diff
@@ -0,0 +1,120 @@
+import { reverseImageSearchResolver } from "../reverseImageSearch"
+import { Readable } from "stream"
+
+const _ = {}
+
+describe("reverseImageSearchResolver", () => {
```
mzikherman, line 6: nice specs!

mzikherman (approved):
Overall pretty vanill-ish code(!), sweet!

Re: latency on `/api/v1/me`, the 900ms in the video is a lot.

Datadog shows 100-150ms https://app.datadoghq.com/apm/resource/gravity/rack.request/56114a75d452f587?query=env%3Aproduction%20service%3Agravity%20operation_name%3Arack.request%20resource_name%3A%22Api%3A%3AV1%3A%3AMeEndpoint%23%2F%22&env=production&index=apm-search&topGraphs=latency%3Alatency%2CbreakdownAs%3Aabsolute%2Cerrors%3Acount%2Chits%3Acount&start=1651864807088&end=1651879207088&paused=false , which at least feels better.

Wondering if we need to verify the token, if so I agree there might be a 'faster' way to do that, but it could be a staging red-herring. Based on the datadog information, verifying using `meLoader` would add 100-150ms of latency to the request (not sure if that's acceptable or not).

nickskalkin (approved):
looks great 👍

src/schema/v2/reverseImageSearch.ts
dimatretyak: I don't think that `results` should be nullable. We should return an empty array (as it happens now) if there are no matches

src/schema/v2/__tests__/reverseImageSearch.test.ts
dimatretyak: thank you!

src/schema/v2/reverseImageSearch.ts
dimatretyak: 🤘🤘🤘

dimatretyak:
> I wonder if Metaphysics could start verifying user tokens

As the simplest solution, we can decode JWT token without verifying its signature (for example, using [decodeUnverifiedJWT](https://github.com/artsy/metaphysics/blob/main/src/lib/decodeUnverifiedJWT.ts#L3)). In this way, we can check the basic info (for example, we can find out when the token expires). **But** it will be unsafe, because user can pass any JWT token with any payload information (for example, user can pass token that has not expired yet, but which has not been generated by one of the artsy services).

If we want to make this solution secure, we need to verify the token signature, but for this we need to know the `secret` value that was used when generating the token. To get this `secret`, we need to make a request to `gravity` side

kathytafel:
I was thinking about this yesterday (dangerous!), and just want to call out any wins for latency are good as the use case for this will be fairs that tend to maybe be in big concrete convention centers with unpredictable connectivity. It's gotten much better in the last few years so maybe this isn't a concern but just thought I'd put this somewhere. No blocking from me, happy to see progress.

mdole (approved):
@gkartalis and I just chatted about this - we think it should be merged as-is! We should definitely QA and observe the latency (including simulating slower connections) once it's on production to see what the real-world impact of the extra hop to Gravity is. But it seems like there's not an obvious way to securely reduce that latency. So let's ship it, see if it's a problem, and then iterate if we feel the additional latency of confirming the request is executed by a logged-in user is too severe.

Nice work @dimatretyak !

src/schema/v2/reverseImageSearch.ts
mzikherman: (in that case maybe `new GraphQLNonNull(new GraphQLList(...` is the right invocation for the typing)

PR #4030, opened 2022-05-04
feat: Add artworksForUser root level query

jonallured:
This PR adds a new root query that can be used like so:

```gql
query ArtworksForUserAccessToken {
  artworksForUser {
    edges {
      node {
        title
      }
    }
  }
}
```

Or:

```gql
query ArtworksForUserSignedAppToken {
  artworksForUser(userId:"587aef0ea09a67787600115b") {
    edges {
      node {
        title
      }
    }
  }
}
```

Note that attempting to use it without either a user token or a signed app token will result in an error.

In terms of what this is doing - it moves the `newWorksByInterestingArtists` query up out of the `me` namespace and allows for two kinds of graphql loaders to hit vortex:

* one that uses the existing access token
* one that falls back to the app token being sent in

Note: depends on this one https://github.com/artsy/vortex/pull/446.

/cc @artsy/grow-devs

### DIFF
--- a/src/lib/loaders/loaders_without_authentication/vortex.ts
+++ b/src/lib/loaders/loaders_without_authentication/vortex.ts
@@ -1,0 +1,30 @@
+import factories from "../api"
+
+interface LoaderArgs {
+  query: string
+  variables?: any
+}
+
+export default (opts) => {
+  const { vortexLoaderWithAuthenticationFactory } = factories(opts)
+
+  const setup = (appToken) => {
+    return vortexLoaderWithAuthenticationFactory(() =>
+      Promise.resolve(appToken)
+    )
+  }
+
+  return {
+    vortexGraphqlLoaderFactory: (appToken) => {
+      return ({ query, variables }: LoaderArgs) => {
+        return setup(appToken)(
+          "/graphql",
+          { query, variables: JSON.stringify(variables) },
+          {
+            method: "POST",
+          }
+        )
+      }
+    },
+  }
+}
--- a/src/schema/v2/artworksForUser.ts
+++ b/src/schema/v2/artworksForUser.ts
@@ -11,30 +11,17 @@ const MAX_ARTISTS = 50
 const MAX_ARTWORKS = 100
 const MIN_AFFINITY_SCORE = 0.5
 
-export const artworksForUser: GraphQLFieldConfig<void, ResolverContext> = {
-  description: "A connection of artworks for a user.",
-  type: artworkConnection.connectionType,
-  args: pageable({
-    page: { type: GraphQLInt },
-    userId: { type: GraphQLString },
-  }),
-  resolve: async (
-    _root,
-    args: CursorPageable,
-    {
-      vortexGraphqlLoader,
-      vortexGraphqlLoaderFactory,
-      artworksLoader,
-      appToken,
-    }
-  ) => {
-    if (!artworksLoader) return
+export const getArtistAffinities = async (
+  args: CursorPageable,
+  context: ResolverContext
+): Promise<string[]> => {
+  const { appToken, vortexGraphqlLoader, vortexGraphqlLoaderFactory } = context
 
-    const graphqlLoader =
-      vortexGraphqlLoader || vortexGraphqlLoaderFactory(appToken)
+  const graphqlLoader =
+    vortexGraphqlLoader || vortexGraphqlLoaderFactory(appToken)
 
-    const vortexResult = await graphqlLoader({
-      query: gql`
+  const vortexResult = await graphqlLoader({
+    query: gql`
         query artistAffinitiesQuery {
           artistAffinities(
             first: ${MAX_ARTISTS}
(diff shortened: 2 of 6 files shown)


src/lib/loaders/loaders_without_authentication/vortex.ts
```diff
@@ -0,0 +1,30 @@
...
+
+interface LoaderArgs {
+  query: string
+  variables?: any
+}
+
+export default (opts) => {
+  const { vortexLoaderWithAuthenticationFactory } = factories(opts)
+
+  const partial = (appToken) => {
```
damassi, line 11: `setup`?

src/schema/v2/artworksForUser.ts
```diff
@@ -0,0 +1,129 @@
...
+        })
+      }
+
+      // TODO: get count from artworks loader to optimize pagination
+      const count = artworks.length === 0 ? 0 : MAX_ARTWORKS
+
+      return {
+        totalCount: count,
+        pageCursors: createPageCursors({ ...args, page, size }, count),
+        ...connectionFromArraySlice(artworks, args, {
```
damassi, line 76: all of this connection stuff can be simplified using our connection helper patterns: https://github.com/artsy/metaphysics/blob/main/docs/pagination.md

src/lib/loaders/loaders_without_authentication/vortex.ts
jonallured: 👍

src/schema/v2/artworksForUser.ts
jonallured: I just copy/pasted this part from the original query here:

https://github.com/artsy/metaphysics/blob/main/src/schema/v2/me/newWorksByInterestingArtists.ts#L70-L78

src/schema/v2/artworksForUser.ts
```diff
@@ -0,0 +1,86 @@
...
+    const count = artworks.length === 0 ? 0 : MAX_ARTWORKS
+
+    return {
+      totalCount: count,
+      pageCursors: createPageCursors({ ...args, page, size }, count),
+      ...connectionFromArraySlice(artworks, args, {
+        arrayLength: count,
+        sliceStart: offset ?? 0,
+      }),
+    }
```
damassi, line 84: Can we leverage our pagination helpers here (which will simplify all of this connection code)? https://github.com/artsy/metaphysics/blob/main/docs/pagination.md

damassi (commented):
There's a good amount of tricky logic stuff in here. Could we add some tests? Happy to merge and come back to this to unblock, but we should def follow up.

src/schema/v2/artworksForUser.ts
damassi: We should update to use pagination helpers! it will make this code easier to read and simpler and in line with new patterns.

dzucconi:
I kind of wonder if `me` could accept a `token` argument? (Do these tokens provide a general kind of auth or is it scoped to this specific field)

src/schema/v2/artworksForUser.ts
jonallured: Is this what you had in mind @damassi? [857af3e](https://github.com/artsy/metaphysics/pull/4030/commits/857af3ec288275b1b96aa9f175efd0df0082096c)

jonallured:
> I kind of wonder if me could accept a token argument? (Do these tokens provide a general kind of auth or is it scoped to this specific field)

It's a good question @dzucconi - the way I generated this token is like so:

```ruby
braze_staging = ClientApplication.find_by(name: "Braze Staging")
vortex_staging = ClientApplication.find_by(name: "vortex-staging")
expires_in = 10.hours.from_now
app_token = ApplicationTrust.create_for_token_authentication(vortex_staging, { expires_in: expires_in, subject_application: braze_staging})
```

So what this means is that the token I'm suppling in the case of Braze would be using it as the subject and vortex as the audience app. I don't know if that answers your question but that's what I got.

jonallured:
@damassi do you think this looks right? [5345a08](https://github.com/artsy/metaphysics/pull/4030/commits/5345a08b1e201cebd9a835f6d69b818a4658f2f7)? Without these fallbacks and the headers true bit I was not always getting the headers. But when I add that header true bit then all sorts of error pop up:

https://app.circleci.com/pipelines/github/artsy/metaphysics/12528/workflows/a1299240-51fe-47a9-bfcb-b44980c43391/jobs/26038

damassi:
Ah @jonallured yeah, i didn't realize we'd need to change the artistLoader gravity invocation in that way, which would reasonably impact each of the consumers since the return is different. In that case, no problem punting on this. (yay to typescript)

jonallured:
Ok this morning I reset back to my original diff and force pushed. I'm happy to return to this pagination stuff but I feel like I need to land this so that I can unblock the email template and web page streams of work.

So! Reset moment @damassi - how does this PR look now? haha

damassi (approved):
Thank you for adding a test! I'll leave [this thread](https://artsy.slack.com/archives/C01ADJNCS5D/p1652218720104729?thread_ts=1652218488.528239&cid=C01ADJNCS5D) to ponder, but feel free to merge at your discretion 👍

olerichter00:
I wonder how the new query `artworksForUser` relates to the existing recommendation queries `artworkRecommendations`, `artistRecommendations` and `newWorksByInterestingArtist`. Is the new query just moving `newWorksByInterestingArtist` out of the `me` namespace or will there be more changes as the name change indicates?

PR #4031, opened 2022-05-05
feat: Launch Gravity backed collections

mdole:
This PR makes the default value of ENABLE_GRAVITY_BACKED_COLLECTIONS `true`, so all clients will receive the updated schema and requests will flow to Gravity by default.

If we run into a problem after launch, we can still route requests back to KAWS by flipping the feature flag to `false` using `hokusai production env set ENABLE_GRAVITY_BACKED_COLLECTIONS=false && hokusai production refresh`


src/config.ts
```diff
@@ -161,7 +161,7 @@ export default {
   ENABLE_GEOLOCATION_LOGGING: ENABLE_GEOLOCATION_LOGGING === "true",
   ENABLE_GRAPHQL_UPLOAD: ENABLE_GRAPHQL_UPLOAD === "true",
   ENABLE_GRAVITY_MARKETING_COLLECTIONS:
-    ENABLE_GRAVITY_MARKETING_COLLECTIONS === "true",
+    ENABLE_GRAVITY_MARKETING_COLLECTIONS === "true" ?? true,
```
dblandin: **suggestion**: How about `!== "false"` as the updated condition?

mdole: hah, that works too! I overthought it

nickskalkin (approved):
👍

PR #4042, opened 2022-05-11
Fix: address bugs with isHangable [FX-3950]

mdole:
A gallery partner [asked](https://artsy.slack.com/archives/C7LEJU1QU/p1651828938959449) why one of their works did not have the "view in room" button on the desktop artwork page.

The [only thing](https://github.com/artsy/force/blob/6609be28a4657e9d65f435b74c7de59edbe4fcb0/src/v2/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx#L165) that determines whether we show or hide that button in the Force UI is a computed field `isHangable` in Metaphysics.

Looking into how `isHangable` is computed, I noticed that we were comparing the wrong category names, so we were effectively never making use of a "is2DCategory" and "is3DCategory" check. More digging also uncovered some bugs with how we parsed depth (see below for breakdown of what was happening and what's changing), and with how our specs were set up. I worked on splitting out the specs so that we had more test coverage, and on correcting the surprising behavior.

~In my opinion, this logic could probably be simplified to simply be "This work isHangable if it is not one of these 3D categories that we are confident are never hangable, and it has reasonable height, width, and depth" (basically just combining the existing `isTooBig` and `isTwoDimensional` functions). However, based on the conversation in [this ticket](https://artsyproduct.atlassian.net/browse/NX-2719), the use of category as part of the criteria is quite intentional, and I've already spent enough time on this bug that I don't feel like opening a whole 'nother can of worms 🙃~ Edit: I've come around on this a bit. Seems like it's at least helpful to not require or check `depth` for a lot of the "we're pretty sure this is 2D" categories, since we don't always have perfect information input by partners on artworks. In a perfect world, we'd have exact dimensions and could just rely on those...but requirements in CMS have been updated many times over the years, and we don't always have depth information (or it's wrong because we were forcing partners to enter depth and they would put things like `10cm` for a paper-thin work, just to have _something_ in there). So, seems good to still have the 2D/3D category split for now, though it is still confusing. 

### Criteria for a work to display the "view in room" button, before this PR:
- It did not matter what category the artwork was in because we were comparing the wrong category string, so we treated all artwork categories as if they were neither 2D or 3D categories
- Thus the criteria was:
  - Must have height > 0 and < 50ft/15.23m
  - Must have width > 0 and < 50ft/15.23m
  - Must not have diameter
  - Depth is either **null** (was never entered; I don't think we supported this intentionally but we were indeed supporting it) or must be less than 3 inches or centimeters (Note: we did not differentiate between the two, so a work with metric set to `cm` and `depth: 4` would have failed, while a work with metric `in` and `depth: 2` would have passed even though 2 inches is ~5 cm)
    - Edit: just [came across](https://artsy.slack.com/archives/C02K5KBPG9W/p1651873273193649?thread_ts=1651873072.176499&cid=C02K5KBPG9W) one other interesting `depth` bug: works like [this one](https://www.artsy.net/artwork/macauley-norman-the-great-bambino) that have depth set to something like `7/8in` end up not being eligible for View in Room because `parseFloat("7/8")` is `7` (it hits the / and gives up). This is also solved in this PR by using `depth_cm` instead of just `depth`.

### Criteria for a work to display the "view in room" button, after this PR:
- If artwork is in one of the following "3D categories", we never show the "View in Room" button:
  - "Architecture"
  - "Books and Portfolios"
  - "Design/Decorative Art"
  - "Fashion Design and Wearable Art"
  - "Installation"
  - "Jewelry"
  - "NFT"
  - "Performance Art"
  - "Video/Film/Animation"
- If artwork is in one of the following "2D categories", we will show the "View in Room" button as long as we have a width and height OR a diameter (to cover round + flat works like [this one](https://www.artsy.net/artwork/takashi-murakami-flower-parent-and-child-rum-pum-pum-2)) and the work is no larger than 50ft/15.23m in any dimension. **Note that this means we ignore depth for these categories**! If a partner lists a "Painting" with a depth of 50 cm, we would still allow it to be used with View in Room:
  - "Drawing, Collage or other Work on Paper"
  - "Painting"
  - "Photography"
  - "Posters"
  - "Print"
- If the artwork is in neither of those groups of categories (which means its category is `"Mixed Media", "Ephemera or Merchandise", "Sculpture", "Reproduction", "Textile arts", or "Other"`), our criteria are as follows:
  - width > 0 and < 50ft/15.23m
  - height > 0 and < 50ft/15.23m
  - depth > 0 and <= ~4in/10 cm. Note that this means that **we would no longer support null values, so any works (that are not in one of our 2D categories above) without a depth set will be excluded from "View in Room"**
  - Must not have diameter (it must be null or 0)

~Assigning myself to this PR because I want to make sure we're good with the above changes before shipping!~

Jira ticket: https://artsyproduct.atlassian.net/browse/FX-3950

### DIFF
--- a/src/schema/v2/artwork/__tests__/utilities.test.ts
+++ b/src/schema/v2/artwork/__tests__/utilities.test.ts
@@ -1,0 +1,149 @@
+import { Artwork } from "types/runtime/gravity"
+import { isTooBig, isTwoDimensional } from "../utilities"
+
+describe("isTwoDimensional", () => {
+  let artwork: Artwork
+
+  beforeEach(() => {
+    artwork = {
+      ...artwork,
+      width: null,
+      height: null,
+      diameter: null,
+      depth: null,
+      metric: null,
+      width_cm: null,
+      height_cm: null,
+      depth_cm: null,
+      diameter_cm: null,
+    }
+  })
+
+  describe("artwork has diameter", () => {
+    it("returns false", () => {
+      artwork.diameter_cm = 1
+      artwork.width_cm = 1
+      artwork.height_cm = 1
+      artwork.depth_cm = 1
+
+      expect(isTwoDimensional(artwork)).toBe(false)
+    })
+  })
+
+  describe("artwork has null depth", () => {
+    it("returns false", () => {
+      artwork.width_cm = 1
+      artwork.height_cm = 1
+      artwork.depth_cm = null
+
+      expect(isTwoDimensional(artwork)).toBe(false)
+    })
+  })
+
+  describe("artwork is too deep", () => {
+    it("returns false", () => {
+      artwork.width_cm = 1
+      artwork.height_cm = 1
+      artwork.depth_cm = 30
+
+      expect(isTwoDimensional(artwork)).toBe(false)
+    })
+  })
+
+  describe("artwork has no width", () => {
+    it("returns false", () => {
+      artwork.height_cm = 1
+      artwork.depth_cm = 1
+
+      expect(isTwoDimensional(artwork)).toBe(false)
+    })
+  })
+
+  describe("artwork has no height", () => {
+    it("returns false", () => {
+      artwork.width_cm = 1
+      artwork.depth_cm = 1
+
+      expect(isTwoDimensional(artwork)).toBe(false)
+    })
+  })
+
+  describe("artwork is two-dimensional", () => {
+    it("returns true", () => {
+      artwork.width_cm = 1
+      artwork.height_cm = 1
+      artwork.depth_cm = 1
+      expect(isTwoDimensional(artwork)).toBe(true)
+    })
+  })
+})
+
+describe("isTooBig", () => {
+  let artwork: Artwork
+
+  beforeEach(() => {
+    artwork = {
+      ...artwork,
+      width: null,
+      height: null,
+      metric: null,
+    }
+  })
+
+  describe("artwork has null width", () => {
+    it("returns true", () => {
+      artwork.height = "1"
+      artwork.metric = "cm"
+
+      expect(isTooBig(artwork)).toBe(true)
+    })
+  })
+
+  describe("artwork has null height", () => {
+    it("returns true", () => {
+      artwork.width = "1"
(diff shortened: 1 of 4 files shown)


src/schema/v2/artwork/__tests__/utilities.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      metric: null,
+      width_cm: null,
+      height_cm: null,
+      depth_cm: null,
+      diameter_cm: null,
+    }
+  })
+
+  describe("artwork has diameter", () => {
+    it("returns false", () => {
```
anandaroop, line 23: Just to confirm: a work in a 2D category doesn't _automatically_ get rejected if it has a diameter, right? 

(Many legit cases like that e.g. https://www.artsy.net/artwork/takashi-murakami-flower-parent-and-child-rum-pum-pum-2)

anandaroop (dismissed):
Wow, thanks for working through this. LGTM. Just one lil confirmation below...

nickskalkin (approved):
looks good to me 👍

src/schema/v2/artwork/__tests__/utilities.test.ts
mdole: Yes correct! We [skip](https://github.com/artsy/metaphysics/blob/c744241bdd4d59b4b0700716807e19144a10b4fd/src/schema/v2/artwork/index.ts#L617-L618) the `isTwoDimensional` check for works in one of our 2D categories

mdole: aha, BUT: you did just make me realize there's something I missed. We still do the `isTooBig` check for 2D works, and that check currently fails unless a work has width and height. I'll add specs and update the code to check for width/height OR diameter 👍

mdole: Addressed in https://github.com/artsy/metaphysics/pull/4042/commits/b9b54690f657f1adab17456c08f42d7056125f63

anandaroop (approved):
🚢

PR #4043, opened 2022-05-11
GRO-1009: Switch from FreeGeoIp to IpBase

The-Beez-Kneez:
This PR is in support of the following ticket [GRO-1009].

We are switching to IpBase from FreeGeoIp in order to maintain our GDPR compliant sign up flow, and because we have exceeded the limit on FreeGeoIp. FreeGeoIp was failing without sending us any notifications so it was determined a paid service would be worth the expense.

[GRO-1009]: https://artsyproduct.atlassian.net/browse/GRO-1009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jonallured:
ooooops! looks like some unrelated changes made it in!

The-Beez-Kneez:
I will post here once I upgrade our subscription!

PR #4051, opened 2022-05-13
feat: unstitch KAWS [FX-3937]

mdole:
After [shipping](https://github.com/artsy/metaphysics/pull/4031) the change to resolve collections from Gravity, we can fully remove KAWS from Metaphysics!

We've confirmed in DataDog that KAWS is only receiving health checks and no other traffic. Should be safe to remove from Metaphysics.
Screen Shot 2022-05-13 at 11 28 01 AM
Screen Shot 2022-05-13 at 11 28 20 AM

Jira ticket: https://artsyproduct.atlassian.net/browse/FX-3937


damassi:
🥂

PR #4065, opened 2022-05-18
feat: add additional payment methods 

kiryl-zubarau:
This PR adds a list of additional payment methods for the order query. 
JIRA - [TX-331]

Screenshot from 2022-05-20 14-01-17

[TX-331]: https://artsyproduct.atlassian.net/browse/TX-331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/partner.ts
+++ b/src/schema/v2/partner.ts
@@ -749,11 +749,6 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLBoolean,
         resolve: ({ verified_seller }) => verified_seller,
       },
-      wireTransferEnabled: {
-        description: "Indicates the partner supports wire transfer",
-        type: GraphQLNonNull(GraphQLBoolean),
-        resolve: ({ wire_transfer_enabled }) => wire_transfer_enabled ?? false,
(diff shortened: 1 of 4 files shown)


AlicanAkyuz (dismissed):
lgtm! 

holding the merge until exchange and kinetic PRs are merged.

src/schema/v2/partner.ts
```diff
@@ -749,6 +749,11 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLBoolean,
         resolve: ({ verified_seller }) => verified_seller,
       },
+      wireTransferEnabled: {
```
starsirius, line 752: This might be not necessary for the Vibrations work because it directly consumes Gravity through Kinetic and hasn't integrated with Metaphysics yet.

For the checkout UI on the buyer side to get applicable payment methods, I wonder if we can expose something directly on an order itself.

iskounen (approved):

PR #4066, opened 2022-05-18
feat: Support backfill for artwork recs GRO-1033

jonallured:
This PR updates the `artworksForUser` query with a flag called `includeBackfill` to add support for showing backfill when a user does not have enough recs. My approach was to extract out a function for the affinity artworks and then write a new one for the backfill artworks. I was able to test these functions in jest for unit coverage but technically I do not have an integration test - maybe I should? I'm thinking about a test that actually sends this query and ensures all the pieces are connected correctly.

I have a couple questions that I'll ask inline.

https://artsyproduct.atlassian.net/browse/GRO-1033

/cc @artsy/grow-devs


src/schema/v2/artworksForUser.ts
```diff
@@ -47,33 +52,83 @@ export const getArtistAffinities = async (
...
+
+  return body.slice(0, remainingSize)
+}
+
 export const artworksForUser: GraphQLFieldConfig<void, ResolverContext> = {
   description: "A connection of artworks for a user.",
   type: artworkConnection.connectionType,
   args: pageable({
+    // how do i make this not null with a false default?
+    includeBackfill: { type: GraphQLBoolean },
```
jonallured, line 106: I can't remember and couldn't easily search for how we are supposed to do this - what I want is a non-null field that defaults to false.

src/schema/v2/artworksForUser.ts
```diff
@@ -47,33 +52,83 @@ export const getArtistAffinities = async (
...
+  context: ResolverContext
+): Promise<any[]> => {
+  if (!includeBackfill || remainingSize < 1) return []
+
+  const { artworksLoader, setItemsLoader } = context
+
+  // should this be injected in the env? i could also send yet another api call
+  // to get this id via the key of artwork-backfill. could also consider
+  // caching this...
+  const backfillSetId = "SHHH"
```
jonallured, line 90: I'm interested in ideas here. What I really wish I could do is pass `artwork-backfill` like a slug but the Gravity API does not support this. I guess I should inject this using ENV vars, right?

damassi: It feels weird to pass things along as an ENV var, but we do do it elsewhere. I think if anything we should update the  gravity endpoint to accept a new optional arg (if it makes sense) first.

src/schema/v2/artworksForUser.ts
damassi: You can do `new GraphQLNonNull(GraphQLBoolean)` - but you don't need a default because by explicitly stating that its non-null, one can't send a query without a value, it will error out at the graphql level.

damassi (dismissed):
Looks great @jonallured. As far as "integration tests" for the query goes, I shall again point to [this thread](https://github.com/artsy/metaphysics/pull/4030#pullrequestreview-968440154) :wink:

src/schema/v2/artworksForUser.ts
jonallured: Cool updated!

src/schema/v2/artworksForUser.ts
jonallured: More chatter here but I ended up going with an ENV var:

https://artsy.slack.com/archives/CP9P4KR35/p1652907794787929

jonallured: Update: I did not go with an ENV var haha. I was able to figure out that I could call the setsLoader with the key of `artwork-backfill` and get the id. Those two calls are not authenticated so that means they are cached. 😎

PR #4068, opened 2022-05-19
fix(NX-3138): add conversation to CommerceBuyOrder in stitching

tam-kis:
Solves: [NX-3138](https://artsyproduct.atlassian.net/browse/NX-3138)

Added Conversation to CommerceBuyOrder that was causing issues for the "Purchase" cta redirection in Conversations.


araujobarret (approved):
Looks good, thanks for the quick action 🍪

PR #4073, opened 2022-05-20
fix: Allow My Collection artworks without artist

olerichter00:
[Slack Thread](https://artsy.slack.com/archives/C01B2P6LJUU/p1652404071482149)

## Description

This change prevents the My Collection connection query from failing if the list of artworks contains an artwork without an artist. Since we do not validate the presence of artists, this case is to be expected.


olerichter00:
Merging this fix to move forward in Eigen.

PR #4075, opened 2022-05-20
chore: check schema drift on CI

starsirius:
Surfaced (but not directly related to) this [thread](https://artsy.slack.com/archives/C02BC3HEJ/p1652989950505809), we currently use a [git pre-commit hook](https://github.com/artsy/metaphysics/blob/21698481018fa8e3b4d7f2df0daf8d8f2d308f72/package.json#L166) to update the schema, but we don't have a corresponding check/block on CI. It's possible the hook is skipped and outdated schema lands in the main branch.

This introduces a new schema check on CI for PRs, and blocks merge when schema is not up-to-date.

Screen Shot 2022-05-20 at 10 07 11 AM

### DIFF
--- a/scripts/ensure-schema-update.sh
+++ b/scripts/ensure-schema-update.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+set -e
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NO_COLOR='\033[0m'
+
+MSG_CLEAN="\n${GREEN}Schema is up-to-date.${NO_COLOR}"
+MSG_DRIFT="\n${RED}Schema is outdated. You might want to run \`yarn dump:staging\` and commit the changes.${NO_COLOR}"
+
+yarn dump:staging
+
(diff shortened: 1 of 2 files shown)


scripts/ensure-schema-update.sh
```diff
@@ -0,0 +1,18 @@
...
+set -e
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NO_COLOR='\033[0m'
+
+MSG_CLEAN="\n${GREEN}Schema is up-to-date.${NO_COLOR}"
+MSG_DRIFT="\n${RED}Schema is outdated. You might want to run \`yarn dump:staging\` and commit the changes.${NO_COLOR}"
+
+yarn dump:staging
```
starsirius, line 12: Simply dump the schema and check if there is a change.

joeyAghion (approved):
This might also be an opportunity to nudge whichever contributor was pushing without the necessary hooks to update their setup.

starsirius:
Merging this and giving it a try!

PR #4080, opened 2022-05-23
feat: Switch query that powers artworks for user

jonallured:
This one relies on https://github.com/artsy/vortex/pull/461 and isn't quite complete but wanted to round out the effort by starting it so as the pieces fall into place I can work through all the things.

/cc @artsy/grow-devs


jonallured:
Ok this is ready! The key change here is that rather than calling `artistAffinities` this PR updates the logic to call `newForYouRecommendations`. This puts Vortex in the driver's seat and is now where the recs are coming from. This allowed us to implement artist-level ranking so that the results are one artwork per artist.

jonallured:
Oops, forgot to erase some of the new vortex schema that I checked in.

damassi (approved):
LGTM 👍

PR #4084, opened 2022-05-24
feat: add has market price insights flag to the artwork model

MounirDhahri:
This PR adds `hasMarketPriceInsights` to the artwork model.

Screenshot 2022-05-24 at 16 54 32


dariakoko (approved):
Thanks Mounir!
I've been working on it too, but you've been faster 😅

MrSltun (approved):
Thank you Mounir!  🙌  🌟 
but I have a question, should we do something about the schema warning above 👆 ? maybe merge the main branch and sync the schema?

MounirDhahri:
@MrSltun no, I don't think we should be doing anything about it. I actually believe we should probably delete this warning from peril and only keep it for merge PRs

PR #4086, opened 2022-05-24
WIP: order.paymentMethod union field with types for bank account and wire tranfer

erikdstock:
More to do here to make it all work. Also may need to be more specific whether a bank account is US or something else. Also don't know if we need an empty wire transfer type.

The current state of the PR allows a query like so (to replace our `order#credit_card` field:
image

cc @rquartararo


src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -185,6 +185,55 @@ export const exchangeStitchingEnvironment = ({
...
+          fieldName: "creditCard",
+          args: {
+            id: creditCardId,
+          },
+          context,
+          info,
+          transforms: exchangeSchema.transforms,
+        })
+        // note: us_bank_account with underscore *is* the current value from exchange
+      } else if (paymentMethod === "us_bank_account") {
```
erikdstock, line 214: Can we rename this string on exchange to be consistent with the space-cased strings for credit card and wire transfer?

src/schema/v2/bank_account.ts
```diff
@@ -0,0 +1,43 @@
...
+// fields: https://github.com/artsy/gravity/blob/main/db/schema.rb
+export const BankAccountType = new GraphQLObjectType<any, ResolverContext>({
+  name: "BankAccount",
+  fields: () => ({
+    ...InternalIDFields,
+    bankName: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: "Bank name",
+      resolve: ({ bank_name }) => bank_name,
+    },
```
erikdstock, line 19: Defined a minimal bank account type - we should be clear on which fields need to be added (unless it is all of them).

src/schema/v2/bank_account.ts
```diff
@@ -0,0 +1,43 @@
...
+  description: "A user's bank account",
+  args: {
+    id: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: "The ID of the bank account",
+    },
+  },
+  resolve: (_root, { id }, { bankAccountLoader }) =>
+    bankAccountLoader ? bankAccountLoader(id) : null,
+}
```
erikdstock, line 43: defined a minimal field configuration which will be added to our root v2 schema.

src/schema/v2/payment_method_union.ts
```diff
@@ -0,0 +1,27 @@
...
+  resolveType: (value: any) => {
+    if (Boolean(value["bank_name"])) {
+      return BankAccountType
+    } else if (value["_isManualPayment"]) {
+      return ManualWirePaymentType
+    } else {
+      return CreditCardType
+    }
+  },
+})
```
erikdstock, line 27: A union type with an empty type for manual wire transfer payments. perhaps there is a better way to do this.

src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -185,6 +185,55 @@ export const exchangeStitchingEnvironment = ({
...
+        creditCardId
+        bankAccountId
+        paymentMethod
+      }
+    `,
+
+    resolve: async (parent, _args, context, info) => {
+      const { creditCardId, bankAccountId, paymentMethod } = parent
+
+      if (paymentMethod === "credit card") {
```
erikdstock, line 200: And finally in this file we add the stitched and enriched payment method to the schema.

src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -185,6 +185,55 @@ export const exchangeStitchingEnvironment = ({
...
+          },
+          context,
+          info,
+          transforms: exchangeSchema.transforms,
+        })
+      } else if (paymentMethod === "wire transfer") {
+        return {
+          // PaymentMethodUnionType relies on this property
+          // not sure if this is the best way to model a manual payment TBH
+          _isManualPayment: true,
```
damassi, line 231: Why the underscore?

src/schema/v2/bank_account.ts
```diff
@@ -0,0 +1,43 @@
...
+export const BankAccount: GraphQLFieldConfig<void, ResolverContext> = {
+  type: BankAccountType,
+  description: "A user's bank account",
+  args: {
+    id: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: "The ID of the bank account",
+    },
+  },
+  resolve: (_root, { id }, { bankAccountLoader }) =>
```
damassi, line 41: This is nitpicky, but it's always good to avoid implicit returns unless it's like `(x, y) => x + y`. I suspect this resolver will in the future need to be inspected somehow, which would require unwinding the implicit return to add a console log or more code or a `debugger` statement.

src/schema/v2/payment_method_union.ts
```diff
@@ -0,0 +1,27 @@
...
+      resolve: () => true,
+    },
+  },
+})
+
+export const PaymentMethodUnionType = new GraphQLUnionType({
+  name: "PaymentMethodUnion",
+  types: [BankAccountType, CreditCardType, ManualWirePaymentType],
+  resolveType: (value: any) => {
+    if (Boolean(value["bank_name"])) {
```
damassi, line 19: `value.bank_name`

src/schema/v2/payment_method_union.ts
```diff
@@ -0,0 +1,27 @@
...
+      return BankAccountType
+    } else if (value["_isManualPayment"]) {
```
damassi, line 21: `value.isManualPayment`

src/schema/v2/bank_account.ts
erikdstock: ah yeah 100 percent agree

src/lib/stitching/exchange/v2/stitching.ts
erikdstock: my logic in the moment: in this file i am returning an object that will be used to identify the type as a manual wire transfer. in the other file i am using this internal property to identify the type, which happens to have a single field of the same name. Yes, i agree it doesn't make sense.

damassi: I know it seems silly, but maybe do `__isManualPayment`. `_foo` -- one underscore -- means "private variable", where's the double underscore (`__whatever`) is used for marking internal / otherwise clashing implementation details

erikdstock:
@rquartararo some questions i'd like your opinion on:
- I only exposed a few fields in the BankAccount type. do you know which we need for our frontend needs?
- I also called the one field `last4` because that is its backend name - CreditCard uses `lastDigits`. Maybe we would prefer the same name but it can work either way.
- What we should name the payment method string vs the payment method type

kiryl-zubarau:
Maybe name it just `payment` (union type) and keep `paymentMethod` as an enum value.

erikdstock:
@rquartararo 
> As for the payment method naming, more descriptive might be better -- so my vote is for paymentMethod and paymentMethodType.

I think this is my preference as well, but i'll have to mess around with it a bit - the `-Type` version sounds like the string to me, which means in this PR i would have to both _rename_ `paymentMethod` to `paymentMethodType` and then _create_ the new rich field with the same name. I'm not 100% sure this is possible but it probably is.

_schema.graphql
```diff
@@ -4465,6 +4465,9 @@ type CommerceBuyOrder implements CommerceOrder {
   ): CommerceLineItemConnection
   mode: CommerceOrderModeEnum
   orderHistory: [CommerceOrderEventUnion!]!
+
+  # Whether Artsy collects taxes (e.g., sales tax or VAT) on this order.
```
AlicanAkyuz, line 4469: **chore**: I might be missing something, but I wonder why this comment is made for `paymentMethod` in schemas.

src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -185,6 +185,60 @@ export const exchangeStitchingEnvironment = ({
...
+    fragment: gql`
+      fragment CommerceOrderPaymentMethod on CommerceOrder {
+        creditCardId
+        bankAccountId
+        paymentMethod #TODO: one of these fields should probably have a different name
+      }
+    `,
+
+    resolve: async (parent, _args, context, info) => {
+      const { creditCardId, bankAccountId, paymentMethod } = parent
```
starsirius, line 198: It seems like the Exchange `paymentMethod` hasn't been exposed and used. I wonder if it's possible to extend the `CommerceOrder` interface and just completely replace the Exchange `paymentMethod` string with our new type.

_schema.graphql
erikdstock: ah, i'd ignore this because i manually edited the schema based on work @fladson was doing in exchange, planning to rebase after #4098 was merged.

src/lib/stitching/exchange/v2/stitching.ts
erikdstock: That was my thought, but i wondered if we would want to keep a string representation so we don't have to query a bank account or credit card from gravity.

erikdstock: For now it is a new field - `paymentDevice`. @damassi, do you know if it is possible to simply replace the exchange string enum field?

_schemaV2.graphql
```diff
@@ -2858,6 +2858,23 @@ type BackupSecondFactors {
...
+  bankName: String!
+
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
+
+  # Last four characters of the account identifier
+  last4: String!
```
erikdstock, line 2875: I only included a minimal set of bank account data. This is what we currently have available from gravity v1: 
```
{
  "id": "13039038-68c9-44ef-973c-ab6f269646e0",
  "customer_account_id": "",
  "external_id": "pm_1L88BJGK3Gnpfa3OC421s4sv",
  "account_holder_name": "",
  "account_holder_type": "individual",
  "bank_name": "STRIPE TEST BANK",
  "country": "",
  "currency": "",
  "routing_number": "110000000",
  "last4": "6789",
  "fingerprint": "Qf0RvlKZ2z3hh89Z",
  "status": "",
  "created_at": "2022-06-07T19:35:16.203Z",
  "updated_at": "2022-06-07T19:35:16.203Z",
  "one_time_use": false,
  "user_id": "57da9676a09a6720300017cb",
  "account_type": "checking",
  "deactivated_at": null
}
```

_schemaV2.graphql
```diff
@@ -9331,6 +9351,10 @@ type LotStanding {
   saleArtwork: SaleArtwork
 }
 
+type ManualPayment {
+  isManualPayment: Boolean!
+}
```
erikdstock, line 9356: This is a new, manually-defined type that only exists as a sentinel value. There is no manual payment information to add yet.

src/schema/v2/payment_device_union.ts
```diff
@@ -12,23 +12,12 @@ export const ManualPaymentType = new GraphQLObjectType({
...
 })
 
 export const PaymentDeviceUnionType = new GraphQLUnionType({
   name: "PaymentDeviceUnion",
   types: [BankAccountType, CreditCardType, ManualPaymentType],
-  resolveType: (value) => {
-    if (value.foo === 42) {
-      return ManualPaymentType
-    }
-  },
```
erikdstock: I was never able to get this function firing, even when explicitly returning something without a `__typename` in the stiched resolver (see https://github.com/artsy/metaphysics/pull/4086/commits/bd59fa135dfbc2118451b7a269f627945b7e4740#diff-c1775973b51cf561f4b812a9a296de0f15a8829b65a608ad42642cb67aabf1c9L201). Likewise for the `isManualPayment` field above. So i ultimately decided to remove this. I suppose stitched resolvers are expected to provide their own type name and fields

src/schema/v2/payment_device_union.ts
```diff
@@ -0,0 +1,23 @@
...
+} from "graphql"
+import { BankAccountType } from "schema/v2/bank_account"
+import { CreditCardType } from "schema/v2/credit_card"
+
+export const ManualPaymentType = new GraphQLObjectType({
+  name: "ManualPayment",
+  fields: () => ({
+    isManualPayment: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      resolve: () => true,
```
erikdstock, line 15: Do we need to have this field at all? I don't know if objects that only have a `__typename` are allowed (I seem to recall they weren't in the past but i may be confusing things). 

Alternately it could be `isWireTransfer: true` to allow for other types of manual payment in the future - thus giving us a _reason_ to have the field.

erikdstock (commented):
Ok this is ready for review and merge 🙈

_schemaV2.graphql
erikdstock: See my other comment [here](https://github.com/artsy/metaphysics/pull/4086#discussion_r891762542) - I wonder if the `ManualPayment` type should have a descriptor inside it for wire vs other future manual payment types. This would be more useful in the literal sense than a property that duplicates the type name.

starsirius: How about we split by the actual payment method and have `isManual` a property in it.
```GraphQL
type WireTransfer {
  isManual: Boolean!
}
```

This makes it clear for clients what exactly the payment method is and we can add more future manual payment methods.

src/lib/stitching/exchange/v2/stitching.ts
starsirius: Not sure if it's a better name, but how about `paymentMethodDetails`?

starsirius (dismissed):
I have some non-blocking comments on the type for wire transfer and the naming of the enriched payment method type. Looks great to me!! 🎉

narikazu (approved):
LGTM

erikdstock:
#squashongreen

PR #4089, opened 2022-05-25
chore: update exchange schema

fladson:
follow up work from https://github.com/artsy/exchange/pull/1171 and https://github.com/artsy/exchange/pull/1182


erikdstock:
> Input field 'bankAccountId' was removed from input object type 'CommerceSetPaymentInput'

Why remove `bankAccountId`?

fladson:
> > Input field 'bankAccountId' was removed from input object type 'CommerceSetPaymentInput'
> 
> Why remove `bankAccountId`?

Please ignore this PR for now, we should merge this one instead https://github.com/artsy/metaphysics/pull/4098/files until the deprecation is addressed.

kiryl-zubarau (dismissed):

src/schema/v1/ecommerce/types/enums/payment_method_enum.ts
```diff
@@ -0,0 +1,27 @@
+import { GraphQLEnumType } from "graphql"
+
+export const PaymentMethodEnum = new GraphQLEnumType({
```
kiryl-zubarau, line 3: Are we really need this enum?

fladson: Do you know another way to define it? I followed the project convention for enums.

starsirius: When Exchange schema is merged into Metaphysics, we [rename and prefix](https://github.com/artsy/metaphysics/blob/ca3ca865c173d956c551a60db9d70d371c75cd9e/src/lib/stitching/exchange/schema.ts#L27-L29) the types with `Commerce...`. So the Exchange payment method enum is [already in MP](https://github.com/artsy/metaphysics/blob/ca3ca865c173d956c551a60db9d70d371c75cd9e/_schemaV2.graphql#L4710-L4725) (for free!) as `CommercePaymentMethodEnum` and that's what we'll use for MP. So I think this addition is not needed.

fladson: good to know, but what about v1? I think that is needed for v1

starsirius: We want to deprecate v1 and I believe we have migrated all BNMO related code to use v2. So I think we can ignore updating v1 schema here (which can be useful speeding up the deprecation). If I'm wrong, we'll find out! 😛

fladson: Do you mean omitting `_schema` changes?

src/schema/v1/ecommerce/set_order_payment_mutation.ts
```diff
@@ -47,7 +47,8 @@ export const SetOrderPaymentMutation = mutationWithClientMutationId<
       mutation setOrderPayment($orderId: ID!, $creditCardId: String!) {
         ecommerceSetPayment(input: {
           id: $orderId,
-          creditCardId: $creditCardId,
+          paymentMethod: CREDIT_CARD,
+          paymentMethodId: $creditCardId,
```
starsirius, line 51: Updating the implementation of v1 so the tests pass with this breaking change. This is a good reminder that we should see where we are at with deprecating/retiring v1.

PR #4091, opened 2022-05-25
feat: Add pagination functionality to myCollectionConnection

olerichter00:
## Description

This change is needed to make pagination for My Collection work in Force. I just used the same helper we already [use for saved artworks](https://github.com/artsy/metaphysics/blob/0f90d50899db09d0ca63f584b0a5c3c01d438e2a/src/schema/v2/me/savedArtworks.ts#L76).

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -9557,6 +9557,8 @@ type Me implements Node {
     excludePurchasedArtworks: Boolean = false
     first: Int
     last: Int
+    page: Int
+    size: Int
     sort: MyCollectionArtworkSorts
   ): MyCollectionConnection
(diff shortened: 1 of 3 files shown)


MounirDhahri (commented):
@olerichter00 this looks good to me, I only have one question before approving. This won't break any existing functionality on Eigen right?

olerichter00:
Exactly, this should break pagination in Eigen. I'll double-check after this is deployed on staging 🤞

olerichter00:
My mistake. I meant: This should **not** break Eigen :)

MounirDhahri:
ah cool 👍

PR #4095, opened 2022-05-26
feat: Extract ipbase loader GRO-1047

jonallured:
This PR refactors how we make our api calls for ip lookup. We were using node fetch before which means we didn't get the advantages of the data loader pattern including caching!

So I followed the example of the Greenhouse loaders and now when a request for a given ip address comes through it'll be cached for...well it'll be cached. 😉 

https://artsyproduct.atlassian.net/browse/GRO-1047

/cc @artsy/grow-devs


damassi (dismissed):
This looks good! One follow-up would be ensuring that the memcached TTL is long enough for our needs (in this instance).

jonallured:
Ok after some pondering I was able to see that I could add an ENV var for ipbase so that the TTL on the cache can be really long. These things don't really change that much.

Set in both staging and production:

```
$ MS_IN_DAY=$(expr 60 \* 60 \* 24 \* 1000)
$ hokusai staging env set IPBASE_REQUEST_THROTTLE_MS=$MS_IN_DAY
$ hokusai production env set IPBASE_REQUEST_THROTTLE_MS=$MS_IN_DAY
```

damassi (approved):

PR #4100, opened 2022-05-31
chore(video): add mute query param to video urls

laurabeth:
Requested during initial video QA. Adding query to vimeo url to fetch muted videos.


damassi (approved):

dzucconi (approved):
Is fine. Alternatively you can expose this over an arg (which is what we'll want to do if this becomes an actual feature)

PR #4103, opened 2022-05-31
fix(requestlocation): includes id and cached fields

dzucconi:
OK this was a bit confusing but:
* Adds an `id` field that's just the base64 encoded IP address so that Relay can throw it in the cache store
* Exposes the `cached` field (and updates the corresponding type)

### DIFF
--- a/src/schema/v2/requestLocation.ts
+++ b/src/schema/v2/requestLocation.ts
@@ -1,11 +1,14 @@
 import {
   GraphQLFieldConfig,
+  GraphQLID,
   GraphQLNonNull,
   GraphQLObjectType,
   GraphQLString,
 } from "graphql"
 import { ResolverContext } from "types/graphql"
+import cached from "schema/v2/fields/cached"
 import config from "config"
+import { base64 } from "lib/base64"
 
 const logRateHeaders = (headers) => {
   const headerKeys = [...Object.keys(headers)]
@@ -20,6 +23,8 @@ const logRateHeaders = (headers) => {
 export const RequestLocationType = new GraphQLObjectType<any, ResolverContext>({
   name: "RequestLocation",
   fields: () => ({
+    cached,
+    id: { type: new GraphQLNonNull(GraphQLID) },
     country: { type: GraphQLString },
     countryCode: { type: GraphQLString },
   }),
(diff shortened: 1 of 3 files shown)


src/schema/v2/requestLocation.ts
```diff
@@ -34,14 +39,16 @@ export const RequestLocationField: GraphQLFieldConfig<void, ResolverContext> = {
     },
   },
   resolve: async (_root, args, { requestLocationLoader }) => {
-    const { body, headers } = await requestLocationLoader({ ip: args.ip })
+    const { body, headers, cached } = await requestLocationLoader({
```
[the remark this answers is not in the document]
dzucconi: I only know it was there because I had initially included it on all requests to ease debugging when I first built MP :)

https://github.com/artsy/metaphysics/blob/048107b32ec677e7922422f84d6dab74de00d49a/src/lib/cache.ts#L143-L149

src/schema/v2/requestLocation.ts
```diff
@@ -20,6 +23,8 @@ const logRateHeaders = (headers) => {
 export const RequestLocationType = new GraphQLObjectType<any, ResolverContext>({
   name: "RequestLocation",
   fields: () => ({
+    cached,
+    id: { type: new GraphQLNonNull(GraphQLID) },
```
damassi, line 27: @dzucconi - i just ran into this when using the fragment spread technique for updating a view post mutation and would have had no clue how to achieve had this PR not been opened the other day... How did you know to do this? Like are there some docs we could point at somewhere?

dzucconi: I guess there's this: https://relay.dev/docs/principles-and-architecture/thinking-in-graphql/#caching-a-graph

I actually thought that Relay was using `${_typename}:${id}` like Apollo but instead it relies on globally unique IDs. Which means we should actually use this instead of just base64 encoding the IP address: https://github.com/artsy/metaphysics/blob/0e7202ed5899e9dee465ddded88950138ee7e5f3/src/schema/v2/object_identification.ts#L166-L178

damassi: nice, ty 👍

PR #4134, opened 2022-06-10
feat(admin): Add unleash feature flag support 

damassi:
This adds support for our Feature Flag management service [Unleash](https://www.getunleash.io/) to Force. 

See the API docs here: 
- https://docs.getunleash.io/api/admin/feature-toggles-v2

Related: 
- https://github.com/artsy/forque/pull/206 

Update ENV vars:
- [x] Set `UNLEASH_ADMIN_TOKEN` on staging 
- [x] Set `UNLEASH_ADMIN_TOKEN` on prod  

## Query 

```graphql
{
  admin {
    featureFlags {
      name
      description
      createdAt(format: "MMM DD, YYYY")
      type
      stale
      impressionData
      environments {
        enabled
        name
      }
      variants {
        name
        stickiness
        weight
        weightType
      }
    }
  }
}
```

#### Create a Feature Flag 

```graphql
mutation {
  adminCreateFeatureFlag(input:{
    name:"my-feature-flag",
    description:"some description",
    impressionData:true,
    type:RELEASE,
    strategy:{
      strategyType: FLEXIBLE_ROLLOUT,
      rollOut:100
    }
    variants:[{
      name: "control",
      weight: 200
      weightType:VARIABLE
    }, {
      name: "experiment",
      weight:500,
      weightType:VARIABLE
    }]
  }) {
    featureFlag {
      name
      enabled
      description
      impressionData
      type
      variants {
        name
        stickiness
        weight
        weightType
      }
    }
  }
}
```

### Delete Feature Flag 

```graphql 
mutation {
  adminDeleteFeatureFlag(input: {name:"my-feature-flag"}) {
    success
  }
}
```

### Toggle Feature Flag 

```graphql 
mutation {
  adminToggleFeatureFlag(input: {
    name: "my-feature-flag", 
    enabled: true, 
    environment: PRODUCTION
  }) {
    featureFlags {
      name
      environments {
        enabled
        name
      }
      stale
      enabled
    }
  }
}
```

#### Update Feature Flag

> ⚠️ Updating has limited support and does not cover variants. (We currently link directly out to Unleash for edits.) 

```graphql
mutation {
  adminUpdateFeatureFlag(input:{
    name:"my-feature-flag", 
    description:"updated description"
  }) {
    featureFlag {
      name
    }
  }
}
```


src/lib/apis/fetch.ts
```diff
@@ -20,9 +20,26 @@ export const constructUrlAndParams = (method, url): URLAndRequestBodyParams => {
...
+        .every((key) => !Number.isNaN(key) && typeof key === "number")
+
+      if (isArray) {
+        const arrayParams = Object.entries(parsedParams).map(
+          ([_, entry]) => entry
+        )
+        body = arrayParams
+      } else {
+        body = parsedParams
+      }
```
damassi, line 40: Example of issue: 

```tsx
const dataToSend = ["a", "b"]
fetch(..., dataToSend) 
```

src/lib/loaders/loaders_with_authentication/unleash.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export const unleashLoaders = (accessToken, opts) => {
+  const accessTokenLoader = () => Promise.resolve(accessToken)
+  const { unleashLoaderWithAuthenticationFactory } = factories(opts)
+
+  const unleashLoader = unleashLoaderWithAuthenticationFactory(
+    accessTokenLoader
+  )
+
+  // TODO: Check for admin role on me before executing loader
```
damassi, line 11: Waiting for @joeyAghion to get back with a role to guard on.

src/schema/v2/admin/mutations/createFeatureFlagMutation.ts
```diff
@@ -0,0 +1,167 @@
...
+  GraphQLInt,
+  GraphQLEnumType,
+  GraphQLNonNull,
+} from "graphql"
+import { mutationWithClientMutationId } from "graphql-relay"
+import { omit } from "lodash"
+import { ResolverContext } from "types/graphql"
+import { FeatureFlags, FeatureFlagType } from "../featureFlags"
+
+export const FeatureFlagInputFields = {
```
damassi, line 15: Pulled this out for now as we can share it for Update (if we decide to add that directly into Forque; not sure its necessary since there's links out to Unleash)

src/lib/apis/fetch.ts
damassi: I can also pull this out into a separate PR.

src/schema/v2/admin/featureFlags.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export const FeatureFlagType = new GraphQLObjectType<any, ResolverContext>({
+  name: "FeatureFlag",
+  fields: {
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
+      resolve: (data) => {
+        return base64(data.name)
+      },
+    },
```
damassi, line 25: Relay [GlobalID cache support](https://relay.dev/graphql/objectidentification.htm)

dzucconi: Maybe augment the object with `id` in the resolver and then just use the global ID field? https://github.com/artsy/metaphysics/blob/ca3ca865c173d956c551a60db9d70d371c75cd9e/src/schema/v2/object_identification.ts#L166-L178

src/schema/v2/admin/featureFlags.ts
```diff
@@ -0,0 +1,131 @@
...
+          name: "FeatureFlagEnvironments",
+          fields: {
+            name: {
+              type: GraphQLString,
+            },
+            displayName: {
+              type: GraphQLString,
+            },
+            enabled: {
+              type: GraphQLBoolean,
```
dzucconi, line 41: Lists + their members should always be non-nullable.
Booleans should always be non-nullable.
And I'd check to see if name and displayName can be too?

src/schema/v2/admin/featureFlags.ts
```diff
@@ -0,0 +1,131 @@
...
+            },
+          },
+        })
+      ),
+    },
+    type: {
+      type: GraphQLString,
+    },
+    variants: {
+      type: new GraphQLList(
```
dzucconi, line 51: `new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(WhateverType)))`

src/schema/v2/admin/mutations/createFeatureFlagMutation.ts
```diff
@@ -0,0 +1,167 @@
...
+          },
+        },
+      })
+    ),
+  },
+}
+
+export const createFeatureFlagMutation = mutationWithClientMutationId<
+  any,
+  any,
```
dzucconi, line 104: I generally will throw types into https://quicktype.io/

dzucconi (dismissed):
LGTM; though I'd see what I could make non-nullable, throughout.

src/lib/apis/fetch.ts
```diff
@@ -20,9 +20,26 @@ export const constructUrlAndParams = (method, url): URLAndRequestBodyParams => {
   if (method === "PUT" || method === "POST" || method === "DELETE") {
     const [path, queryParams] = url.split("?")
     const parsedParams = parse(queryParams)
+    let body
 
     if (isExisty(parsedParams)) {
-      opts.body = parsedParams
+      // If the query params are formatted in a 0, 1, 2 way we know it's an array
+      // disguised as an object. Convert it back to an array.
+      // TODO: Extract this into own PR and write a test
```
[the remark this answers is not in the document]
damassi: Will triple check, but `stringify` will always convert to an object; the various settings deal with arrays nested within the object.

src/schema/v2/admin/featureFlags.ts
damassi: For real 😅 ?

dzucconi: Yeah: if it's an empty array; that's still technically a non-null array of non-null members. It should never be `null` or `[null]`.

src/lib/apis/fetch.ts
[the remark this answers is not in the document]
dzucconi: This sounds like a bug in qs then? What's the query string look like?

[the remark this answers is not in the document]
damassi: - Given `["foo", "bar"]
- `qs.stringify`, which [expects an object](https://github.com/ljharb/qs#stringifying), will transform this into `0=foo&1=bar`
- which then parses into `{ 0: "foo", 1: "bar" }`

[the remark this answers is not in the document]
damassi: Doesn't look like it, because this was clearly a bug! There's no way we could have gone from `0=foo&1=bar&2=baz` to a successful POST or PUT to an endpoint that is expecting an array.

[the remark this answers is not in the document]
dzucconi: Why would we pass a bare array as a query string (inherently an object representation), ever?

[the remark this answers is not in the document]
damassi: This is how we [get data back and forth](https://github.com/artsy/metaphysics/blob/main/src/lib/loaders/api/loader_interface.ts#L21-L23) between our api-constructed loader endpoints and our underlying `fetch` wrapper. Whether or not this is correct is another matter.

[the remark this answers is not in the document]
dzucconi: I guess I don't see how that would ever result in us serializing a bare array onto the query string or why we would have any services that would expect that? What bug is this actually fixing?

src/schema/v2/admin/mutations/createFeatureFlagMutation.ts
damassi: done

src/lib/apis/fetch.ts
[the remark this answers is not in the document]
dzucconi: But, in that case though it'd never be a query string since it's a PUT, it can just be the body.

damassi:
Going to merge and can follow up with any other feedback tomorrow 👍

PR #4151, opened 2022-06-14
feat: Add sorting by most recent insight update to My Collection artwork connection

olerichter00:
Addresses [CX-2503]

🔴 Depends on https://github.com/artsy/vortex/pull/473

## Description

In order to get the most recent price insight updates for My Collection artworks, this PR adds the argument `sortedByMostRecentInsightUpdates` to `myCollectionConnection`. If this argument is provided, the resolver fetches all (limited to 100) artworks from the user's collection and sorts them by the most recent price insight update (which comes from **Vortex**).

```graphql
{ 
  me {
    myCollectionConnection(first: 18, sortByLastAuctionResultDate: true) {
      edges {
        node {
          medium
          artist {
            name
          }
          marketPriceInsights {
            demandRank
	    annualValueSoldCents
	    annualLotsSold
            lastAuctionResultDate
          }
        }
      }
    }
  }
}
```

[CX-2503]: https://artsyproduct.atlassian.net/browse/CX-2503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/loadBatchPriceInsights.ts
+++ b/src/lib/loadBatchPriceInsights.ts
@@ -4,12 +4,23 @@ import { StaticPathLoader } from "./loaders/api/loader_interface"
 
 const MAX_MARKET_PRICE_INSIGHTS = 50
 
+type MarketPriceInsight = {
+  artistId: string
+  medium: string
+  demandRank: number
+  lastAuctionResultDate: string
+  annualLotsSold: number
+  annualValueSold: number
+}
+
 export const loadBatchPriceInsights = async (
   artistIDMediumTuples: { artistId: string; medium: string }[],
   vortexGraphQLLoader: ({ query, variables }) => StaticPathLoader<any>
 ) => {
   try {
-    const vortexResult = await vortexGraphQLLoader({
+    // TODO: Add lastAuctionResultDate
+
+    const vortexResult: any = await vortexGraphQLLoader({
       query: gql`
         query MarketPriceInsightsBatchQuery($artistIDMediumTuples: [ArtistIdMediumTupleType!]!) {
           marketPriceInsightsBatch(input: $artistIDMediumTuples, first: ${MAX_MARKET_PRICE_INSIGHTS}) {
@@ -19,6 +30,8 @@ export const loadBatchPriceInsights = async (
                 artistId
                 medium
                 demandRank
+                annualLotsSold
+                annualValueSoldCents
               }
             }
           }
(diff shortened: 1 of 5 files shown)


MrSltun (dismissed):
Thank you Ole! 🙌 
I only left a question about typing :)

src/lib/loadBatchPriceInsights.ts
```diff
@@ -4,12 +4,23 @@ import { StaticPathLoader } from "./loaders/api/loader_interface"
...
+
 export const loadBatchPriceInsights = async (
   artistIDMediumTuples: { artistId: string; medium: string }[],
   vortexGraphQLLoader: ({ query, variables }) => StaticPathLoader<any>
 ) => {
   try {
-    const vortexResult = await vortexGraphQLLoader({
+    // TODO: Add lastAuctionResultDate
+
+    const vortexResult: any = await vortexGraphQLLoader({
```
MrSltun, line 23: **question:** why do we type `vortexResult` as `any`?

MrSltun (commented):
forgot to click start review 🌚

MrSltun (approved):

src/schema/v2/me/myCollection.ts
```diff
@@ -70,6 +63,12 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
       description:
         "Exclude artworks that have been purchased on Artsy and automatically added to the collection.",
     },
+    sortByLastAuctionResultDate: {
+      type: GraphQLBoolean,
+      defaultValue: false,
+      description:
+        "Sort by most recent price insight updates and filter out artworks without insights.",
+    },
```
MounirDhahri, line 71: I think we should be probably changing this to use the same pattern for sorting as in other connections for a more future-proof approach. I suggest we update this to be an enum with the possible sorting methods. In this case, it will be probably just an enum of one value `LAST_AUCTION_RESULT_DATE_DESC`.

olerichter00: Very good idea! I kept it as a flag because it is actually doing two things (sorting and filtering (if we don't filter, we get issues in the frontend with pagination)) and there is just this one specific use case. 

I don't see us sorting this the other way around in the future. I'd either keep it like this or if you feel strong about this, use an enum with the types `DESC` and ASC` (because the name already says what will be sorted here).

MounirDhahri: cool, I understand your rational and it makes sense to me now. thanks for explaining

PR #4156, opened 2022-06-15
feat(artist dupes): wire up artist merger mutation to Gravity

anandaroop:
https://artsyproduct.atlassian.net/browse/PLATFORM-4225

This adds a `mergeArtist` mutation that sends the artist merger request up to Gravity and returns the response

- In the success case, this will be the artist record that was retained and updated

- In the error case, we implement the typical [data-or-errors union pattern](https://artsy.github.io/blog/2018/10/19/where-art-thou-my-error/). (For this to work, we will require a tweak to the `api/v1/artists/merge` endpoint in Gravity, which is in https://github.com/artsy/gravity/pull/15454)

## On success

Screen Shot 2022-06-15 at 3 41 36 PM

## On error

Screen Shot 2022-06-15 at 3 42 58 PM

### DIFF
--- a/src/schema/v2/artists/mergeArtistsMutation.ts
+++ b/src/schema/v2/artists/mergeArtistsMutation.ts
@@ -1,0 +1,95 @@
+import {
+  GraphQLList,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLString,
+  GraphQLUnionType,
+} from "graphql"
+import { mutationWithClientMutationId } from "graphql-relay"
+import {
+  formatGravityError,
+  GravityMutationErrorType,
+} from "lib/gravityErrorHandler"
+import { ResolverContext } from "types/graphql"
+import { ArtistType } from "../artist"
+
+const SuccessType = new GraphQLObjectType<any, ResolverContext>({
+  name: "MergeArtistsSuccess",
+  isTypeOf: (data) => data.id,
+  fields: () => ({
+    artist: {
+      type: ArtistType,
+      resolve: (artist) => artist,
+    },
+  }),
+})
+
+const FailureType = new GraphQLObjectType<any, ResolverContext>({
+  name: "MergeArtistsFailure",
+  isTypeOf: (data) => data._type === "GravityMutationError",
+  fields: () => ({
+    mutationError: {
+      type: GravityMutationErrorType,
+      resolve: (err) => err,
+    },
+  }),
+})
+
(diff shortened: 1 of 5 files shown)


src/schema/v2/artists/mergeArtistsMutation.ts
```diff
@@ -0,0 +1,95 @@
...
+import {
+  formatGravityError,
+  GravityMutationErrorType,
+} from "lib/gravityErrorHandler"
+import { ResolverContext } from "types/graphql"
+import { ArtistType } from "../artist"
+
+const SuccessType = new GraphQLObjectType<any, ResolverContext>({
+  name: "MergeArtistsSuccess",
+  isTypeOf: (data) => data.id,
```
ovasdi, line 18: **Question**: I wanted to ask regarding this usage as I am working on a similar PR and keep having issues with types in my tests. I see that in a number of _SuccessTypes_ we return the `.id` for `isTypeOf`. Do you happen to know why do we use `.id` here instead of `_type` as we do in the _FailureType_?

ovasdi (approved):
LGTM 🦖

src/schema/v2/artists/mergeArtistsMutation.ts
anandaroop: Good question. I sort of breezed over this as I tried to learn from existing mutations in MP.

But my understanding is that the strategy that `isTypeOf` uses basically just depends on the shape of the underlying data.

So in the error case `isTypeOf` will be handed an object that looks like:

```json
{
  "type": "error",
  "_type": "GravityMutationError",
  "message": "Artist Not Found",
}
```

So `(data) => data._type === "GravityMutationError"` serves as a simple test to identify this blob as an instance of a Gravity error.

Whereas in the success case it will be handed on object that looks like:

```json
{
  "id":  "abc123",
  "name": "Some Artist",
  "nationality": "Some Country",
  "birthday": 1942,
  "…": "…"
}
```

And the `(data) => data.id` test returns a truthy value, telling us that we were handed an instance of a database record, thus we presumably succeeded. 

Does that make sense?

ovasdi: It does. Thanks for the explanation.

mc-jones (approved):

src/schema/v2/artists/mergeArtistsMutation.ts
artsyjian: thanks @ovasdi and @anandaroop . i learned!

PR #4168, opened 2022-06-17
fix: Adjust averageSalePriceDisplayText calculation

olerichter00:
## Description

Adjust averageSalePriceDisplayText calculation.


olerichter00:
#squashongreen

PR #4175, opened 2022-06-21
feat(FX-4048): Add isReverseImageSearchEnabled field in show and fair queries

gkartalis:
## Description

This PR resolves [FX-4048](https://artsyproduct.atlassian.net/browse/FX-4048)

### Context
We created a new `isReverseImageSearchEnabled: boolean` field and included it in fair and show types.

For now the field will be `true` if the show or fair's slug is one of a comma-separated list of slugs stored in an envornment variable (eg REVERSE_IMAGE_SEARCH_ENABLED_FAIR_SLUGS=slug1,slug2 or REVERSE_IMAGE_SEARCH_ENABLED_SHOW_SLUGS=slug1,slug2)

Then we can manually add a fair or show slug to the environment variable with hokusai production env set REVERSE_IMAGE_SEARCH_ENABLED_FAIR_SLUGS=liste-art-fair-basel-2022, which we would only do after manually indexing the fair/show artworks to Tineye. (see [FX-4047](https://artsyproduct.atlassian.net/browse/FX-4047))

In the future, we can resolve that field using information from Gravity (something like tineye_index_threshold > 90 to check if at least 90% of the works in a show/fair were successfully indexed to TinEye).

### Examples

```graphql
     show(id: "amazing-show") {
       id
       isReverseImageSearchEnabled
     }

     fair(id: "amazing-fair") {
       id
       isReverseImageSearchEnabled
     }
```


dzucconi (approved):

src/schema/v2/fair.ts
```diff
@@ -192,6 +193,16 @@ export const FairType = new GraphQLObjectType<any, ResolverContext>({
           return now.isAfter(activeStart)
         },
       },
+      isReverseImageSearchEnabled: {
+        type: new GraphQLNonNull(GraphQLBoolean),
+        description: "Have we indexed this fair's artworks to tineye?",
+        resolve: ({ id }) => {
```
gkartalis, line 199: changed this from `slug` -> `id` because for some reason inside the resolver when we run the server locally (pointing to staging gravity) the slug is **ALWAYS** undefined. 🤔 

@dzucconi maybe you know why? Am I missing something? is it a historical thing that we're missing?

src/schema/v2/__tests__/fair.test.js
```diff
@@ -762,6 +763,69 @@ describe("Fair", () => {
...
+
+        const data = await runQuery(query, context)
+
+        expect(data).toEqual({
+          fair: {
+            isReverseImageSearchEnabled: false,
+          },
+        })
+      })
+    })
```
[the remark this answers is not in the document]
gkartalis: Added the extra tests for these cases for both fairs and shows! Let me know what you think!

src/schema/v2/fair.ts
dzucconi: Ah right. Gravity returns the slug under `id` and the internal ID under `_id`

PR #4178, opened 2022-06-22
fix: (PX-5094) account for artsy shipping logic in onlyShipsDomestically function

jpotts244:
Addresses this bug: https://artsyproduct.atlassian.net/browse/PX-5094

MP's `onlyShipsDomestically` field was only checking flat rate shipping fees to determine if an artwork could be shipping domestically vs internationally.

This was causing users to not be able to check out adding a new `NEW` address if the work was international artsy shipping enabled but had a domestic shipping fee cent value

The form would lock and not allow international addresses to be added by collectors:
Screen Shot 2022-06-22 at 12 50 30 PM

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -2574,50 +2574,204 @@ describe("Artwork type", () => {
         }
       }
     `
-    it("is true when domestic_shipping_fee_cents is present and international_shipping_fee_cents is null", () => {
-      artwork.domestic_shipping_fee_cents = 1000
-      artwork.international_shipping_fee_cents = null
-      return runQuery(query, context).then((data) => {
-        expect(data).toEqual({
-          artwork: {
-            onlyShipsDomestically: true,
-          },
+    describe("when artsy domestic shipping enabled", () => {
+      beforeEach(() => {
+        artwork.process_with_artsy_shipping_domestic = true
+      })
+
+      describe("when domestic_shipping_fee_cents and international_shipping_fee is null", () => {
+        beforeEach(() => {
+          artwork.domestic_shipping_fee_cents = null
+          artwork.international_shipping_fee_cents = null
+        })
+
+        it("returns true", () => {
+          return runQuery(query, context).then((data) => {
+            expect(data).toEqual({
+              artwork: {
+                onlyShipsDomestically: true,
+              },
+            })
+          })
+        })
+      })
+
+      describe("when domestic_shipping_fee_cents is present", () => {
+        beforeEach(() => {
+          artwork.domestic_shipping_fee_cents = 100
+          artwork.international_shipping_fee_cents = null
+        })
+
+        it("returns true", () => {
+          return runQuery(query, context).then((data) => {
+            expect(data).toEqual({
+              artwork: {
+                onlyShipsDomestically: true,
+              },
+            })
+          })
+        })
+      })
+
+      describe("when only international_shipping_fee_cents is present", () => {
+        beforeEach(() => {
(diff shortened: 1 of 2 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -893,8 +893,10 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         description: "Is this work only available for shipping domestically?",
         resolve: (artwork) => {
           return Boolean(
+            // should we factor in artwork.artsy_shipping_domestic ?? probably yeah?
             artwork.domestic_shipping_fee_cents &&
```
oxaudo, line 897: I believe this needs to be (artwork.domestic_shipping_fee_cents != null || artwork.artsy_shipping_domestic) && !(international_shipping_fee_cents !== null || artsy_shipping_international).

Or we can be even more explicit:
```
const domesticShippingDefined? = artwork.domestic_shipping_fee_cents != null || artwork.artsy_shipping_domestic
const internationalShippingDefined? = artwork.international_shipping_fee_cents != null || artwork.artsy_shipping_international
return domesticShippingDefined? && !internationalShippingDefined?
```


The point is we need to account for artsy shipping in both domestic and international and all the mix/match cases I believe.

One question is if no shipping defined at all - everything is nil/false - this method will return `false`. That is current and intended behaviour I think~

oxaudo (commented):
I think we need to check for both domestic and international Arta as well here @jpotts244

leamotta (dismissed):
that's some thorough testing! 🎖️

src/schema/v2/artwork/index.ts
jpotts244: > One question is if no shipping defined at all - everything is nil/false - this method will return false. That is current and intended behaviour I think~

Sounds like this is correct and inline with the previous behavior.

Addressed and added specs!!!

src/schema/v2/artwork/index.ts
```diff
@@ -892,9 +892,15 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLBoolean,
         description: "Is this work only available for shipping domestically?",
         resolve: (artwork) => {
+          const domesticShippingSupported =
+            artwork.domestic_shipping_fee_cents ||
```
oxaudo, line 896: Would the free domestic shipping (0) return true here as we want? I was not 100% sure if we need it to be `!=` null as well.

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -2574,50 +2574,186 @@ describe("Artwork type", () => {
...
+                  onlyShipsDomestically: false,
+                },
+              })
+            })
+          })
+        })
+
+        describe("when free shipping worldwide", () => {
+          beforeEach(() => {
+            artwork.domestic_shipping_fee_cents = 0
```
oxaudo, line 2744: So yea, I want to see a spec with `artwork.domestic_shipping_fee_cents = 0 and artwork.international_shipping_fee_cents = nil` return `true` here.

src/schema/v2/artwork/index.ts
jpotts244: excellent catch, adjusted

jpotts244: Will always forget this 😭 

0 truthy in ruby
```ruby
irb(main):002:0> !!0
=> true
```

0 falsy in JS
```js
> !!0
=> false
```

oxaudo: Impossible to remember as it has 0(true/false) logical sense! Basically random!

oxaudo:
Think this is what we want! Merging this:)

PR #4183, opened 2022-06-24
fix: sort figures the same way as images

mdole:
Because `figures` and `images` were ordered differently, we were getting some buggy behavior on Force where clicking an image would take you to a different image's deep zoom view.

This PR should fix the problem by introducing the same sorting and normalization to images in `figures` as in `images`.

[Slack context](https://artsy.slack.com/archives/C2TQ4PT8R/p1656063626023679)


dzucconi (approved):
Thanks much!

PR #4197, opened 2022-06-30
fix: Replace Continental Europe with European Union

leamotta:
Following up with the European Union shipping zone after the release of Artsy Shipping


jpotts244 (approved):
🏅

PR #4200, opened 2022-07-01
fix: Pass user ID to myCollectionInfo context

olerichter00:
## Description

We need to pass the user ID to the `myCollectionInfo` context to then pass it as an argument to the `collectedArtistsConnection` loader.

This avoids the following error:
Screenshot 2022-07-01 at 11 27 41


MrSltun (approved):
👏👏

PR #4205, opened 2022-07-04
feat: Filter artists' auction results with a sale end date in the future

olerichter00:
Addresses [CX-2654]

🔴 Depends on https://github.com/artsy/diffusion/pull/619

## Description

Filtering artists' auction results with a sale end date in the future for now. This hack should be removed when we find a proper solution to filter out these auction results in Diffusion.

Apart from refactoring, [this](https://github.com/artsy/metaphysics/pull/4205/files#diff-fbc9f27e95d788d42f0cab6546d248560e876101b04647f6ce43a1b066db62deR280) is the only relevant change in this PR.

[CX-2654]: https://artsyproduct.atlassian.net/browse/CX-2654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/schema/v2/artist/index.ts
```diff
@@ -275,36 +277,43 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
...
+
+          const { total_count, _embedded } = await auctionLotsLoader(
+            diffusionArgs
+          )
+
+          // TODO: Filter upcoming auction results (sale_end_date > today) in Diffusion.
+          const filteredAuctionResults = _embedded.items.filter(
+            (auctionResult) =>
+              !moment(auctionResult.sale_end_date).isAfter(moment())
+          )
```
olerichter00, line 289: This is the only relevant change.

src/schema/v2/artist/index.ts
```diff
@@ -275,36 +277,46 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
...
+
+          // TODO: Filter upcoming auction results (sale_end_date > today) in Diffusion.
+          const filteredAuctionResults = _embedded.items.filter(
+            (auctionResult) => {
+              const now = moment.utc()
+
+              return moment.utc(auctionResult.sale_end_date).isBefore(now)
+            }
+          )
```
olerichter00, line 292: This is the only relevant change.

src/schema/v2/artist/__tests__/index.test.js
```diff
@@ -108,25 +120,45 @@ describe("Artist type", () => {
     })
   })
 
-  it("returns the number of auction results for an artist", () => {
+  it("returns the auction results and number of auction results for an artist", () => {
```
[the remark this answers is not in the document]
olerichter00: Good point! I've not added a test for this case because we will hopefully remove this hack very soon. But you're right and I've added another lot to the test to check if it gets filtered out 👍

MounirDhahri:
Thanks for implementing this @olerichter00. I am wondering though if it would make more sense to do this from diffusion at the endpoint level and by sending an extra argument with the `auctionLotsLoader`. My main thinking here, is to keep page sizes the same from metaphysics. In case for example all returned auction results are for active sale_artworks, we will probably have no auction results returned for example for the first page

olerichter00:
Exactly, this is the problem with this quick fix. @egdbear [this is the ticket](https://artsyproduct.atlassian.net/browse/DO-234) for a proper solution (filtering in Diffusion, adding parameters and missing database indexes, checking if we need to filter out these auction results in other queries as well).

I'll double-check if there are cases that return no results at all and if we have to roll back this change.

MounirDhahri:
Cool - thank for sharing this ticket Ole. Feel free to touch the lots endpoint as well in diffusion if you feel this is needed.

PR #4213, opened 2022-07-06
feat: Add artistSparklines root field from Vortex

olerichter00:
Addresses [CX-2644]

## Description

Add artistSparklines root field from Vortex.

### DIFF
--- a/src/lib/stitching/vortex/schema.ts
+++ b/src/lib/stitching/vortex/schema.ts
@@ -1,12 +1,12 @@
-import { createVortexLink } from "./link"
-import {
-  makeRemoteExecutableSchema,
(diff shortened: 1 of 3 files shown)


src/lib/stitching/vortex/schema.ts
```diff
@@ -24,7 +24,6 @@ export const executableVortexSchema = ({
     "BigInt",
     "artistAffinities",
     "artistRecommendations",
-    "artistSparklines",
```
egdbear, line 27: I assume this was the old way of stitching the vortex schema?

egdbear (approved):
LGTM

PR #4254, opened 2022-07-26
fix: Send only valid vortex mediums when fetching price insights

olerichter00:
## Description

Addresses [CX-2677]

Currently, we have some My Collection artworks where `category` and `medium` are interchanged. 

Until we fix all of these artworks, this change will add logic to check which one is the category that we need to send to Vortex in order to fetch price insights.

[CX-2677]: https://artsyproduct.atlassian.net/browse/CX-2677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/loaders/loaders_with_authentication/vortex.ts
```diff
@@ -34,15 +36,98 @@ export default (accessToken, opts) => {
...
   return {
     vortexTokenLoader,
-    vortexGraphqlLoader: ({ query, variables }: LoaderArgs) =>
-      vortexLoader(
-        "/graphql",
-        { query, variables: JSON.stringify(variables) },
-        {
-          method: "POST",
+    vortexGraphqlLoader,
+    marketPriceInsightsBatchLoader: async (
```
olerichter00, line 51: I have moved the logic from `loadBatchPriceInsights` here because the functionality is basically just a loader.

src/lib/loaders/loaders_with_authentication/vortex.ts
```diff
@@ -34,15 +36,98 @@ export default (accessToken, opts) => {
...
+          variables: { artistIDMediumTuples },
+        })()
+
+        const priceInsightNodes: MarketPriceInsight[] = extractNodes(
+          vortexResult.data?.marketPriceInsightsBatch
+        )
+
+        return priceInsightNodes
+      } catch (e) {
+        console.error(e)
```
lordkiz, line 88: Log to Sentry?

src/lib/loaders/loaders_with_authentication/vortex.ts
```diff
@@ -34,15 +36,98 @@ export default (accessToken, opts) => {
...
+              node {
+                artistId
+                medium
+                demandRank
+                annualLotsSold
+                annualValueSoldCents
+                lastAuctionResultDate
+              }
+            }
+          }
```
lordkiz, line 76: @olerichter00 I am curious about how this might resolve with older versions of Eigen that do not currently do not send the category param when requesting for insights. Maybe I am missing something here.

olerichter00: In this case, the client doesn't have to send the category at all. When using `marketPriceInsightsBatchLoader` Metaphysics has already loaded the artwork (with medium and category) and can use the helper `getVortexMedium(artist.medium, artist.category)` to determine which one is the string that needs to be send to Vortex.

olerichter00: For example [here](https://github.com/artsy/metaphysics/blob/6dab4f604e7aa345c22339571749ce31ea102802/src/schema/v2/me/myCollection.ts#L291).

lordkiz: Ahaaaaa!!!! That was what I was expecting. But I can't figure out at what point in the entire code that happens. I see a default resolution for all insights 👍 , but can't find the logic for a single insight 🙏

src/lib/loaders/loaders_with_authentication/vortex.ts
olerichter00: Yes! I've moved the try-catch block outside of the loader to the function [enrichArtworksWithPriceInsights](https://github.com/artsy/metaphysics/blob/b7de1a3118724ec0b480f703b880e9c297b8be59/src/schema/v2/me/myCollection.ts#L272) and added a `Sentry.captureException`.

src/lib/loaders/loaders_with_authentication/vortex.ts
olerichter00: Yeah, actually we always use this batch loader at the moment for performance reasons...

olerichter00: Sorry, there I've refactored the code a lot...

MounirDhahri (approved):
nice. thanks for all the refactoring

PR #4255, opened 2022-07-26
feat(user): Add completedOnboardingAt mutation

damassi:
This adds a new `completedOnboarding` mutation:

```graphql
mutation {
  completedOnboarding(input:{ userID:"58828b1e9c18db30f3002fba",, timestamp:"222"}) {
    user {
      completedOnboardingAt     
    }
  }
}
```

Related: https://github.com/artsy/gravity/pull/15544

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -5321,6 +5321,17 @@ type CommerceUser {
   id: String!
 }
 
+input CompletedOnboardingMutationInput {
+  clientMutationId: String
+  timestamp: String!
+  userID: String!
+}
+
+type CompletedOnboardingMutationPayload {
+  clientMutationId: String
+  user: User!
+}
+
 type ConditionReportRequest {
   internalID: ID!
   saleArtworkID: ID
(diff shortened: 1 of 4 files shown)


jonallured:
Hrm! I guess I expected for this to be included in a more general mutation that updates other parts of a user - is there a reason to have an entirely separate mutation? 🤔

damassi:
@jonallured - Why? mutations are cheap and specific functions scoped by action

damassi:
(Gonna merge to get this into force but happy to refactor further if strong opinions)

jonallured:
I wouldn't say I have strong opinions just that taken to the extreme this philosophy would cause us to have a mutation for every field on every model, no? Why not have a single user mutation? Oh we do actually already have one here:

https://github.com/artsy/metaphysics/blob/main/src/schema/v2/users/updateUserMutation.ts

I used that in Forque semi-recently. Seems like it's better to add a field to that mutation over adding an entirely new mutation. 🤷

damassi:
Ah, I see. I didn't look under the `users` folder, just `me`. Will update 👍

PR #4257, opened 2022-07-26
feat(user): Update updateUserMutation with completedOnboardingAt

damassi:
Follow-up to https://github.com/artsy/metaphysics/pull/4255 where we decided to move this to a more sensible spot.


jonallured (approved):
LGTM!! 🎉

PR #4273, opened 2022-08-02
feat: Beginner Friendly GraphQL Doc

Mitch-Henson:
This PR attempts to add some more backend / beginner friendly documentation on how to add a GraphQL endpoint for an existing Gravity endpoint. This was previously added in a [previous PR](https://github.com/artsy/metaphysics/pull/4269) but is being moved to its own PR to allow for more discussion without blocking any feature work.

As a first time user of GraphQL I definitely struggled even with the most basic of operations and hopefully this will help other future devs with their first PR.

### DIFF
--- a/docs/adding_a_new_gravity_endpoint_into_metaphysics.md
+++ b/docs/adding_a_new_gravity_endpoint_into_metaphysics.md
@@ -1,0 +1,76 @@
+## Adding a gravity endpoint into Metaphysics
+
+This doc is aimed at helping backend devs or beginner front end devs on how to add an existing gravity endpoint into Metaphysics.
+
+This article assumes that you already have your local machine setup and have your graphiql.app or other app (Postman, Insomnia etc) setup. If you need help with that, you can start [here](https://github.com/artsy/metaphysics#setting-up-your-local-graphiql)
+
+### Steps to add an existing endpoint
+
+1. Find an existing PR or commit which adds something similar to what you want. If you are looking for a PUT endpoint, then [this](https://github.com/artsy/metaphysics/pull/3918) is a great example.
+1. Follow the structure of the PR to add your own endpoint and input/outputs. Some further tips can be found below.
+1. After the code has been written, you can start your local Metaphysics server with `yarn start`. This will generate the new schema files from the new changes and will continue to do so whilst the process is running. Should you have any issues with the schema generation, you can always run `yarn dump:local` in the Metaphysics directory which manually generates the schema changes. This script (and other useful scripts) can be found in the `package.json` file.
+1. With your favourite graphql app, you should be able to hit the local Metaphysics server to refresh the schema. In insomnia, you can click on the "schema" button in the GraphQL panel to manually refresh this, if you don't have auto-refresh turned on (also found in this list). Sidenote: During development Metaphysics should be pointed at Gravity staging (unless otherwise configured in `.env.shared`)
+1. Work through the errors, making sure to read the message in its entirety. Also don't hesitate to add it to the errors section below if useful.
+
+### Tips on writing and debugging locally
+
+- You don't need to restart the server once you make changes. The updates get triggered automatically once you save your changes locally and re-run your query in GraphiQL (or whatever app you use). You should see some lines similar to these below, which shows that it has been updated
+
+```
+[@artsy/express-reloadable] File /Users/mitchellhenson/dev/metaphysics/src/schema/v2/partnerArtworksMutation.ts has changed.
+[V1] [FEATURE] Enabling Schema Stitching
+[V2] [FEATURE] Enabling Schema Stitching
+```
+
+- You should be able to see the request that you are making to the gravity REST endpoint in the GraphQL output. This can potentially help pinpoint errors, if it is related to the endpoint URL. This code below is taken from the bottom of a GraphQL request made locally.
+
+```
+"extensions": {
+		"requests": {
+			"gravity": {
+				"requests": {
+					"partner/5a0323e68b0c146d7634f081/artworks?artsy_shipping_domestic=true": {
+						"time": "0s 495.276ms",
+						"cache": false,
+						"length": "0 B"
+					}
+				}
+			}
+		},
+		"requestID": "72c50460-119f-11ed-9554-25745ca04f3b"
+	}
+```
+
+- GraphQL is strongly typed and needs to know the exact inputs and outputs of the endpoint as it mediates between the clients and the REST endpoint. This is why you need to set up a union type which instructs GraphQL on what both responses will look like. This allows GraphQL tools to know if the request you are making (in Insomnia, for example) is valid, even before you send the request.
+- You can always add some more console logs to any resolver to help see what is coming back from the REST endpoint, e.g.
+
+```typescript
+  outputFields: {
+    partnerArtworksOrError: {
+      type: UpdatePartnerArtworksMutationType,
+      resolve: (result) => result,
+    },
+  },
+```
+
+can be changed to
+
+```typescript
+outputFields: {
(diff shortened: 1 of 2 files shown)


kajatiger (commented):
That's a nice addition to the existing docs, thanks for adding that! 
What helped me was also the docs that you can find in some projects for example the ones in [exchange](https://github.com/artsy/exchange#did-you-change-graphql-schema) for the preparation of the metaphysics change.

araujobarret (approved):
Loved the initiative, all look good from my side! 🎸 

I would maybe add a section to say briefly about data loaders and provide a link to the other doc https://github.com/artsy/metaphysics/blob/main/docs/dataloaders.md, people working in newer features might bump in a need of a data loader that was never implemented before

damassi (commented):
I agree, this is a really good first step! 

Similar to @araujobarret's comment, I think before this is merged we should step through some of the basics just so a totally new GraphQL dev doesn't arrive at this document and get confused. Meaning: There's the setup steps (good!), the code running steps (also good), and then one suddenly lands upon more advanced topics like handling errors via complex patterns (union types) that aren't quite idiomatic to the wider GraphQL community (and are scary, even for those who have been using GraphQL for a while). 

And so I would suggest two possible paths:
- We could rename this file to something along the lines of `error_handling_patterns_in_gravity` and strip out some of the setup steps or move them into the main README; or 
- Develop this doc further to be more comprehensive and _lead_ into error handling patterns as a secondary concern 

I'm hesitant to approve this as it is just because GraphQL is such an intimidating technology at first, and so being overly cautious about baby steps is crucial. The current document's name is also somewhat conflicting because there are no actual steps here that talk about adding a new endpoint, as one would expect changes to `lib/loaders/loaders_with/without_authentication/gravity.ts` file, and then the corresponding wire-up in a schema file (and so on). 

So, if wanting to go with the latter option, here are some suggestions (which would also be a good exercise to work through docs-wise just because you're new to it):
1. Start small. Begin with setup, as you're currently doing, and then move into simple `GET`s
1. Document how to add a new `REST` endpoint to [`lib/loaders/loaders_without_authentication/gravity.ts`](https://github.com/artsy/metaphysics/blob/main/src/lib/loaders/loaders_with_authentication/gravity.ts#L39)
1. Show how to access (and resolve) that data by [adding a new field](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/artwork/index.ts#L1227-L1230) to a slice of the schema -- say, `artwork` or `artist.ts`
1. With the absolute must-have basics now out of the way (which shows how to add a new Gravity Endpoint to MP, per the document's aim), now show how to do the next most crucial operation, [a `POST`](https://github.com/artsy/metaphysics/blob/main/src/lib/loaders/loaders_with_authentication/gravity.ts#L75-L79) 
1. Show how to write a simple mutation, following [this example](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/me/linkAuthenticationMutation.ts#L1) from the `src/schema/me` folder. This format linked here handles most everything that we need, from handling errors from gravity, to typing inputs with TypeScript. It's how we should introduce new devs to mutations, and is a gentle approach and the one that developers will most commonly use going forward (as opposed to the union-type example herein). 
1. Finally, once adding a new endpoint and writing a basic mutation is covered, we can then handle advanced concepts like our union-type mutations, but not before explaining the reason _why_ we would want to use this pattern. (If we can't adequately explain why a dev should use this approach, it should be omitted for the sake of teaching.) 

I don't mean to blow up the scope of this document 🙏 but I also want to really encourage us to think about what's most relevant to a new developer's needs and to introduce them slowly to the concepts. Totally understand that a lot of the things in this document reflect your first experience in MP and could be valuable for others; however, practically-speaking it is a rare day when a new GraphQL dev needs to dive into complex topics like unions and start writing mutations, where alternatively, it's very common for someone to need to add a Gravity `GET` endpoint, and then access it from a new field that resolves said data via the newly-created dataLoader. 

Getting the prerequisites simply expressed will yield the greatest impact from an organizational perspective, so it's worth spending the time doing it right -- especially because we currently don't have a comprehensive doc that goes over this adequately. 

If you feel inspired to flesh this out further, totally down to pair on this and fill in any gaps 👍

Mitch-Henson:
Thanks for everyone's feedback! I've made considerable updates, though it is not completely finished.

I would love to take you up on your pairing offer @damassi, but as I'm in Berlin, I think it is almost impossible. I've taken your very thorough feedback to heart and have moved a lot of things around. Let me know if the current _Adding a GET request_ section is more of what you had in mind!

The _Adding a POST request_ and _Writing a mutation_ sections are still both empty, and I will have to pair with someone to make sure that I don't spread misinformation in this doc!

docs/adding_a_new_gravity_endpoint_into_metaphysics.md
```diff
@@ -0,0 +1,131 @@
...
+
+With your favourite graphql app, you should be able to hit the local Metaphysics server to refresh the schema. In insomnia, you can click on the "schema" button in the GraphQL panel to manually refresh this, if you don't have auto-refresh turned on (also found in this list). Sidenote: During development Metaphysics should be pointed at Gravity staging (unless otherwise configured in `.env.shared`)
+
+### Adding a GET request
+
+In order to learn how to add a GET request, this section will walk you through an existing GET request and editing an existing field. The request that we want to look at is the artwork request.
+
+<!-- This should render the code inline -->
+
+https://github.com/artsy/metaphysics/blob/79a387cd837d48485bf6a0aa00e8999ad323ff23/src/lib/loaders/loaders_with_authentication/gravity.ts#L39
```
damassi, line 19: I'd suggest using the `loaders_without_authentication/gravity`  file instead, just so new devs can practice without needing to be authenticated (meaning, without needing to add an additional `X-Access-Token` header into their GraphiQL app)

docs/adding_a_new_gravity_endpoint_into_metaphysics.md
```diff
@@ -0,0 +1,131 @@
...
+
+As seen in the code, this `gravityLoader` takes in an `id` which it passes to the artwork URL. This loader function calls the `gravityLoaderWithAuthenticationFactory` function which eventually calls the gravity api connector which shows how the request will be made to gravity.
+
+<!-- This should render the code inline -->
+
+https://github.com/artsy/metaphysics/blob/79a387cd837d48485bf6a0aa00e8999ad323ff23/src/lib/apis/gravity.ts#L14-L43
+
+If you have worked with Gravity before, this code should be recogniseable. It takes the previously given partial path and makes a call to Gravity for this endpoint. This Gravity loader is then [matched](https://github.com/artsy/metaphysics/blob/79a387cd837d48485bf6a0aa00e8999ad323ff23/src/schema/v2/Match.ts#L33) with its specific type, the Artwork type. This shows GraphQL the structure and type all of the potential fields that can come back from the Gravity endpoint.
```
damassi, line 27: We should probably end this paragraph here: 

> If you have worked with Gravity before, this code should be recogniseable. It takes the previously given partial path and makes a call to Gravity for this endpoint.

Short and sweet. The reasoning is, in all my years working in MP i've never once opened this `Match` file!

damassi (commented):
These updates are looking absolutely fantastic @Mitch-Henson 💯. Nice and clear and covers all the basics well. Just a couple minor comments. 

I would argue that once some of the placeholders are filled in, even without the mutation section this is mergable 👍 Having the basic mutation overview would be a (helpful) bonus, though. Short of that, maybe a section called `Writing Mutations` with a few bullet points with links pointing to good examples would be sufficient. First bullet point could be the example I pointed out above, and the second could be the one with the union type (along with a link to our blog post). 

Thanks again for working through these updates --

PR #4277, opened 2022-08-03
Bulk update mutation tweaks

lilyfromseattle:
@ansor4 and I made some tweaks to the bulk update mutation to simplify the structure and make the naming more consistent. If this looks all good to you I can continue with my front end changes once its merged and released! If there are any issues or questions we can discuss at the sync I scheduled for tomorrow morning.


ansor4:
Just a note on that peril warning above ☝️ since volt is the only consumer of these fields, this should be a-ok

_schemaV2.graphql
```diff
@@ -3172,6 +3172,44 @@ type BidIncrementsFormatted {
...
+}
+
+union BulkUpdatePartnerArtworksMutationType =
+    BulkUpdatePartnerArtworksMutationFailure
+  | BulkUpdatePartnerArtworksMutationSuccess
+
+type BulkUpdatePartnerArtworksResponse {
+  count: Int
+  ids: [String]
+}
```
Mitch-Henson, line 3211: Shouldn't `errors` also be here? It is a part of the response and we seem to be using it in the `formattedReturn`

src/schema/v2/bulkUpdatePartnerArtworksMutation.ts
```diff
@@ -124,7 +113,19 @@ export const updatePartnerArtworksMutation = mutationWithClientMutationId<
...
-      return await updatePartnerArtworksLoader(id, gravityOptions)
+      const gravityResponse = await updatePartnerArtworksLoader(
+        id,
+        gravityOptions
+      )
+
+      // In the future it could be helpful to have a list of successfully opted in ids, can add this to gravity at a later date
+
+      const formattedReturn = {
+        updatedPartnerArtworks: { count: gravityResponse.success, ids: [] },
```
Mitch-Henson, line 124: The array of IDs are actually those of artworks that are _not_ updated. If the artwork was successfully updated, then that is what is what was expected and those that weren't updated are the anomaly, hence their IDs as we may be able to use them in a call to action in the future.

Some more context for the endpoint can be found in [this comment](https://github.com/artsy/metaphysics/pull/4269#discussion_r935331223) and in the [gravity code](https://github.com/artsy/gravity/blob/39caa1fca2398a0468273f7aa67230b47d1f0b17/app/services/partner_artwork_service.rb#L32-L47).

Because of that, I would expect this to be coupled with the `gravityResponse.errors` in `skippedPartnerArtworks`

Mitch-Henson (changes requested):
The naming changes look much better than what I came up with and are really nicely standardised. I think there needs to be some small changes relating to the gravity response but otherwise looks great

src/schema/v2/bulkUpdatePartnerArtworksMutation.ts
```diff
@@ -124,7 +113,19 @@ export const updatePartnerArtworksMutation = mutationWithClientMutationId<
...
+        skippedPartnerArtworks: gravityResponse.errors,
```
ansor4, line 125: @lilyfromseattle maybe let's be explicit here so future devs don't have to go digging into the gravity response format
```
skippedPartnerArtworks: { count: gravityResponse.errors.count, ids: gravityResponse.errors.ids }
```

src/schema/v2/bulkUpdatePartnerArtworksMutation.ts
Mitch-Henson: I misread this. I thought there was mapping from the `errors.ids` to here, but it is just an empty _ids_ array to make sure that the `updatedPartnerArtworks` and `skippedPartnerArtworks` follow the same format!

ansor4 (approved):

PR #4299, opened 2022-08-12
fix(CX-2748): show category instead of medium

MounirDhahri:
We are filtering using the medium instead of the category and this results on the mediums displayed in the insights tab. 
Screenshot_20220811_162442 (2)

After changes locally 
Simulator Screen Shot - iPhone 13 Pro - 2022-08-12 at 11 54 45

Note: the category is displayed twice because we run uniqWith also applied to medium, we will need to delete that in few weeks


olerichter00 (approved):
Makes a lot of sense! Thanks for the fix 🙇

PR #4308, opened 2022-08-16
feat: collectorProfile type + mutations for non-me context [PLATFORM-4426]

mdole:
Update: This PR expanded a bit to include the mutations for updating collector profile. It turns out that the existing `updateCollectorProfile` mutation in MP is actually hitting Gravity's `PUT /me/collector_profile`, and in the case of Forque, I feel that it makes more sense to be hitting `PUT /collector_profile/:id` since the user performing the update is not `me`.

To that end, moved a lot of the `collectorProfile` types/resolvers etc. into their own folder (which I think also makes sense since it's its own model and endpoint in Gravity that is not always scoped to `/me/`) and created a new mutation to hit that endpoint.

---

So...this just works out of the box! How nice is that!

Added company name and website to the available fields from Grav. IDK if we still want/need them for Forque, but they're in Torque and now we have them if we want them 🤷 

Also did consider adding tests, but as with many things in MP, it feels a little silly - like "yes, when I asked for mocked data I got mocked data back." Open to suggestions for a better test though!

Jira ticket: https://artsyproduct.atlassian.net/browse/PLATFORM-4426

Sample query with all of the fields:
```graphql
{
	user(id: "5744b8888b3b8145d7005304") {
		collectorProfile {
			collectorLevel
			companyName
			companyWebsite
			email
			id
			institutionalAffiliations
			intents
			internalID
			loyaltyApplicantAt
			name
			privacy
			professionalBuyerAppliedAt
			professionalBuyerAt
			selfReportedPurchases
			confirmedBuyerAt
			userInterests {
				internalID
				body
				category
				id
				__typename
				interest {
					__typename
				}
			}
		}
	}
}
```


src/schema/v2/user.ts
```diff
@@ -10,6 +10,7 @@ import { InternalIDFields } from "./object_identification"
 import { LocationType } from "schema/v2/location"
 import { ResolverContext } from "types/graphql"
 import { connectionWithCursorInfo } from "./fields/pagination"
+import CollectorProfile from "./me/collector_profile"
```
mdole, line 13: feels a bit bad to import this from `me`, but I'm not seeing a good pattern in MP for "two different types use this same type/resolver combo".

We could:
- Copy/paste from `me` to `user` so that they can change independently (meh)
- Create a new folder `shared` for things shared by multiple base types
- Leave as-is

Any feelings @mc-jones?

mc-jones: Agree that it feels a bit a anti-patern. I believe the way we generally handle this is to resolve the loader itself, similar to how do in [IdentityVerifications](https://github.com/artsy/metaphysics/blob/dd15c1c8176a3464dbf6b852e47e77625b669880/src/schema/v2/identityVerification.ts#L172-L178).

mdole: I ended up creating a separate `collectorProfile` folder/type and referencing it in both places - I'm happy with how it turned out, but lmk if you think it's the wrong approach @mc-jones !

mc-jones: I think this approach makes sense given the `collector_profile/:id` endpoint!

mc-jones (approved):

PR #4309, opened 2022-08-16
feat(PX-5114): Artwork page: Update buyer-facing shipping copy

lilyfromseattle:
[PX-5114](https://artsyproduct.atlassian.net/browse/PX-5114)
paired with @jpotts244 

Background: We updated the domestic/international logic for Europe countries as part of the ARTA Europe/UK launch. We should update the buyer-facing domestic/international copy on the artwork page, to match.

New copy
For works located in Ireland, Greece, and Malta: “Shipping: XX within the European Union, XX rest of world”

For works located in Switzerland: “Shipping: XX within Switzerland, XX rest of world”

For works located in Norway: “Shipping: XX within Norway, XX rest of world”

Q: What about other countries? Ex: USA. A: Let’s use the name of the country all the time.


damassi:
@lilyfromseattle - note that we should be using conventional-commit formatting at least in our PR titles (i need to update [danger here](https://github.com/artsy/force/pull/10753#issuecomment-1215915690))

src/schema/v2/artwork/index.ts
```diff
@@ -919,6 +919,11 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         description:
           "The string that describes domestic and international shipping.",
         resolve: (artwork) => {
+          const fullCountryName =
+            COUNTRIES[
+              artwork.shipping_origin[artwork.shipping_origin.length - 1]
```
ansor4, line 924: info: This seems like a concise but potentially brittle way of getting the country name from that map. Can we expect artwork to always have a `shipping_origin` key on it? Can we expect that `artwork.shipping_origin[artwork.shipping_origin.length - 1]` will always resolve to a key in `COUNTRIES`?

ansor4 (approved):

PR #4312, opened 2022-08-17
feat: add marketPriceInsights to the artwork interface

MounirDhahri:
**Description:**

The `marketPriceInsights` interface was broken for artwork and was always returning `null`. The problem was that there was no resolver supplied to it.

I also added some of the logic of the display that we need for the artist insights inside Force. The next step will be to use those new field in Eigen since Eigen is now doing that logic locally.

**Jira ticket:** https://artsyproduct.atlassian.net/browse/CX-2768

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -17,7 +17,7 @@ describe("Artwork type", () => {
     {
       artistId: "artist-id",
       demandRank: 0.64,
-      medium: "Print",
+      medium: "print",
       annualLotsSold: 25,
       annualValueSoldCents: 577662200012,
       lastAuctionResultDate: "2022-06-15T00:00:00Z",
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -13,6 +13,19 @@ describe("Artwork type", () => {
   let artwork = null
   let context = null
 
+  const artworkInsights = [
+    {
+      artistId: "artist-id",
+      demandRank: 0.64,
+      medium: "Print",
+      annualLotsSold: 25,
+      annualValueSoldCents: 577662200012,
+      lastAuctionResultDate: "2022-06-15T00:00:00Z",
+      medianSalePriceLast36Months: 577662200012,
+      liquidityRank: 0.9,
+    },
+  ]
+
   const artworkImages = [
     {
       position: 2,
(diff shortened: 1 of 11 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -208,6 +245,24 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+            {
+              artistId: artist._id,
+              medium,
+              category,
+            },
+          ])
+
+          if (marketPriceInsightNodes.length === 0) return null
+          return marketPriceInsightNodes[0]
+        },
```
olerichter00, line 265: **(issue):** Wouldn't this cause a request per artwork when loading multiple artworks? 

The reason why we introduced the batch load in the first place was to be able to [load all price insights at once](src/schema/v2/me/myCollection.ts).

We could try to do the same in the [artwork resolver](src/schema/v2/artwork/index.ts) and load price insights whenever `marketPriceInsights` is requested for artworks.

olerichter00 (changes requested):
Although this works, I think this will cause a lot of requests when loading a list of artworks (see comment for details).

src/schema/v2/artwork/index.ts
MounirDhahri: true, i am actually working on fixing that right now and how I can avoid it. thanks for also pointing it

MounirDhahri: The thing is that I would like to also be able to request the price insights for an artwork without using the global marketPriceInsights interface and therefore the need to get the medium first and make the query and ending up nesting query renderers

olerichter00: I've fixed the links in the comment above ⬆️ .

You could load the artwork first and then enrich with the price insights (if the field is requested) [here](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/artwork/index.ts#L1460)

yarn.lock
```diff
@@ -4912,6 +4912,14 @@ graphql-middleware@1.2.6:
   dependencies:
     graphql-tools "^3.0.2"
 
+graphql-parse-resolve-info@^4.12.3:
```
MounirDhahri, line 4915: this looks like the defacto solution to see if a field exists in a query, i looked in the project and couldn't find any other other helpers so i created one

src/schema/v2/artwork/index.ts
```diff
@@ -1451,13 +1489,51 @@ const Artwork: GraphQLFieldConfig<void, ResolverContext> = {
...
+          category: gravityArtworkRes.category,
+        },
+      ])
+
+      if (marketPriceInsight.length > 0) {
+        enrichedArtwork.marketPriceInsights = marketPriceInsight[0]
+      } else {
+        enrichedArtwork.marketPriceInsights = null
+      }
+    }
```
olerichter00, line 1534: Can we use [enrichArtworksWithPriceInsights](https://github.com/artsy/metaphysics/blob/47c071848ab12900d5996d08d44faa33fae1ba67/src/schema/v2/me/myCollection.ts#L272)? Maybe we need to modify it but it would be nice to reuse the logic we already have.

olerichter00 (approved):
This is amazing work!

I've just added a comment about using the helper `enrichArtworksWithPriceInsights` instead of implementing the logic again.

src/schema/v2/artwork/index.ts
MounirDhahri: good idea

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -17,7 +17,7 @@ describe("Artwork type", () => {
     {
       artistId: "artist-id",
       demandRank: 0.64,
-      medium: "Print",
+      medium: "print",
```
MounirDhahri, line 20: this was breaking the comparison here https://github.com/artsy/metaphysics/blob/0ea0fbc063d1c29a5ea38f68cc30d4800af2b19a/src/lib/fillers/enrichArtworksWithPriceInsights.ts-L32

src/schema/v2/artwork/index.ts
```diff
@@ -1506,34 +1507,23 @@ const Artwork: GraphQLFieldConfig<void, ResolverContext> = {
...
+        [artwork],
+        marketPriceInsightsBatchLoader
+      )
 
-      if (marketPriceInsight.length > 0) {
-        enrichedArtwork.marketPriceInsights = marketPriceInsight[0]
-      } else {
-        enrichedArtwork.marketPriceInsights = null
-      }
+      return enrichedArtworks && enrichedArtworks[0]
```
MounirDhahri, line 1523: this is probably not necessary, but in case it occurs we will be safe

olerichter00: You could simplify this to `enrichedArtworks?.[0]`.

olerichter00 (dismissed):

src/schema/v2/artwork/index.ts
MounirDhahri: 😅
true

MrSltun (approved):
Looks great! Thank you Mounir! 🚀 🚀

PR #4313, opened 2022-08-17
feat: expose partner owned data

brainbicycle:
adds some endpoints and updates others for partner app use cases:

<details><summary>Allows getting the shows managed by a partner filtered by an artistID</summary>

old-folio-artist-shows

</details>

<details><summary>Allows getting the artworks in a show including unpublished artworks</summary>

old-folio-show-artworks

</details>

<details><summary>Allows getting artworks managed by a partner including unpublished works for a specific artist</summary>

old-folio-artist-works

</details>

<details><summary>Allows getting the count of artworks for a given artist managed by the partner</summary>

old-folio-all-artists

</details>


src/schema/v2/partner.ts
```diff
@@ -200,7 +204,11 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
             has_published_artworks: args.hasPublishedArtworks,
           }
 
-          return allViaLoader(partnerArtistsForPartnerLoader, {
+          // If we are authenticated / loader exists use the all loader
+          const partnerArtistsLoader =
+            folioPartnerArtistsAllLoader ?? partnerArtistsForPartnerLoader
```
brainbicycle, line 209: this is quite weird and I am hoping there is a better way to do this:

the field on partner artist, artworks_count that is needed to populated the managedArtworks count field I have added here is not returned unless I explicitly request all fields from gravity which is the reason for the other loader:
`partner/${id}/partner_artists/all`

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -314,6 +314,29 @@ export default (accessToken, userID, opts) => {
       {},
       { headers: true }
     ),
+    folioPartnerArtistsAllLoader: gravityLoader(
```
damassi, line 317: This can just be `partnerArtistsAllLoader`; another app might want to come along and use the partner loader for something else

src/schema/v2/partner.ts
```diff
@@ -284,6 +292,87 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
+        description:
+          "A connection of artworks owned by a Partner for a given artist.",
+        type: artworkConnection.connectionType,
+        args: pageable({
+          artistID: {
+            type: GraphQLString,
+          },
+          ...artworksArgs,
+        }),
+        resolve: ({ id }, args, { folioPartnerArtistArtworksLoader }) => {
```
damassi, line 305: ditto

src/schema/v2/show.ts
```diff
@@ -112,8 +112,69 @@ export const ShowType = new GraphQLObjectType<any, ResolverContext>({
           return artists
         },
       },
+      allArtworksConnection: {
```
damassi, line 115: Can this be simply `artworksConnection`?

src/lib/loaders/loaders_with_authentication/gravity.ts
brainbicycle: yeah the only reason I had the prefix was because in the unauthenticated loaders there is another loader that has the same name but hits a different endpoint which I found confusing: https://github.com/artsy/metaphysics/blob/483f4fdd95f8b5efc5e089b1a2488df880999b68/src/lib/loaders/loaders_without_authentication/gravity.ts#L117
but it is a bad name and was meant to be a placeholder

src/schema/v2/show.ts
brainbicycle: there is already an `artworksConnection` but it only returns unpublished artworks as far as I can tell so I named this allArtworksConnection to distinguish the 2

damassi: Perhaps we could [update the arguments](https://github.com/artsy/metaphysics/blob/339aca96e75135a42e7e137ae7f0bfc308ad75fa/src/schema/v2/show.ts#L87), or adjust the loader so there's no need for duplication

src/schema/v2/partner.ts
starsirius: It's because `artworks_count` is only included when [requesting `:all` properties](https://github.com/artsy/gravity/blob/db8d902085f39a13d2a888611aeddbe35d16bffa/app/models/domain/partner_artist.rb#L66), which is what the `/all` endpoint [requests](https://github.com/artsy/gravity/blob/db8d902085f39a13d2a888611aeddbe35d16bffa/app/api/v1/partners_artists_endpoint.rb#L121) (but not the other public one). I think the reason is that the count includes unpublished artworks.

src/schema/v2/partner.ts
```diff
@@ -284,6 +292,87 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
           )
         },
       },
+      partnerArtistArtworksConnection: {
```
starsirius, line 295: If we haven't had a recommended pattern, this could be a good opportunity to establish one in Metaphysics for querying data based on different roles/authorization. Here are my 0.2 cents:

For example, this new field and the [existing one](https://github.com/artsy/metaphysics/blob/483f4fdd95f8b5efc5e089b1a2488df880999b68/src/schema/v2/partner_artist.ts#L127-L162) could be confusing:

```graphql
{
  partnerArtist(...) {
    artworksConnection # public, returns public/less data
  }
  partner(...) {
    partnerArtistArtworksConnection # requires authorized partner or admin, returns more data
  }
}
```

They both get artworks for a partner artist, but authorize differently and return different data. It's not immediately clear which one to use for a certain scenario. It would be great we have a consistent pattern to make our future life easier. I don't have a good answer but naively some options:

- **Use a single field for multiple purposes**. This [`artworksConnection` under partner](https://github.com/artsy/metaphysics/blob/62c985ab60f8ce348e53342571855691de0eb153/src/schema/v2/partner.ts#L287-L371) is an example. I think this is not ideal. The logic is complex. The arguments overlap but not always applicable. There are also more round trips.
- **Use a separate field for each use case**. This might be closer to this PR. I think it's cleaner implementation, and potentially can be less confusing to developers if, say, we name the fields consistently in certain way and/or write a clear field description.
- **Use a root field**. The [`me` root field](https://github.com/artsy/metaphysics/blob/604043ac24afbe724a10202e8dde88441ef4f142/src/schema/v2/schema.ts#L222) is clear that one has to be authorized as herself/himself to access her/his data in the me type. I wonder if we can borrow similar concept for accessing partner owned data.

I don't think this should be blocking. I thought it's a good timing to discuss and come up with a pattern for partner use cases in MP.

starsirius: I realized the example above wasn't accurate. The `partnerArtist { artworksConnection }` actually returns [`partner_artist_artworks`](https://github.com/artsy/metaphysics/blob/0e7202ed5899e9dee465ddded88950138ee7e5f3/src/lib/loaders/loaders_without_authentication/gravity.ts#L122) rather than `artworks`. But the idea should be the same.

src/schema/v2/partner.ts
```diff
@@ -200,7 +204,11 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
             has_published_artworks: args.hasPublishedArtworks,
           }
 
-          return allViaLoader(partnerArtistsForPartnerLoader, {
+          // If we are authenticated / loader exists use the all loader
```
mzikherman, line 207: Is that correct? The 'all' loader will exist for all requests that come from a logged-in user (since that will include the `X-Access-Token` header, and the presence of the header is what Metaphysics considers as 'authenticated' for loader purposes in https://github.com/artsy/metaphysics/blob/594022480e18cf881ad215c404855c3f0d38acda/src/lib/loaders/index.ts#L82-L84 - there's no verification happening).

So that means if this field is being used in Force/Eigen somewhere, with this change the authenticated loader will be used for all logged-in users, which might then fail as gravity enforces verification on that endpoint: https://github.com/artsy/gravity/blob/db8d902085f39a13d2a888611aeddbe35d16bffa/app/api/v1/partners_artists_endpoint.rb#L99

brainbicycle: thanks @mzikherman! that makes sense and this felt sketchy while I was writing it, thank you for clarifying how this works 🙌 
I think instead then I will need a separate endpoint on partner for the all loader? is that what you would suggest?

brainbicycle: I think I will take a page from the artwork endpoint and add an explicit param to use this loader 🤔 
`artwork(id: "test-7", includeUnlisted:  true)`
maybe `includePartnerFields` or something, let me know if it doesn't make sense

src/schema/v2/partner.ts
brainbicycle: yeah you are right. I made this an explicit param `includeAllFields` to keep it backwards compatible, let me know what you think!

src/schema/v2/show.ts
brainbicycle: I think we can and this extra field is totally unnecessary, thanks!

src/lib/loaders/loaders_without_authentication/gravity.ts
```diff
@@ -114,7 +114,7 @@ export default (opts) => {
     >(
       ({ artist_id, partner_id }) => `partner/${partner_id}/artist/${artist_id}`
     ),
-    partnerArtistArtworksLoader: gravityLoader<
+    partnerArtistPartnerArtistArtworksLoader: gravityLoader<
```
brainbicycle, line 117: this is a clunky but more accurate name for this loader since it returns PartnerArtistArtworks and the other loader returns just Artworks

src/schema/v2/partner.ts
brainbicycle: Thanks Chung Yi! after looking around a bit more it actually seems like most places are using the same field in most cases, e.g. your option 1 and when you hit the endpoint in an authenticated state it uses the authenticated loader and returns more data depending on your access. I agree this is not the most intuitive but since it seems like that is the standard I think I should stick with that, for this case it looks like I _can_ actually reuse the existing artworksConnection and pass some other args e.g. `published: false` So I've removed this extra connection and added some extra args and an authenticated loader to the existing field. 

For me it would make sense to have a separate field if the args for the authenticated request were significantly different or the authenticated request had to use a different endpoint. Let me know what you think!

brainbicycle:
Alright this is ready to go from my perspective, I will do some testing tomorrow before merging but any additional change requests please let me know! Thanks for the thoughtful reviews this is much cleaner now 🙈

src/schema/v2/partner.ts
```diff
@@ -74,6 +74,10 @@ const artworksArgs: GraphQLFieldConfigArgumentMap = {
     type: new GraphQLList(GraphQLString),
     description: "Return only artwork(s) included in this list of IDs.",
   },
+  artistID: {
```
brainbicycle, line 77: this should be an array similar to artworkIDs for future compatibility, thanks @mdole

brainbicycle: jk gravity only supports passing 1 to this endpoint

src/lib/loaders/loaders_without_authentication/gravity.ts
mzikherman: possibly `partnerArtistArtworksLoader` ? (avoiding the double `partnerArtist` thing - but i hear you this naming is a bit tough and verbose)

_schemaV2.graphql
```diff
@@ -11617,6 +11623,9 @@ type Partner implements Node {
     # Return artworks that are missing priority metadata
     missingPriorityMetadata: Boolean
 
+    # If false return both published and unpublished artworks, requires auth
+    published: Boolean
```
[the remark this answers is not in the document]
oxaudo: +1 that `includeUnpublished` describes much better that both published and unpublished are returned here.

Interestingly I was just adding `include_deleted` attribute on one of the endpoints in Gravity :) So we already have this pattern `include_xxx` used in some endpoints in Gravity.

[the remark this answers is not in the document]
brainbicycle: yeah that is a much better name 🙌 https://github.com/artsy/metaphysics/pull/4372

src/lib/loaders/loaders_without_authentication/gravity.ts
brainbicycle: that was what it was originally called but the reason for the rename was because I need another loader which would more accurately be named that in the authenticated loaders:
https://github.com/artsy/metaphysics/blob/d99b86824d9ae4a07acdd95f3d58fee86f7297d4/src/lib/loaders/loaders_with_authentication/gravity.ts#L322

this loader hits this endpoint:
`partner/${partnerID}/artist/${artistID}/partner_artist_artworks`

the one I need hits this endpoint:
`partner/${partnerID}/artist/${artistID}/artworks/all`

so the name thinking was:

```
partnerArtist -> an artist managed by this partner
+ 
PartnerArtistArtworks -> return type partnerArtistArtworks
```

and the other one:

```
partnerArtist -> an artist managed by this partner
+ 
Artworks -> return type Artworks
```

PR #4316, opened 2022-08-18
feat: expose notificationsConnection using existing /feed structure

dblandin:
Spike on exposing notifications via notificationsConnection query field.

cc/ @evaschilken @nickskalkin 

**Query**

```graphql
{
  notificationsConnection(first: 1) {
    edges {
      node {
        internalID
        title
        message
        createdAt
        isUnread
        targetHref
        artworksConnection(first: 5) {
          edges {
            node {
              title
            }
          }
        }
      }
    }
  }
}
```

**Response**

```json
{
  "data": {
    "notificationsConnection": {
      "edges": [
        {
          "node": {
            "internalID": "62fe270e071fb90007594c12",
            "title": "Gina Beavers",
            "message": "1 Work Added",
            "createdAt": "2022-08-18T11:48:30.000Z",
            "isUnread": true,
            "targetHref": "/artist/damien-hirst/works-for-sale",
            "artworksConnection": {
              "edges": [
                {
                  "node": {
                    "title": "Florescent Lip"
                  }
                }
              ]
            }
          }
        }
      ]
    }
  }
}
```
```

### DIFF
--- a/src/schema/v2/notifications/index.ts
+++ b/src/schema/v2/notifications/index.ts
@@ -1,0 +1,125 @@
+import {
+  GraphQLBoolean,
+  GraphQLEnumType,
+  GraphQLFieldConfig,
+  GraphQLList,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLString,
+} from "graphql"
+import { convertConnectionArgsToGravityArgs } from "lib/helpers"
+import { pageable } from "relay-cursor-paging"
+import {
+  connectionWithCursorInfo,
+  createPageCursors,
+} from "schema/v2/fields/pagination"
+import { ResolverContext } from "types/graphql"
+import { IDFields, NodeInterface } from "../object_identification"
+import { date } from "schema/v2/fields/date"
+import { artworkConnection } from "../artwork"
+import { connectionFromArray, connectionFromArraySlice } from "graphql-relay"
+import numeral from "../fields/numeral"
+import { pick } from "lodash"
+
+const NotificationTypesEnum = new GraphQLEnumType({
+  name: "NotificationTypesEnum",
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/schema/v2/notifications/index.ts [hunk 1]
```diff
@@ -0,0 +1,105 @@
...
+    href: {
+      type: new GraphQLNonNull(GraphQLString),
+      resolve: ({ object }) => `/artist/${object.artist.id}/works-for-sale`,
+    },
+    artworksConnection: {
+      type: artworkConnection.connectionType,
+      args: pageable(),
+      resolve: async ({ object_ids: ids }, args, { artworksLoader }) => {
+        const { page, size } = convertConnectionArgsToGravityArgs(args)
+        return artworksLoader({ ids, batched: true }).then((body) => {
```
anandaroop, line 49: I haven't seen `batched` in use before — what does this do for us?

anandaroop (commented):
😎

src/schema/v2/notifications/index.ts
dblandin: There's some context in this PR: https://github.com/artsy/metaphysics/pull/2537

and it might make sense to use a similarly named variable here to better communicate the intent. The gist is that by passing `batched: true` we're instructing the data loader to pass the id arguments in the exact order that we're specifying here rather than sorting the params to increase the likelihood of cache hits.

anandaroop: Got it, thanks!

nickskalkin (approved):
Looks great! Approving PR, but one small question - you haven't added `targetURL`, is this intentional? (maybe you want Force to resolve target URLs after all).

PR #4319, opened 2022-08-19
chore: return liquidity rank with market insights

MounirDhahri:
- add liquidityRank to the query we use to get data from vortex.
- Follow-up on https://github.com/artsy/metaphysics/pull/4317


MounirDhahri:
cc @MrSltun

PR #4330, opened 2022-08-22
feat: folio additional artwork fields

brainbicycle:
Adds a few fields required to display the artwork page in the partner app:
https://artsyproduct.atlassian.net/browse/MOPLAT-343

- internalDisplayPrice on the artwork
- internalDisplayPrice on editionSets
- confidentialNotes on the artwork

These fields return as null when accessed by a user without access to the artwork.


rajsam003 (approved):
LGTM

src/schema/v2/artwork/index.ts
```diff
@@ -358,6 +358,12 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           existyValue(collecting_institution),
       },
       comparableAuctionResults: ComparableAuctionResults,
+      confidentialNotes: {
+        type: GraphQLString,
+        resolve: ({ confidential_notes }) => existyValue(confidential_notes),
```
ovasdi, line 363: **Question**: what is special about this field to require `existyValue` to be used? Never came across this usage before so would like to understand this. Is it that the field may come as an empty object or empty string and if so, don't even set it? Thanks!

ovasdi (approved):
LGTM

src/schema/v2/artwork/index.ts
brainbicycle: 🙈 I copied this because I saw the field above using it, I don't think it is necessary the other fields just return null when they are not present without this which I think is fine, I will remove it

PR #4334, opened 2022-08-24
feat: push schema updates to energy

MounirDhahri:
This PR makes sure that we update the schema inside energy to match the one in metaphysics to avoid manual updates like this one https://github.com/artsy/energy/pull/143

cc @rajsam003


MounirDhahri:
I will merge this since it's the only way to test it out. This makes no schema changes, so we should be safe

brainbicycle:
Woot! thanks Mounir!

patrinoua:
@MounirDhahri I'd love some more context on this if you have time 🙏

MounirDhahri:
sure, I will reach out

PR #4343, opened 2022-08-26
chore: expose `totalCount` field for `notificationsConnection`

dimatretyak:
Initial PR: artsy/metaphysics#4316

### Description
Small changes that allow to return the correct value of `totalCount`

#### Before
before

#### After
after

### DIFF
--- a/src/schema/v2/notifications/__tests__/index.test.ts
+++ b/src/schema/v2/notifications/__tests__/index.test.ts
@@ -28,6 +28,7 @@ describe("notificationsConnection", () => {
     const query = gql`
       {
         notificationsConnection(first: 10, notificationTypes: [ARTWORK_ALERT]) {
+          totalCount
           counts {
             total
             unread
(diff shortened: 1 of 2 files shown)


src/schema/v2/notifications/__tests__/index.test.ts
```diff
@@ -28,6 +28,7 @@ describe("notificationsConnection", () => {
     const query = gql`
       {
         notificationsConnection(first: 10, notificationTypes: [ARTWORK_ALERT]) {
+          totalCount
```
dblandin, line 31: Does this differ from the total count field already exposed?

```graphql
{
  notificationsConnection(first: 5) {
    counts {
      total
    }
  }
}
```

dblandin: Ahhh, I see the other form is already exposed (via the helper we're using) but not threading the right value. I suppose it's fine to fix here and let the client choose how to best access.

dblandin (approved):

src/schema/v2/notifications/__tests__/index.test.ts
dimatretyak: > I see the other form is already exposed (via the helper we're using)

Yes, that's right

PR #4354, opened 2022-08-31
fix: Return image square placeholder when no geometry is provided

olerichter00:
Addresses [CX-2854]

## Description

In cases when no geometry is provided for an image, we return a square image by default. This PR fixes the placeholder field that was returning "NaN%" instead of "100%" if the original image width and height are `0`.

The placeholder is important to correctly display the images in the frontend. Without this value being set correctly, My Collection images are not displayed correctly shortly after an artwork has been created.

[CX-2854]: https://artsyproduct.atlassian.net/browse/CX-2854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/image/__tests__/index.test.js
+++ b/src/schema/v2/image/__tests__/index.test.js
@@ -74,6 +74,24 @@ describe("Image type", () => {
     })
   })
 
+  describe("#placeholder", () => {
+    const query = `{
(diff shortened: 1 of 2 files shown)


MounirDhahri (commented):
😎

PR #4369, opened 2022-09-09
feat: Segment indexed images by environment for reverse image search

olerichter00:
Addresses [FX-4202]

## Description

**Reverse Image Search:**

Segment indexed [TinEye](https://tineye.com/) images by environment using a new env variable (`process.env.NODE_ENV` is set to `production` in both staging and prod).

**Todo:**

- [ ] Reindex all images

[FX-4202]: https://artsyproduct.atlassian.net/browse/FX-4202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/config.ts
```diff
@@ -79,6 +79,7 @@ const {
   STATSD_PORT,
   TINEYE_API_PASSWORD,
   TINEYE_API_USERNAME,
+  TINEYE_ENVIRONMENT,
```
olerichter00, line 82: Wondering if we should give this env variable a more generic name to make it reusable for similar cases. Something like `SYSTEM_ENVIRONMENT` could make sense.

anandaroop: 👍🏽

src/config.ts
```diff
@@ -214,6 +215,7 @@ export default {
   STATSD_PORT: Number(STATSD_PORT) || 8125,
   TINEYE_API_PASSWORD,
   TINEYE_API_USERNAME,
+  TINEYE_ENVIRONMENT,
```
anandaroop, line 218: Let's default to `development`

anandaroop (commented):
Nice! Thanks for the change and the migration ✅

PR #4374, opened 2022-09-09
fix(reverse image search): parsing of TinEye artwork result

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-4202

Follow-up to https://github.com/artsy/metaphysics/pull/4369

The recent change to the convention for the TinEye filepath that we use as an identifier…

|which version|filepath convention|
|---|---|
|old|`artwork/<artwork id>/image/<image token>`|
|new|`<current env>/artwork/<artwork id>/image/<image token>`|

…means that we need to change the way we parse that identifier in order to extract the artwork id — it's now the third segment in that path, not the second.

There wasn't an existing test to catch that, which is why it slipped through #4369 and resulted in broken image matching.

This PR adds the test and updates the implementation.

### DIFF
--- a/src/schema/v2/__tests__/reverseImageSearch.test.ts
+++ b/src/schema/v2/__tests__/reverseImageSearch.test.ts
@@ -58,7 +58,7 @@ describe("reverseImageSearch", () => {
         "reverseImageSearch": Object {
           "results": Array [
             Object {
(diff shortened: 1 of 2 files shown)


dimatretyak (approved):
Thank you for these changes

PR #4378, opened 2022-09-13
feat: Filter notifications by notification type

olerichter00:
Addresses [FX-4200]

## Description

Filter notifications by notification type using the new param introduced in https://github.com/artsy/gravity/pull/15643.

[FX-4200]: https://artsyproduct.atlassian.net/browse/FX-4200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/notifications/index.ts
+++ b/src/schema/v2/notifications/index.ts
@@ -7,19 +7,24 @@ import {
   GraphQLObjectType,
   GraphQLString,
 } from "graphql"
+import { connectionFromArray, connectionFromArraySlice } from "graphql-relay"
 import { convertConnectionArgsToGravityArgs } from "lib/helpers"
+import { pick } from "lodash"
 import { pageable } from "relay-cursor-paging"
+import { date } from "schema/v2/fields/date"
 import {
   connectionWithCursorInfo,
   createPageCursors,
 } from "schema/v2/fields/pagination"
 import { ResolverContext } from "types/graphql"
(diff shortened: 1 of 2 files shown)


src/schema/v2/notifications/index.ts
```diff
@@ -7,19 +7,24 @@ import {
...
-import { pick } from "lodash"
+import { IDFields, NodeInterface } from "../object_identification"
+
+const GRAVITY_NOTIFICATION_TYPE_MAPPING = {
+  artwork_alert: "SavedSearchHitActivity",
+  artwork_published: "ArtworkPublishedActivity",
+}
 
 const NotificationTypesEnum = new GraphQLEnumType({
   name: "NotificationTypesEnum",
```
dblandin, line 30: **question**: Instead of defining a new mapping - what do you think of updating these enum values to `SavedSearchHitActivity`, and `ArtworkPublishedActivity`? Clients will still refer to `ARTWORK_ALERT` and `ARTWORK_ALERT` but the underlying value could match what Gravity expects.

olerichter00: Good idea! It might be confusing because in most places in the code base the value is very similar to the key of an enum, but I find this solution better than introducing the mapping. 

I've updated the code and adjusted the values.

dblandin (approved):
Nice 👍🏼

PR #4379, opened 2022-09-13
feat(FX-4181): Create a mutation to mark all unread notifications as read

dimatretyak:
Jira ticket: [FX-4181]

### Mutation
```graphql
mutation {
  markAllNotificationsAsRead(input: {}) {
    responseOrError {
      ... on MarkAllNotificationsAsReadSuccess {
        success
      }
      
      ... on MarkAllNotificationsAsReadFailure {
        mutationError {
          error
        }
      }
    }
  }
}
```

### Mutation success response
```graphql
markAllNotificationsAsRead: {
  responseOrError: {
    success: true
  }
}
```

### Mutation failure response
```graphql
{
  "responseOrError": {
    "mutationError": {
      "error": "...",
      "message": "..."
      ...
    }
  }
}
```

### Debug info
```javascript
{
  "data": {
    "markAllNotificationsAsRead": {
      "responseOrError": {
        "success": true
      }
    }
  },
  "extensions": {
    "requests": {
      "gravity": {
        "requests": {
          "me/notifications?status=read": {
            "time": "0s 524.472ms",
            "cache": false,
            "length": "0 B"
          }
        }
      }
    },
    "requestID": "6d1956a0-34dc-11ed-94de-1dd82adc179a"
  }
}
```

[FX-4181]: https://artsyproduct.atlassian.net/browse/FX-4181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/loaders/loaders_with_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_with_authentication/gravity.ts
@@ -298,6 +298,11 @@ export default (accessToken, userID, opts) => {
       {},
       { method: "DELETE" }
     ),
+    updateNotificationsLoader: gravityLoader(
+      "me/notifications",
+      {},
+      { method: "PUT" }
+    ),
     notificationsFeedLoader: gravityLoader("me/notifications/feed"),
     partnerAllLoader: gravityLoader((id) => `partner/${id}/all`),
     partnerArtistDocumentsLoader: gravityLoader<
--- a/src/schema/v2/me/mark_all_notifications_as_read_mutation.ts
+++ b/src/schema/v2/me/mark_all_notifications_as_read_mutation.ts
@@ -1,0 +1,36 @@
+import { GraphQLBoolean } from "graphql"
+import { mutationWithClientMutationId } from "graphql-relay"
+import { ResolverContext } from "types/graphql"
+
+export const markAllNotificationsAsReadMutation = mutationWithClientMutationId<
+  any,
+  any,
+  ResolverContext
+>({
+  name: "MarkAllNotificationsAsRead",
+  description: "Mark all unread notifications as read",
+  inputFields: {},
+  outputFields: {
+    success: {
+      type: GraphQLBoolean,
+      resolve: (result) => result.success,
+    },
+  },
+  mutateAndGetPayload: async (_, { updateNotificationsLoader }) => {
+    if (!updateNotificationsLoader) {
+      throw new Error("You need to be signed in to perform this action")
+    }
+
+    try {
+      await updateNotificationsLoader({ status: "read" })
+
+      return {
+        success: true,
+      }
+    } catch {
+      return {
+        success: false,
+      }
+    }
+  },
+})
--- a/src/schema/v2/me/mark_all_notifications_as_read_mutation.ts
+++ b/src/schema/v2/me/mark_all_notifications_as_read_mutation.ts
@@ -1,6 +1,41 @@
+import { GraphQLObjectType } from "graphql"
+import { GraphQLUnionType } from "graphql"
 import { GraphQLBoolean } from "graphql"
 import { mutationWithClientMutationId } from "graphql-relay"
+import {
+  formatGravityError,
+  GravityMutationErrorType,
+} from "lib/gravityErrorHandler"
 import { ResolverContext } from "types/graphql"
+
+const SuccessType = new GraphQLObjectType<any, ResolverContext>({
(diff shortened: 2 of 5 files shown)


dimatretyak:
**question**: `NotificationService.mark_with_status` marks notifications as read only for **the last 7 days** as read by default (see the [since](https://github.com/artsy/gravity/blob/472728b7a7af2da4d2ff9f73e7e0c95f23823133/app/models/util/notification_service.rb#L60) parameter). Should we change this behavior?

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -298,6 +298,11 @@ export default (accessToken, userID, opts) => {
       {},
       { method: "DELETE" }
     ),
+    updateNotificationsLoader: gravityLoader(
+      "me/notifications",
+      {},
+      { method: "PUT" }
```
gkartalis, line 304: **question:** why put method? (out of curiosity)

gkartalis (dismissed):

src/lib/loaders/loaders_with_authentication/gravity.ts
dimatretyak: such HTTP Method is used for [this endpoint](https://github.com/artsy/gravity/blob/main/app/api/v1/me_notifications_endpoint.rb#L58)

src/schema/v2/me/mark_all_notifications_as_read_mutation.ts
```diff
@@ -0,0 +1,36 @@
...
+      throw new Error("You need to be signed in to perform this action")
+    }
+
+    try {
+      await updateNotificationsLoader({ status: "read" })
+
+      return {
+        success: true,
+      }
+    } catch {
```
dimatretyak, line 30: **question**: should we call `Sentry.captureException` here? otherwise the error will not be logged anywhere and only [warning will be outputted](https://github.com/artsy/metaphysics/blob/86b39d687fb655079bd287fb83b31f67aa07812f/src/lib/loaders/api/loader_with_authentication_factory.ts#L68).

src/schema/v2/me/mark_all_notifications_as_read_mutation.ts
```diff
@@ -0,0 +1,36 @@
...
+      return {
+        success: false,
+      }
+    }
```
dimatretyak, line 34: or is it even better to use a common response format for mutations (for example, as in the screenshot), as is done for all mutations in Metaphysics?

image

src/schema/v2/me/mark_all_notifications_as_read_mutation.ts
dimatretyak: or is it even better to use a common response format for mutations (for example, as in the screenshot), as is done for all mutations in Metaphysics?

image

mzikherman: I don't think we currently even use Sentry for Metaphysics as it is super noisy, and duplicates errors logged elsewhere in the request. The MP project in Sentry is even empty - https://sentry.io/organizations/artsynet/issues/?project=299996

In terms of mutation success/error patterns, yea I think the `... on MutationSuccess` and `... on MutationFailure` style is the way to go.

dimatretyak: > In terms of mutation success/error patterns, yea I think the ... on MutationSuccess and ... on MutationFailure style is the way to go

@dblandin what do you think about it? or is it better to keep the current solution with `success: true | false`?

dblandin: Sticking with the established pattern sounds good to me! We don't have much in terms of success attributes or anticipated errors to return, but I suppose this lets us easily add those later on within the response union type.

dimatretyak: Done

dblandin (approved):
Looks great 👍🏼

src/schema/v2/me/mark_all_notifications_as_read_mutation.ts
```diff
@@ -0,0 +1,75 @@
+import { GraphQLObjectType } from "graphql"
```
damassi, line 1: FYI new files don't need to be snake_case

PR #4393, opened 2022-09-19
chore(PX-5236): remove unlisted and includeUnlisted

ansor4:
Removing `unlisted` field on `artwork` and the `includeUnlisted` argument on the `artwork` field.

Basically reverting [this PR](https://github.com/artsy/metaphysics/pull/2428)

-----------------------------

Expected failure when querying with`includeUnlisted` param

2022-09-19 at 1 35 01 PM

Expected failure when querying for `unlisted` field

2022-09-19 at 1 34 17 PM

A successful query

2022-09-19 at 1 34 39 PM

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2165,9 +2165,6 @@ type Artwork implements Node & Searchable & Sellable {
   submissionId: String
   title: String
 
-  # Whether this artwork is unlisted or not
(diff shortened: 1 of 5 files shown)


oxaudo (approved):
Should be safe I think as it should not have been really used anywhere.

PR #4398, opened 2022-09-22
feat[FX-4244]: use activity_type from Gravity

mdole:
Lil PR to return notificationType using values passed from Gravity instead of inferring it based on message text/other parts of the Gravity response.

Jira ticket: https://artsyproduct.atlassian.net/browse/FX-4244

@dblandin anything else that needs to be done for that ticket? Didn't see other mentions of notificationType/activity_type that needed to be updated, but LMK if I missed things!


src/schema/v2/notifications/__tests__/index.test.ts
```diff
@@ -13,6 +13,7 @@ describe("notificationsConnection", () => {
           date: "2022-08-22T21:15:49.000Z",
           object_ids: ["63036fafbe5cfc000cf358e3", "630392514f13a5000b55ecec"],
           object: { artist: { id: "damien-hirst" } },
+          activity_type: "ARTWORK_ALERT",
```
dblandin, line 16: **suggestion**: I think this should actually be `SavedSearchHitActivity`. The enum keys are used at the GraphQL layer but from Gravity, we use the underlying string value.

mdole: aha yes thank you, good catch! updated

src/schema/v2/notifications/__tests__/index.test.ts
```diff
@@ -12,7 +12,11 @@ describe("notificationsConnection", () => {
           status: "unread",
           date: "2022-08-22T21:15:49.000Z",
           object_ids: ["63036fafbe5cfc000cf358e3", "630392514f13a5000b55ecec"],
-          object: { artist: { id: "damien-hirst" } },
+          object: {
+            artist: { id: "damien-hirst", _id: "4d8b926a4eb68a1b2c0000ae" },
```
mdole, line 16: I'm honestly not sure how these tests were passing before 🤔 this was failing both locally and CI because it thought we should always have an ID for the artist

dblandin (approved):
LGTM 👍🏼

PR #4401, opened 2022-09-22
feat(vimeo): Add Vimeo API / loader

damassi:
Addresses [GRO-1305]

This adds the Vimeo API and associated loader. 

See:
- https://developer.vimeo.com/api/reference

Also refactors and adds a new `accessTokenByRoleLoader` helper that formally defines a way to use roles as a gate-keeper for admin-ish API tasks to 3rd parties. 

Example: 
- In order to retrieve all feature flags you need to have `team` role in JWT; or, 
- In order to upload video to vimeo you must have `creator` role, etc. 

This could of course work for other things outside of 3rd parties, but its most useful in a 3rd party context because the access tokens generated by the 3rd party API are generally admin-permissive and designed to be embedded within MP, but we don't want to give random MP users admin-level access without the proper role.

cc @artsy/grow-devs 

[GRO-1305]: https://artsyproduct.atlassian.net/browse/GRO-1305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/loaders/loaders_with_authentication/vimeo.ts
```diff
@@ -0,0 +1,22 @@
+import factories from "../api"
+import { accessTokenByRoleLoader } from "./utils/accessTokenByRoleLoader"
+
+export const vimeoLoaders = (accessToken, opts) => {
+  const { vimeoLoaderWithAuthenticationFactory } = factories(opts)
+
+  const { accessTokenLoader } = accessTokenByRoleLoader(
+    // TODO: What role do partners updating artworks have?
+    "team",
```
damassi, line 9: @jonallured - not sure. Do you know?

jonallured:
Huh, so this is a loader for MP to hit the Vimeo API. I guess I assumed that this would be happening in Gravity but maybe all Gravity will have is an identifier? Will Gravity also have a Vimeo API client? How does Gravity get that Vimeo identifier? Do you anticipate Volt hitting MP in order to create the video at Vimeo? Lol lots of questions from me this morning! 😝

src/lib/loaders/loaders_with_authentication/vimeo.ts
jonallured: I think that they don't have a special role, they have access to do things on PartnerArtwork records that they create but I'd have to do more research to be sure.

jonallured: Shit which reminds me - will we be adding the Vimeo fields to `Artwork` records or `PartnerArtwork` records...

jonallured:
For posterity we had some chatter on Slack here:

https://artsy.slack.com/archives/C01ADJNCS5D/p1663950939075709

And then we caught up IRL after standup so I now understand more about the direction here! 👍

PR #4410, opened 2022-09-27
feat: Add high auction record to artist insights GRO-841

jonallured:
This PR moves calculating an artist's high auction record up into MP from Force. Right now this is done like here:

https://github.com/artsy/force/blob/9fb0f70c7abda2db06a79d80d4a77dea68c53334/src/Apps/Artist/Components/ArtistInsights/ArtistInsightBadges.tsx#L75

I think this is better because it pulls logic out of the client which is always a win but it also makes the mental model easier: if you want insights they are on the `artist.insights` field full stop. No more having to think about where they are.

And so then this sets me up to add both critically acclaimed and Artsy vanguard as the next two artist insights. I'll follow up with that work as I get things in place.

https://artsyproduct.atlassian.net/browse/GRO-841

/cc @artsy/grow-devs

### DIFF
--- a/src/schema/v2/artist/__tests__/helpers.test.ts
+++ b/src/schema/v2/artist/__tests__/helpers.test.ts
@@ -123,4 +123,40 @@ describe("getArtistInsights", () => {
       })
     })
   })
+
+  describe("string insight fields", () => {
+    const fields = [
+      {
+        key: "highAuctionRecord",
+        kind: "HIGH_AUCTION_RECORD",
+        value: "£18.6m, Sotheby's, 2021",
+      } as any,
+    ]
+
+    fields.forEach((field) => {
+      it(`returns an empty array of entities when the ${field.key} has a value and sets the description`, () => {
+        const artist = {
+          [field.key]: field.value,
+        }
+
+        const insights = getArtistInsights(artist)
+        const insight = insights.find((insight) => insight.kind === field.kind)!
+
+        expect(insight.count).toEqual(0)
+        expect(insight.entities).toEqual([])
+        expect(insight.description).toEqual(field.value)
+      })
+
+      it(`returns an empty artist object when the ${field.key} has no value`, () => {
+        field.value = null
+        const artist = {
+          [field.key]: field.value,
+        }
+
+        const insights = getArtistInsights(artist)
+        const insight = insights.find((insight) => insight.kind === field.kind)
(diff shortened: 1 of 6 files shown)


src/schema/v2/artist/__tests__/helpers.test.ts
```diff
@@ -0,0 +1,162 @@
...
+      },
+      {
+        key: "biennials",
+        kind: "BIENNIAL",
+        value,
+      },
+    ]
+
+    fields.forEach((field) => {
+      it(`returns an array of ${field.key} entities split by pipe`, () => {
```
damassi, line 42: Ooh, i've never considered putting an `it` within a loop, but why not?

damassi (approved):
Looks great

src/schema/v2/artist/__tests__/helpers.test.ts
laurabeth: `it.each` would do what you want here w/i the rtl paradigm

TMcMeans:
im soo happy about this!!

TMcMeans (approved):

src/schema/v2/artist/__tests__/helpers.test.ts
jonallured: I tried this but wasn't able to get it to work the way I expected. I followed the docs (I think!) but was not able to get the test titles to do string interp.

```
    pipe delimited insight fields
      ✓ returns an array of $variable.key entities split by pipe
      ✓ returns an array of $variable.key entities split by pipe
      ✓ returns an array of $variable.key entities split by pipe (1ms)
      ✓ returns an array of $variable.key entities split by pipe
```

For whatever reason that `$variable.key` construct from the docs doesn't expand?

https://jestjs.io/docs/api#testeachtablename-fn-timeout

src/schema/v2/artist/helpers.ts
```diff
@@ -0,0 +1,112 @@
+import { compact } from "lodash"
```
TMcMeans, line 1: I think it makes a ton of sense to move a bunch of this logic into a helper file and out of the insights file, nice work!

jonallured:
Ok I'm going to merge this and move onto the Gravity side of things to add the critically acclaimed and vanguard year fields. 👍

PR #4412, opened 2022-09-27
chore:add visibility level to artwork

nspinazz89:
This PR adds visibility level to artwork per [PX-5249]. It also has a related [gravity ticket ](https://github.com/artsy/gravity/pull/15693)that needs to be merged first.

[PX-5249]: https://artsyproduct.atlassian.net/browse/PX-5249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2172,7 +2172,7 @@ type Artwork implements Node & Searchable & Sellable {
   vatRequirementComplete: Boolean
 
   # The visibility level of the artwork
-  visibilityLevel: String
+  visibilityLevel: Visibility
 
   # If the category is video, then it returns the href for the (youtube/vimeo) video, otherwise returns the website from CMS
   website: String
@@ -17563,6 +17563,12 @@ input ViewingRoomSubsectionInput {
   delete: Boolean = false
   image: ARImageInput
   internalID: ID
+}
+
+enum Visibility {
+  DRAFT
+  LISTED
+  UNLISTED
 }
 
 type WireTransfer {
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2171,6 +2171,9 @@ type Artwork implements Node & Searchable & Sellable {
   # Based on artwork location verify that VAT info for the partner is complete.
   vatRequirementComplete: Boolean
 
+  # The visibility level of the artwork
+  visibilityLevel: String
+
   # If the category is video, then it returns the href for the (youtube/vimeo) video, otherwise returns the website from CMS
   website: String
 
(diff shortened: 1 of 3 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -1446,6 +1446,11 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
         },
       },
+      visibilityLevel: {
+        description: "The visibility level of the artwork",
+        type: GraphQLString,
+        resolve: ({ visibility_level }) => visibility_level,
+      },
```
jpotts244, line 1453: Might be nice to add a spec to capture the new field

Something simple like:
https://github.com/artsy/metaphysics/blob/main/src/schema/v2/artwork/__tests__/artwork.test.js#L692-L714 for example!

_schemaV2.graphql
```diff
@@ -2171,6 +2171,9 @@ type Artwork implements Node & Searchable & Sellable {
   # Based on artwork location verify that VAT info for the partner is complete.
   vatRequirementComplete: Boolean
 
+  # The visibility level of the artwork
+  visibilityLevel: String
+
```
jpotts244, line 2176: Curious what others think here of typing?

I wonder if it is helpful at all to specify an enum type here (well in the artwork.index page) similar to how it is handled in [artwork import source](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/artwork/index.ts#L323-L335) for example. Then it is a little more explicit in the schema that the values for visibilityLevel are expected to only be X,Y,Z. Example [here](https://github.com/artsy/metaphysics/pull/3886/files)

nspinazz89: So I based this on how we handle `cost_currency_code` which is also an enum field on `Artwork` in gravity, but definitely open to changing it!

src/schema/v2/artwork/index.ts
nspinazz89: cool cool will do!

_schemaV2.graphql
ansor4: blocking: +1 to adding an enum for this. Part of adding an enum to a graphql schema is that client will know that the field is an enum and not just a string. I can see why we might have done that for `cost_currency_code`, but honestly that should be a graphql enum as well.

ansor4 (changes requested):
Let's create the enum that @jpotts244 mentioned! Happy to pair/mob on it!

_schemaV2.graphql
nspinazz89: alrighty I'm convinced and I switched it to an ENUM!

ansor4 (approved):
thanks!

PR #4416, opened 2022-09-29
feat: Support two new artist insights GRO-916

jonallured:
This PR updates the Artist schema to include two new artist insight fields from Gravity:

* criticallyAcclaimed
* vanguardYear

But then I also went to the insights field and updated that to use these fields to return the formatted bits and bobs. Once this is merged I can turn to Force and start in on that last piece of the puzzle.

https://artsyproduct.atlassian.net/browse/GRO-841
https://artsyproduct.atlassian.net/browse/GRO-916

/cc @artsy/grow-devs

### DIFF
--- a/src/schema/v2/artist/helpers.ts
+++ b/src/schema/v2/artist/helpers.ts
@@ -12,6 +12,8 @@ export const ARTIST_INSIGHT_KINDS = {
   BIENNIAL: { value: "BIENNIAL" },
   ACTIVE_SECONDARY_MARKET: { value: "ACTIVE_SECONDARY_MARKET" },
   HIGH_AUCTION_RECORD: { value: "HIGH_AUCTION_RECORD" },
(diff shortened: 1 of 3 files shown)


src/schema/v2/artist/helpers.ts
```diff
@@ -72,12 +85,13 @@ export const getArtistInsights = (artist) => {
   ][]
 
   const insights = mappings.map((mapping) => {
-    const [kind, { getDescription, getEntities, label }] = mapping
+    const [kind, { getDescription, getEntities, getLabel }] = mapping
```
jonallured, line 88: Given enough time everything in javascript ends up being a function 😝

TMcMeans: you're a big fan of mapping functions! Love that!

TMcMeans (approved):
this looks good, no notes! 
LGTM 👍🏾 🥇

PR #4419, opened 2022-09-30
feat(artwork): Update figure with new video

damassi:
Addresses https://artsyproduct.atlassian.net/browse/GRO-1306

This updates video support in the artwork type in support of the `setVideoAsCover` prop, which will push the video to the front. 

cc @artsy/grow-devs

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -74,7 +74,13 @@ import { ArtworkHighlightType } from "./highlight"
 import ArtworkLayer from "./layer"
 import ArtworkLayers, { artworkLayers } from "./layers"
 import Meta, { artistNames } from "./meta"
-import { embed, isEmbeddedVideo, isTooBig, isTwoDimensional } from "./utilities"
+import {
+  embed,
+  getFigures,
+  isEmbeddedVideo,
+  isTooBig,
+  isTwoDimensional,
+} from "./utilities"
 
 const has_price_range = (price) => {
   return new RegExp(/-/).test(price)
@@ -1327,7 +1333,7 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       series: markdown(),
-      setVideoAsCover: {
+      isSetVideoAsCover: {
         type: GraphQLBoolean,
         description: "Should the video be used as the cover image",
         resolve: ({ set_video_as_cover }) => set_video_as_cover,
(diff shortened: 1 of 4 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -1515,10 +1519,22 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+                },
+              ]
             : []
-          return [...sortedTypedImages, ...typedVideos]
+
+          if (set_video_as_cover) {
+            return [...typedVideos, ...sortedTypedImages]
+          } else {
+            return [...sortedTypedImages, ...typedVideos]
+          }
```
jonallured, line 1537: **nitpick [non-blocking]:** add missing test

I think this little bit here is worth a test just because it's a bit odd so maybe future us will have a better time remembering what's going on here with a spec for before/after and whether the video array is empty or not. What do you think?

damassi: Will add!

src/schema/v2/artwork/index.ts
```diff
@@ -1294,6 +1293,11 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       series: markdown(),
+      setVideoAsCover: {
```
mzikherman, line 1296: #minor should we be prefixing boolean fields with `is` ?

src/schema/v2/artwork/index.ts
```diff
@@ -1515,10 +1519,22 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
             _.sortBy(typedImages, "position")
           )
 
-          const typedVideos = VIDEOS[id]
-            ? [{ ...VIDEOS[id], type: "Video" }]
+          const typedVideos = external_video_id
```
mzikherman, line 1522: #minor since this intentionally just supports a single video, maybe this should be `const typedVideo` and be an object (vs. an array) - doesn't need to be spread below.

src/schema/v2/artwork/index.ts
damassi: It's less a check against whether a prop is true, but rather setter: 

Screen Shot 2022-09-30 at 11 22 13 AM

I tend to understand `isFoo` is more of a one way street, a getter that refers to something like 

```tsx
const isThing = !!thing.length > 5
```

damassi: OH i think i understand what you mean, since this is MP: we're not setting anything over here. I can update 👍

src/schema/v2/artwork/index.ts
damassi: Will fix 👍

PR #4428, opened 2022-10-05
feat: expand user partner access field

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-4595

In support of the linked ticket, this updates `PartnerAccessField` (making it `UserPartnerAccessField`) and changes what shape is returned to the client.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11887,6 +11887,11 @@ type Partner implements Node {
   website: String
 }
 
+type PartnerAccessPropertyField {
+  id: String!
+  name: String!
+}
+
 # A partner agreement
 type PartnerAgreement {
   # Datetime the agreement was accepted
@@ -16285,8 +16290,8 @@ type User implements Node {
   # The paddle number of the user
   paddleNumber: String
 
-  # The Parnter or Profile access granted to the user
-  partnerAccess: [String]!
+  # The partner access granted to the user
+  partnerAccess: [UserPartnerAccess]!
 
   # The given phone number of the user.
   phone: String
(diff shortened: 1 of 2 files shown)


ovasdi:
After getting a bit more familiar with the modeling on the backend (which I found not very straightforward) and discussing the options with @mc-jones, I wanted to invite @mzikherman to discuss (and share his opinion on) whether this change would be better implemented as a `connection`.

I haven't looked at how prevalent it is when users have access to more than 20 partners (or profiles) but regardless, these are lists and so I suspect pagination could be useful in some cases.

As it stands, we retreive the data via `userAccessControlLoader` which, please correct me if im wrong, makes [AccessControl](https://github.com/artsy/gravity/blob/7d9a83c1e069d0fd9ad99986a55989e5946d90a1/app/models/domain/access_control.rb#L1) the edge. So the _nodes_ are [User](https://github.com/artsy/gravity/blob/main/app/models/domain/user.rb), [Partner](https://github.com/artsy/gravity/blob/main/app/models/domain/partner.rb), [Profile](https://github.com/artsy/gravity/blob/main/app/models/domain/profile.rb) and the relationship between these would be AccessControl?

In order to get the data we want, a [model type](https://github.com/artsy/gravity/blob/main/app/api/v1/users_access_controls_endpoint.rb#L33) must be specified. Gravity will return different shapes (by my understanding) based on the _type_ we request and so I reckon we would have _separate_ connections for each type and those would use the _same_ edge type, `UserAccessControlEdge`?

As far as the task goes, I only need one piece of data, a partner [slug](https://github.com/artsy/gravity/blob/main/app/models/domain/partner.rb#L659) to build a link to the admin partner page (vibrations) however, if there is a strong case for using a connection here as opposed to keeping the current pattern in place, perhaps this is an opportunity to switch the existing implementation?

ovasdi:
Closing this in favor of https://github.com/artsy/metaphysics/pull/4434

PR #4437, opened 2022-10-10
fix: divide value by 100 vo cut the cents

dariakoko:
We used cants as if it was dollars, which caused the Annual Value Sold to be 100 times bigger than it should be.

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -110,7 +110,7 @@ const ArtworkPriceInsightsType = new GraphQLObjectType<any, ResolverContext>({
       type: GraphQLString,
       description: 'The annual value of the work sold "in USD "',
       resolve: ({ annualValueSoldCents }) => {
-        return `$${formatLargeNumber(annualValueSoldCents)}`
+        return `$${formatLargeNumber(annualValueSoldCents / 100)}` // dividing by 100 because we don't need the value in cents
       },
     },
     annualLotsSold: {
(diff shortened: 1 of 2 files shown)


olerichter00 (dismissed):
Thanks a lot for the fix 🙏

MounirDhahri:
Let's merge this as soon as we fix the broken tests

src/schema/v2/artwork/index.ts
```diff
@@ -110,7 +110,7 @@ const ArtworkPriceInsightsType = new GraphQLObjectType<any, ResolverContext>({
       type: GraphQLString,
       description: 'The annual value of the work sold "in USD "',
       resolve: ({ annualValueSoldCents }) => {
-        return `$${formatLargeNumber(annualValueSoldCents)}`
+        return `$${formatLargeNumber(annualValueSoldCents / 100)}` // dividing by 100 because we don't need the value in cents
```
joeyAghion, line 113: We should avoid doing currency math in multiple, local, view-oriented places. Better touse the money helpers that are (I think) already available in this project, and avoid currency differences, rounding, floating-point precision issues, etc.

sweir27: +1 to this! Baking in `/ 100` makes it much harder to support different currencies in the future.

MounirDhahri: @dariakoko let's fix this and use the money helper then.

PR #4438, opened 2022-10-10
fix: use errorOrResponse pattern for user role mutations

mc-jones:
This PR makes two fixes: 

1. Updates `addUserRoleMutation` and `deleteUserRoleMutation` to use our `errorOrResponse` pattern
- As part of this change we are now returning the user parent object upon success instead of just the `roles` array.
2. corrects the associated loaders to use the proper naming convention (end with `loader`)

### considerations
This PR includes breaking changes, as we are now returning the parent `user` object instead of just the user `roles`. This should be safe, as `forque` should be (will confirm again before merging) the only client using this mutation. Similarly, we change the input from `role` to `role_type` to better match what is expected in `gravity`.

This supports `forque` changes to display the error message coming from gravity to the user. 

[PLATFORM-4648]

[PLATFORM-4648]: https://artsyproduct.atlassian.net/browse/PLATFORM-4648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


artsyjian (approved):
lgtm

ovasdi (approved):
Nice refactoring to match the existing conventions! 🙌

PR #4443, opened 2022-10-12
feat: Add support for remote images to My Collection artwork mutations

olerichter00:
Addresses [CX-2983]

## Description

In order to support adding images from a remote URL (not S3) to a My Collection artwork, this PR adds the functionality to the function that computes the image source. 

This PR prepares the endpoint for https://github.com/artsy/convection/pull/1348 and uses Gravity's optional parameter [remote_image_url](https://github.com/artsy/gravity/blob/bcf19fd80bc908a802191e80366685c346bc3f53/app/api/v1/artworks_images_endpoint.rb#L49) (`POST /api/v1/artwork/:artwork_id/image`)

[CX-2983]: https://artsyproduct.atlassian.net/browse/CX-2983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
+++ b/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
@@ -255,7 +255,15 @@ export const computeImageSources = (externalImageUrls) => {
   const imageSources = externalImageUrls.map((url) => {
     const match = url.match(externalUrlRegex)
 
-    if (!match) return
+    if (!match) {
+      if (url.startsWith("http")) {
+        return {
+          remote_image_url: url,
+        }
+      } else {
+        return
+      }
+    }
 
     const { sourceBucket, sourceKey } = match.groups
 
--- a/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
+++ b/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
@@ -16,7 +16,7 @@ import { ArtworkImportSourceEnum } from "../artwork"
 import { MyCollectionArtworkMutationType } from "./myCollection"
 import { EditableLocationFields } from "./update_me_mutation"
 
-const externalUrlRegex = /https:\/\/(?<sourceBucket>.*).s3.amazonaws.com\/(?<sourceKey>.*)/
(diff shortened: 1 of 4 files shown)


src/schema/v2/me/update_me_mutation.ts
```diff
@@ -260,3 +260,21 @@ export default mutationWithClientMutationId<any, any, ResolverContext>({
     }
   },
 })
+
+export const computeImageSources = (externalImageUrls) => {
```
olerichter00, line 264: We don't need support for remote image URLs here and it's maybe better if this mutation uses it's own `computeImageSources` function.

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
```diff
@@ -255,7 +255,15 @@ export const computeImageSources = (externalImageUrls) => {
   const imageSources = externalImageUrls.map((url) => {
     const match = url.match(externalUrlRegex)
 
-    if (!match) return
+    if (!match) {
+      if (url.startsWith("http")) {
```
MounirDhahri, line 259: **thought:** do you intentionally have this as `http` instead of `https` ?

MounirDhahri (commented):
LGTM, let's check this out before rolling it out

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
olerichter00: Not really, my intention was to keep it as simple as possible and I don't see a need to restrict to `https`. We could check for `https` or even use a URL regex (for example `/^https?:\/\/(?:www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b(?:[-a-zA-Z0-9()@:%_\+.~#?&\/=]*)$/`)

MounirDhahri (approved):

PR #4452, opened 2022-10-17
feat(FX-4313): Add the ability to query a formatted date for notification

dimatretyak:
Jira ticket: [FX-4313]

### Description
Add a new field for [Notification](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/notifications/index.ts#L34) type to be able to query a formatted date

### Acceptance Criteria
* Formatted date labels
  * `Today` when notification was published today
  * `Yesterday` when notification was published yesterday
  * `x days ago` when the notification was published 2 or more days ago
* It is necessary to take into account the user's time zone

[FX-4313]: https://artsyproduct.atlassian.net/browse/FX-4313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/notifications/index.ts
+++ b/src/schema/v2/notifications/index.ts
@@ -56,16 +56,17 @@ export const NotificationType = new GraphQLObjectType<any, ResolverContext>({
     publishedAt: {
       type: new GraphQLNonNull(GraphQLString),
       args: {
-        formatted: {
-          type: GraphQLBoolean,
+        format: {
+          type: GraphQLString,
           description:
-            'The human-friendly date (e.g. "Today", "Yesterday", "5 days ago")',
+            'pass `RELATIVE` to display the human-friendly date (e.g. "Today", "Yesterday", "5 days ago")',
         },
       },
-      resolve: ({ date }, { formatted }, { defaultTimezone }) => {
+      resolve: ({ date }, { format }, { defaultTimezone }) => {
         const timezone = defaultTimezone ?? DEFAULT_TZ
+        const dateFormat = format ?? "YYYY-MM-DDTHH:mm:ss[Z]"
 
-        if (formatted) {
+        if (format === "RELATIVE") {
           const today = moment.tz(moment(), timezone).startOf("day")
           const createdAt = moment.tz(date, timezone).startOf("day")
           const days = today.diff(createdAt, "days")
@@ -81,7 +82,7 @@ export const NotificationType = new GraphQLObjectType<any, ResolverContext>({
           return `${days} days ago`
         }
 
-        return formatDate(date, "YYYY-MM-DDTHH:mm:ss[Z]", timezone)
+        return formatDate(date, dateFormat, timezone)
       },
     },
     targetHref: {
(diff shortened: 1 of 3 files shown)


src/schema/v2/notifications/index.ts
```diff
@@ -48,6 +49,24 @@ export const NotificationType = new GraphQLObjectType<any, ResolverContext>({
       resolve: ({ status }) => status === "unread",
     },
     createdAt: date(({ date }) => date),
+    formattedCreatedAt: {
```
dimatretyak, line 52: **question**: should we leave such a name or is it better `formattedPublishedAt`?

dblandin: Good question. I think we can use `publishedAt` instead. Maybe while you're in here, we can mark `createdAt` as deprecated and expose a new `publishedAt` field which matches the current implementation?

dimatretyak: Done

image

dimatretyak: **question**: maybe it will be useful to provide an opportunity to query a formatted or raw date? for example, something like this

```graphql
{
  notificationsConnection(first: 10) {
    edges {
      node {
        publishedAt(formatted: true)
      }
    }
  }
}
```

gkartalis: What do you both think of supporting both like:
```graphql
{
  notificationsConnection(first: 10) {
    edges {
      node {
        publishedAt // returns "2022-10-17T00:00:00+03:00"
        formattedPublishedAt // returns "Today", "Yesterday", "x days ago"
      }
    }
  }
}
```

nickskalkin (approved):
👍

gkartalis (dismissed):

src/schema/v2/notifications/index.ts
dimatretyak: > I would prefer the first option (with an optional formatted: argument),

@gkartalis I'm also more prefer this

gkartalis: feel free to go ahead with this, don't have any strong preference here

dblandin: Another option might be to extend the existing `date` field args which already support `format` and `timezone` args.

Maybe if `format === 'RELATIVE'`, we trigger the Today / Yesterday / N days ago format, otherwise we pass the format string (e.g. `YYYY-MM-DDTHH:mm:ss[Z]`) to moment as other fields do?

```graphql
{
  notificationsConnection(first: 10) {
    edges {
      node {
        publishedAt // "2022-10-17T00:00:00+03:00"
        publishedAt(format: "YYYY-MM-DD") // "2022-10-17"
        publishedAt(format: "RELATIVE") // "Yesterday"
      }
    }
  }
}
```

dimatretyak: @dblandin added changes

image

nickskalkin (approved):

PR #4453, opened 2022-10-18
chore: add comment about the artwork price insights type

MounirDhahri:
This PR adds some description to the artwork insights type to mention that it is not available for all artwork connections.


MounirDhahri:
cc @olerichter00 @dariakoko @MrSltun @lordkiz

PR #4456, opened 2022-10-18
refactor: UpdateCollectorProfileWithID mutation

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-4674

This refactors the `UpdateCollectorProfileWithID` _mutation_ adding in the `ResponseOrErrorType` pattern. The primary reason for this refactor is to expose and handle `GravityMutationError` on the client via response payload.

Code search [shows](https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Aartsy+updateCollectorProfileWithID) that this mutation is used only in _Forque_ and so I am updating in place instead of opting for a more graceful switch. Please let me know if I should instead roll this out differently.

<details>
<summary>Click to expand</summary>

Screen Shot 2022-10-18 at 4 33 04 PM

</details>

---

The affected Forque change is [here](https://github.com/artsy/forque/pull/459).

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -15706,6 +15706,10 @@ type UpdateAppSecondFactorPayload {
   secondFactorOrErrors: AppSecondFactorOrErrorsUnion!
 }
 
+type UpdateCollectorProfileFailure {
+  mutationError: GravityMutationError
+}
+
 input UpdateCollectorProfileInput {
   # List of affiliated auction house ids, referencing Galaxy.
   affiliatedAuctionHouseIds: [String]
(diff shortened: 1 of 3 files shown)


ovasdi:
> The V2 schema in this PR has breaking changes with Force. Remember to update the Force schema if necessary.

RE: the warning, this should only affect Forque. https://github.com/artsy/force/search?q=updateCollectorProfileWithID

mc-jones:
> Please let me know if I should instead roll this out differently.

I believe the `forque` changes will need to be released first, or `metaphysics` will fail the schema change push as I [recently discovered](https://artsy.slack.com/archives/CP9P4KR35/p1665593981272599) when making a very similar change. In between the `forque` and `metaphysics` releases, the client will technically be broken (i.e. the mutation would fail), which we would normally want to coordinate the release as Chung-Yi [calls out](https://artsy.slack.com/archives/CP9P4KR35/p1665595947806819?thread_ts=1665593981.272599&cid=CP9P4KR35). In this instance it is probably OK to accept the minimal downtime in forque as the `Users` page is still not technically released or being used other than for development purposes.

mc-jones (approved):

ovasdi:
> > Please let me know if I should instead roll this out differently.
> 
> I believe the `forque` changes will need to be released first, or `metaphysics` will fail the schema change push as I [recently discovered](https://artsy.slack.com/archives/CP9P4KR35/p1665593981272599) when making a very similar change. In between the `forque` and `metaphysics` releases, the client will technically be broken (i.e. the mutation would fail), which we would normally want to coordinate the release as Chung-Yi [calls out](https://artsy.slack.com/archives/CP9P4KR35/p1665595947806819?thread_ts=1665593981.272599&cid=CP9P4KR35). In this instance it is probably OK to accept the minimal downtime in forque as the `Users` page is still not technically released or being used other than for development purposes.

Thanks for the heads up on all of that. 🙇 

_If_ this was a release that touched surfaces being actively used, would I then have to introduce an entirely new mutation, release that, update the client to use it and the follow up in MP with removal of the old one?

mc-jones:
> If this was a release that touched surfaces being actively used, would I then have to introduce an entirely new mutation, release that, update the client to use it and the follow up in MP with removal of the old one?

That would definitely be one approach. Also depending on the change, if possible, we could also look to rewrite the existing mutation in a backwards-compatible way, as Chung-Yi mentioned: 

>So I think typically we'll avoid breaking changes like this by, instead of renaming a field directly, adding a new optional field, migrating clients to use that while supporting the old one, and deprecating the old field if applicable.

PR #4462, opened 2022-10-19
fix: Adjust logic to get the default artwork images to not return broken images without URL

olerichter00:
Addresses [CX-3058]

## Description

This PR adjusts the logic to get the default artwork images to not return broken images without URL which happened in the case described in the Jira ticket. 

[CX-3058]: https://artsyproduct.atlassian.net/browse/CX-3058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/image/__tests__/index.test.js
+++ b/src/schema/v2/image/__tests__/index.test.js
@@ -8,17 +8,31 @@ describe("getDefault", () => {
   it("returns the default image", () => {
     expect(
       getDefault([
-        { id: "foo", is_default: false },
-        { id: "bar", is_default: true },
-        { id: "baz", is_default: false },
+        { id: "foo", image_url: "a-url", is_default: false },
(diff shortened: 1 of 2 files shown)


MounirDhahri (approved):
LGTM

MounirDhahri:
Would you mind check if this, along with the other changes you made, fix the issue on the user collection and shows them images. Since the image is there, already but it's just not showing up properly.

lordkiz (approved):
LGTM

olerichter00:
> Would you mind check if this, along with the other changes you made, fix the issue on the user collection and shows them images. Since the image is there, already but it's just not showing up.

This should fix the issue for the user mentioned in the ticket and the images will show up in the My Collection overview screen correctly 👍

PR #4468, opened 2022-10-20
feat: Artwork#isEligibleForOnPlatformTransaction

dblandin:
As we continue to work on the artwork page/screen and associated details, we've realized that it would be useful to have a consistent definition for when an artwork is eligble to transact on-platform. We have a lot of checks for `isAcquireable || isOfferable` in clients. We've recently introduced the notion of "offerable by inquiry" which expands the cases for which an artwork supports on-platform commercial action. Beyond simplifying these client conditionals today, this also gives us a single source of truth we can adjust as this definition changes in the future.

cc/ @anandaroop @nickskalkin @dimatretyak @gkartalis


dimatretyak (approved):
LGTM 👍

PR #4477, opened 2022-10-27
feat(FX-4212): add receivePartnerShowNotification user preference

nickskalkin:
[Jira](https://artsyproduct.atlassian.net/browse/FX-4412)

Recently we introduced a new user preference - receive_partner_show_notification. We want to add functionality to add this setting in Eigen. For this purpose, we need to add a mutation in Metaphysics to alter this field.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10544,6 +10544,9 @@ type Me implements Node {
   # This user should receive outbid notifications
   receiveOutbidNotification: Boolean
 
+  # This user should receive partner show notifications
+  receivePartnerShowNotification: Boolean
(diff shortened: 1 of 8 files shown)


dblandin (approved):
This LGTM. To clarify, this is actually modifying an existing mutation and not defining a new mutation, correct?

dimatretyak (approved):

nickskalkin:
@dblandin yes, you are right. This is similar how it is going to be used on eigen side: https://github.com/artsy/eigen/commit/e09d9f37f46264bf08bf9a85519362e413007bf4

PR #4480, opened 2022-10-28
feat: Expose is_personal_artist field

olerichter00:
## Description

Expose is_personal_artist field.


lordkiz (approved):
Nice work. Will merge asap so you can check if you can get the gravity is_personal_artist.

PR #4489, opened 2022-11-02
feat: add partner.documentsConnection query field

rajsam003:
This PR resolves: https://artsyproduct.atlassian.net/browse/MOPLAT-477

To fetch documents of the partner by its ID(s)

- [x] modify endpoint in gravity - https://github.com/artsy/gravity/pull/15808

**Query**

```graphql
{
  partner(id: "partnerID") {
    documentsByID: documentsConnection(first: 10, documentIDs: ["4faae457c056c20001000113", "4fc3b2b06c53ad0001000883"]) {
      edges {
        node {
          title
        }
      }
    }

    documentsByArtist: documentsConnection(first: 10, artistID: "abe-ajax") {
      edges {
        node {
          title
        }
      }
    }

    documentsByShow: documentsConnection(first: 10, showID: "folio-test-partner-mega-show-with-lots-of-artworks") {
      edges {
        node {
          title
        }
      }
    }
  }
}
```

**Response**

```json
{
  "partner": {
    "documentsByID": {
      "edges": [
        {
          "node": {
            "title": "hi"
          }
        },
        {
          "node": {
            "title": "hi"
          }
        }
      ]
    },
    "documentsByArtist": {
      "edges": [
        {
          "node": {
            "title": "Bla bl;a"
          }
        }
      ]
    },
    "documentsByShow": {
      "edges": [
        {
          "node": {
            "title": "test"
          }
        }
      ]
    }
  }
}
```

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11811,13 +11811,13 @@ type Partner implements Node {
   displayWorksSection: Boolean
   distinguishRepresentedArtists: Boolean
 
-  # Retrieve all partner documents for a given partner
+  # Return partner documents if current user has CMS access.
   documentsConnection(
     after: String
     artistID: String
     before: String
 
-    # Return only document(s) included in this list of IDs.
+    # Filter documents by ID.
     documentIDs: [String]
     first: Int
     last: Int
@@ -13571,6 +13571,7 @@ type Query {
     partnerID: String!
     size: Int
   ): PartnerArtistDocumentsConnectionConnection
+    @deprecated(reason: "Prefer `partner.documentsConnection`")
 
   # A list of Artworks for a partner
   partnerArtworks(
@@ -13640,6 +13641,7 @@ type Query {
     showID: String!
     size: Int
   ): PartnerDocumentConnection
+    @deprecated(reason: "Prefer `partner.documentsConnection`")
 
   # Phone number information
   phoneNumber(phoneNumber: String!, regionCode: String): PhoneNumberType
@@ -17318,6 +17320,7 @@ type Viewer {
     partnerID: String!
     size: Int
   ): PartnerArtistDocumentsConnectionConnection
+    @deprecated(reason: "Prefer `partner.documentsConnection`")
 
   # A list of Artworks for a partner
   partnerArtworks(
@@ -17387,6 +17390,7 @@ type Viewer {
     showID: String!
     size: Int
   ): PartnerDocumentConnection
+    @deprecated(reason: "Prefer `partner.documentsConnection`")
 
   # Phone number information
   phoneNumber(phoneNumber: String!, regionCode: String): PhoneNumberType
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11810,6 +11810,19 @@ type Partner implements Node {
   displayFullPartnerPage: Boolean
   displayWorksSection: Boolean
   distinguishRepresentedArtists: Boolean
+
+  # A connection of documents at a partner.
+  documentsConnection(
+    after: String
+    before: String
+
+    # Return only document(s) included in this list of IDs.
+    documentIDs: [String]
+    first: Int
+    last: Int
+    page: Int
+    size: Int
+  ): PartnerDocumentConnection
   featuredShow: Show
 
   # Artworks Elastic Search results
@@ -12254,6 +12267,40 @@ type PartnerCounts {
     format: String
     label: String
   ): FormattedNumber
+}
+
+type PartnerDocument {
+  filename: String!
+
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
+  publicUrl: String!
+  size: Int!
+  title: String!
+  uri: String!
+}
+
+# A connection to a list of items.
+type PartnerDocumentConnection {
+  # A list of edges.
+  edges: [PartnerDocumentEdge]
+  pageCursors: PageCursors!
+
+  # Information to aid in pagination.
+  pageInfo: PageInfo!
+  totalCount: Int
+}
+
+# An edge in a connection.
+type PartnerDocumentEdge {
+  # A cursor for use in pagination
+  cursor: String!
+
+  # The item at the end of the edge
+  node: PartnerDocument
 }
 
 # An edge in a connection.
@@ -13581,6 +13628,19 @@ type Query {
     # The slug or ID of the PartnerCategory
     id: String!
   ): PartnerCategory
+
+  # Retrieve all partner documents for a given partner
+  partnerDocumentsConnection(
+    after: String
+    before: String
+    first: Int
+    last: Int
+    page: Int
+
+    # The slug or ID of the Partner
+    partnerID: String!
+    size: Int
+  ): PartnerDocumentConnection
 
   # A list of Partners
   partnersConnection(
@@ -17328,6 +17388,19 @@ type Viewer {
     # The slug or ID of the PartnerCategory
     id: String!
   ): PartnerCategory
+
+  # Retrieve all partner documents for a given partner
+  partnerDocumentsConnection(
+    after: String
+    before: String
+    first: Int
+    last: Int
+    page: Int
+
+    # The slug or ID of the Partner
+    partnerID: String!
+    size: Int
+  ): PartnerDocumentConnection
 
   # A list of Partners
   partnersConnection(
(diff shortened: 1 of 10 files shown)


src/schema/v2/partner.ts
```diff
@@ -482,6 +484,50 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLString,
         resolve: ({ default_profile_id }) => default_profile_id,
       },
+      documentsConnection: {
+        description: "A connection of documents at a partner.",
+        type: partnerDocumentsConnection.connectionType,
+        args: pageable({
+          partnerID: {
```
dblandin, line 491: **question**: If we're nesting within the partner type, I don't think we'll need to support this explicit partnerID query field argument. Should be able to grab the partner ID from the `_root` parameter in the resolve function.

src/schema/v2/partner.ts
```diff
@@ -482,6 +484,50 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
+            type: new GraphQLNonNull(GraphQLString),
+            description: "The slug or ID of the Partner",
+          },
+          page: {
+            type: GraphQLInt,
+          },
+          size: {
+            type: GraphQLInt,
+          },
```
dblandin, line 500: **suggestion**: I think our connection type helpers expose `first/last/after/before` arguments for you. `page/size` can be used when translating the request for Gravity's API but I don't think they're normally exposed here as query field arguments.

src/schema/v2/partnerDocumentsConnection.ts
```diff
@@ -0,0 +1,89 @@
...
+    filename: {
+      type: new GraphQLNonNull(GraphQLString),
+    },
+    title: {
+      type: new GraphQLNonNull(GraphQLString),
+    },
+    size: {
+      type: new GraphQLNonNull(GraphQLInt),
+    },
+    publicUrl: {
```
dblandin, line 34: **suggestion**: `publicURL`

dblandin: **question**: Are all of these fields used in the design? Wondering if we could get by with just `internalID`, `title`, and `publicURL` to start...

dblandin (commented):
Looking good!

dblandin:
Noting that I didn't realize before that there were two other examples of partner-facing query fields for documents scoped to shows and artists. Discussed in Slack [here](https://artsy.slack.com/archives/C02NAJ2CGLW/p1667483560622489?thread_ts=1667409253.637859&cid=C02NAJ2CGLW).

dblandin:
Some notes for tomorrow's pairing session:

- I think an unintended yarn.lock upgrade is causing some relay compilation errors
- Switch to SlugAndInternalIDFields
- Can we define the PartnerDocument/Connection types once and reuse it?
- publicUrl > publicURL
- partnerId > partnerID
- Check if size/page arguments are used or if first/after are sufficient
- Reference pageCursors and helpers in Force/Eigen

cc/ @rajsam003

_schemaV2.graphql
```diff
@@ -17298,7 +17319,8 @@ type Viewer {
     # The slug or ID of the Partner
     partnerID: String!
     size: Int
-  ): PartnerArtistDocumentConnection
+  ): PartnerArtistDocumentsConnectionConnection
```
brainbicycle, line 17322: there is a double `Connection` here I wonder if we can update the types so it is only one?

_schemaV2.graphql
```diff
@@ -11810,6 +11810,19 @@ type Partner implements Node {
...
+    after: String
+    artistID: String
+    before: String
+
+    # Filter documents by ID.
+    documentIDs: [String]
+    first: Int
+    last: Int
+    showID: String
+  ): PartnerDocumentsConnectionConnection
```
brainbicycle, line 11825: double `Connection` in the type name, can we get rid of this?

src/schema/v2/partnerDocumentsConnection.ts
```diff
@@ -0,0 +1,143 @@
...
+
+const FILTER_KEYS = ["documentIDs", "artistID", "showID"]
+
+export const PartnerDocumentsConnection: GraphQLFieldConfig<
+  any,
+  ResolverContext
+> = {
+  description: "Return partner documents if current user has CMS access.",
+  type: connectionWithCursorInfo({
+    name: "PartnerDocumentsConnection",
```
brainbicycle, line 62: I suspect this is where the double `Connection` is coming from, maybe this should be just `PartnerDocuments` and the `Connection` is added automatically?

brainbicycle (changes requested):
this looks great and simpler than before! 🎉 one small change request, the double `Connection` in the type name. I also had a question about consolidating / changing the types for the existing endpoints but I am not sure if it is actually a problem.

src/schema/v2/partnerShowDocumentsConnection.ts
```diff
@@ -14,42 +7,18 @@ import {
...
-})
+import { PartnerDocumentType } from "./partnerDocumentsConnection"
 
 export const PartnerShowDocumentsConnection: GraphQLFieldConfig<
   void,
   ResolverContext
 > = {
   type: connectionWithCursorInfo({
-    nodeType: PartnerShowDocumentType,
+    nodeType: PartnerDocumentType,
```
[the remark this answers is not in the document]
dblandin: Good question! The internal names shouldn't matter to clients unless the client is doing something specific with the `__typename` internal field or dealing with union types via (`... on SomeSpecificType {}`). The more common breaking changes are when in-use fields/arguments are removed/renamed in-place. We've been careful to mark in-use fields as deprecated until Folio (which I believe is the only client using these fields) is updated .

src/schema/v2/partnerDocumentsConnection.ts
dblandin: Updated with https://github.com/artsy/metaphysics/pull/4489/commits/c4ecc372fa614158a3cf9b1f9705c2b249c67b11!

_schemaV2.graphql
dblandin: Updated with https://github.com/artsy/metaphysics/pull/4489/commits/c4ecc372fa614158a3cf9b1f9705c2b249c67b11!

_schemaV2.graphql
dblandin: Updated with https://github.com/artsy/metaphysics/pull/4489/commits/c4ecc372fa614158a3cf9b1f9705c2b249c67b11!

dblandin:
Note that there are a few deprecated fields which will be removed in a subsequent PR after Folio is updated:
- partnerArtistDocumentsConnection > prefer partner.documentsConnection
- partnerShowDocumentsConnection > prefer partner.documentsConnection
- PartnerDocument.size > prefer filesize
- PartnerDocument.publicUrl > prefer publicURL

src/schema/v2/partnerDocumentsConnection.ts
```diff
@@ -0,0 +1,142 @@
...
+        { showID, partnerID },
+        gravityArgs
+      )
+    } else {
+      if (documentIDs) {
+        gravityArgs.document_ids = flatten([args.documentIDs])
+      }
+
+      response = await partnerDocumentsLoader!(partnerID, gravityArgs)
+    }
```
dblandin, line 128: FWIW: I think this resolver could be simplified quite a bit by adding additional `artist_id` and `show_id` filters to the existing `/api/v1/partner/:id/documents` endpoint. Then we could always use the partnerDocumentsLoader and get ride of the switching altogether.

I don't think we're gaining much benefit from having three different endpoints that have very similar behavior:

- /api/v1/partner/:id//documents ([GitHub](https://github.com/artsy/gravity/blob/0c439fed720e6add00fb802fa10881d246a56872/app/api/v1/partners_documents_endpoint.rb))
- /api/v1/partner/:id/artist/:artist_id/documents ([GitHub](https://github.com/artsy/gravity/blob/0c439fed720e6add00fb802fa10881d246a56872/app/api/v1/partners_shows_documents_endpoint.rb))
- /api/v1/partner/:id/show/:show_id/documents ([GitHub](https://github.com/artsy/gravity/blob/0c439fed720e6add00fb802fa10881d246a56872/app/api/v1/partners_artists_documents_endpoint.rb))

dblandin (approved):

brainbicycle (approved):

brainbicycle:
nice one! 💪

PR #4490, opened 2022-11-02
feat(art-quiz) Add DislikedArtwork mutation and REST loaders

TMcMeans:
The type of this PR is: **feat**

This PR solves [GRO-1289]

### Description
The intent of this PR is to hook up dislikes to MP schema, to be utilized for negative signals on user taste in the Art Quiz on the frontend. Please look closely at the specs to verify they are good! Currently as far as REST loaders go, there is only going to be the `DELETE` and `POST` requests for now. Pulling all disliked artworks by user isn't in scope for this project at the moment. 

The overall goal is to begin gathering negative signals from the user about what art they don't like and holding onto that data for future analysis.

can check if artwork is isDisliked 
Screen Shot 2022-11-11 at 11 46 27 AM

can run a mutation to remove/add disliked artworks
Screen Shot 2022-11-11 at 11 44 39 AM

[GRO-1289]: https://artsyproduct.atlassian.net/browse/GRO-1289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -118,6 +123,27 @@ export default (accessToken, userID, opts) => {
       {},
       { method: "PUT" }
     ),
+    dislikeArtworkLoader: gravityLoader(
+      (id) => `collection/disliked-artwork/artwork/${id}`,
+      {},
+      { method: "POST" }
+    ),
+    dislikedArtworkLoader: trackedEntityLoaderFactory(
```
dzucconi, line 131: What's `trackedEntityLoaderFactory` do?

src/schema/v2/artwork/index.ts
```diff
@@ -861,6 +861,15 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           return savedArtworkLoader(_id).then(({ is_saved }) => is_saved)
         },
       },
+      isDisliked: {
+        type: GraphQLBoolean,
```
dzucconi, line 865: I'd mark this as non-nullable. We _never_ want a 3-state boolean (`true|false|null`). You can resolve with `!!is_disliked` to ensure.

damassi (commented):
Nice work on this @TMcMeans! A little round the world in MP.

src/schema/v2/me/dislikedArtworks.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    try {
+      const { body, headers } = await collectionArtworksLoader(
+        COLLECTION_ID,
+        gravityOptions
+      )
+
+      const totalCount = parseInt(headers["x-total-count"] || "0", 10)
+
+      return paginationResolver({ totalCount, offset, page, size, body, args })
```
dzucconi, line 77: 👍

src/schema/v2/artwork/index.ts
dzucconi: (Also I realize none of the other booleans are non-null: they should be — but don't worry about that with this PR)

src/schema/v2/me/dislikeArtworkMutation.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export default mutationWithClientMutationId<any, any, ResolverContext>({
+  name: "DislikeArtwork",
+  description: "Add (or remove) an artwork to (from) a users dislikes.",
+  inputFields: {
+    artworkID: {
+      type: GraphQLString,
+    },
+    remove: {
+      type: GraphQLBoolean,
```
dzucconi, line 14: These arguments look like they should be required with a non-null

src/schema/v2/me/dislikeArtworkMutation.ts
```diff
@@ -0,0 +1,36 @@
...
+  outputFields: {
+    artwork: {
+      type: ArtworkType,
+      resolve: ({ artwork_id }, _, { artworkLoader }) =>
+        artworkLoader(artwork_id),
+    },
+  },
+  mutateAndGetPayload: (
+    { artworkID: artwork_id, remove },
+    /* TODO */
```
dzucconi, line 26: What's the TODO?

laurabeth (commented):
Looks great! I second damon's comments about the non-null booleans. Once those are fixed we can hit go on this.

src/schema/v2/user.ts
```diff
@@ -239,6 +239,44 @@ export const UserType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLBoolean,
         resolve: async ({ data_transfer_opt_out }) => data_transfer_opt_out,
       },
+      dislikedArtworksConnection: {
```
jonallured, line 242: **question [non-blocking]:** unused connection

I don't believe we need to expose this connection because I don't think we have a use-case for showing a user their disliked artworks. But lmk if I'm missing something!

src/lib/loaders/loaders_with_authentication/gravity.ts
TMcMeans: it produces a loader that checks if an entity is followed/saved by batching API requests... but now I'm wondering if this won't actually work for dislikes actually

TMcMeans: if its just a data loader that bundles batch API requests then its fine but if its specifically for follows and saves.... dunno

TMcMeans: gotta look into this more! glad this PR is in WIP for now while I work out the kinks

src/schema/v2/user.ts
TMcMeans: no your right, we only need to collect and store dislikes for now!

src/schema/v2/artwork/index.ts
TMcMeans: ok gotcha, gotta run through and add non-nullables on a bunch of data types for sure

TMcMeans: ok gotcha, gotta run through and add non-nullables on a bunch of data types for sure

src/lib/loaders/loaders_with_authentication/gravity.ts
TMcMeans: so it looks like this loader just bundles API requests and doesn't do much outside of making requests for checks on saves or follows more performant. I've tested this out and it seems fine, so I'm going to keep using it.

src/schema/v2/user.ts
damassi: Yup, this kind of thing should live on `me`! All the goodies are there

src/lib/loaders/loaders_with_authentication/gravity.ts
damassi: Sounds good - it follows existing patterns!

damassi (approved):
Fantastic 💯

TMcMeans:
Please squash and merge when ready!

src/schema/v2/me/dislikeArtworkMutation.ts
```diff
@@ -0,0 +1,41 @@
...
+      type: new GraphQLNonNull(GraphQLBoolean),
+    },
+  },
+  outputFields: {
+    artwork: {
+      type: ArtworkType,
+      resolve: ({ artwork_id }, _, { artworkLoader }) =>
+        artworkLoader(artwork_id),
+    },
+  },
```
dzucconi, line 23: This is actually minor but you're gonna want to output `me` as well. Generally speaking you always want to return the thing you're mutating on since we use fragments to update state in the Relay cache.

src/schema/v2/me/dislikeArtworkMutation.ts
```diff
@@ -0,0 +1,41 @@
...
+  ) => {
+    if (!deleteDislikedArtworkLoader || !dislikeArtworkLoader) {
+      return new Error("You need to be signed in to perform this action")
+    }
+
+    const loader = remove ? deleteDislikedArtworkLoader : dislikeArtworkLoader
+    try {
+      return loader(artwork_id, { user_id: userID }).then(() => ({
+        artwork_id,
+      }))
```
dzucconi, line 36: So this try and catch doesn't do anything without awaiting on the loader. This actually would never throw (unless something internal to the loader thru, not the actual promise rejection)

src/schema/v2/me/dislikeArtworkMutation.ts
```diff
@@ -0,0 +1,41 @@
...
+    } catch (error) {
+      throw error
+    }
```
dzucconi, line 39: This doesn't actually do anything since you're just catching the error to throw it again.

dzucconi: This still just re-throws the error — which is maybe fine; but probably not: how do we want this to error? With what message? What's useful during the mutation?

TMcMeans: is this blocking? please clarify the issue here

dzucconi: I would just completely remove the try/catch here

laurabeth (dismissed):

PR #4494, opened 2022-11-03
chore: return `null` in TaxInfo when artwork is NOT eligible for on-platform transaction

dimatretyak:


src/schema/v2/artwork/utilities.ts
```diff
@@ -111,3 +111,15 @@ export const getFigures = ({
     return [...sortedImages, ...videos]
   }
 }
+
+export const isEligibleForOnPlatformTransaction = ({
```
dblandin, line 115: Nice! This is definitely a better place for this utility function 💯

dblandin (approved):
LGTM 👍🏼

PR #4500, opened 2022-11-07
feat(CX-3053): define logic to show the embedded carousel within metaphysics

dariakoko:
Added a new field to HomePageType
image


src/schema/v2/home/home_page_my_collection_onboarding_module.ts
```diff
@@ -0,0 +1,52 @@
...
+import { ResolverContext } from "types/graphql"
+
+const MyCollectionOnboardingModuleType = new GraphQLObjectType<
+  any,
+  ResolverContext
+>({
+  name: "HomePageMyCollectionOnboardingModule",
+  fields: {
+    showMyCollectionCard: {
+      type: GraphQLBoolean,
```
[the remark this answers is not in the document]
MounirDhahri: we will need then to return false in line 23 in case the artworks count is bigger than 3

src/schema/v2/home/home_page_my_collection_onboarding_module.ts
```diff
@@ -0,0 +1,52 @@
...
+          return collectionLoader("my-collection", {
+            user_id: me.id,
+            private: true,
+          }).then((res) => {
+            return res.artworks_count > 3
+          })
+        })
+      },
+    },
+  },
```
MounirDhahri, line 44: **issue(blocker):** I think you copied the code here without adjusting it, for now we need to return true always according to product specs

MounirDhahri (changes requested):
Thanks for almost wrapping up this work Daria. I left some comments, please let me know if you need any support. Also, would you mind adding a screenshot from Insomnia showing the query to make it easier for other folks to get what the PR does quickly

src/schema/v2/home/home_page_my_collection_onboarding_module.ts
dariakoko: Hm, no I didn't
Will revisit this tomorrow

MounirDhahri: True, but I believe that's the logic only for my collection onboarding card. The swa card is supposed to be always visible I believe

src/schema/v2/home/__tests__/home_page_my_collection_onboarding_module.test.js
```diff
@@ -0,0 +1,38 @@
...
+        Promise.resolve({
+          artworks_count: 10,
+        }),
+    })
+
+    const results = {
+      homePage: {
+        onboardingModule: {
+          showMyCollectionCard: response,
+          showSWACard: response,
```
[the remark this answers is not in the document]
dariakoko: Oh wow yes, you're right, the test is not doing anything useful actually

MounirDhahri (approved):
Nice let's get this merged

PR #4505, opened 2022-11-07
fix(artist): fixes related shows pagination

dzucconi:
Switches to use our `paginationResolver` and adds support for a `page` param.

### DIFF
--- a/src/schema/v2/artist/shows.ts
+++ b/src/schema/v2/artist/shows.ts
@@ -4,15 +4,13 @@ import {
   GraphQLString,
   GraphQLBoolean,
   GraphQLFieldConfig,
-  GraphQLFieldConfigArgumentMap,
 } from "graphql"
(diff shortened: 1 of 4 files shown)


src/schema/v2/artist/shows.ts
```diff
@@ -34,92 +33,55 @@ export function showsWithDenyListedPartnersRemoved(shows) {
...
-            pageCursors: createPageCursors(pageOptions, totalCount),
-            totalCount,
-          },
-          connectionFromArraySlice(body, args, {
-            arrayLength: totalCount,
-            sliceStart: offset,
-          }),
-          {
-            pageInfo: {
-              hasPreviousPage: page > 1,
```
dzucconi, line 118: Oh interesting, it was manually computing this value. It's funny how this makes intuitive sense. I wonder why they didn't union type this thing.

dzucconi: They = Facebook

PR #4506, opened 2022-11-07
fix(video): fix ref to `url` in resolver

damassi:
Fixes an issue with a field rename in the Video resolver.

### DIFF
--- a/src/schema/v2/types/Video.ts
+++ b/src/schema/v2/types/Video.ts
@@ -8,28 +8,37 @@ import {
 import uuid from "uuid/v5"
 import { ResolverContext } from "types/graphql"
 
-export const VideoType = new GraphQLObjectType<any, ResolverContext>({
-  name: "Video",
-  description: "An object containing video metadata",
-  fields: {
-    id: {
-      type: new GraphQLNonNull(GraphQLID),
-      resolve: ({ url }) => {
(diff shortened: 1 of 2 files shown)


src/schema/v2/types/Video.ts
```diff
@@ -8,28 +8,37 @@ import {
...
-    id: {
-      type: new GraphQLNonNull(GraphQLID),
-      resolve: ({ url }) => {
-        return uuid(url, uuid.URL)
+interface VideoTypeProps {
+  id: string
+  playerUrl: string
+  height: number
+  width: number
+}
```
dzucconi, line 16: It'd be good to annotate the return of the function that constructs this object with this type. (Or import the `ReturnType` of that function here instead.)

damassi: Not sure that I follow? Like, the return type of `uuid`? (Can follow-up)

dzucconi: `getFigures` returns:

dzucconi: so make that like: `NormalizedImageData | VideoData` and then import `VideoData` here.

damassi: Oh I see! Will do 👍

PR #4513, opened 2022-11-10
feat(FX-4254): recognize PartnerShowPublishedActivity events

nickskalkin:
- [Jira](https://artsyproduct.atlassian.net/browse/FX-4254)
- [Gravity PR](https://github.com/artsy/gravity/pull/15818)

Introducing new Activity panel notification type

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11541,6 +11541,7 @@ input NotificationPreferenceInput {
 enum NotificationTypesEnum {
   ARTWORK_ALERT
   ARTWORK_PUBLISHED
+  PARTNER_SHOW_PUBLISHED
   VIEWING_ROOM_PUBLISHED
 }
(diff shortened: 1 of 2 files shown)


nickskalkin:
@dimatretyak am I right that we don't need to add any changes to Force / Eigen? I couldn't find anything

dimatretyak (dismissed):

dimatretyak:
> @dimatretyak am I right that we don't need to add any changes to Force / Eigen? I couldn't find anything

Yes, that's right. We would need to add changes if we added a new tab that only displayed `PARTNER_SHOW_PUBLISHED` notifications (for example, how do [we do this](https://github.com/artsy/force/blob/main/src/Components/Notifications/NotificationsList.tsx#L226-L228) for "Alerts" tab)

PR #4514, opened 2022-11-10
feat(video): gets video response from vimeo api

dzucconi:
I need to flesh out some specs and test this with incomplete status works to see how the response changes but, this PR changes the resolver to favor fetching from the Vimeo API vs constructing a response based on the URL we've saved.

Note that we'll need an authenticated scope API token to access these videos. I don't think there's any real risk here but just noting this about the `VIMEO_TOKEN`.


jonallured:
I'm surprised to see this here in MP rather than in Gravity? 🤔

damassi:
The way I see it this isn't necessary with the work we're doing over in Gravity (which should be fairly quick).

dzucconi:
Yeah this was to fix the bleeding of the extraneous whitespace (It's very bad on mobile).

If we think that won't take more than a few days then this isn't needed (but is also fine IMO). I suppose I'd rather not add a service integration if we don't need it.

PR #4521, opened 2022-11-15
feat(FX-4437): Add curated marketing collections query field

nickskalkin:
[Jira](https://artsyproduct.atlassian.net/browse/FX-4437)

Adds query field to for new Artsy curated collection section

Screenshot 2022-11-16 at 13 39 38

# Sample Query

```graphql
{
    curatedMarketingCollections(first: 5) {
        slug
        title
        thumbnailImage {
	    resized(width: 200) {
	        url
	    }
        }
    }
}
```

# Response

```json
{
	"data": {
		"curatedMarketingCollections": [
			{
				"slug": "photography",
				"title": "Photography",
				"thumbnailImage": {
					"resized": {
						"url": "https://d7hftxdivxxvm.cloudfront.net?resize_to=width&width=200&quality=80&src=https%3A%2F%2Fd32dm0rphc51dk.cloudfront.net%2F_zWRgEXoyJtTVaM6NznROw%2Fnormalized.jpg"
					}
				}
			},
			{
				"slug": "painting",
				"title": "Painting",
				"thumbnailImage": null
			},
			{
				"slug": "artists-on-the-rise",
				"title": "Artists on the Rise",
				"thumbnailImage": null
			},
			{
				"slug": "trending-this-week",
				"title": "Trending This Week",
				"thumbnailImage": null
			},
			{
				"slug": "trove-editors-picks",
				"title": "Trove: Editors' Picks",
				"thumbnailImage": null
			}
		]
	}
}
```

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10075,6 +10075,7 @@ type MarketingCollection {
   showHeaderArtworksRail: Boolean!
   slug: String!
   thumbnail: String
+  thumbnailImage: Image
   title: String!
   updatedAt: ISO8601DateTime!
 }
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -13289,12 +13289,7 @@ type Query {
     # The ID of the Credit Card
     id: String!
   ): CreditCard
-  curatedMarketingCollections(
-    after: String
-    before: String
-    first: Int
-    last: Int
-  ): [MarketingCollection]
+  curatedMarketingCollections(size: Int): [MarketingCollection]
   departments: [Department!]!
 
   # Get an image info
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -13288,6 +13288,12 @@ type Query {
     # The ID of the Credit Card
     id: String!
   ): CreditCard
+  curatedMarketingCollections(
+    after: String
+    before: String
+    first: Int
+    last: Int
+  ): [MarketingCollection]
   departments: [Department!]!
 
   # Get an image info
--- a/src/lib/stitching/gravity/v2/stitching.ts
+++ b/src/lib/stitching/gravity/v2/stitching.ts
@@ -169,6 +169,10 @@ export const gravityStitchingEnvironment = (
         ): [MarketingCollection]
       }
 
+      extend type Query {
+        curatedMarketingCollections(first: Int, last: Int, after: String, before: String): [MarketingCollection]
+      }
+
       extend type System {
         algolia: Algolia
       }
@@ -763,6 +767,37 @@ export const gravityStitchingEnvironment = (
               context,
               info,
             })
+          },
+        },
+      },
+      Query: {
+        curatedMarketingCollections: {
+          fragment: gql`
+            ...on Viewer {
+              __typename
+            }
+          `,
+          resolve: async (_parent, _args, context, info) => {
+            try {
+              return await info.mergeInfo.delegateToSchema({
+                schema: gravitySchema,
+                operation: "query",
+                fieldName: "marketingCollections",
+                args: {
+                  slugs: [
+                    "trending-this-week",
+                    "artists-on-the-rise",
+                    "trove-editors-picks",
+                    "painting",
+                    "photography",
+                  ],
+                },
+                context,
+                info,
+              })
+            } catch (error) {
+              return []
+            }
           },
         },
       },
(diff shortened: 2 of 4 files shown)


src/lib/stitching/gravity/v2/stitching.ts
```diff
@@ -287,6 +292,31 @@ export const gravityStitchingEnvironment = (
...
+          fragment: gql`
+          ... on MarketingCollection {
+            image_url: thumbnail
+          }
+          `,
+          resolve: async ({ image_url }, args, context, info) => {
+            if (image_url) {
+              context.imageData = {
+                image_url,
+              }
```
dimatretyak, line 305: **question**: maybe it would be a good idea to use `normalizeImageData` by analogy as [here](https://github.com/artsy/metaphysics/blob/a8f7199c29da02ad89233ca01d4aabaa253b98fc/src/schema/v2/article/index.ts#L385)?

dimatretyak (dismissed):

_schemaV2.graphql
```diff
@@ -10075,6 +10075,7 @@ type MarketingCollection {
   showHeaderArtworksRail: Boolean!
   slug: String!
   thumbnail: String
+  thumbnailImage: Image
```
dblandin, line 10078: **suggestion**: Awesome that we can upgrade the image URL here. What do you think of marking `thumbnail` as deprecated in favor of `thumbnailImage.url` or `thumbbailImage.imageURL`? Is there anyway to do that in Metaphysics or would that have to be a Gravity GraphQL schema change?

_schemaV2.graphql
```diff
@@ -13288,6 +13289,12 @@ type Query {
     # The ID of the Credit Card
     id: String!
   ): CreditCard
+  curatedMarketingCollections(
+    after: String
+    before: String
+    first: Int
+    last: Int
+  ): [MarketingCollection]
```
dblandin, line 13297: **suggestion**: Since we're returning an array of MarketingCollection entries here, it seems disingenuous to use the after/before/first/last connection-style args. I think we should expose `size: Int` instead and pass that through to the delegated query.

src/lib/stitching/gravity/v2/stitching.ts
```diff
@@ -287,6 +292,31 @@ export const gravityStitchingEnvironment = (
...
+            } else {
+              context.imageData = null
+            }
+
+            return info.mergeInfo.delegateToSchema({
+              args,
```
dblandin, line 311: **question**: Do these args contain any data? Or is the image type fulfilled entirely by the context manipulation above?

src/lib/stitching/gravity/v2/stitching.ts
```diff
@@ -766,6 +796,37 @@ export const gravityStitchingEnvironment = (
...
         },
       },
+      Query: {
+        curatedMarketingCollections: {
+          fragment: gql`
+            ...on Viewer {
+              __typename
+            }
+          `,
+          resolve: async (_parent, _args, context, info) => {
```
dblandin, line 806: **suggestion**: I think that we want to pass these args to the delegated query below so that `size` works as expected (see args comment above).

dblandin (commented):
Looking good! Left a few comments. Overall I think it's fine to model this is an array response based on the Gravity query fields but it might be useful to discuss providing connection-support for these fields in the future.

_schemaV2.graphql
nickskalkin: good idea to deprecate old field. I couldn't find a way how to deprecate a field in the stitching code. I think we can do some manual transformations, but I would better submit a PR to gravity. How does it sound?

_schemaV2.graphql
nickskalkin: makes sense, I think I wanted to still follow connection conventions, but it doesn't makes sense I guess

src/lib/stitching/gravity/v2/stitching.ts
nickskalkin: removed args

src/lib/stitching/gravity/v2/stitching.ts
nickskalkin: I don't know how it works, but size limiting was working even without explicitly passing `args` to `delegateToSchema`. I've added `...args` though to make is explicit

_schemaV2.graphql
dblandin: Sounds good 👍🏼

dblandin (approved):
Looks great!! 💯

dimatretyak (approved):

PR #4522, opened 2022-11-15
feat: add a curated trending artists connection

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-4451

This adds a new `curatedTrendingArtists` root-level field in order to expose a timely list of Artsy-suggested artists for use in various client-side contexts — starting with search suggestions in Eigen.

The list of artists:

- is sourced from the [Trending This Week](https://www.artsy.net/collection/trending-this-week) marketing collection
  - for an idea of what that would feel like, see [here](https://artsy.slack.com/archives/C9SATFLUU/p1668542926931289)
- is exposed as a GraphQL connection
- has a built-in daily shuffle, such that each user will get a different sampling of artists every day
  - the shuffle attempts to be time zone aware, so that the shuffle happens at midnight in the user's time zone

Since seeded random numbers are [not a thing](https://artsy.slack.com/archives/CP9P4KR35/p1668534209604179) in vanilla JS, this PR borrows some patterns from Force to do stabled seeded random number generation and shuffling (https://github.com/artsy/metaphysics/pull/4522/commits/90a996401f797b47acc009970a6c6d4156878c38)

It adds on top of that a date-stable, timezone-aware shuffling method for convenience (https://github.com/artsy/metaphysics/pull/4522/commits/d6ea35ff672c4b3f551255dcc520161454a7c1ce)

### Request

```graphql
{
  curatedTrendingArtists(first: 3) {
    edges {
      node {
        slug
        name
      }
    }
  }
}
```

### Response

```json
{
  "data": {
    "curatedTrendingArtists": {
      "edges": [
        {
          "node": {
            "slug": "devin-troy-strother",
            "name": "Devin Troy Strother"
          }
        },
        {
          "node": {
            "slug": "andy-warhol",
            "name": "Andy Warhol"
          }
        },
        {
          "node": {
            "slug": "damien-hirst",
            "name": "Damien Hirst"
          }
        }
      ]
    }
  }
}
```

<details>
  <summary>
    <strong>See here (and hidden comments below) for earlier discussion about alternative fields, schema design etc</strong>
  </summary>

*Very WIP-y at the moment for the reasons outlined below.*

Adds a new root-level `trendingArtists` field.

This new field implements the proposed strategy of sourcing a list of trending artists from the [Trending This Week collection](https://www.artsy.net/collection/trending-this-week), which is actively curated on a weekly basis.

In doing this I discovered that there are several different versions of a "trending artist" concept floating around in Metaphysics.

<details>
  <summary>
    <strong>Open me up for a tour of the ~4 such existing fields…</strong>
  </summary>

### 1. An obsolete root-level `trendingArtists` field

- Looks like it dates back to the v1 schema, 
- Commented out in the v2 schema (which is why the root-level `trendingArtists` field name was available for me to squat on)
- Data is/was sourced directly from Delta via a `deltaLoader` (not to be confused with [beltalowdas](https://expanse.fandom.com/wiki/Beltalowda))

### 2. A home page module at `homePage.artistModule(key: TRENDING)`

- Data is sourced via the `trendingArtistsLoader`, which hits Gravity's `artist/trending` endpoint

### 3. A gene-specific  `gene.trendingArtists` 

- Data is also sourced via the `trendingArtistsLoader`, which hits Gravity's `/artists/trending` endpoint, and passes a `gene` param into that endpoint

### 4. A `-trending` sort on the root-level `artists` and `artistsConnection` fields

- Data is sourced via the `artistsLoader`, which hits Gravity's `/artists` endpoint with a custom `-trending` sort

</details>

…So we've got four fields kinda named "trending artists," sourced in four somewhat different ways (though I think they all back up onto [artsy/delta](https://github.com/artsy/metaphysics/pull/4522))

And we are going to add a fifth, sourced differently 😅

### Questions

- What is the **best name** for this new field? We've been calling it "trending artists" so far, but that name seems a bit overloaded? Are there good alternatives? "Curated trending artists"?

- What is the **best location** in the graph for this new field? At the root? I'm not sure if any other existing nesting would make sense e.g. `{ trending { artists } }`

- Wild left-field thought — should we consider actually leaning into the existing trending artist microservice (https://github.com/artsy/delta) but add more signals to it, such as the fact that an artist is in Trending This Week (which we might favor with a huge boost)?

</details>

### TODO

- [x] settle field name
- [x] settle field location
- [x] add date-stable shuffling ([a tip](https://artsy.slack.com/archives/CP9P4KR35/p1668534209604179))
- [x] specs

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -13249,6 +13249,14 @@ type Query {
     id: String!
   ): CreditCard
   curatedMarketingCollections(size: Int): [MarketingCollection]
+
+  # A list of trending artists. Inferred from a manually curated collection of trending artworks.
+  curatedTrendingArtists(
+    after: String
+    before: String
+    first: Int
+    last: Int
+  ): ArtistConnection
   departments: [Department!]!
 
   # Get an image info
@@ -17057,6 +17065,14 @@ type Viewer {
     # The ID of the Credit Card
     id: String!
   ): CreditCard
+
+  # A list of trending artists. Inferred from a manually curated collection of trending artworks.
+  curatedTrendingArtists(
+    after: String
+    before: String
+    first: Int
+    last: Int
+  ): ArtistConnection
   departments: [Department!]!
 
   # Get an image info
--- a/src/lib/__tests__/shuffle.test.ts
+++ b/src/lib/__tests__/shuffle.test.ts
@@ -53,3 +53,57 @@ describe("seededShuffle", () => {
     })
   })
 })
+
+describe("dailyShuffle", () => {
+  describe("with no time zone specified", () => {
+    it("produces a stable shuffle for an entire UTC day", () => {
+      const veryEarly = new Date("2022-11-15T01:00:00Z").valueOf()
+      const veryLate = new Date("2022-11-15T23:00:00Z").valueOf()
+      const nextDay = new Date("2022-11-16T01:00:00Z").valueOf()
+
+      // early in the day - same
+      jest.spyOn(global.Date, "now").mockImplementation(() => veryEarly)
+      expect(dailyShuffle([1, 2, 3, 4, 5])).toEqual([4, 2, 5, 3, 1])
+
+      // late in the day - same
+      jest.spyOn(global.Date, "now").mockImplementation(() => veryLate)
+      expect(dailyShuffle([1, 2, 3, 4, 5])).toEqual([4, 2, 5, 3, 1])
+
+      // next day - different
+      jest.spyOn(global.Date, "now").mockImplementation(() => nextDay)
+      expect(dailyShuffle([1, 2, 3, 4, 5])).toEqual([3, 4, 1, 5, 2])
+    })
+  })
+
+  describe("with a valid time zone specified", () => {
+    it("produces a stable shuffle for an entire day in that time zone", () => {
+      const veryEarly = new Date("2022-11-15T01:00:00-05:00").valueOf()
+      const veryLate = new Date("2022-11-15T23:00:00-05:00").valueOf()
+      const nextDay = new Date("2022-11-16T01:00:00-05:00").valueOf()
+      const nycTime = "America/New_York"
+
+      // early in the day - same
+      jest.spyOn(global.Date, "now").mockImplementation(() => veryEarly)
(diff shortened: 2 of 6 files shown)


dblandin:
Some loose thoughts:
- which of the supported events documented in Delta's readme are actively firing?
- if we were to take a Delta approach would we have to emit an event per artist per [time window]? For example per day that an artist is featured in a targeted manually curated collection?
- I like the idea of utilizing the existing artistsConnection `-trending` sort if possible.
- Note that I think there is yet another version of this which we were experimenting with for artworks

anandaroop:
> which of the supported events documented in Delta's readme are actively firing?

I wondered the same, and tbh I don't see much in Gravity. There's [this code](https://github.com/artsy/gravity/blob/main/app/models/domain/fair_action.rb#L45) for example, in a model I hadn't known about at all 🤷🏽 

> if we were to take a Delta approach would we have to emit an event per artist per [time window]? For example per day that an artist is featured in a targeted manually curated collection?

Yeah that would be part of the complexity — synthesizing some kind of event stream out of the curation activity for that specific collection. Could prob go in many directions, such as what you suggest. 

> I like the idea of utilizing the existing artistsConnection `-trending` sort if possible.

I like this too, but I guess this is complicated by the fact that we have such a small (~80 right now) list of of artists. I feel like we would expect the `artistsConnection`, invoked without any filters, to eventually return all artists, rather than just a subset?

---

Delta in general feels very blackbox to me. Could be a worthy project to get to know it better, vet the incoming signals / outgoing results, bring it up to date etc. But that feels like a longer-term project.

In the meantime, we've got some stakeholder [excitement about the newly sourced data](https://artsy.slack.com/archives/C9SATFLUU/p1668592701018679?thread_ts=1668542926.931289&cid=C9SATFLUU), so I'm inclined to ship this, despite the ambiguity in the schema, which I hope to minimize.

So for now, I'm leaning toward something like…

```graphql
curatedTrendingArtists(first: 5) {
  edges {
    node {
      slug
      name
      # etc
    }
  }
}
```

That is:

- root-level `curatedTrendingArtists` field
- named to denote that this is the result of manual curation
- modeled as connection rather than list, for future-proofing

dblandin (approved):
Very cool 💯

anandaroop:
Merging!

PR #4523, opened 2022-11-15
chore(partner): consolidate partner files in /partner folder

damassi:
Small cleanup PR that moves all partner-related files into a `partner` folder and out of the root schema directory. 

cc @artsy/negotiate-devs


araujobarret (dismissed):
🙏

PR #4532, opened 2022-11-17
chore: expose `totalCount` field

dimatretyak:
### Demo
Artsy - Metaphysics (Development) – Sandbox 2022-11-17 20-35-38


dblandin (approved):
Good catch 💯

PR #4535, opened 2022-11-17
feat(sendConversationMessage): Expand mutation to include full API

damassi:
Addresses [NX-3326]

This updates the `sendConversationMessage` mutation API to include all available fields, and is backwards compatible. 

Added the `User-Agent: graphiql-app;Volt/cms.artsy.net` header and things worked well, too. 

```graphql
mutation convo($input: SendConversationMessageMutationInput!) {
  sendConversationMessage(input: $input) {
    conversation {
      from {
        email
      }
    }
  }
}
```

**Payload:**

```json
{
  "input":{
    "attachments": [{
      "id": "foo",
      "name": "artwork.pdf",
      "size": "20",
      "type": "application/pdf",
      "url": "https://yahoo.com"
    }],
    "bodyText": "testing mutation", 
    "bodyHTML": "<b>some html</b>",
    "from":"damassi.pappas@gmail.com", 
    "fromId": "553ffd847261691a86190a00",
    "id":"89370", 
    "replyAll": true,
    "replyToMessageID":"94786",
    "to: ["foo@bar.com"]
  }
}
```

Lil attachment example image:  
Screenshot 2022-11-17 at 4 02 37 PM

The one issue i'm noticing above is that when the `to` field is included in the payload it triggers [this radiation error](https://github.com/artsy/radiation/blob/66f5277920e329bbecb352fe3a20c3b9ae7261f9/app/services/conversation_service.rb#L42). This is due to the conditions described in [this slack thread](https://artsy.slack.com/archives/C9YNS4X32/p1668728080830159), so things should be ok in practice. 

cc @artsy/negotiate-devs 

[NX-3326]: https://artsyproduct.atlassian.net/browse/NX-3326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/index.ts
```diff
@@ -199,7 +199,7 @@ const graphqlServer = graphqlHTTP((req, res, params) => {
     }),
     validationRules,
     extensions: ({ document, result }) =>
-      createExtensions(document, result, requestID),
+      createExtensions(document, result, requestID, userAgent),
```
damassi, line 202: This adds some more useful debug information to GraphiQL output

src/schema/v2/conversation/send_message_mutation.ts
```diff
@@ -60,35 +117,53 @@ export default mutationWithClientMutationId<any, any, ResolverContext>({
...
+      .then(({ id: newMessageID }) => {
+        return conversationLoader(args.id).then((updatedConversation) => {
+          return {
+            conversation: updatedConversation,
+
+            // Because Impulse does not have the full new message object available
+            // immediately, we return an optimistic response so the mutation can
+            // return it too.
+            newMessagePayload: {
+              attachments: args.attachments,
```
mzikherman, line 151: `args.attachments` might be `null` or `undefined`, maybe needs a `|| []`?

src/schema/v2/conversation/send_message_mutation.ts
```diff
@@ -1,46 +1,103 @@
...
   name: "SendConversationMessageMutation",
   description: "Appending a message to a conversation thread",
   inputFields: {
-    id: {
+    attachments: {
+      description: "Attachments to the message",
+      type: new GraphQLList(
+        new GraphQLInputObjectType({
+          name: "ConversationMessageAttachmentInput",
+          fields: {
```
mzikherman, line 46: I assume this schema makes sense as something coming from Impulse (like `id` isn't required _seems_ maybe off but if it matches Impulse)

mzikherman (dismissed):
Looks good!

src/schema/v2/conversation/send_message_mutation.ts
```diff
@@ -1,46 +1,103 @@
...
+      description: "Sender user id",
+      type: GraphQLString,
+    },
+    id: {
+      description: "The id of the conversation to be updated",
       type: new GraphQLNonNull(GraphQLString),
     },
+    replyAll: {
+      description: "Reply to all",
+      type: GraphQLBoolean,
```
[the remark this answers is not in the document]
damassi: NonNull will break clients, and has a `defaultValue` of true!

src/schema/v2/conversation/send_message_mutation.ts
damassi: Oh good idea, will fix

src/schema/v2/conversation/send_message_mutation.ts
damassi: All that's strictly needed are the the fields required here, though there might be some updating here in near-term.

src/index.ts
damassi: Screenshot 2022-11-17 at 4 14 10 PM

PR #4549, opened 2022-11-23
chore: update gravity schema

oxaudo:
Thanks for the pointer @starsirius


src/data/gravity.graphql
```diff
@@ -2706,6 +2706,11 @@ type ViewingRoom {
   Datetime after which the viewing room is no longer viewable
   """
   endAt: ISO8601DateTime
+
+  """
+  Datetime when viewing room first viewable
+  """
+  firstLiveAt: ISO8601DateTime
```
starsirius, line 2713: You'll also need to run `yarn dump:local` to include this change in MP's own schema.

PR #4553, opened 2022-11-24
feat(volt-v2): add to schema push

damassi:
This adds `volt-v2` to schema push list so that it receives schema changes when MP changes. 

cc @artsy/negotiate-devs


damassi:
Ah! just now opening my PR viewer: https://github.com/artsy/metaphysics/pull/4551

scripts/push-schema-changes.js
```diff
@@ -76,6 +76,7 @@ async function main() {
       { repo: "volt-v2" },
```
patrinoua, line 76: @damassi it looks like this is already here?

araujobarret: He already reverted 😬

patrinoua: perfect!

PR #4567, opened 2022-12-01
fix(partner): do not return invalid href

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-4484

Auction artwork pages were rendering links to auction profile pages that were unreachable and 404-ing

In https://github.com/artsy/force/pull/11465 we addressed that on the client side.

This goes a step further and ensures that we don't return a partner `href` for partner types that are not eligible for public pages in the first place, such as Auction, Demo, Private Dealer, etc.

(Note that CI is failing on some python build step at the moment; tests seem ok.)


dblandin (approved):
Nice! LGTM 👍🏼

PR #4579, opened 2022-12-06
chore: access sale_date from user purchase

kajatiger:
We want to display the date of the purchase on the User-Purchases tab on the forque side and need to access the data through metaphysics.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -17044,6 +17044,7 @@ type UserPurchasesEdge {
   # The item at the end of the edge
   node: Artwork
   ownerType: String
+  saleDate: String
   salePrice: Float
   source: String
 }
--- a/src/schema/v2/__tests__/user.test.ts
+++ b/src/schema/v2/__tests__/user.test.ts
@@ -477,6 +477,7 @@ describe("User", () => {
         {
           owner_type: "SaleArtwork",
           sale_price: 1000.9999,
+          sale_date: "2018-09-26 04:00:00 UTC",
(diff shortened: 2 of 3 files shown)


ovasdi (changes requested):
👋, As far as mapping the field this seems _mostly_ correct, however, the type is wrong for a `date` field.

**Suggestion**: Test changes locally using a graphql client. It would have revealed the following: `Float cannot represent non numeric value: \"2022-06-11T20:00:00+00:00\""`

The tests may have also revealed the type issue but, they aren't complete. May be worth checking if tests would have thrown the expected error.

See this [github (MP) search](https://github.com/search?q=repo%3Aartsy%2Fmetaphysics+date%28%28&type=code) for how date fields are mapped. It might also work with `GraphQLString` but I think using `date()` is the convention.

---

A couple of points regarding the test:
* `sale_date` was added in the first fixture (_SaleArtwork_) but not the second (_ArtworkInquiryRequest_). May be worth adding to both?
* [the test query](https://github.com/artsy/metaphysics/blob/9a1795e2496b68885c6bdf02f0017d30955c2059/src/schema/v2/__tests__/user.test.ts#L450-L453) was not updated to include the new field
* [the expectations](https://github.com/artsy/metaphysics/blob/9a1795e2496b68885c6bdf02f0017d30955c2059/src/schema/v2/__tests__/user.test.ts#L542-L578) do not include the new field with value

ovasdi (approved):
LGTM

PR #4580, opened 2022-12-06
feat: allow filtering for upcoming lots

MounirDhahri:
This PR resolves [CX-3222]

Related Diffusion changes: https://github.com/artsy/diffusion/pull/653

**What's still required:**
- Update params according to the changes in Diffusion

@egdbear I am making some changes to the API. thanks for checking this out

[CX-3222]: https://artsyproduct.atlassian.net/browse/CX-3222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10371,6 +10371,9 @@ type Me implements Node {
     # Filter auction results by earliest created at year
     earliestCreatedYear: Int
     first: Int
+
+    # Include upcoming auction results
+    includeUpcoming: Boolean = true
     last: Int
 
     # Filter auction results by latest created at year
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1329,9 +1329,6 @@ type Artist implements EntityWithFilterArtworksConnectionInterface & Node & Sear
     earliestCreatedYear: Int
     first: Int
 
-    # include upcoming auction results
-    includeUpcoming: Boolean = true
-
     # Filter by artwork title or description keyword search
     keyword: String
     last: Int
@@ -1344,6 +1341,9 @@ type Artist implements EntityWithFilterArtworksConnectionInterface & Node & Sear
 
     # When true, will only return records for allowed artists.
     recordsTrusted: Boolean = false
+
+    # Show only upcoming auction results, set to false to show only past results, by default both are shown
+    showUpcoming: Boolean
 
     # Filter auction results by Artwork sizes
     sizes: [ArtworkSizes]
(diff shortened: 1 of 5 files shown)


_schemaV2.graphql
```diff
@@ -1342,6 +1342,9 @@ type Artist implements EntityWithFilterArtworksConnectionInterface & Node & Sear
     # When true, will only return records for allowed artists.
     recordsTrusted: Boolean = false
 
+    # Show only upcoming auction results, set to false to show only past results, by default both are shown
+    showUpcoming: Boolean
```
olerichter00, line 1346: Interesting, this param can have three different values 🤔 Is this really required? I wonder if it could be confusing that `false` and `undefined` mean completely different things

olerichter00 (dismissed):
Looks good! I just wonder if we really need these three possibilities (only upcoming, only previous, all).

_schemaV2.graphql
MounirDhahri: Actually the designs in Eigen want to include either - All (past + upcoming) or only past auction results

egdbear: All right if we specify "&upcoming=true" parameter we'll get only upcoming, otherwise all :)

src/schema/v2/me/myCollectionAuctionResults.ts
```diff
@@ -93,9 +93,21 @@ const MyCollectionAuctionResults: GraphQLFieldConfig<any, ResolverContext> = {
...
       }
 
-      const { total_count, _embedded } = await auctionLotsLoader(
-        auctionLotsLoaderArgs
-      )
+      let optionalDiffusionArgs = {}
+
+      // We should only inject upcoming if it's explicitly set to true or false
+      // See https://github.com/artsy/diffusion/pull/653
+      if (options.showUpcoming === true || options.showUpcoming === false) {
```
egdbear, line 100:
```suggestion
      if (options.showUpcoming === true) {
```

_schemaV2.graphql
```diff
@@ -1329,6 +1329,9 @@ type Artist implements EntityWithFilterArtworksConnectionInterface & Node & Sear
     earliestCreatedYear: Int
     first: Int
 
+    # Include upcoming auction results
+    includeUpcoming: Boolean = true
```
egdbear, line 1333: Do we want by default to have included upcoming?

MounirDhahri: yes, was going to make this change right now

egdbear (dismissed):

PR #4583, opened 2022-12-06
chore(conversations): Address follow-up PR feedback

damassi:
This addresses follow-up PR feedback from https://github.com/artsy/metaphysics/pull/4536


mzikherman (approved):
Sweet!

PR #4590, opened 2022-12-07
feat: add trim option to strip away lot label past first alphanumeric chunk

dblandin:
We've encountered cases where sale artworks in Gravity have non-numeric characters to store metadata related to the lot. This isn't a recommended use of the lot label field. As an alternative, we've recommended that colleagues use the `blurb` field which is also exposed on the same surfaces across web and app ([Slack](https://artsy.slack.com/archives/C02RAMW9QQJ/p1664364783118159)).

This PR updates the lotLabel resolver to accept an optional `trim` argument (default=false) to strip away anything past the first alphanumeric chunk from the Gravity-backed input. We'll use this in clients that need a shorter display string but we might consider adding this formatting and perhaps validation upstream in Gravity.

**Desired design**:

Screenshot 2022-12-07 at 19 33 36

**Example of undesired use of the `lotLabel` field**:

### DIFF
--- a/src/schema/v2/sale_artwork.ts
+++ b/src/schema/v2/sale_artwork.ts
@@ -315,7 +315,16 @@ export const SaleArtworkType = new GraphQLObjectType<any, ResolverContext>({
       },
       lotLabel: {
         type: GraphQLString,
-        resolve: ({ lot_label }) => lot_label,
+        args: {
+          trim: {
+            type: GraphQLBoolean,
+            description:
+              "Whether to trim anything past the first alphanumeric chunk",
+            defaultValue: false,
+          },
+        },
+        resolve: ({ lot_label }, { trim }) =>
+          trim ? lot_label?.match(/\S+/)?.shift() : lot_label,
       },
       lotID: {
         type: GraphQLString,
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -14513,7 +14513,10 @@ type SaleArtwork implements ArtworkEdgeInterface & Node {
   isWatching: Boolean
   isWithReserve: Boolean
   lotID: String
-  lotLabel: String
+  lotLabel(
+    # Whether to trim anything past the first alphanumeric chunk
+    trim: Boolean = false
+  ): String
   lotState: CausalityLotState
   lowEstimate: SaleArtworkLowEstimate
   minimumNextBid: SaleArtworkMinimumNextBid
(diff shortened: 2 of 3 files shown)


src/schema/v2/sale_artwork.ts
```diff
@@ -315,7 +315,7 @@ export const SaleArtworkType = new GraphQLObjectType<any, ResolverContext>({
       },
       lotLabel: {
         type: GraphQLString,
-        resolve: ({ lot_label }) => lot_label,
+        resolve: ({ lot_label }) => lot_label?.match(/\d+/)?.shift(),
```
anandaroop, line 318: Two things:

- I notice there [already is a `lot_number` field](https://github.com/artsy/gravity/blob/796e90e4492e5667c75907d54d5b3964e8e2c191/app/models/domain/sale_artwork.rb#L19) which appears to be the `lot_label` but [parsed](https://github.com/artsy/gravity/blob/796e90e4492e5667c75907d54d5b3964e8e2c191/app/models/domain/sale_artwork.rb) as an int 

- My  hesitation is that I think alphanumeric lot numbers may be legitimate (maybe even crucial?) in rare circumstances?

```rb

pp SaleArtwork.order(created_at: :desc).limit(1000).
  pluck(:lot_label, :lot_number).
  select{ |label, number| label != number.to_s }

[["79b", 79],
 ["79a", 79],
 ["256a", 256],
 ["281a", 281],
 ["257a", 257],
 ["93A", 93],
 ["221a", 221],
 ["203a", 203],
 ["190a", 190],
 ["182a", 182],
 ["171a", 171],
 ["217a", 217],
 ["269a", 269]]
```

anandaroop: I don't really know how `lot_number` is used, maybe it's involved in sorting somewhere? (Though it would be dicey for that if `79a` and `79b` both become `79` 😅 )

**suggestion (non-blocking):** A trickier regex could work. But then I wonder if should be an option rather than a default, controlled via e.g. a `trim` argument? That way we could more gradually opt in to this behavior where it's needed, like the artwork screen redesign.

dblandin: Good thought on an optional argument. Sounds like a safer iteration than swapping in a new resolution for all clients right away.

Went ahead and pushed that change if you'd like to take a look 👍🏼

anandaroop (approved):
LGTM!

PR #4594, opened 2022-12-08
fix: return isForSale=false if artwork is sold

dblandin:
Patches a backend data issue for which an artwork can be both available for sale _and_ sold at the same time.

See [this Slack thread](https://artsy.slack.com/archives/CP9P4KR35/p1669977785252689) for more context.


gkartalis (approved):
amazing! Will wait for this to be merged to use it on https://github.com/artsy/eigen/pull/7817

PR #4596, opened 2022-12-08
Refactor: Use of reserved word ID in ConsignmentInquiry type

lordkiz:
### Description
- Refactor to use internalID as the key for id in ConsignmentInquiry type since id is a reserved word in Relay. Clients using Relay such as Eigen are unable to query for the field id of a consignment inquiry.


olerichter00 (approved):
Good catch! 🚂

PR #4600, opened 2022-12-09
fix(filter): force undefined when partners filters are not present

araujobarret:
Related to #4530 

We notice Impulse returning errors when we don't pass these args, now forcing them to `undefined` if they're not present solves the issue.

cc @artsy/negotiate-devs


lean8086 (approved):
LGTM

damassi:
Ah, nice 👍

PR #4602, opened 2022-12-09
refactor(gemini): uses @artsy/img to mint urls

dzucconi:
With this we shouldn't see any changes to the URL output (so specs should pass) since these were already being alphabetized.

### DIFF
--- a/src/schema/v2/image/services/gemini.ts
+++ b/src/schema/v2/image/services/gemini.ts
@@ -1,40 +1,21 @@
-import urljoin from "url-join"
-import { isExisty, toQueryString } from "lib/helpers"
 import config from "config"
+import { configureImageServices, ResizeMode } from "@artsy/img"
 
 const { GEMINI_ENDPOINT } = config
+
+const services = configureImageServices({
+  gemini: {
+    endpoint: GEMINI_ENDPOINT!,
+  },
+})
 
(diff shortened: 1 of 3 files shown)


src/schema/v2/image/services/gemini.ts
```diff
@@ -1,37 +1,18 @@
-import urljoin from "url-join"
-import { isExisty, toQueryString } from "lib/helpers"
 import config from "config"
+import { configure, Mode } from "@artsy/img"
```
damassi, line 2: This is silly and minor, but as a third party lib it would be nice if these setup functions were better described and less generic:

```
configureImageProvider
``` 
or 
```
ImageProviderMode`
```

Shorthand names like this feel more suitable for locally-scoped operations on a smaller scale.

damassi: At the very least, the exported types should be descriptive. Looking in force for example: 

Screenshot 2022-12-09 at 10 59 34 AM

dzucconi: Yeah, I see your point (and do agree about `Mode` at least). Alright might as well just cut a quick breaking change so this gets to 1.0.0 :)

damassi: 🙏

PR #4603, opened 2022-12-09
fix(updateMessageMutation): Fix type caps case

damassi:
Fixes issue where some types were lowercase rather than caps case.


damassi:
something is up with github...

PR #4605, opened 2022-12-09
feat: Use curated trending for app home feed artist GRO-1381

jonallured:
This PR steals all of the good work done by @anandaroop and uses it on the home feed in the app. It'll require a PR over on Eigen to actually use this `TRENDING` flavor of artists but this is the first step.

https://artsyproduct.atlassian.net/browse/GRO-1381

/cc @artsy/grow-devs

### DIFF
--- a/src/schema/v2/artists/curatedTrending.ts
+++ b/src/schema/v2/artists/curatedTrending.ts
@@ -6,7 +6,7 @@ import { dailyShuffle } from "lib/shuffle"
 import { ArtworksAggregation } from "../aggregations/filter_artworks_aggregation"
 import { artistConnection } from "../artist"
 
-const getCuratedArtists = async (context): Promise<[typeof Artist]> => {
+export const getCuratedArtists = async (context): Promise<[typeof Artist]> => {
   const { artistsLoader, filterArtworksLoader, defaultTimezone } = context
   const artworks = await filterArtworksLoader({
     size: 0,
--- a/src/schema/v2/artists/curatedTrending.ts
+++ b/src/schema/v2/artists/curatedTrending.ts
@@ -1,10 +1,30 @@
 import { GraphQLFieldConfig } from "graphql"
 import type { ResolverContext } from "types/graphql"
 import { connectionArgs, connectionFromArray } from "graphql-relay"
-
+import Artist from "schema/v2/artist"
 import { dailyShuffle } from "lib/shuffle"
 import { ArtworksAggregation } from "../aggregations/filter_artworks_aggregation"
 import { artistConnection } from "../artist"
+
+const getCuratedArtists = async (context): Promise<[typeof Artist]> => {
+  const { artistsLoader, filterArtworksLoader, defaultTimezone } = context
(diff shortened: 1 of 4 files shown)


src/schema/v2/artists/curatedTrending.ts
```diff
@@ -1,11 +1,31 @@
...
+import Artist from "schema/v2/artist"
 import { dailyShuffle } from "lib/shuffle"
 import { ArtworksAggregation } from "../aggregations/filter_artworks_aggregation"
 import { artistConnection } from "../artist"
 
+export const getCuratedArtists = async (context): Promise<[typeof Artist]> => {
+  const { artistsLoader, filterArtworksLoader, defaultTimezone } = context
+  const artworks = await filterArtworksLoader({
+    size: 0,
+    marketing_collection_id: "trending-this-week",
```
anandaroop, line 13: **thought (non-blocking):** Reminds me that this is a bit of a weak link. If anything happens to this `trending-this-week` collection, there is an increasingly wide blast radius. Would be good to add some kind of admin-facing note in the admin UI for this, somehow.

anandaroop (approved):
Nice!

src/schema/v2/home/home_page_artist_module.ts
```diff
@@ -56,7 +57,7 @@ export const HomePageArtistModuleTypes: {
   TRENDING: {
     description: "The trending artists.",
     display: () => Promise.resolve(true),
```
anandaroop, line 59: One follow-up, after reviewing the Eigen side of this.

**thought:** My 2¢ is that we could consider adding a new key (a la `CURATED_TRENDING`) instead of redefining `TRENDING` here.

In my earlier PR work that I felt the "trending" nomenclature was sort of reserved for our algorithmic concept of trending artists (whether artsy/delta today or something else in the future) and therefore tried to distinguish this human-curated version with the naming of the field. 

Do with that what you will!

src/schema/v2/home/home_page_artist_module.ts
```diff
@@ -56,7 +57,7 @@ export const HomePageArtistModuleTypes: {
...
-    resolve: ({ trendingArtistsLoader }) => trendingArtistsLoader(),
```
anandaroop, line 59: One follow-up, after reviewing the Eigen side of this.

**thought:** My 2¢ is that we could consider adding a new key (a la `CURATED_TRENDING`) instead of redefining `TRENDING` here.

In my earlier PR work that I felt the "trending" nomenclature was sort of reserved for our algorithmic concept of trending artists (whether artsy/delta today or something else in the future) and therefore tried to distinguish this human-curated version with the naming of the field. 

Do with that what you will!

jonallured: Ok I have added a new key with [8e098c6](https://github.com/artsy/metaphysics/pull/4605/commits/8e098c6de91fd00810df38f9b4b2e75e9282a4e4).

jonallured:
Ok this is green going to merge thanks for the thoughtful review @anandaroop!! ❤️

PR #4612, opened 2022-12-13
chore: format sale date string into date

kajatiger:
I used the helper `date` to send a better formated version of the sale date to forque, because over in forque we don't have moment.js and formatting there would mean creating a new helper in forque.

### DIFF
--- a/src/schema/v2/userPurchases.ts
+++ b/src/schema/v2/userPurchases.ts
@@ -8,6 +8,7 @@ import {
 import { IDFields } from "./object_identification"
 import { ArtworkType } from "./artwork"
 import { connectionWithCursorInfo } from "schema/v2/fields/pagination"
+import { date } from "./fields/date"
 
 export const edgeFields: Thunk<GraphQLFieldConfigMap<
   any,
@@ -16,7 +17,7 @@ export const edgeFields: Thunk<GraphQLFieldConfigMap<
   ...IDFields,
   ownerType: { type: GraphQLString, resolve: ({ owner_type }) => owner_type },
   salePrice: { type: GraphQLFloat, resolve: ({ sale_price }) => sale_price },
-  saleDate: { type: GraphQLString, resolve: ({ sale_date }) => sale_date },
+  saleDate: date(({ sale_date }) => sale_date),
   source: { type: GraphQLString },
 })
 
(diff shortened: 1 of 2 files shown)


src/schema/v2/userPurchases.ts
```diff
@@ -16,7 +17,7 @@ export const edgeFields: Thunk<GraphQLFieldConfigMap<
   ...IDFields,
   ownerType: { type: GraphQLString, resolve: ({ owner_type }) => owner_type },
   salePrice: { type: GraphQLFloat, resolve: ({ sale_price }) => sale_price },
-  saleDate: { type: GraphQLString, resolve: ({ sale_date }) => sale_date },
+  saleDate: date(({ sale_date }) => sale_date),
```
kajatiger, line 20: The weird thing is that his still returns the same type of string: `"saleDate": "2022-06-11T20:00:00+00:00",` when running the query or the test.

src/schema/v2/__tests__/user.test.ts
```diff
@@ -497,7 +497,7 @@ describe("User", () => {
         {
           owner_type: "ArtworkInquiryRequest",
           sale_price: 600.1234,
-          sale_date: "2018-09-26 04:00:00 UTC",
+          sale_date: "Sep 26th, 2018",
```
kajatiger, line 500: I am curious why the test returns `invalid date` here when they seemed to have had the same string before 🤔

ovasdi: Perhaps this is because the stub has changed. It went from `sale_date: "2018-09-26 04:00:00 UTC"` to `sale_date: "Sep 26th, 2018",`.

The block defined starting [L:477](https://github.com/artsy/metaphysics/blob/kajatiger/PLATFORM-4793-format-sale_date-string-to-be-date-in-metaphysics/src/schema/v2/__tests__/user.test.ts#L477-L517) `purchases` is the stubbed shape coming in from gravity.

I think you changed the format in the wrong place. The date as stubbed to be coming from gravity doesn't change, but the expected shape after it has gone through metaphysics should have changed now that the query specifies a format.

So [L:547](https://github.com/artsy/metaphysics/pull/4612/files#diff-62c68948ab97aaf79e25a22d7344f59bd6ff8bb78af7fa6fe9ef020920408d5dR547) should probably be `Sep 26th, 2018`

kajatiger: Ah okay, makes sense! Thanks @ovasdi !!

PR #4613, opened 2022-12-13
[WIP] spike: learn about order stitching

iskounen:
This is a spike to learn about how order stitching happens on the `me` query.

### DIFF
--- a/src/lib/stitching/exchange/v2/stitching.ts
+++ b/src/lib/stitching/exchange/v2/stitching.ts
@@ -518,7 +518,7 @@ export const exchangeStitchingEnvironment = ({
             __typename
           }`,
           resolve: async (_source, args, context, info) => {
-            return await info.mergeInfo.delegateToSchema({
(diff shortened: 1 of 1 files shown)


src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -518,14 +518,20 @@ export const exchangeStitchingEnvironment = ({
             __typename
           }`,
           resolve: async (_source, args, context, info) => {
-            return await info.mergeInfo.delegateToSchema({
+            const orderResult = await info.mergeInfo.delegateToSchema({
               schema: exchangeSchema,
               operation: "query",
               fieldName: "commerceMyOrders",
```
iskounen, line 524: **thought:** Alternatively, we could make this change in [Exchange's `myOrders` type](https://github.com/artsy/exchange/blob/c4d6bb4af8da22de05aa15eb4f9724194e10145c/app/graphql/types/query_type.rb#L70-L78).

PR #4619, opened 2022-12-15
refactor: improve auctionResultsConnection interface [DO NOT MERGE]

MounirDhahri:
This is going to be a non-breaking change to our frontend since we still did not roll this change out to App.

What's needed before merging this
1. ✅ Revert PR on Eigen: https://github.com/artsy/eigen/pull/7867
2. ✅ Revert PR on Force: https://github.com/artsy/force/pull/11575#pullrequestreview-1220973769

### DIFF
--- a/src/schema/v2/me/myCollectionAuctionResults.ts
+++ b/src/schema/v2/me/myCollectionAuctionResults.ts
@@ -12,49 +12,49 @@ import { pageable } from "relay-cursor-paging"
 import { createPageCursors } from "schema/v2/fields/pagination"
 import { ResolverContext } from "types/graphql"
 import ArtworkSizes from "../artwork/artworkSizes"
-import { auctionResultConnection, AuctionResultSorts } from "../auction_result"
+import {
+  auctionResultConnection,
+  AuctionResultSorts,
+  AuctionResultsState,
+} from "../auction_result"
 import { MAX_ARTISTS } from "./myCollectionInfo"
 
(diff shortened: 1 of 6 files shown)


src/schema/v2/me/myCollectionAuctionResults.ts
```diff
@@ -12,49 +12,49 @@ import { pageable } from "relay-cursor-paging"
...
-      description: "Filter auction results by organizations",
-    },
-    sizes: {
-      type: new GraphQLList(ArtworkSizes),
-      description: "Filter auction results by Artwork sizes",
+    allowEmptyCreatedDates: {
+      type: GraphQLBoolean,
+      defaultValue: true,
+      description:
+        "Filter auction results by empty artwork created date values",
```
olerichter00, line 30: **suggestion:** Not quite sure if the description is matching the param name and describing what the param does. (allow vs. filtering).

olerichter00 (approved):
Thanks for the adjustment 🙏

src/schema/v2/me/myCollectionAuctionResults.ts
MounirDhahri: this was there for a while though, I only updated the order of the args

olerichter00 (approved):
Nice 🌟

MounirDhahri:
This PR has breaking changes - this shouldn't break any existing clients since we already removed the fields usage from Force and we haven't yet released the changes in App

PR #4626, opened 2022-12-16
feat(quiz): exposes saved artworks field

dzucconi:
Quick little optimization since we're doing it in a few places. Doesn't need to be a connection since it's a bounded list that won't be paginated.

### DIFF
--- a/src/schema/v2/quiz.ts
+++ b/src/schema/v2/quiz.ts
@@ -1,11 +1,12 @@
-import { GraphQLFieldConfig } from "graphql"
+import { GraphQLFieldConfig, GraphQLList } from "graphql"
 import { GraphQLNonNull, GraphQLObjectType } from "graphql"
 import { InternalIDFields } from "schema/v2/object_identification"
 import { quizArtworkConnection } from "./quizArtworkConnection"
 import { ResolverContext } from "types/graphql"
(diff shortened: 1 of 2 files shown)


src/schema/v2/quiz.ts
```diff
@@ -1,18 +1,39 @@
...
+      ),
+      resolve: async ({ quiz_artworks }, _, { savedArtworkLoader }) => {
+        if (!savedArtworkLoader) return []
+
+        try {
+          const artworks = await Promise.all(
+            quiz_artworks.map(({ artwork_id }) => {
+              return savedArtworkLoader(artwork_id)
+            })
+          )
```
dzucconi, line 29: This is bundled as a single request because of the dataloader pattern:

src/schema/v2/quiz.ts
```diff
@@ -1,18 +1,39 @@
...
+
+          return artworks.filter(({ is_saved }) => is_saved)
```
dzucconi, line 31: The loader returns full artwork results for the saved artworks; then `{ id, is_saved }` for the not saved.

src/schema/v2/quiz.ts
jonallured: That's wild that you get that behavior without being aware of it!!

PR #4631, opened 2022-12-19
feat: only non-expired credit cards are qualified

erikdstock:
Completes [TX-854](https://artsyproduct.atlassian.net/jira/software/c/projects/TX/boards/114?modal=detail&selectedIssue=TX-854)

This pr updates our `Me.hasQualifiedCreditCards` field to filter out expired cards, which are not 'qualified.'

I verified that cards are [still good during their expiration month](https://www.americanexpress.com/en-us/credit-cards/credit-intel/what-happens-when-credit-card-expires/), so only check whether the current date is after that.

I tried to take the quickest route to implementation and defaulted to measuring by utc. I considered adding a 1-day buffer to the expiration date to account for time zone weirdness, but that seemed like it might be confusing later, and really whether your card is considered qualified or not in this short period isn't very important since the field isn't currently used to process a charge immediately.

### DIFF
--- a/src/schema/v2/me/index.ts
+++ b/src/schema/v2/me/index.ts
@@ -59,6 +59,7 @@ import { SavedArtworks } from "./savedArtworks"
 import { ShowsByFollowedArtists } from "./showsByFollowedArtists"
 import { WatchedLotConnection } from "./watchedLotConnection"
 import { quiz } from "../quiz"
+import moment from "moment"
 
 /**
  * @deprecated: Please use the CollectorProfile type instead of adding fields to me directly.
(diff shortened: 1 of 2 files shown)


src/schema/v2/me/index.ts
```diff
@@ -210,8 +211,22 @@ export const meType = new GraphQLObjectType<any, ResolverContext>({
...
-          ({ body }) => {
-            return body && body.length > 0
+          ({ body: cards = [] }) => {
+            const unexpiredCards = cards.filter((card) => {
+              const {
+                expiration_month: expMonth,
+                expiration_year: expYear,
+              } = card
+
+              // Moment months are 0-indexed
```
starsirius, line 221: 😱

src/schema/v2/me/index.ts
```diff
@@ -210,8 +211,22 @@ export const meType = new GraphQLObjectType<any, ResolverContext>({
...
+              const expirationMoment = moment({
+                year: expYear,
+                month: expMonth - 1,
+              }).endOf("month")
```
starsirius, line 225: #non-blocking. I think `moment(...)` will be local timezone. This exact moment may be a few hours different depending on the timezone. MP servers might be configured in the UTC timezone so it's still UTC, but potentially we can explicitly set the end of month moment to be UTC. Maybe something like
```
moment.utc(...)
```

starsirius (dismissed):
🥲  yay finally!

src/schema/v2/me/index.ts
erikdstock: good catch!

erikdstock:
#squashongreen

joeyAghion:
Metaphysics is an unexpected place to impose this logic. The API implements logic [for whether users or cards are `qualified_for_bidding?`](https://github.com/artsy/gravity/search?q=qualified_for_bidding), and would be the more natural place to reflect a business rule change such as this.

PR #4639, opened 2022-12-21
feat(graphql): deprecate emailConfirmed field

araujobarret:
This PR addresses this [request](https://github.com/artsy/metaphysics/pull/4628#discussion_r1053407946) deprecating `emailConfirmed` and adding `isEmailConfirmed` as an option.

We'll have PRs to update the clients to use the new field, then we can retire the field.


mzikherman (commented):
🔥 great!

PR #4641, opened 2022-12-21
feat(graphql): deprecate identityVerified field

araujobarret:
related to #4639 
Deprecates also `identityVerified` in favor of `isIdentityVerified`.


MrSltun (approved):
🌟

PR #4642, opened 2022-12-21
feat(FX-4497): Article featured artist activity

nickskalkin:
Recognize ArticleFeaturedArtistActivity events

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11924,6 +11924,7 @@ input NotificationPreferenceInput {
 }
 
 enum NotificationTypesEnum {
+  ARTICLE_FEATURED_ARTIST
   ARTWORK_ALERT
   ARTWORK_PUBLISHED
   PARTNER_SHOW_OPENED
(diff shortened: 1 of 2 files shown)


dblandin:
Just to confirm — this allows us to query for the new event type via Metaphysics but will not affect the default set of exposed events, which is determined [here in Gravity](https://github.com/artsy/gravity/blob/07bf67a0d253b9a21acf53040b81bb1531e2c56a/app/api/v1/me_notifications_endpoint.rb#L50).

dblandin (approved):
Left one clarifying comment. LGTM!

nickskalkin:
> Just to confirm — this allows us to query for the new event type via Metaphysics but will not affect the default set of exposed events, which is determined [here in Gravity](https://github.com/artsy/gravity/blob/07bf67a0d253b9a21acf53040b81bb1531e2c56a/app/api/v1/me_notifications_endpoint.rb#L50).

@dblandin yes, you are right

PR #4649, opened 2022-12-22
Boost coverage of high auction record badge to 750 artists

mbransfield:
Adding an additional 250 artist ids to the list of "trusted artists" so that we feature their high auction records on the artist and artwork pages.


egdbear (approved):

egdbear:
Done with https://github.com/artsy/metaphysics/pull/4654/files

PR #4650, opened 2022-12-23
feat: add isUpcoming to auction result interface

MounirDhahri:
This PR adds `isUpcoming` to the auction result interface. 
**Notes:** 
- ⭐️This is important to avoid doing this computation on the frontend side.
- ⭐️⭐️ We intentionally compare to the beginning of the day instead of the exact moments 

Screenshot 2022-12-23 at 12 02 37

### DIFF
--- a/src/schema/v2/__tests__/auction_result.test.ts
+++ b/src/schema/v2/__tests__/auction_result.test.ts
@@ -9,7 +9,7 @@ const mockAuctionResult = {
   dimension_text: "20 x 20",
   organization: "Christie's",
   category_text: "an old guitar",
-  sale_date: "yesterday",
+  sale_date: "2022-01-19T20:00:00.000Z",
   images: [
     {
       thumbnail: "https://path.to.thumbnail.jpg",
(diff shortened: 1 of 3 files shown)


src/schema/v2/__tests__/auction_result.test.ts
```diff
@@ -9,7 +9,7 @@ const mockAuctionResult = {
   dimension_text: "20 x 20",
   organization: "Christie's",
   category_text: "an old guitar",
-  sale_date: "yesterday",
```
MounirDhahri, line 12: this was wrong 😄

src/schema/v2/auction_result.ts
```diff
@@ -96,6 +96,15 @@ const AuctionResultType = new GraphQLObjectType<any, ResolverContext>({
       type: GraphQLString,
       resolve: ({ category_text }) => category_text,
     },
+    isUpcoming: {
+      type: GraphQLBoolean,
+      resolve: ({ sale_date }) => {
+        if (!sale_date) return null
+        const utc = new Date().toJSON().slice(0, 10).replace(/-/g, "/")
+
+        return new Date(sale_date) >= new Date(utc)
```
olerichter00, line 105: **suggestion:** Maybe we can move this to a helper function. Like this it would be more clear what the resolver does.

```suggestion
        return isInFuture(sale_date)
```

olerichter00 (approved):
Looks good 🌟 

I've just added a suggestion about the resolver.

src/schema/v2/auction_result.ts
MounirDhahri: good suggestion, I will

olerichter00 (approved):

PR #4657, opened 2022-12-28
fix: Set includePersonalArtists to true for artwork form query

olerichter00:
Addresses [CX-3285]

## Description

Because `includePersonalArtists` defaults to `false`, we need to use a hack for older clients that don't send the parameter introduced in https://github.com/artsy/eigen/pull/7904. The only query we need to set the param to true is part of the My Collection artwork form (which sets `size` to `100`). 

[CX-3285]: https://artsyproduct.atlassian.net/browse/CX-3285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/me/myCollectionInfo.ts
+++ b/src/schema/v2/me/myCollectionInfo.ts
@@ -184,8 +184,11 @@ export const myCollectionInfoFields = {
       // TODO: Remove this once all clients pass includePersonalArtists correctly
       // This is a hack we need to return the correct results for older cleints that don't send the param `includePersonalArtists`.
       // With this solution we are defaulting to true if the query comes from the My Collection artwork form (which is the only query that passes `size: 100`)
+      const SIZE_ARG_VALUE_FOR_MY_COLLECTION_ARTWORK_FORM = 100
       const includePersonalArtists =
-        args.size === 100 ? true : args.includePersonalArtists
+        args.size === SIZE_ARG_VALUE_FOR_MY_COLLECTION_ARTWORK_FORM
+          ? true
+          : args.includePersonalArtists
 
       const { body, headers } = await collectionArtistsLoader("my-collection", {
         size,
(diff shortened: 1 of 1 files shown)


src/schema/v2/me/myCollectionInfo.ts
```diff
@@ -181,14 +181,19 @@ export const myCollectionInfoFields = {
         args
       )
 
+      // TODO: Remove this once all clients pass includePersonalArtists correctly
+      // This is a hack and currently we are defaulting to true if the query comes from the artwork form (passing a size of 100)
+      const includePersonalArtists =
+        args.size === 100 ? true : args.includePersonalArtists
```
olerichter00, line 187: Open to suggestions to avoid this hack.

lordkiz: I think allowing old clients to stay `false` would be the proper way to go. Since this can easily be fixed on Force, the issue would only be on Eigen and that can easily be remedied by having a version that has been tested thoroughly for this behaviour instead of trying to make it compatible with those versions. What do you think? Because users on these versions are already using the app without issues regarding this. Instead let us ensure the next release sends this arg.

olerichter00: The issue is that with the current change users of the app would only see artists that are not personal (non-public) and currently we need to show all artists for one query (insights) and only non-personal (public) artists for another query (artwork form). 

If we don't use a hack like this, we would need to wait with this change until all users have updated their apps to the new version which will be released next year.

Or is there another way I don't see? 🤔

lordkiz: I cannot think of any other way. Since this is documented and there is a followup ticket to update this, I think we can live with this for now.

PR #620, opened 2017-04-21
Adds a Relay connection for a user's saved artworks

orta:
This is definitely not done, as it's getting _all_ artworks right now instead of the specified collection, and there's no tests, but hey - it works in terms of getting all of a user's saved artworks. I think.

screen shot 2017-04-21 at 17 35 06

PR #4674, opened 2023-01-10
feat: add collection field under `me`

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-4512

This adds a `me.collection` field such that the following query for a user's given collection can be resolved via a new authenticated loader:

### Query

```graphql
query {
  me {
    collection(id: "55de7e8172616975e80009dd") {
      name
      default
      saves
    }
  }
}
```

### Success Response

```json
{
  "data": {
    "me": {
      "collection": {
        "name": "Saved Artwork",
        "default": true,
        "saves": true
      }
    }
  }
}
```

### Unauthorized Response 

If there is a mismatch between the `me` user and the user associated with the current access token

```json
{
  "data": {
    "me": {
      "collection": null
    }
  },
  "errors": [
    {
      "message": "… {\"error\":\"Collection Not Found\"}",
      "locations": "…",
      "path": "…",
      "stack": "…"
    }
  ]
}
```


src/schema/v2/collection.ts
```diff
@@ -1,114 +0,0 @@
...
-} from "graphql"
-import { ResolverContext } from "types/graphql"
-import { includesFieldsOtherThanSelectionSet } from "lib/hasFieldSelection"
-
-// Note to developers working on collections, the staging server does not get a copy
-// of all artwork saves, so you will need to add some each week in order to have data
-// to work with.
-
-// TODO: Is the root collection field really used and do we still need this?
-export const CollectionType = new GraphQLObjectType<any, ResolverContext>({
```
anandaroop, line 27: This original implementation was [initially added in 2017](https://github.com/artsy/metaphysics/pull/620) and then [pruned in 2019 as part of the MP v2 effort](https://github.com/artsy/metaphysics/pull/1920). 

(The old v1 api itself was [dropped in its entirety a few months ago](https://github.com/artsy/metaphysics/pull/4262), though this unused code was left in place.)

I opted to drop this old, orphaned implementation and start fresh with a schema design that more closely reflects our product need and leans less on some of the tricky patterns seen here. I hope we don't end up re-inventing these wheels. Let me know if you have any concerns about this.

_schemaV2.graphql
```diff
@@ -3549,6 +3549,25 @@ type CitySponsoredContent {
   ): ShowConnection
 }
 
+# A collection of artworks
+type Collection {
+  # True if this is the default collection for this user, i.e. the default Saved Artwork collection.
+  default: Boolean!
```
anandaroop, line 3555: This is _almost_ always true, and serves a reminder that we should probably proceed with some [legacy data cleanup](https://www.notion.so/artsy/Wishlists-5b45360488f540f79212c0dd00d66466#a8b0090795a1482aa4560df99b345d65) to ensure that this is in fact always true.

src/schema/v2/object_identification.ts
```diff
@@ -55,7 +55,6 @@ const SupportedTypes: any = {
     "./show",
     "./fair",
     "./sale",
-    "./collection",
```
anandaroop, line 58: [Tricky!](https://artsy.slack.com/archives/CP9P4KR35/p1673974840356529)

nickskalkin (approved):
👍

src/schema/v2/me/collection.ts
```diff
@@ -0,0 +1,54 @@
...
+      default: {
+        type: new GraphQLNonNull(GraphQLBoolean),
+        description:
+          "True if this is the default collection for this user, i.e. the default Saved Artwork collection.",
+      },
+      saves: {
+        type: new GraphQLNonNull(GraphQLBoolean),
+        description:
+          "True if this collection represents artworks explicitly saved by the user, false otherwise.",
+      },
```
dblandin, line 32: **suggestion**: Do we have any immediate user-facing need for either `default` or `saves` attributes on this new type? Not sure how they would surface in the UI at the moment. Might be easy to leave them off for now and introduce either when we can quickly associate with a UI need.

Based on what I think is included in the [JSON representation at the moment](https://github.com/artsy/gravity/blob/main/app/models/domain/collection.rb#L172-L194), the following attributes seem immediately useful:
- name
- internalID
- image - via full Image type supporting cropped/resized
- Maybe artworksCount - not sure if we'd prefer to get this from the artworksConnection.

dblandin (dismissed):
LGTM! Left one non-blocking comment about the attributes we're prioritizing.

PR #4686, opened 2023-01-16
feat: Expose images in create artwork mutation

olerichter00:
Addresses [CX-3333]

## Description

Expose images in create artwork mutation.

### DIFF
--- a/src/schema/v2/image/normalize.ts
+++ b/src/schema/v2/image/normalize.ts
@@ -1,16 +1,16 @@
 import {
-  pick,
-  values,
-  first,
   assign,
   compact,
+  curry,
+  find,
+  first,
   flow,
   includes,
-  last,
   isArray,
   isString,
-  find,
-  curry,
+  last,
+  pick,
+  values,
 } from "lodash"
 import { DEFAULT_SRCSET_QUALITY } from "./services/config"
 
@@ -36,7 +36,7 @@ export const setVersion = (
 
 const normalizeImageUrl = (image) => {
   const image_url = grab(image, ["url", "image_url"])
-  if (!image_url) return null
+  // if (!image_url) return null
   return assign({ image_url }, image)
 }
 
--- a/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
+++ b/src/schema/v2/me/myCollectionCreateArtworkMutation.ts
@@ -227,11 +227,13 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
 
       const imageSources = computeImageSources(externalImageUrls)
 
-      for (const imageSource of imageSources) {
-        await createArtworkImageLoader(artworkId, imageSource)
-      }
+      const images = await Promise.all(
+        imageSources.map((imageSource) =>
+          createArtworkImageLoader(artworkId, imageSource)
+        )
+      )
 
-      return response
+      return { ...response, images }
     } catch (error) {
       const formattedErr = formatGravityError(error)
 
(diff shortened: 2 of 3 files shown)


src/schema/v2/image/normalize.ts
```diff
@@ -36,7 +36,7 @@ export const setVersion = (
 
 const normalizeImageUrl = (image) => {
   const image_url = grab(image, ["url", "image_url"])
-  if (!image_url) return null
+
```
olerichter00, line 39: When the image is created,  the is no `imageURL` until Gemini has processed the image. To return the images from the mutation we therefor shouldn't filter images anymore.

MounirDhahri: **praise:** Nice, good catch.

MounirDhahri: **suggestion:** I am wondering if we should only enable this only when a flag is specified so the behavior is still the same in the frontend

MounirDhahri: I mean something like:
```typescript
  if (!image_url && !includeProcessing ) return null
```

MounirDhahri: I understand though that this is not probably necessarily needed since we don't use returned images anywhere now but being explicit about it might be useful

src/schema/v2/me/myCollectionCreateArtworkMutation.ts
```diff
@@ -227,11 +227,13 @@ export const myCollectionCreateArtworkMutation = mutationWithClientMutationId<
 
       const imageSources = computeImageSources(externalImageUrls)
 
-      for (const imageSource of imageSources) {
-        await createArtworkImageLoader(artworkId, imageSource)
-      }
+      const images = await Promise.all(
```
MounirDhahri, line 230: **praise:** that's a really nice improvement

src/schema/v2/image/normalize.ts
lordkiz: +1 to @MounirDhahri 's suggestion, especially if we are using normalizeImageUrl elsewhere besides handling MyCollectionImages.

lordkiz (approved):
👏 LGTM. Will wait for resolution on the suggestion + failing tests before merging 👍

src/schema/v2/image/normalize.ts
olerichter00: You're right. I'll double-check if this could lead to issues in other places. Probably, it's a good idea to add the `includeProcessing` parameter (or `includeAll` - failed images wouldn't have an `imageURL` either) because we only need to show these images in the MyCollection context.

MounirDhahri (approved):
Nice! thanks

src/schema/v2/me/__tests__/myCollectionCreateArtworkMutation.test.ts
```diff
@@ -285,43 +285,6 @@ describe("myCollectionCreateArtworkMutation", () => {
 
       expect(message).toEqual(serverError)
     })
-
-    it("creates additional images in sequence to avoid a gravity race condition", async () => {
```
MounirDhahri, line 289: This test's presence make me wonder if we need the optimisation above on using `Promise.all`. It looks like it was an intentional decision from David to go with it. I would suggest reverting this unless we can see a significant increase in performance with no side effects

MounirDhahri: we thought about the same thing at the same time! thanks @olerichter00

MounirDhahri: feel free to merge afterwards

MounirDhahri (dismissed):

MounirDhahri (approved):

PR #4698, opened 2023-01-19
feat(NX-3371): Expose message's to and cc

lean8086:
This PR solves part of [NX-3371](https://artsyproduct.atlassian.net/browse/NX-3371)

# Description

Expose `message` recipients on `to` and `cc`, so we can determine who has opened an email on Volt's Conversations.

cc @artsy/negotiate-devs 

[NX-3371]: https://artsyproduct.atlassian.net/browse/NX-3371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/conversation/message.ts
+++ b/src/schema/v2/conversation/message.ts
@@ -134,5 +134,13 @@ export const MessageType = new GraphQLObjectType<any, ResolverContext>({
       resolve: ({ metadata }) => isInvoiceMessage(metadata),
     },
     createdAt: date(),
+    to: {
+      description: "Masked emails w/ display name of the recipients.",
+      type: new GraphQLList(GraphQLString),
+    },
+    cc: {
+      description: "Masked emails w/ display name of the recipients in copy.",
+      type: new GraphQLList(GraphQLString),
+    },
   },
 })
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11176,6 +11176,9 @@ type Message implements Node {
 
   # Unaltered text if possible, otherwise `body`: a parsed/sanitized version from Sendgrid.
   body: String
+
+  # Masked emails w/ display name of the recipients in copy.
+  cc: [String]
   createdAt(
     format: String
 
@@ -11211,6 +11214,9 @@ type Message implements Node {
     @deprecated(
       reason: "Payment Request was deprecated. The field was kept for legacy client support. [Will be removed in v2]"
     )
+
+  # Masked emails w/ display name of the recipients.
+  to: [String]
 }
 
 # A connection to a list of items.
--- a/src/schema/v2/conversation/__tests__/__snapshots__/message.test.js.snap
+++ b/src/schema/v2/conversation/__tests__/__snapshots__/message.test.js.snap
@@ -64,23 +64,31 @@ Object {
         Object {
           "node": Object {
             "body": "Loved some of the works at your fair booth!",
+            "cc": Array [],
             "from": Object {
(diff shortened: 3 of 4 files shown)


lidimayra (dismissed):
👏 👏 👏

src/schema/v2/conversation/message.ts
```diff
@@ -134,5 +134,13 @@ export const MessageType = new GraphQLObjectType<any, ResolverContext>({
       resolve: ({ metadata }) => isInvoiceMessage(metadata),
     },
     createdAt: date(),
+    to: {
```
araujobarret, line 137: shouldn't we cover `to` and `cc` in the tests too?

lean8086: I wasn't too sure about what to test. I added `to` and `cc` to the snapshot on https://github.com/artsy/metaphysics/pull/4698/commits/8c2824aa58f963b21a128c8fe93680be53b0337e. Not sure if it makes a huge difference, though. Thoughts?

araujobarret: I'm not a huge fan of snapshots but if was already here let's leave it like this, thanks for addressing it!

araujobarret (approved):

src/schema/v2/conversation/message.ts
```diff
@@ -134,5 +134,13 @@ export const MessageType = new GraphQLObjectType<any, ResolverContext>({
...
+      description: "Masked emails w/ display name of the recipients.",
+      type: new GraphQLList(GraphQLString),
+    },
+    cc: {
+      description: "Masked emails w/ display name of the recipients in copy.",
+      type: new GraphQLList(GraphQLString),
```
dzucconi, line 143: These types could and should be changed to `new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(GraphQLString))))`

lean8086: I thought [those two can be `null`](https://github.com/artsy/radiation/blob/main/db/schema.rb#L133-L134), no?

dzucconi: Should they just be empty array if they are `null`?

lean8086: yeah, I think `new GraphQLList(new GraphQLNonNull(GraphQLString)))` would suit them better, right?

dzucconi: Generally; if you're expecting empty array then marking both the array and the array contents as non-nullable will make this type accurate and much easier to work with. If you're actually expecting null; then null; though in this case it looks like empty array would be preferred.

lean8086: > "if you're expecting empty array then marking both the array and the array contents as non-nullable will make this type accurate and much easier to work with"

I would've thought that marking both as non-nullable would not allow for an empty array. I might be reading this sentence wrong, though.

> "If you're actually expecting null; then null; though in this case it looks like empty array would be preferred."

I think an empty array would be preferred, yep.

dzucconi: Consider the difference between: `[]` and `[null, 'whatever']`

There's nothing about empty array that means you need to mark anything as nullable. `const foo: string[] = []` is totally valid right?

lean8086: Makes total sense. Thanks for putting it into perspective 👍  Will make the change tomorrow morning.

PR #4701, opened 2023-01-20
chore(NX-3371): make message's to and cc non-nullable

lean8086:
Solves part of [NX-3371](https://artsyproduct.atlassian.net/browse/NX-3371)

## Description

As a follow-up of https://github.com/artsy/metaphysics/pull/4698, define `message`'s `to` and `cc` as non-nullable since they can be empty arrays. More context [here](https://github.com/artsy/metaphysics/pull/4698#discussion_r1081643489).

cc @artsy/negotiate-devs 

[NX-3371]: https://artsyproduct.atlassian.net/browse/NX-3371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11178,7 +11178,7 @@ type Message implements Node {
   body: String
 
   # Masked emails w/ display name of the recipients in copy.
-  cc: [String]
+  cc: [String!]!
   createdAt(
     format: String
 
@@ -11216,7 +11216,7 @@ type Message implements Node {
     )
 
   # Masked emails w/ display name of the recipients.
-  to: [String]
+  to: [String!]!
 }
 
 # A connection to a list of items.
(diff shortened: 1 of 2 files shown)


_schemaV2.graphql
```diff
@@ -11178,7 +11178,7 @@ type Message implements Node {
   body: String
 
   # Masked emails w/ display name of the recipients in copy.
-  cc: [String]
+  cc: [String!]!
```
araujobarret, line 11181: this doesn't look like accurate, I would think something like `[String]!`, if we use the way it is now we're considering we always have an element `String`, don't we?

lean8086: I had the same thought [here](https://github.com/artsy/metaphysics/pull/4698#discussion_r1081704058).

dzucconi: @araujobarret I know it seems a bit counter intuitive but it _is_ accurate.

Consider the difference between:

```ts
const a: string[] = []
const b: (string | null)[] = []
const c: (string | null)[] | null = []
```

dzucconi (approved):

PR #4707, opened 2023-01-23
Revert "feat(CX-2847): GraphQLLong"

dzucconi:
Reverts artsy/metaphysics#4702


damassi:
@dzucconi - please update PR description and add reason for reverting 🙏

lordkiz:
@damassi I had missed updating MyCollection Artwork Update Input type for pricePaidCents with the new Long type. I didn't want to continue having a deploy block in place as that blocked other people. So I reverted to fix instead.

PR #4711, opened 2023-01-24
feat: Add includeAll param to artwork figures field

olerichter00:
## Description

This PR adds the `includeAll` param to artwork's `figures` field.


src/schema/v2/artwork/utilities.ts
```diff
@@ -78,12 +78,16 @@ export const getFigures = ({
   images,
   external_video_id,
   set_video_as_cover,
+  includeAll = false,
```
olerichter00, line 81: It's been already added for `images`. For the artwork detail page we also need `figures` to return all images (including processing images).

MounirDhahri (approved):

PR #4722, opened 2023-01-26
feat(NX-3406): add date formatter to arstyUserSince field

araujobarret:
[NX-3406]

Add date format helper helper to the field `collectorProfile.artsyUserSince`

cc @artsy/negotiate-devs 

[NX-3406]: https://artsyproduct.atlassian.net/browse/NX-3406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/CollectorProfile/collectorProfile.ts
+++ b/src/schema/v2/CollectorProfile/collectorProfile.ts
@@ -1,4 +1,4 @@
-import date from "schema/v2/fields/date"
+import date, { date as dateFormatter } from "schema/v2/fields/date"
 import { InternalIDFields } from "schema/v2/object_identification"
 import {
   GraphQLID,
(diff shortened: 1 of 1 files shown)


src/schema/v2/CollectorProfile/collectorProfile.ts
```diff
@@ -59,7 +59,7 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
 
   // moved InquirerCollectorProfileFields here
   location: { type: myLocationType },
-  artsyUserSince: date,
```
jo-rs, line 62: Out of ignorance: will this be an issue with the already existing usages of this field in the FE?

araujobarret: you can still query the field without the format filter, it's optional so won't break 😉 
I tested locally too to be extra safe 😂

jo-rs (approved):

patrinoua:
🎉

PR #4727, opened 2023-01-26
feat(wishlists): add deleteCollection mutation

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-4530

Adds a mutation to delete the current user's collection by id.

### Mutation

```graphql
mutation DeleteCollection {
  deleteCollection(input: { id: "ee941f42-9d1e-41e2-ad47-6f5b53315fc9" }) {
    responseOrError {
      ... on DeleteCollectionSuccess {
        collection {
          internalID
          name
        }
      }
      ... on DeleteCollectionFailure {
        mutationError {
          statusCode
          message
        }
      }
    }
  }
}
```

### Success Response

```json
{
  "data": {
    "deleteCollection": {
      "responseOrError": {
        "collection": {
          "internalID": "dc24d286-4621-4613-bb63-c709e97ee789",
          "name": "Delete me"
        }
      }
    }
  }
}
```

### Error Response 

If there is a mismatch between the collection owner and the user associated with the current access token

```json
{
 "data": {
    "deleteCollection": {
      "responseOrError": {
        "mutationError": {
          "message": "Collection Not Found"
        }
      }
    }
  }
}
```

### DIFF
--- a/src/lib/loaders/loaders_with_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_with_authentication/gravity.ts
@@ -127,6 +127,11 @@ export default (accessToken, userID, opts) => {
       {},
       { method: "DELETE" }
     ),
+    deleteCollectionLoader: gravityLoader(
+      (id) => `collection/${id}`,
+      { user_id: userID },
+      { method: "DELETE" }
+    ),
     deleteCollectorProfileIconLoader: gravityLoader(
       "me/collector_profile/icon",
       {},
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -7710,6 +7710,30 @@ type DeleteBankAccountPayload {
   me: Me
 }
 
+type DeleteCollectionFailure {
+  mutationError: GravityMutationError
+}
+
+input deleteCollectionInput {
+  clientMutationId: String
+  id: String!
+}
+
+type deleteCollectionPayload {
+  clientMutationId: String
+
+  # On success: the deleted collection
+  responseOrError: DeleteCollectionResponseOrError
+}
+
+union DeleteCollectionResponseOrError =
+    DeleteCollectionFailure
+  | DeleteCollectionSuccess
+
+type DeleteCollectionSuccess {
+  collection: Collection
+}
+
 type DeleteConversationFailure {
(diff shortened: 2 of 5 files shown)


nickskalkin (approved):
👍

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -126,6 +126,11 @@ export default (accessToken, userID, opts) => {
       {},
       { method: "DELETE" }
     ),
+    deleteCollectionLoader: gravityLoader(
+      (id) => `collection/${id}`,
+      { user_id: userID },
```
dblandin, line 131: **question**: I see that many of the restored API endpoints require a `user_id` query param (example [here](https://github.com/artsy/gravity/blob/aedb621f398c0ae8bbb69271ad90d6c3c3c53dcb/app/api/v1/collections_endpoint.rb#L116-L119) 🔒 ). Is there any context around that requirement? Since these are all authenticated endpoints, I would expect that `current_user` is always available for user-scoped collection queries and authorization checks. Seems a bit strange to take this in as an input arg here and pass that back to Gravity which should already have easy access to the user via authentication headers.

dblandin (approved):
LGTM! Left one question around the user_id param sent back to Gravity.

src/lib/loaders/loaders_with_authentication/gravity.ts
anandaroop: Good question, and I'm not sure about the original motivation — I figured that general pattern was to allow for the flexibility to have users interact with _other_ users' collections. Maybe this was even part of the ancient collections implementation we had  several years ago?

dblandin: Maybe we can take another look in Gravity and simplify things there. This implementation should remain compatible, even if we remove the backend's reliance on this user_id query param. Non-blocking so feel free to merge this in when you're ready.

anandaroop: Yeah, in today's demo we noticed a few other things in those old endpoints that warrant another look, I'll get these ideas into a ticket.

PR #4729, opened 2023-01-27
fix: Pass includeAll arg correctly to artwork figures resolver

olerichter00:
Pass includeAll arg correctly to artwork figures resolver

### DIFF
--- a/src/schema/v2/artwork/utilities.ts
+++ b/src/schema/v2/artwork/utilities.ts
@@ -74,12 +74,10 @@ export const embed = (website, { width, height, autoplay }) => {
   }
 }
 
-export const getFigures = ({
-  images,
-  external_video_id,
-  set_video_as_cover,
-  includeAll = false,
-}) => {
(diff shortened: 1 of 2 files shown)


src/schema/v2/artwork/utilities.ts
```diff
@@ -74,12 +74,10 @@ export const embed = (website, { width, height, autoplay }) => {
...
-export const getFigures = ({
-  images,
-  external_video_id,
-  set_video_as_cover,
-  includeAll = false,
-}) => {
+export const getFigures = (
+  { images, external_video_id, set_video_as_cover },
+  { includeAll = false }
+) => {
```
olerichter00, line 80: The params get passed in the second argument and not in the first 🤦

MounirDhahri: thanks for the quick fix

MounirDhahri (approved):

PR #4743, opened 2023-02-02
fix: Match gravity's snake_case for time_zone in show event resolver

jpotts244:
While investigating a tangential [bug](https://artsyproduct.atlassian.net/browse/PX-5385), came across an issue here where `timezone` on show events was always returning `undefined` from gravity. If timezone is undefined, we fallback to UTC

That is because we are using the wrong attribute name in the resolver, `timezone` vs `time_zone` (the [ShowEvent](https://github.com/artsy/gravity/blob/main/app/models/domain/partner_show_event.rb#L15) model in gravity is using the snake case)

PS youll notice is gravity we use a mix of the different spellings which makes this interface a little more confusing, both [timezone](https://github.com/artsy/gravity/blob/fa87f2b5a51e01bd3777218b468bb45bb5360ce6/app/models/domain/partner_location.rb#L28) and [time_zone](https://github.com/artsy/gravity/blob/main/app/models/domain/partner_show_event.rb#L15)

🐛 Ex of before this change:
Screen Shot 2023-02-01 at 4 32 41 PM

Ex after this change:
Screen Shot 2023-02-01 at 4 32 13 PM

### DIFF
--- a/src/schema/v2/__tests__/show_events.test.ts
+++ b/src/schema/v2/__tests__/show_events.test.ts
@@ -3,56 +3,101 @@ import gql from "lib/gql"
 import { ResolverContext } from "types/graphql"
 
 describe("date resolving", () => {
-  const showData = {
-    id: "helwaser-gallery-anton-ginzburg-views",
-    displayable: true,
-    partner: {
-      id: "helwaser-gallery",
-    },
-    events: [
-      {
-        event_type: "Opening Reception",
-        start_at: "2019-03-28T18:00:00+00:00",
-        end_at: "2019-03-28T21:00:00+00:00",
-      },
-    ],
-  }
-
-  const mockLoader = jest.fn(() => Promise.resolve(showData))
-
+  let context: Partial<ResolverContext>
   const query = gql`
     {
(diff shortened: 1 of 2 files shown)


src/schema/v2/__tests__/show_events.test.ts
```diff
@@ -3,56 +3,101 @@ import gql from "lib/gql"
...
+
+      const data = await runQuery(query, context)
+      expect(data).toEqual({
+        show: {
+          events: [
+            {
+              endAt: "2019-03-28T21:00:00+00:00",
+              startAt: "2019-03-28T18:00:00+00:00",
+              dateTimeRange: "Thu, Mar 28, 2019 from 2:00 – 5:00pm EDT",
+            },
```
jpotts244, line 97: Wondering if this spec might break when we roll into of daylights savings time (EST -> EDT)? Can test for that

Mitch-Henson (approved):
Nice sleuthing 🕵️‍♀️

dblandin (approved):
Nice find! 👌🏼

jpotts244:
Thanks all for review, wiped it due to a quick rebase, going to merge now!

PR #4749, opened 2023-02-03
feat(artwork lists): support 'saves' and 'default' args

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-4596

Adds support for `default` and `saves` fields, which are already support at the upstream Gravity endpoint, e.g.:

```graphql
{
  me {
    defaultSaves: collection(id: "saved-artwork") {
      name
      artworksCount
    }

    #                                                👇🏽              👇🏽 
    artworkLists: collectionsConnection(first: 10, default: false, saves: true) {
      edges {
        node {
          name
          artworksCount
        }
      }
    }
  }
}
```

This would return every collection that is a user-defined saves collection, but not the default saved-artworks collection:

```json
{
  "data": {
    "me": {
      "defaultSaves": {
        "name": "Saved Artwork",
        "artworksCount": 19
      },
      "artworkLists": {
        "edges": [
          {
            "node": {
              "name": "For the kitchen",
              "artworksCount": 3
            }
          },
          {
            "node": {
              "name": "For the living room",
              "artworksCount": 7
            }
          }
        ]
      }
    }
  }
}
```

### DIFF
--- a/src/schema/v2/me/collectionsConnection.ts
+++ b/src/schema/v2/me/collectionsConnection.ts
@@ -5,7 +5,7 @@ import {
 import { convertConnectionArgsToGravityArgs } from "lib/helpers"
 import { CollectionType } from "./collection"
 import { pageable } from "relay-cursor-paging"
-import { GraphQLFieldConfig, GraphQLInt } from "graphql"
+import { GraphQLBoolean, GraphQLFieldConfig, GraphQLInt } from "graphql"
 import { ResolverContext } from "types/graphql"
 
 export const CollectionsConnectionType = connectionWithCursorInfo({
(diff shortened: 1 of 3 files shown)


dimatretyak (approved):
LGTM 👍

src/schema/v2/me/collectionsConnection.ts
```diff
@@ -18,16 +18,22 @@ export const CollectionsConnection: GraphQLFieldConfig<any, ResolverContext> = {
...
     if (!collectionsLoader) return null
 
     const { page, size, offset } = convertConnectionArgsToGravityArgs(args)
+    const { default: default_, saves } = args // 'default' is a reserved word in JS
+
     const { body, headers } = await collectionsLoader({
       user_id: meID,
       private: true,
+      default: default_,
+      saves,
```
dimatretyak, line 36: **suggestion**: what do you think about this?

```javascript
const { body, headers } = await collectionsLoader({
  user_id: meID,
  private: true,
  default: args.default,
  saves: args.saves,
  size,
  offset,
  total_count: true,
})
```

anandaroop: Yeah, that's nicer — changed in 4f34425

dimatretyak (approved):

PR #4760, opened 2023-02-07
feat: skip InquiryRequest creation for in_review offers

mdole:
Per the [tech plan](https://www.notion.so/artsy/Prevent-sellers-receiving-conversations-associated-with-in_review-offers-9d43b056735f489788a17ce4357cfc08), we'll skip creating conversations at offer creation time when the associated order is held for review.

For the time being, this will never actually trigger (even if we were to launch putting orders with offers `in_review`) because of the changes in [this PR](https://github.com/artsy/exchange/pull/1464). To summarize, when a buyer-facing client (Force, Exchange) requests the `state` field exposed by Exchange's GraphQL API, we return `submitted` whenever the order is `in_review` to maintain the illusion that the order is not being held for review, and to avoid having to update clients to know how to display `in_review` orders.

We're planning to discuss https://github.com/artsy/exchange/pull/1503/files in platform practice this week, which is one direction that might help us solve this problem. Exposing & requesting separate buyer/seller-facing states would obviate the need to change the actual `state` and we could use that internally as needed.

Jira ticket: https://artsyproduct.atlassian.net/browse/PRO-59

### DIFF
--- a/src/lib/stitching/exchange/__tests__/stitching.test.ts
+++ b/src/lib/stitching/exchange/__tests__/stitching.test.ts
@@ -515,6 +515,56 @@ describe("submitOfferOrderWithConversation", () => {
     expect(context.submitArtworkInquiryRequestLoader).not.toHaveBeenCalled()
   })
 
+  it("does not call submitArtworkInquiryRequestLoader if order is in_review", async () => {
+    const { resolvers } = await getExchangeStitchedSchema()
+    const resolver = resolvers.Mutation.submitOfferOrderWithConversation.resolve
+    const args = {
+      input: {
+        offerId: "offer-id",
+      },
+    }
+    const order = {
+      orderOrError: {
+        order: {
+          state: "IN_REVIEW",
+          source: "artwork_page",
+          internalID: "order-id",
+          myLastOffer: {
+            note: "test note",
+          },
+          lineItems: {
+            edges: [
+              {
+                node: {
+                  artworkId: "artwork-id",
+                },
+              },
+            ],
+          },
+        },
+      },
+    }
+    mergeInfo.delegateToSchema.mockResolvedValue(order)
+
+    const result = await resolver({}, args, context, { mergeInfo })
+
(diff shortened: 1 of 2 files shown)


mdole:
oh meant to cc @erikdstock on this one - this is the first case I've run into so far where our "just change the `state` field on the way out of Exchange" approach has caused a concrete problem. Looking forward to discussing your PR and maybe moving toward a more complete solution!

iskounen:
**suggestion:**  Can we add a field to the return value from Exchange (or create a new error type like the one used for SCA validation) that lets Metaphysics know whether it should continue with the conversation creation?

mdole:
> suggestion: Can we add a field to the return value from Exchange (or create a new error type like the one used for SCA validation) that lets Metaphysics know whether it should continue with the conversation creation?

That's an interesting idea! IMO it starts to feel like we're stacking workarounds on our workarounds at that point though. Like the thing we care about in this moment is order `state` - so that's what we should use to determine our route. Adding an additional field only for this use-case would certainly be fast and easy, but it feels to me like it would mostly be tech debt and would not solve the larger problem, which is that the complexity and number of our states is increasing and we have fragmented logic for handling that across clients.

So I think the best move is something in the direction of Erik's PR. Hopefully we'll have a fruitful discussion in platform practice on Thursday!

src/lib/stitching/exchange/__tests__/stitching.test.ts
```diff
@@ -515,6 +515,56 @@ describe("submitOfferOrderWithConversation", () => {
...
+        },
+      },
+      fieldName: "commerceSubmitOrderWithOffer",
+      operation: "mutation",
+      schema: expect.anything(),
+      context: expect.anything(),
+      info: expect.anything(),
+      transforms: [expect.anything()],
+    })
+    expect(result).toEqual(order)
```
iskounen, line 564: **nitpick (non-blocking):** I think this is just testing that the mock resolver is working correctly. (We could save a few milliseconds on tests runs.)

iskounen (approved):

PR #4765, opened 2023-02-09
fix(artwork lists): collection artworks pagination

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-4593

Adds missing `page` arg for `collection.artworksConnection`

```graphql
{
  me {
    collection(id: "42427e8172616975e8004242") {
                         👇🏽
      artworksConnection(page: 2) {
        edges {
          node {
            slug
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -3567,7 +3567,6 @@ type Collection {
     first: Int
     last: Int
     page: Int = 1
-    size: Int = 10
     sort: CollectionArtworkSorts
   ): ArtworkConnection
 
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -3566,6 +3566,8 @@ type Collection {
     before: String
     first: Int
     last: Int
+    page: Int = 1
+    size: Int = 10
     sort: CollectionArtworkSorts
   ): ArtworkConnection
 
--- a/src/schema/v2/me/__tests__/collection.test.ts
+++ b/src/schema/v2/me/__tests__/collection.test.ts
@@ -151,7 +151,7 @@ describe("artworksConnection", () => {
         query {
           me {
             collection(id: "123-abc") {
-              artworksConnection(page: 2, size: 3, sort: SAVED_AT_DESC) {
+              artworksConnection(page: 2, sort: SAVED_AT_DESC) {
                 totalCount
                 edges {
                   node {
@@ -170,7 +170,6 @@ describe("artworksConnection", () => {
         context.collectionArtworksLoader as jest.Mock
       ).toHaveBeenCalledWith(mockGravityCollection.id, {
         page: 2,
-        size: 3,
         sort: "-created_at",
         user_id: "user-42",
         private: true,
--- a/src/schema/v2/me/__tests__/collection.test.ts
+++ b/src/schema/v2/me/__tests__/collection.test.ts
@@ -144,4 +144,38 @@ describe("artworksConnection", () => {
       },
     })
   })
+
+  describe("pagination", () => {
+    it("supports page and size params", async () => {
+      query = gql`
+        query {
+          me {
+            collection(id: "123-abc") {
+              artworksConnection(page: 2, size: 3, sort: SAVED_AT_DESC) {
+                totalCount
+                edges {
+                  node {
+                    slug
+                  }
+                }
+              }
+            }
+          }
+        }
+      `
+
+      await runAuthenticatedQuery(query, context)
+
+      expect(
+        context.collectionArtworksLoader as jest.Mock
+      ).toHaveBeenCalledWith(mockGravityCollection.id, {
+        page: 2,
+        size: 3,
+        sort: "-created_at",
+        user_id: "user-42",
+        private: true,
+        total_count: true,
+      })
+    })
+  })
 })
--- a/src/schema/v2/me/collection.ts
+++ b/src/schema/v2/me/collection.ts
@@ -31,7 +31,6 @@ export const CollectionType = new GraphQLObjectType<any, ResolverContext>({
             defaultValue: CollectionArtworkSorts.getValue("SAVED_AT_DESC")!
               .value,
           },
-          size: { type: GraphQLInt, defaultValue: 10 },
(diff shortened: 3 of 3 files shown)


dimatretyak:
I removed the `size` argument as it duplicates the functionality of the `first` argument

dimatretyak (approved):

anandaroop:
Thanks for merging, but I'm interested in the `size` thing —

My understanding is that `page` + `size` go together for "traditional" pagination, whereas `first` + `after` and `last` + `before` go together for [cursor-based connection pagination](https://www.apollographql.com/blog/graphql/explaining-graphql-connections/).

So `size` and `first` don't really represent the same concept.

In fact many if not most of our connections support both styles of pagination, and thus accept `size` as well as `first`?

Examples [here](https://github.com/artsy/metaphysics/blob/99a12ac1c449379f128e5ad01794f64437cadb7a/_schemaV2.graphql#L1315-L1323), [here](https://github.com/artsy/metaphysics/blob/99a12ac1c449379f128e5ad01794f64437cadb7a/_schemaV2.graphql#L1444-L1477) and [here](https://github.com/artsy/metaphysics/blob/99a12ac1c449379f128e5ad01794f64437cadb7a/_schemaV2.graphql#L9428-L9431)

My 2¢ is that we should add that back in, what do you think?

dimatretyak:
@anandaroop yes, maybe you're right. I have nothing against `size'. it just seemed like it wasn't necessary to use it here

dimatretyak:
> In fact many if not most of our connections support both styles of pagination, and thus accept size as well as first?

to tell the truth, it seemed to me that the `size` argument is not particularly used and for this reason I decided to remove it

image

damassi:
Size is a not-too-connection-y concept which is a bit confusing when mixed with `first`, imo. If `size` is being used anywhere at the field level its likely tech debt.

anandaroop:
Yeah agree that neither `size` should not be mixed with `first` — but nor should `page` be mixed with `first` either.

If you specify `page`, how else would you control the page size without a `size` argument?

anandaroop:
Oh h/t @dzucconi for writing this up:

https://github.com/artsy/metaphysics/blob/main/docs/pagination.md#2-pageable-arguments

We can still consider this a best-practice reference right?

dzucconi:
Yep, you'd want to include `size` if you're including `page`. We can't do unions in input types unfortunately, so yes: technically you can pass both and technically that's incorrect: don't do that obviously.

dzucconi:
And `size` is not tech debt. `first` is just a different concept for cursor based arguments.

You wouldn't mix `first` and `last`, for instance too btw.

damassi:
Well I mean, its typically not used with our connection-based approaches. When we utilize pageable and the helper and pass in `first`, etc, its converted to size which we then pass to gravity. It is uncommon to see it actually used at the field level, and that's what i mean about it being debt; its old and perhaps misconfigured for our current intents and purposes. Do correct me if I'm wrong here!

dzucconi:
We don't actually do real cursor based pagination. We support numeric pages. `page` and `size` are the natural arguments to support that. `page` and `first` does not make any sense.

dzucconi:
And it's not uncommon in our schema either, every piece of UI that uses numeric pages is using `page` + `size`

anandaroop:
Going back to the use case that originated this ticket, we currently see examples like this in client code:

https://github.com/artsy/force/blob/6aad98f3bbf84b40f4afd9731d36ae8ee08218f9/src/Apps/CollectorProfile/Routes/Saves2/Components/SavesArtworks.tsx#L107

Two points:

- To me that is an improper combination of `page` and `first` — we mixing two different pagination metaphors there
- Not clear if this should be doing page+size or first+after as a best-practice — I don't think we are nearly consistent about this across the codebase

dzucconi:
Personally: I'd find it a lot more annoying to have to fetch the page cursors, associate those with the numeric pages, pull out the correct cursor on click and query param just so I can use `first/after`, when I could just use `page/size` and not have to think about that. I see no problem with supporting both.

dzucconi:
And despite it being 'incorrect' to use `size/first` interchangeably.... it also doesn't matter when both are supported.

anandaroop:
Hmm but are they really interchangeable? I guess if you are on page 1 of an unchanging list then yes, but otherwise the point of first/after cursor pagination is to be resilient to changing lists. 

I recognize it may "just work" most of the time, I just want to save myself from a footgun by keeping the two pagination metaphors nice and separate.

dzucconi:
first after isn't resilient to this because they aren't real cursors though.

anandaroop:
Ah, yeah that is an important point

Okay, moving this to #graphql-yelling

Thanks for all the input though, very helpful.

mzikherman:
We support + prefer the 'proper' connection-style arguments of `first` and `after`. Let's not even think about `last` and `before` b/c i don't think we do 'backwards' scrolling anywhere, or we can relegate that to a more niche mobile-specific sitation.

So basically, use `first` and `after` almost everywhere/always.

There's one use-case where we _do_ want to support `page` and `size` as connection-level arguments to a connection. Whether a little custom validation needs to be written that checks you are using either/or style, for a connection that supports both, would be gravy.

The reason _some_ connections might want to support `page` and `size` is to allow a user to load an arbitrary page for the first time. This is exclusively for web, and basically handles someone sharing a link of the form `?page=10`. If your connection isn't going to be used in that context you can stay with `first`/`last`.

dzucconi:
Here's a question: why even bother supporting `first` and `after`?

mzikherman:
We don't have to, really. I do recall some (early) thing with Relay where it straight up complains if you don't provide some connection-approved arguments to a connection (or maybe the outdated lib we use in MP that brings some Relay/connection helpers).

mzikherman:
One more advantage of using `first/after` is that it's opaque to the client and _if we ever changed that implementation and was more cursor-like_ clients may continue to work seamlessly.

However, given that we've not ever done that, I'm willing to say that's a purely hypothetical advantage.

damassi:
+1 in regarding Relay and graphql idioms. The compiler is finnicky, still.

PR #4771, opened 2023-02-13
feat: expose isInquireable on partner

gkartalis:
## Description

This PR resolves [FX-4584](https://artsyproduct.atlassian.net/browse/FX-4584)

Exposes isInquireable on partner

Screenshot 2023-02-13 at 12 02 42

[FX-4584]: https://artsyproduct.atlassian.net/browse/FX-4584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/partner/partner.ts
+++ b/src/schema/v2/partner/partner.ts
@@ -576,6 +576,11 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLBoolean,
         resolve: ({ default_profile_public }) => default_profile_public,
       },
+      isInquireable: {
+        type: GraphQLBoolean,
+        description: "If the partner supports inquiries",
+        resolve: ({ inquireable }) => inquireable,
+      },
       isLinkable: {
         type: GraphQLBoolean,
         resolve: ({ default_profile_id, default_profile_public, type }) =>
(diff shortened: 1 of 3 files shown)


dimatretyak (approved):
LGTM 👍

src/schema/v2/partner/partner.ts
```diff
@@ -576,6 +576,11 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
         type: GraphQLBoolean,
         resolve: ({ default_profile_public }) => default_profile_public,
       },
+      isInquireable: {
+        type: GraphQLBoolean,
```
dblandin, line 580: **suggestion / non-blocking**: We generally prefer to make resolver types non-nullable if possible although I notice that many types in this Partner type are already nullable. Since the definition of `inquireable?` in Gravity ([source](https://github.com/artsy/gravity/blob/9fed3fe7e8cb20e50fb5bf268f9c456d149038aa/app/models/domain/partner.rb#L356-L358) 🔒) always returns a boolean and this JSON property should always be present in the JSON response, what do you think of upgrading this field to non-null?

```ruby
isInquireable: {
  type: new GraphQLNonNull(GraphQLBoolean),
```

dblandin (approved):
LGTM! Left one non-blocking comment 👍🏼

src/schema/v2/partner/partner.ts
gkartalis: that makes total sense, updating

dimatretyak (approved):

PR #4781, opened 2023-02-14
Feat(CX-3426): Allow MyCollection Users add Notes to Artworks

lordkiz:
This PR resolves [CX-3426]

### Description
- Allows users add personal notes to their MyCollection Artworks.

[CX-3426]: https://artsyproduct.atlassian.net/browse/CX-3426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2035,7 +2035,7 @@ type Artwork implements Node & Searchable & Sellable {
   ): AuctionResultConnection
   conditionDescription: ArtworkInfoRow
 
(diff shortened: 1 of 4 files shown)


MrSltun (approved):
Looks great! 🌟 
but I have a question, can we query for the field when it is only defined in the mutations?

PR #4790, opened 2023-02-16
fix: added missing field that was needed in stitching logic

iskounen:
When an offer order is submitted, Metaphysics decides whether a Conversation needs to be created for that order. In #4785 we updated this logic, but forgot to request a field from the Exchange schema that is used in the new logic. This PR makes the request for that field.


mdole:
🙏 🙏 🙏

PR #4800, opened 2023-02-20
Fix Artworks not enriched with insights

lordkiz:
### Description
This PR removes the redundant forcing of `size: 100` when using the `sortByLastAuctionResultDate`, which when used populates marketInsights on artworks with nulls.

<table>
  <tr>
    <td>Before</td>
     <td>After</td>
  </tr>
  <tr>
    <td valign="top"></td>
<td valign="top"></td>
  </tr>
 </table>

### DIFF
--- a/src/schema/v2/me/myCollection.ts
+++ b/src/schema/v2/me/myCollection.ts
@@ -86,9 +86,7 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
       return null
     }
 
-    const paginationArgs = options.sortByLastAuctionResultDate
-      ? { size: 100 }
-      : convertConnectionArgsToGravityArgs(options)
+    const paginationArgs = convertConnectionArgsToGravityArgs(options)
 
     const gravityOptions = {
       exclude_purchased_artworks: options.excludePurchasedArtworks,


MounirDhahri (approved):
Impressive! do you know why it does that?

olerichter00 (commented):
Great work!

I have two questions:
1. How many artworks will we fetch without specifying the size to 100. Are we still fetching all necessary artworks?
2. Why fixes not specifying the size the issue? 🤔

src/schema/v2/me/myCollection.ts
```diff
@@ -86,9 +86,7 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
       return null
     }
 
-    const paginationArgs = options.sortByLastAuctionResultDate
-      ? { size: 100 }
-      : convertConnectionArgsToGravityArgs(options)
+    const paginationArgs = convertConnectionArgsToGravityArgs(options)
```
olerichter00, line 89: Would the following work as well? Not sure whether we need the `size: 100` for some reason.

```suggestion
    const paginationArgs = options.sortByLastAuctionResultDate
    ? { ...convertConnectionArgsToGravityArgs(options), size: 100 }
    : convertConnectionArgsToGravityArgs(options)
```

lordkiz:
@MounirDhahri @olerichter00 I have added a ticket to the backlog to revisit how we handle the enriching: https://artsyproduct.atlassian.net/browse/CX-3438

lordkiz:
> Great work!
> 
> I have two questions:
> 
> 1. How many artworks will we fetch without specifying the size to 100. Are we still fetching all necessary artworks?
> 2. Why fixes not specifying the size the issue? 🤔

1 - The size is completely ignored in this case when we specify first or last. So the number of artworks specified in first / last is returned. Clients can specify size so this seems to be redundant.
2 - I really do not know. We only use this `sortByLastAuctionResultDate` arg in MedianPriceRail, where we also specify the `first: 3` making the need for forcing 100 when we have this arg present seem needless. Do you know why we need to force 100 here?

lordkiz:
> Impressive! do you know why it does that?
I really do not know. And I also do not know why we force size: 100 when using this sortByLastAuctionResultDate arg.

lordkiz:
We used to force size 100 on myCollection because we did not support pagination for it. Is that still the case @olerichter00 ?

MounirDhahri:
I guess we can merge this already if it fixes the bug and do more investigation later

MounirDhahri:
@olerichter00 i trust your judgement here, can we do more investigation here to fix this properly.

olerichter00:
I added a comment [here](https://github.com/artsy/metaphysics/pull/4800#discussion_r1112024171) that could be a potential solution. But we should debug the issue to find out why `marketInsights` is not set properly.

src/schema/v2/me/myCollection.ts
lordkiz: This does not fix the bug, unfortunately.
With this approach there is also a high chance of passing conflicting params that are not in sync at all. For eg, requesting for the last 3 while also requesting for a size of 100.
Note, We automatically derive `limit` from these options and that would inform the `size`.

lordkiz:
> I added a comment [here](https://github.com/artsy/metaphysics/pull/4800#discussion_r1112024171) that could be a potential solution. But we should debug the issue to find out why `marketInsights` is not set properly.

That suggestion does not fix the bug, unfortunately.
With that approach there is also a high chance of passing conflicting params that are not in sync at all. For eg, requesting for the last 3 while also requesting for a size of 100.
Note, We automatically derive `limit` from these options and that would inform the `size`.
I still maintain that clients should send the args. Forcing size breaks insights on some accounts, removing that logic works fine with all accounts I have tested.

MounirDhahri:
After talking to product about this, fixing this quickly is not an urge right now. Let's investigate why would this break when the size is specified. In my opinion, specifying a size should not break the API anyway. We already do that everywhere else.

olerichter00:
> removing that logic works fine with all accounts I have tested.

It may work randomly on some accounts, but it breaks logic (see https://github.com/artsy/metaphysics/pull/4800#issuecomment-1437167596). 

> This PR removes the redundant forcing of size: 100 when using the sortByLastAuctionResultDate, which when used populates marketInsights on artworks with nulls.

I cannot reproduce the issue and I can't find a field called `marketInsights` somewhere in the schema. Are you referring to [marketPriceInsights](https://github.com/artsy/metaphysics/blob/37e7b6a4143b4c17ff3b00edf69dc7c6e27c2f65/_schemaV2.graphql#L2205)?

lordkiz:
> It may work randomly on some accounts, but it breaks the logic (see [#4800 (comment)](https://github.com/artsy/metaphysics/pull/4800#issuecomment-1437167596)).

I see. Apparently this forcing of size may have never worked. We do not do any slicing in MedianAuctionRail, neither do we on Metaphysics. If it did we'd be getting more than 3 nodes when we pass the `sortByLastAuctionResultDate` arg into myCollectionConnection, I suppose. It is possible we probably just didn't realise it.

olerichter00:
> I see. Apparently this forcing of size may have never worked. We do not do any slicing in MedianAuctionRail, neither do we on Metaphysics. If it did we'd be getting more than 3 nodes when we pass the `sortByLastAuctionResultDate` arg into myCollectionConnection, I suppose. It is possible we probably just didn't realise it.

The slicing happens [here](https://github.com/artsy/metaphysics/blob/37e7b6a4143b4c17ff3b00edf69dc7c6e27c2f65/src/schema/v2/me/myCollection.ts#L157).

MounirDhahri:
that's possible, we were though the median sale price in some cases, probably this was broken for power users. It's a good reminder to test on accounts with a big collection as well during our QA

olerichter00:
This should fix the issue: https://github.com/artsy/metaphysics/pull/4806

lordkiz:
> This should fix the issue: #4806

Thanks 🙏

PR #4801, opened 2023-02-20
chore: Add comment to explain sortByLastAuctionResultDate size

olerichter00:
## Description

Add comment to explain sortByLastAuctionResultDate size.

### DIFF
--- a/src/schema/v2/me/myCollection.ts
+++ b/src/schema/v2/me/myCollection.ts
@@ -86,6 +86,7 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
       return null
     }
 
+    // We're setting the size to 100 here because we're going to filter and sort these artworks later manualy.
     const paginationArgs = options.sortByLastAuctionResultDate
       ? { size: 100 }
       : convertConnectionArgsToGravityArgs(options)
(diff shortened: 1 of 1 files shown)


src/schema/v2/me/myCollection.ts
```diff
@@ -86,8 +88,10 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
       return null
     }
 
+    // We're setting the size to 100 here because we're going to filter and sort these artworks later manualy.
+    // The idea is to fetch (almost) all artworks and then filter/sort them in memory.
     const paginationArgs = options.sortByLastAuctionResultDate
-      ? { size: 100 }
+      ? { size: MAX_COLLECTION_SIZE }
```
MounirDhahri, line 94: do we necessarily need to remove the other args here?
```suggestion
      ? { ...convertConnectionArgsToGravityArgs(options), size: MAX_COLLECTION_SIZE,  }
```

MounirDhahri (approved):
LGTM. Thanks for adding this comment

src/schema/v2/me/myCollection.ts
olerichter00: This could be a potential solution for this bug: https://github.com/artsy/metaphysics/pull/4801. Let's keep the logic like it was in this chore and work on a solution in https://github.com/artsy/metaphysics/pull/4800.

olerichter00: But yeah, sounds like a good idea to keep the other arguments. But I'm not sure if it has any negative consequences 🤔

MounirDhahri: oki

PR #4803, opened 2023-02-21
feat(NX-3383): Expose the associated inquiry request with a conversation

jo-rs:
Addresses [NX-3383]

Exposes the associated `InquiryRequest` as part of the conversation. This allows us to access the questions and shipping locations stored on Gravity, which mobile app users might fill in on initial inquiries.

cc @artsy/negotiate-devs 

[NX-3383]: https://artsyproduct.atlassian.net/browse/NX-3383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/partner/partnerInquirerCollectorProfile.ts
+++ b/src/schema/v2/partner/partnerInquirerCollectorProfile.ts
@@ -28,24 +28,3 @@ export const PartnerInquirerCollectorProfile: GraphQLFieldConfig<
   type: InquirerCollectorProfileType,
   description: "Inquiry requester's profile",
 }
-
-export const InquiryRequestType = new GraphQLObjectType<any, ResolverContext>({
-  name: "PartnerInquiryRequest",
-  fields: {
-    collectorProfile: {
-      type: InquirerCollectorProfileType,
-      resolve: (
-        { id: inquiryId, partnerId },
-        _args,
-        { partnerInquirerCollectorProfileLoader }
-      ) => {
-        if (!partnerInquirerCollectorProfileLoader) return
-
-        return partnerInquirerCollectorProfileLoader({
-          partnerId,
-          inquiryId,
-        })
-      },
-    },
-  },
-})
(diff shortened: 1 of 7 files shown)


src/schema/v2/partner/partnerInquirerCollectorProfile.ts
```diff
@@ -28,24 +28,3 @@ export const PartnerInquirerCollectorProfile: GraphQLFieldConfig<
...
-        if (!partnerInquirerCollectorProfileLoader) return
-
-        return partnerInquirerCollectorProfileLoader({
-          partnerId,
-          inquiryId,
-        })
-      },
-    },
-  },
-})
```
jo-rs, line 51: This was moved to its separate file since the name scope is not only relevant to the Collector Profile.

src/schema/v2/partner/partnerInquiryRequest.ts
```diff
@@ -0,0 +1,36 @@
...
+  fields: () => ({
+    ...InternalIDFields,
+    shippingLocation: {
+      type: LocationType,
+      resolve: ({ inquiry_shipping_location }) => inquiry_shipping_location,
+    },
+    questions: {
+      type: new GraphQLList(InquiryQuestionType),
+      resolve: ({ inquiry_questions }) => inquiry_questions,
+    },
```
jo-rs, line 19: These are the two added fields; the rest was moved from the `partnerInquirerCollectorProfile.ts` file.

src/schema/v2/partner/partnerInquirerCollectorProfile.ts
jo-rs: I don't know why the Collector Profile needs to be part of this since we can access that separately, but that is out of scope at the moment.

araujobarret (approved):

PR #4805, opened 2023-02-21
fix(NX-3465): attachments array in SendMessageMutation

araujobarret:
[NX-3465]

Attachments were not correctly parsed as`array`, the data loader parses the parameter as query string like `path?attachments[][name]="name.jpg"` instead of providing an index.
Spreading the array inside an object fixes the query string to be like `path?attachments[0][name]="name.jpg"&attachments[1][name]="2nd.png"`.

[NX-3465]: https://artsyproduct.atlassian.net/browse/NX-3465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/conversation/send_message_mutation.ts
+++ b/src/schema/v2/conversation/send_message_mutation.ts
@@ -19,7 +19,7 @@ interface SendConversationMessageMutationProps {
     url: string
     id?: string
     size?: string
-  }
+  }[]
   bodyHTML?: string
   bodyText: string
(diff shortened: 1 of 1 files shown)


lidimayra (approved):
This was an interesting one, nice work on quickly identifying the spot! 🏅

PR #4806, opened 2023-02-21
fix: Filter null values in price insights batch loader input

olerichter00:
## Description

We need to filter out `null` values in the price insights batch loader input because the params `artistId` and `medium` are [required in Vortex](https://github.com/artsy/vortex/blob/34991742b702d00e697399acce769e64373cc5f9/app/graphql/types/artist_id_medium_tuple_type.rb#L6) and the loader will fail silently when we send null values.

Related issue: https://github.com/artsy/metaphysics/pull/4800

### DIFF
--- a/src/lib/loaders/loaders_with_authentication/vortex.ts
+++ b/src/lib/loaders/loaders_with_authentication/vortex.ts
@@ -49,12 +49,14 @@ export default (accessToken, opts) => {
     vortexTokenLoader,
     vortexGraphqlLoader,
     marketPriceInsightsBatchLoader: async (
-      params: { artistId: string; medium: string; category: string }[]
+      params: MarketPriceInsightsBatchLoaderParams
     ) => {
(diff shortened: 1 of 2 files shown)


src/lib/loaders/loaders_with_authentication/vortex.ts
```diff
@@ -49,12 +49,14 @@ export default (accessToken, opts) => {
...
-      const artistIDMediumTuples = params.map((artist) => ({
-        artistId: artist.artistId,
-        medium: artist.category,
-      }))
+      const artistIDMediumTuples = params
+        .map((artist) => ({
+          artistId: artist.artistId,
+          medium: artist.category,
+        }))
+        .filter((tuple) => tuple.artistId && tuple.medium)
```
olerichter00, line 59: This is the actual fix ☝️

MounirDhahri (approved):
nice! thanks for looking further into this

PR #4811, opened 2023-02-21
Revert "fix: Filter null values in price insights batch loader input"

MounirDhahri:
Reverts artsy/metaphysics#4806

Reverting this since it's breaking the mutation https://artsynet.sentry.io/issues/3891971626/?project=5867225&query=is%3Aunresolved&referrer=issue-stream

Let's follow-up on this tomorrow morning


olerichter00 (commented):
Thanks for the quick fix 🙏 

Let's find out tomorrow why it's breaking the mutation.

MounirDhahri:
Sure - anytime

PR #4816, opened 2023-02-22
feat(NX-3441): Expose a formatted first message including questions

jo-rs:
Partly resolves [NX-3441]

Exposes a `formattedFirstMessage` field on the `inquiryRequest`, which includes the first message and the formatted inquiry questions and location. If no questions are present, it'll default to the first message and return only the formatted questions if the opposite is true.

Example:
```
{
   conversation (id: "123") {
      inquiryRequest: {
         message: "Hello world!",
         questions: [
             {
                 "internalID": "shipping_quote",
                 "question": "Shipping"
              }, {
                  "internalID": "condition_and_provenance",
                  "question": "Condition & Provenance"
              }
          ],
         shippingLocation: {
            city: "Montevideo",
            country: "Uruguay"
         },
         formattedFirstMessage: "Hello world!\n\nI would like to request the following information about this artwork:\n• Shipping Quote to Montevideo, Uruguay\n• Condition & Provenance"
      }
   }
}

```
Which can be rendered in the frontend as:

> Hello world!
> 
> I would like to request the following information about this artwork:
> • Shipping Quote to Montevideo, Uruguay
> • Condition & Provenance

cc @artsy/negotiate-devs 

[NX-3441]: https://artsyproduct.atlassian.net/browse/NX-3441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/conversation/__tests__/conversation.test.js
+++ b/src/schema/v2/conversation/__tests__/conversation.test.js
@@ -575,6 +575,7 @@ describe("Me", () => {
                   state
                   city
                 }
+                formattedFirstMessage
               }
             }
           }
@@ -586,11 +587,12 @@ describe("Me", () => {
           }
         )
       })
-      it("returns the questions and shipping location when present", () => {
+      it("returns the formatted first message, questions, and shipping location when present", () => {
         const newContext = {
           ...context,
           partnerInquiryRequestLoader: () =>
             Promise.resolve({
+              message: "Hello world!",
               inquiry_questions: [
                 { id: "shipping_quote", question: "Shipping" },
                 {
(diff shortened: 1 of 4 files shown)


src/schema/v2/conversation/__tests__/conversation.test.js
```diff
@@ -611,6 +613,47 @@ describe("Me", () => {
             expect(inquiryRequest.questions).toHaveLength(2)
             expect(inquiryRequest.questions).toMatchSnapshot()
             expect(inquiryRequest.shippingLocation).toMatchSnapshot()
+            expect(inquiryRequest.formattedFirstMessage).toMatchSnapshot()
```
damassi, line 616: Its totally cool to keep this since there's already snapshots in this file but its better to avoid them completely

damassi (approved):
Looks good! Thanks for moving this out of the client 👍

src/schema/v2/conversation/__tests__/conversation.test.js
jo-rs: Makes sense! I'll merge this for now to unblock the FE work and will block some time on Friday to take out snapshot usage from this file 👍

PR #4819, opened 2023-02-23
Revert "fix(NX-3465): attachments array in SendMessageMutation"

araujobarret:
Reverts artsy/metaphysics#4805


mdole (approved):
🙏 🟢 thanks for investigating!

PR #4820, opened 2023-02-23
fix(sendMessageMutation): attachments parsing

araujobarret:
Related to #4819 

Passing the `headers: true` made the loader return `body` and `header` as response object keys instead of the `body` 
as an object.


jo-rs (approved):
Seems straightforward enough. I assume you've tested on force and v2?

mdole (approved):

araujobarret:
Going to test on v2 before merging, but tested on Force and it works 👍

PR #4824, opened 2023-02-23
Revert "fix: Return all saved artworks from quiz GRO-1547"

MounirDhahri:
Reverts https://github.com/artsy/metaphysics/pull/4746

While investigating [this](https://artsy.slack.com/archives/C02BAQ5K7/p1676982560961129) earlier I noticed that it broke the onboarding on app for, af as I can tell, all users who signed u and didn't skip it the onboarding and have a collection with over 10 artworks (a bit over  1k). I had to sign out early for a personal issue and ask others who were pairing with me to urgently investigate this and revert it but found out it wasn't reverted yet so I'm creating a PR to revert it on my phone while bowling 🎳😁

### DIFF
--- a/src/lib/loaders/loaders_with_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_with_authentication/gravity.ts
@@ -528,7 +528,6 @@ export default (accessToken, userID, opts) => {
         paramKey: "artworks",
         trackingKey: "is_saved",
         entityIDKeyPath: "_id",
-        batchSize: 10,
       }
     ),
     savedArtworksLoader: gravityLoader(


jonallured:
Reverting this change will mean that users taking the Art Quiz on web will land on the results screen and only see the first 10 works they liked. This is a bug but not a crash so if we believe that this revert is urgent enough to warrant that regression then I can see how that tradeoff is worth it.

But I don't feel like I'm in a position to make that call so I'm going to re-assign back over to you @MounirDhahri. More chatter here:

https://artsy.slack.com/archives/C02BAQ5K7/p1677185313301949?thread_ts=1676982560.961129&cid=C02BAQ5K7

MounirDhahri:
Closing this in favour of implementing an infinite scroll view in app

PR #4826, opened 2023-02-24
feat: align alert labels with known artwork medium types

dblandin:
Find corresponding medium type via gene slug and use the defined name. When a corresponding medium type is not found, use `Other (<slug>)` as the display name.

Jira: [FX-4578](https://artsyproduct.atlassian.net/browse/FX-4578) 🔒 

[FX-4578]: https://artsyproduct.atlassian.net/browse/FX-4578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/__tests__/searchCriteriaLabel.test.ts
+++ b/src/schema/v2/__tests__/searchCriteriaLabel.test.ts
@@ -73,114 +73,125 @@ describe("resolveSearchCriteriaLabels", () => {
     ])
   })
 
-  it("formats medium criteria", async () => {
-    const parent = {
-      additionalGeneIDs: [
-        "painting",
-        "photography",
-        "sculpture",
-        "prints",
-        "work-on-paper",
-        "nft",
-        "design",
-        "drawing",
-        "installation",
-        "film-slash-video",
-        "jewelry",
-        "performance-art",
-        "reproduction",
-        "ephemera-or-merchandise",
-      ],
-    }
+  describe("medium criteria", () => {
+    it("formats medium criteria", async () => {
+      const parent = {
+        additionalGeneIDs: [
+          "painting",
+          "photography",
+          "sculpture",
+          "prints",
+          "work-on-paper",
+          "nft",
+          "design",
+          "installation",
+          "film-slash-video",
+          "jewelry",
+          "performance-art",
+          "reproduction",
+          "ephemera-or-merchandise",
+        ],
+      }
 
-    const labels = await resolveSearchCriteriaLabels(parent, _, _, _)
+      const labels = await resolveSearchCriteriaLabels(parent, _, _, _)
 
-    expect(labels).toIncludeAllMembers([
-      {
-        name: "Medium",
-        displayValue: "Painting",
-        value: "painting",
-        field: "additionalGeneIDs",
-      },
-      {
-        name: "Medium",
-        displayValue: "Photography",
-        value: "photography",
-        field: "additionalGeneIDs",
-      },
-      {
-        name: "Medium",
-        displayValue: "Sculpture",
-        value: "sculpture",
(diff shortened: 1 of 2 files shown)


src/schema/v2/__tests__/searchCriteriaLabel.test.ts
```diff
@@ -73,114 +73,125 @@ describe("resolveSearchCriteriaLabels", () => {
...
-        displayValue: "Design",
-        value: "design",
-        field: "additionalGeneIDs",
-      },
-      {
-        name: "Medium",
-        displayValue: "Drawing",
-        value: "drawing",
-        field: "additionalGeneIDs",
-      },
```
anandaroop, line 146: Good thought to consolidate the mappings — just noting that this means:

- the pills for Alerts will now match the medium labels at the _artwork_ level rather than what's exposed in _artwork grid_ medium filters, i.e.

  remap

- **Drawing** (which overlaps with **Works on Paper**) will go from having first-class support to being treated as an "other" medium

If we're okay with sometimes having more verbose labels and "demoting" Drawings somewhat then I'm fine with this change

anandaroop (dismissed):
👍🏽  from me, with the caveats below

src/schema/v2/searchCriteriaLabel.ts
```diff
@@ -146,35 +147,18 @@ function getRarityLabels(attributionClasses: string[]) {
...
-    "film-slash-video": "Film/Video",
-    jewelry: "Jewelry",
-    "performance-art": "Performance Art",
-    reproduction: "Reproduction",
-    "ephemera-or-merchandise": "Ephemera or Merchandise",
-  }
+  return additionalGeneIDs.map((geneID) => {
+    const medium = Object.values(artworkMediums).find(
+      (entry) => entry.mediumFilterGeneSlug === geneID
+    ) || { name: `Other (${geneID})` }
```
dimatretyak, line 153: **suggestion**: :ceiling-cat:

```javascript
const defaultMedium = { name: `Other (${geneID})` }
const artworkMedium = Object.values(artworkMediums).find(
  (entry) => entry.mediumFilterGeneSlug === geneID
)
const medium = artworkMedium ?? defaultMedium
```

src/schema/v2/__tests__/searchCriteriaLabel.test.ts
dblandin: Is there any harm is defining a new "Drawings" entry in the mediumTypes mapping? That way, if an alert is created from a filtered artist grid, the alert pill should still show up as "Drawing" and map to the [drawing](https://www.artsy.net/gene/drawing) gene (vs [work-on-paper](https://www.artsy.net/gene/work-on-paper)).

anandaroop: Meaning [this](https://github.com/artsy/metaphysics/blob/main/src/lib/artworkMediums.ts)? I don't see the harm in that change, other than diluting the meaning of that mapping (since it would no longer be a list of CMS/artwork medium types.

The change in https://github.com/artsy/metaphysics/pull/4826/commits/0d39ff3677caef461f728fd6bfd39963a1331033 is an alternative approach, right? That adds another corner to this logic — but that seems hard to avoid given the medium vs gene indirection.

Seems like a lot of imperfect choices, and I don't have a strong opinion. Could merge this as-is currently.

dblandin: Sounds good. I'll merge this PR as is. I think it's a reasonable compromise for now to extend the CMS/filter mapping and then add an Other fallback.

PR #4838, opened 2023-02-28
feat: newWorksFromGalleriesYouFollow artworks connection

olerichter00:
Addresses [CX-3411]

## Description

Introduce new `newWorksFromGalleriesYouFollow` artworks connection under `me`.

### GraphQL Query

```graphql
{
  me {
    newWorksFromGalleriesYouFollowConnection(
      first: 20
    ) {
      pageInfo {
        endCursor
        hasNextPage
        hasPreviousPage
      }
      edges {
        cursor
        node {
          title
        }
      }
    }
  }
}

```

[CX-3411]: https://artsyproduct.atlassian.net/browse/CX-3411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/me/newWorksFromGalleriesYouFollow.ts
+++ b/src/schema/v2/me/newWorksFromGalleriesYouFollow.ts
@@ -1,0 +1,46 @@
+import { GraphQLFieldConfig } from "graphql"
+import { connectionFromArraySlice } from "graphql-relay"
+import { convertConnectionArgsToGravityArgs } from "lib/helpers"
+import { pageable } from "relay-cursor-paging"
+import { artworkConnection } from "schema/v2/artwork"
+import { ResolverContext } from "types/graphql"
+import { createPageCursors } from "../fields/pagination"
+
+export const newWorksFromGalleriesYouFollow: GraphQLFieldConfig<
+  {},
+  ResolverContext
+> = {
+  type: artworkConnection.connectionType,
+  args: pageable({}),
+  description: "A list of artworks from galleries the user follows.",
+  resolve: async ({}, args, { followedProfilesArtworksLoader }) => {
+    if (!followedProfilesArtworksLoader) return null
+
+    const { page, size, offset } = convertConnectionArgsToGravityArgs(args)
+
+    const { body: artworks, headers } = await followedProfilesArtworksLoader({
+      size,
+      offset,
+    })
+
+    const totalCount = parseInt(headers["x-total-count"] || "0", 10)
+
+    const connection = connectionFromArraySlice(artworks, args, {
+      arrayLength: totalCount,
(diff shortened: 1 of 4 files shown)


src/schema/v2/me/newWorksFromGalleriesYouFollow.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const newWorksFromGalleriesYouFollow: GraphQLFieldConfig<
+  {},
+  ResolverContext
+> = {
+  type: artworkConnection.connectionType,
+  args: pageable({}),
+  description: "A list of artworks from galleries the user follows.",
+  resolve: async ({}, args, { followedProfilesArtworksLoader }) => {
+    if (!followedProfilesArtworksLoader) return null
```
lordkiz, line 17: I think we should consistently throw warnings here. So we can easily diagnose crashes, unauthenticated access attempts etc.

lordkiz (commented):
@olerichter00 This looks good. I just have one concern outlined in the review.

src/schema/v2/me/newWorksFromGalleriesYouFollow.ts
olerichter00: Hmm, I'm not sure we really need to throw a warning. I have a feeling this check is more to silence TypeScript. These authenticated loaders (like `followedProfilesArtworksLoader`) should always be there if the auth token is valid and the user is authenticated.

Returning `null` is the consistent behavior in the codebase right now. We could look together to see if it would be a good idea to raise an error here. What do you think?

olerichter00:
> @olerichter00 This looks good. I just have one concern outlined in the review.

Will merge now. We can look into the comment together, but I wouldn't do it in this PR because it would touch a lot of files.

src/schema/v2/me/newWorksFromGalleriesYouFollow.ts
olerichter00: Perhaps an error is thrown elsewhere when the query tries to access fields without an authenticated user. It would be interesting to find out.

MounirDhahri: we do that I believe in a middleware somewhere I believe. Let me check it out

MounirDhahri: I think it's here https://github.com/artsy/metaphysics/blob/e18262f651dc35419bf317e0e5c5e9f4f9c3ddc0/src/schema/v2/test/utils.ts

PR #4852, opened 2023-03-02
chore(PX-5026): Remove references to artwork availability_hidden

jpotts244:
Addresses: https://artsyproduct.atlassian.net/browse/PX-5026

Removing usage of deprecated concept of `availability_hidden` for artworks and edition sets

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -1134,21 +1134,6 @@ describe("Artwork type", () => {
       })
     })
 
-    it("returns null if work is marked with availability_hidden", () => {
-      artwork.sale_message = "for sale"
-      artwork.availability = "on loan"
-      artwork.availability_hidden = true
-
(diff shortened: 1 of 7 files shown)


Mitch-Henson (approved):
Looks good to me! I guess as part of this epic we need to make [this search for usages](https://github.com/search?q=org%3Aartsy+availability_hidden&type=code&p=1) go down to zero. A lot will be removed as part of this PR and a few will be taken care of from the Gravity PR, but I think there are still a few other places where we can stamp it out! Maybe we make another ticket for the rest or even group it into these 2 tickets?

oxaudo:
This looks great! Think we can merge it first thing on Monday though! Happy weekend :)

oxaudo (approved):

jpotts244:
Thanks for the reviews! I am seeing some refs to `availability_hidden` in Quartz and another repo that seems to be archived + read only so I think we are good to go!

Added a ticket for Quartz clean up

PR #4863, opened 2023-03-09
feat: expose provenance and condition description in artwork versions

starsirius:
https://artsyproduct.atlassian.net/browse/TX-1061

Following the [Gravity work](https://github.com/artsy/gravity/pull/16132/files), this exposes provenance and condition description fields in artwork versions.

Artwork version is only exposed through [stitching](https://github.com/artsy/metaphysics/blob/75106b0c6e05e1543e83a1c0f39feddf3b5965c7/src/lib/stitching/exchange/v2/stitching.ts#L265), and I still want to find a better way of testing it. This PR should unblock UI work first.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2615,6 +2615,9 @@ type ArtworkVersion implements Node {
   # The Artwork Version attribution class
   attributionClass: AttributionClass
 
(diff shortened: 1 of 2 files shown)


AlicanAkyuz (approved):
🙏

PR #4864, opened 2023-03-09
fix: ArtworkRecommendations connection

olerichter00:
## Description

Fix ArtworkRecommendations resolver to calculate `totalCount` correctly.


src/schema/v2/me/artworkRecommendations.ts
```diff
@@ -74,7 +74,7 @@ export const ArtworkRecommendations: GraphQLFieldConfig<
       )
     }
 
-    const totalCount = artworks.length
+    const totalCount = artworkRecommendations.length
```
olerichter00, line 77: This is the fix.

MounirDhahri (approved):

PR #4871, opened 2023-03-10
fix: reverts profile loaders change

dzucconi:
This reverts the one irrelevant change from https://github.com/artsy/metaphysics/pull/4854

I'm assuming this didn't actually intend to be committed but in the future if one ever updates loaders to return headers, you need to find every call site and update it accordingly.


dzucconi:
(This was breaking the followed status of every partner follow button)

PR #4884, opened 2023-03-15
feat(FX-4675): Rename every “Saved Artwork” collection to “All Saves” 

dimatretyak:
### Demo
| Before | After |
| ------------- | ------------- |
| before | after |

### DIFF
--- a/src/schema/v2/me/collection.ts
+++ b/src/schema/v2/me/collection.ts
@@ -85,6 +85,13 @@ export const CollectionType = new GraphQLObjectType<any, ResolverContext>({
       type: new GraphQLNonNull(GraphQLString),
       description:
         "Name of the collection. Has a predictable value for 'standard' collections such as Saved Artwork, My Collection, etc. Can be provided by user otherwise.",
+      resolve: ({ name }) => {
+        if (name === "Saved Artwork") {
+          return "All Saves"
+        }
+
+        return name
+      },
     },
     saves: {
       type: new GraphQLNonNull(GraphQLBoolean),
--- a/src/schema/v2/me/__tests__/collection.test.ts
+++ b/src/schema/v2/me/__tests__/collection.test.ts
@@ -61,6 +61,27 @@ it("returns collection attributes", async () => {
   })
 })
 
+describe("name field", () => {
+  it("should return `All Saves` when collection has `Saved Artwork` name", async () => {
+    context.collectionLoader = jest.fn(() => {
+      return Promise.resolve({
+        ...mockGravityCollection,
+        name: "Saved Artwork",
+      })
+    })
+
+    const response = await runAuthenticatedQuery(query, context)
+
+    expect(response.me.collection.name).toBe("All Saves")
+  })
+
+  it("should return name received from gravity", async () => {
+    const response = await runAuthenticatedQuery(query, context)
+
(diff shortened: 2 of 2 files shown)


dimatretyak:
**question**: should we also rename `Saved Artwork` to `All Saves` in descriptions (like [here](https://github.com/artsy/metaphysics/blob/517c1bf801f7dfd1b0f7fff5b0fa94e58d938ba6/src/schema/v2/me/collection.ts#L82) or [here](https://github.com/artsy/metaphysics/blob/517c1bf801f7dfd1b0f7fff5b0fa94e58d938ba6/src/schema/v2/me/collection.ts#L87))?

src/schema/v2/me/collection.ts
```diff
@@ -85,6 +85,13 @@ export const CollectionType = new GraphQLObjectType<any, ResolverContext>({
       type: new GraphQLNonNull(GraphQLString),
       description:
         "Name of the collection. Has a predictable value for 'standard' collections such as Saved Artwork, My Collection, etc. Can be provided by user otherwise.",
+      resolve: ({ name }) => {
+        if (name === "Saved Artwork") {
+          return "All Saves"
+        }
+
+        return name
+      },
```
dimatretyak, line 94: **question**: should we add an argument and based on it return either the original name or the modified one?

dimatretyak:
```javascript
name: {
  description:
    "Name of the collection. Has a predictable value for 'standard' collections such as Saved Artwork, My Collection, etc. Can be provided by user otherwise.",
  args: {
    defaultCollectionName: {
      description:
        "pass `MODIFIED` to return `All Saves` instead of `Saved Artwork`",
    },
  },
  resolve: ({ name }, { defaultCollectionName }) => {
    if (name === "Saved Artwork" && defaultCollectionName === "MODIFIED") {
      return "All Saves"
    }

    return name
  },
}
```

dimatretyak: https://user-images.githubusercontent.com/3513494/225419765-b1e173cb-72f0-48c3-92b0-86c56b772665.mp4

anandaroop: It's a good thought, but I don't think it will be necessary. This `me.collection` MP field was only added recently, and our new code in Force is its only consumer.

anandaroop (commented):
While this does add a permanent wrinkle to our mental model of the default saves collection, it's hard to argue with the simplicity of this approach, compared to the [alternative](https://artsy.slack.com/archives/C01GXD7AFS5/p1678904717670609).

I think I'd be okay with this. What do you think @nickskalkin @dblandin?

anandaroop (approved):
Let's go ahead and merge this 👍🏽

anandaroop:
Yep, good call.

dimatretyak:
@anandaroop [some feedback](https://artsy.slack.com/archives/C9SATFLUU/p1679493745049649?thread_ts=1678965823.142219&cid=C9SATFLUU) has been received from Erin (yes, I know it hasn't been agreed with Ryan yet, but I don't think this copy will be rejected) and I think we can merge these changes. what do you think about this?

anandaroop:
Yes, works for me. (And easy to change, if needed)

PR #4891, opened 2023-03-17
chore: fix artist interface inside auction result

MounirDhahri:
Screenshot 2023-03-17 at 13 12 13

**Note:**
This change might lead to performance issues in case we query for different artists from different lots but given that we are not doing this in any surface, we won't do the optimisation unless needed.

### DIFF
--- a/src/schema/v2/__tests__/auction_result.test.ts
+++ b/src/schema/v2/__tests__/auction_result.test.ts
@@ -246,6 +246,33 @@ describe("AuctionResult type", () => {
       })
     })
   })
+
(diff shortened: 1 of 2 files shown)


MounirDhahri:
Let's get this merged @dariakoko to let you continue the work. @olerichter00 please feel free to leave your comments of course.

olerichter00 (commented):
This looks good to me. Shouldn't lead to any performance issues in places where we fetch the artists in the resolver of the connection.

@MounirDhahri What do you think about adding a comment about potential performance issues in the description of the field?

PR #4892, opened 2023-03-17
fix(FX-4681):  `createCollection` mutation returns an error if `artworksConnection` is requested

dimatretyak:
Jira ticket: FX-4681

#### Before
before

#### After
after


nickskalkin (approved):
yeah, I like the proposed change 👍

dimatretyak:
Decided to use another approach: https://github.com/artsy/metaphysics/pull/4899

dimatretyak:
Decided to use another approach: https://github.com/artsy/metaphysics/pull/4899

PR #4912, opened 2023-03-28
chore: support Date ASC sorting MP

MounirDhahri:
This PR resolves https://artsyproduct.atlassian.net/browse/CX-3575

Add `date-asc` sorting to metaphysics

### DIFF
--- a/src/schema/v2/auction_result.ts
+++ b/src/schema/v2/auction_result.ts
@@ -37,6 +37,9 @@ export const AuctionResultSorts = {
       DATE_DESC: {
         value: "-sale_date",
       },
+      DATE_ASc: {
+        value: "-sale_asc",
+      },
       PRICE_AND_DATE_DESC: {
         value: "-price_realized_cents_usd,-sale_date",
       },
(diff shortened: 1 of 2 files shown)


src/schema/v2/auction_result.ts
```diff
@@ -37,6 +37,9 @@ export const AuctionResultSorts = {
       DATE_DESC: {
         value: "-sale_date",
       },
+      DATE_ASC: {
+        value: "-sale_date_asc",
```
dariakoko, line 41: **Q:** Where is the value `-sale_date_asc` defined?

MounirDhahri: I will be defining it in Diffusion in a bit 👀 
I am drafting the PR now

dariakoko: Oki :)

joeyAghion: According to our usual conventions, ascending sale date would correspond with a `sale_date` value. This `-` and `...asc` seem to be in conflict.

joeyAghion: Nevermind! I see now this is updated in https://github.com/artsy/metaphysics/pull/4926.

MounirDhahri: Thanks Joey 🙏 and @lordkiz

PR #4920, opened 2023-03-29
chore: adjust isUpcoming logic to match diffusion

MounirDhahri:
cc @egdbear


MounirDhahri:
I am wondering if we can expose a new field in diffusion for lots to tell if they're upcoming or not. That would be more future proof than now

MounirDhahri:
Will merge this since it's affecting prod and we can follow up later with the more robust approach

egdbear:
Cool!

PR #4923, opened 2023-03-29
WIP fetch from contentful

mdole:
Current status: pretty sure auth is correct but we're returning a pending promise so we can't parse it out.

CURL works (will post in Slack).

```graphql
query {
	contentfulArticle(id: "3Rqikff1gXdBsC2xQJvMYe") {
		title
	}
}
```
-->
```graphql
	"data": {
		"contentfulArticle": {
			"title": null
		}
	},
	"extensions": {
		"requests": {
			"contentful": {
				"requests": {
					"entries/3Rqikff1gXdBsC2xQJvMYe": {
						"time": "0s 479.811ms",
						"cache": false,
						"length": "677 B"
					}
				}
			}
		},
	}
}
```


mdole:
Closing! Got as far as "can request some nested content via Insomnia", but wasn't fetching the entry when I tried to hook local Force & MP up. Not planning to spend any more time on this at present, but happy to share what I learned in the future if we go this route!

PR #4934, opened 2023-04-10
refactor: deprecate conversation's fromUser in favor of fromProfile

joeyAghion:
Similar to https://github.com/artsy/metaphysics/pull/4856, let's expose the associated collector profile directly on a conversation, rather than nested within a user. The `fromUser` property was added [just a few months ago](https://github.com/artsy/metaphysics/pull/4561) and is [only employed by volt-v2](https://github.com/search?q=org%3Aartsy+fromUser&type=code). No additional user properties are requested. This paves the way to retire partners' access to the user endpoint.

### DIFF
--- a/src/schema/v2/conversation/index.ts
+++ b/src/schema/v2/conversation/index.ts
@@ -28,6 +28,7 @@ import { MessageType } from "./message"
 import { ResolverContext } from "types/graphql"
 import { UserType } from "../user"
 import { InquiryRequestType } from "../partner/partnerInquiryRequest"
+import { CollectorProfileType } from "../CollectorProfile/collectorProfile"
 
 export const BuyerOutcomeTypes = new GraphQLEnumType({
   name: "BuyerOutcomeTypes",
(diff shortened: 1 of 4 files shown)


src/schema/v2/conversation/index.ts
```diff
@@ -270,9 +271,28 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
...
+        if (!collectorProfilesLoader)
+          throw new Error(
+            "A X-Access-Token header is required to perform this action."
+          )
+
+        const { body: profiles } = await collectorProfilesLoader({
+          user_id: from_id,
+        })
+
+        return profiles[0]
```
araujobarret, line 288: Do we have all the time an element inside the profiles array?

joeyAghion: If for some reason no collector profile exists, I believe this will resolve to `null` (which is the best we can do, right?).

araujobarret: yes, the only problem is if `profiles` returns null, then we might see some `Uncaught ReferenceError`

PR #4937, opened 2023-04-11
feat(pref center): update metaphysics to support channel field for push

TMcMeans:
The type of this PR is: **Feat**

This PR solves [GRO-1650]

### Description
The goal of this PR is to update `notificationPreferences` to support a channel field to be passed to Gravity as part of a first step towards rebuilding the user preference center to include push notifications.

Tech plan: https://www.notion.so/artsy/Notification-Preference-Center-Mobile-Web-294c4c1967e9491984048cef8c91893a 

TODO

- [x] Wrap up tests
- [x] Match Eigen and Force to local MP schema to test SavedSearchAlerts query for breaking changes
- [x] Double check case sensitivity in specs

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-1650]: https://artsyproduct.atlassian.net/browse/GRO-1650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -12561,6 +12561,16 @@ type Notification implements Node {
   title: String!
 }
 
+enum NotificationChannel {
+  EMAIL
+  PUSH
+}
+
+enum NotificationChannelInput {
+  EMAIL
+  PUSH
+}
+
 # A connection to a list of items.
 type NotificationConnection {
   counts: NotificationCounts
@@ -12602,13 +12612,14 @@ type NotificationEdge {
 }
 
 type NotificationPreference {
-  channel: String!
+  channel: NotificationChannel!
   id: String!
   name: String!
   status: SubGroupStatus!
 }
 
 input NotificationPreferenceInput {
+  channel: NotificationChannelInput!
   name: String!
   status: SubGroupInputStatus!
 }
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -12602,7 +12602,7 @@ type NotificationEdge {
 }
 
 type NotificationPreference {
-  #  email | push
+  # email | push
   channel: String!
   id: String!
   name: String!
@@ -12610,8 +12610,8 @@ type NotificationPreference {
 }
 
 input NotificationPreferenceInput {
-  #  email | push
-  channel: String!
+  # email | push
+  channel: String
   name: String!
   status: SubGroupInputStatus!
 }
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -12561,16 +12561,6 @@ type Notification implements Node {
   title: String!
 }
 
-enum NotificationChannel {
-  EMAIL
-  PUSH
-}
-
-enum NotificationChannelInput {
-  EMAIL
-  PUSH
-}
-
 # A connection to a list of items.
 type NotificationConnection {
   counts: NotificationCounts
@@ -12612,14 +12602,16 @@ type NotificationEdge {
 }
 
 type NotificationPreference {
-  channel: NotificationChannel!
+  #  email | push
+  channel: String!
   id: String!
   name: String!
   status: SubGroupStatus!
 }
 
 input NotificationPreferenceInput {
-  channel: NotificationChannelInput!
+  #  email | push
+  channel: String!
   name: String!
   status: SubGroupInputStatus!
 }
--- a/src/schema/v2/notification_preferences.ts
+++ b/src/schema/v2/notification_preferences.ts
@@ -32,7 +32,7 @@ const subGroupFields = {
   // Using an enum will cause type errors on the frontend.
   channel: {
     type: new GraphQLNonNull(GraphQLString),
-    description: " email | push ",
+    description: "email | push",
   },
   status: {
     type: new GraphQLNonNull(
@@ -54,8 +54,8 @@ const subGroupInputFields = {
   // NOTE: The 'channel' property is defined as a string instead of an enum because some frontend code relies on it being a string.
   // Using an enum will cause type errors on the frontend.
   channel: {
-    type: new GraphQLNonNull(GraphQLString),
-    description: " email | push ",
+    type: GraphQLString,
+    description: "email | push",
   },
   status: {
     type: new GraphQLNonNull(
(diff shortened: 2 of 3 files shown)


_schemaV2.graphql
```diff
@@ -12576,13 +12586,14 @@ type NotificationEdge {
 }
 
 type NotificationPreference {
-  channel: String!
+  channel: NotificationChannel!
```
dzucconi, line 12589: Did field get used by anything previously?

src/schema/v2/notification_preferences.ts
```diff
@@ -29,15 +30,23 @@ const subGroupFields = {
...
   },
   status: {
     type: new GraphQLNonNull(
       new GraphQLEnumType({
         name: "SubGroupStatus",
         values: {
-          SUBSCRIBED: { value: "Subscribed" },
-          UNSUBSCRIBED: { value: "Unsubscribed" },
+          SUBSCRIBED: { value: "SUBSCRIBED" },
+          UNSUBSCRIBED: { value: "UNSUBSCRIBED" },
```
dzucconi, line 49: Similarly what effect does changing these values have?

_schemaV2.graphql
TMcMeans: `channel` was described when this was created but it wasn't being utilized until now

src/schema/v2/notification_preferences.ts
dzucconi: Values can be anything. If these were used by something else then this would likely constitute a breaking change.

_schemaV2.graphql
TMcMeans: on the gravity side of things it gets carried along with the rest of the data payload for the subscription group to be sent to braze: https://github.com/artsy/gravity/blob/main/spec/api/v1/notification_preferences_endpoint_spec.rb#L11 

but braze only handles email updates for the time being so it wasn't being utilized to switch between logic branches or anything.

dzucconi: Yeah, we'd only be concerned with any frontend queries/mutations that might have used this.

src/schema/v2/notification_preferences.ts
TMcMeans: there isn't anywhere else in the MP schema that these values are being used outside of notification preferences, so I'm not following?

dzucconi: There's a mutation that uses this already in the `PreferencesApp` — changing the value here is going to change what Gravity receives for those.

dzucconi: That might be fine — it also might not be fine. I do not know, you would have to trace this. But also why bother changing it? GraphQL isn't yelling about values.

_schemaV2.graphql
TMcMeans: there are some areas outside of the preferences app that uses channel, but as along as` channel === "email` where its relevant it shouldn't break anything, but fingers crossed!

dzucconi: Not exactly. This is a breaking schema change so the types will break.

TMcMeans: in the case of breaking, then it would have to be reverted back to a string instead of an enum??

dzucconi: And for what it's worth: this is a change worth making — this should have never been `string` but we need to actually verify that nothing was using it previously, otherwise we'll have to deprecate and replace.

TMcMeans: the primary place where this would be a problem would be in SavedSearchAlerts, from what I see

TMcMeans: This is the other place that I see channel being utilized in both eigen & force: https://github.com/artsy/force/blob/main/src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx#L94

dzucconi: I'd check to see if this is used in Eigen before replacing it because then it gets much harder to deal with. But basically: we'd add a `deprecationReason` to this, add a new enum field, and replace/deploy.

dzucconi: Sigh. So yeah. Shame but the path of least resistance here might be to just leave this alone as a string, document the supported types in the description and hope it doesn't happen again.

TMcMeans: its a copy/paste from force in eigen so yes: https://github.com/artsy/eigen/blob/main/src/app/Scenes/SavedSearchAlert/EditSavedSearchAlert.tsx#L44

src/schema/v2/notification_preferences.ts
TMcMeans: it was definitely yelling for me, but worth taking another look on the case sensitivity issues maybe not really being an issue before this gets merged

TMcMeans: graphql would say things like, got "Subscribed" but expected "SUBSCRIBED"

dzucconi: That's more likely your spec.

_schemaV2.graphql
TMcMeans: @jonallured would love your thoughts on this!

dzucconi: Yeah, I think the best idea is to: run Metaphysics locally and sync your local schema with both Eigen and Force to determine if there are type errors.

src/schema/v2/notification_preferences.ts
TMcMeans: my most recent commits actually solves the issues with case transforming the enum types in the specs and mutation/resolver code. It came up as an error in test queries to MP locally so definitely made sense to take another look and fix. I believe this can be resolved now

_schemaV2.graphql
TMcMeans: Will have to locally schema match MP to make test queries on eigen and force and that's tbd so this issue is currently unresolved.

TMcMeans: okay so after locally matching MP schema w/ force I did discover type errors to SavedSearchAlerts so the best course of action (and the easiest thing to do) is to revert channel back into a string. I added some descriptions for acceptable string values and included a comment about this in the code. I think we can call this conversation resolved btw.

jonallured (approved):

PR #4941, opened 2023-04-12
feat: update pagination for auction results.

kiryl-zubarau:
This PR updates page numbers cap logic to be able to load more than 100 pages for auction results. Also added `page` and `size` to use pagination by page for auction results instead of cursor pagination.

JIRA - [DO-675]

[DO-675]: https://artsyproduct.atlassian.net/browse/DO-675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1365,9 +1365,11 @@ type Artist implements EntityWithFilterArtworksConnectionInterface & Node & Sear
 
     # Filter auction results by organizations
     organizations: [String]
+    page: Int
 
     # When true, will only return records for allowed artists.
     recordsTrusted: Boolean = false
+    size: Int
 
     # Filter auction results by Artwork sizes
     sizes: [ArtworkSizes]
--- a/src/schema/v2/artist/index.ts
+++ b/src/schema/v2/artist/index.ts
@@ -235,13 +235,14 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
           },
           sort: AuctionResultSorts,
           state: AuctionResultsState,
+          page: { type: GraphQLInt },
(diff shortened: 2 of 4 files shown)


egdbear (commented):
This looks neat 🚀 
However, this will change behaviour for the all paginated components. 
Therefore I'd appreciate @dzucconi's review on this 😊  We should ask the product if the previous behaviour was intentional.

dzucconi:
I don't really know anything about our current pagination implementation. I can't imagine there was a legitimate reason for a `PAGE_NUMBER_CAP` that seems deeply weird. cc @mzikherman

dzucconi:
```
// In most cases Gravity caps the pagination results to 100 pages and we may not want to return more than that
// otherwise we'll generate links that do not work. As of writing there are three endpoints that do this:
//
// * https://github.com/artsy/gravity/blob/52635528/app/api/v1/filter_endpoint.rb#L38
// * https://github.com/artsy/gravity/blob/52635528/app/api/v1/filter_endpoint.rb#L79
// * https://github.com/artsy/gravity/blob/52635528/app/api/v1/partners_endpoint.rb#L168
```

Extremely deep sigh.

dzucconi:
But yes, for auction results the limitation isn't relevant and this PR looks good.

joeyAghion:
We detected that a lot of our back end services' time was spent serving extremely high-numbered pages of search results and other lists. Essentially, we were being scraped. The high cost may not be a factor here, but any place that has to materialize a large set of results in order to serve later pages should retain a cap.

PR #4942, opened 2023-04-12
feat: fallback image for marketing collection

rajsam003:
Lots of marketing collections do not have thumbnail images attached. When searching by collection name in the app, a collection without a thumbnail image will have a grey image placeholder instead.

This PR gets the `representativeArtworkId` from gravity and uses its `imageUrl` as the fallback image to display as the thumbnail image of those marketing collections.

for eg:

image

### DIFF
--- a/src/lib/stitching/gravity/__tests__/stitching.test.ts
+++ b/src/lib/stitching/gravity/__tests__/stitching.test.ts
@@ -1080,12 +1080,17 @@ describe("gravity/stitching", () => {
           info: expect.anything(),
         })
       })
+    })
 
+    describe("Thumbnail Image", () => {
       it("resolves the thumbnailImage field with a copy of the request context", async () => {
         const { resolvers } = await getGravityStitchedSchema()
         const { thumbnailImage } = resolvers.MarketingCollection
 
-        const parent = { image_url: "https://www.example.com/image.jpg" }
+        const parent = {
+          image_url: "https://www.example.com/image.jpg",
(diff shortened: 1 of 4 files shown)


src/lib/stitching/gravity/__tests__/stitching.test.ts
```diff
@@ -1102,6 +1107,43 @@ describe("gravity/stitching", () => {
...
+          Promise.resolve({
+            images: [imageData],
+          })
+        const sharedContext = {
+          artworkLoader,
+        }
+
+        await thumbnailImage.resolve(parent, args, sharedContext, info)
+
+        expect(sharedContext).toHaveProperty("artworkLoader")
```
nickskalkin, line 1136: **question**: I think this check will always pass, since you are explicitly adding artworkLoader to sharedContext [here](https://github.com/artsy/metaphysics/pull/4942/files#diff-9bdc788fbe3d15873b3600965e16a61d783e2e9257078688b5d7e6a1c58943aeR1131), am I right?

nickskalkin (commented):
Left one question + let's merge [this PR](https://github.com/artsy/gravity/pull/16250) first and sync schema

src/lib/stitching/gravity/__tests__/stitching.test.ts
rajsam003: Yes, I can remove it

nickskalkin (approved):
thanks @rajsam003 🙌

PR #4947, opened 2023-04-14
chore: rename `All Saves` to `Saved Artworks`

dimatretyak:
### Description
We have to rename `All Saves` to `Saved Artworks`. See this [Figma](https://www.figma.com/file/wSV2KNz97Tyj8HkH2ZYAkB/Save-%2B-Lists?node-id=1961-232743&t=GnSm3kVwUDGayQux-0) file

### DIFF
--- a/src/schema/v2/me/collection.ts
+++ b/src/schema/v2/me/collection.ts
@@ -100,7 +100,7 @@ export const CollectionType = new GraphQLObjectType<any, ResolverContext>({
         "Name of the collection. Has a predictable value for 'standard' collections such as Saved Artwork, My Collection, etc. Can be provided by user otherwise.",
       resolve: ({ name }) => {
         if (name === "Saved Artwork") {
-          return "All Saves"
+          return "Saved Artworks"
         }
 
         return name
--- a/src/schema/v2/me/__tests__/collection.test.ts
+++ b/src/schema/v2/me/__tests__/collection.test.ts
@@ -63,7 +63,7 @@ it("returns collection attributes", async () => {
 })
 
 describe("name field", () => {
-  it("should return `All Saves` when collection has `Saved Artwork` name", async () => {
+  it("should return `Saved Artworks` when collection has `Saved Artwork` name", async () => {
     context.collectionLoader = jest.fn(() => {
       return Promise.resolve({
         ...mockGravityCollection,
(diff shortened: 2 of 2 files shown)


src/schema/v2/me/collection.ts
```diff
@@ -100,7 +100,7 @@ export const CollectionType = new GraphQLObjectType<any, ResolverContext>({
         "Name of the collection. Has a predictable value for 'standard' collections such as Saved Artwork, My Collection, etc. Can be provided by user otherwise.",
       resolve: ({ name }) => {
         if (name === "Saved Artwork") {
```
dblandin, line 102: **suggestion**: Could this instead check the `saves` and `default` attributes to determine when to override the name? If `saves=true` and `default=true`, I think that means we should use the default collection name of "Saved Artworks", regardless of what the underlying database name is.

dimatretyak: Added changes (https://github.com/artsy/metaphysics/pull/4947/commits/b0ea60241efe6499ac4187cdaa5918ae43496a5e)

src/schema/v2/me/collection.ts
```diff
@@ -98,9 +98,10 @@ export const CollectionType = new GraphQLObjectType<any, ResolverContext>({
       type: new GraphQLNonNull(GraphQLString),
       description:
         "Name of the collection. Has a predictable value for 'standard' collections such as Saved Artwork, My Collection, etc. Can be provided by user otherwise.",
-      resolve: ({ name }) => {
-        if (name === "Saved Artwork") {
-          return "All Saves"
+      resolve: ({ default: isDefault, saves, name }) => {
```
dimatretyak, line 101: `default` is reserved name in JS

dblandin (approved):

nickskalkin (approved):

src/schema/v2/me/collection.ts
dblandin: You can check this by executing a similar GraphQL query for any user. There should always only be one collection returned, the "standard" collection:

```graphql
{
  me {
    internalID
    collectionsConnection(first: 5, saves: true, default: true) {
      edges {
        node {
          default
          saves
          name
        }
      }
    }
  }
}
```

PR #4952, opened 2023-04-17
feat(NX-3437): add unread fields for Partner and Collector

araujobarret:
[NX-3437]

- Adds unreadByPartner and unreadByCollector fields to the conversation
- deprecates `unread` in favor of `unreadByCollector` to improve the general understanding of it

[NX-3437]: https://artsyproduct.atlassian.net/browse/NX-3437?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/conversation/index.ts
+++ b/src/schema/v2/conversation/index.ts
@@ -329,6 +329,13 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
       type: GraphQLString,
       resolve: ({ from_last_viewed_message_id }) => from_last_viewed_message_id,
     },
+    toLastViewedMessageID: {
+      type: GraphQLString,
+      resolve: ({ from_last_viewed_message_id, ...rest }) => {
+        console.log("### rest", rest)
+        return from_last_viewed_message_id
+      },
+    },
     initialMessage: {
       deprecationReason:
         "This field is no longer required. Prefer the first message from the MessageConnection.",
@@ -463,7 +470,8 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
     messagesConnection,
     unread: {
       type: GraphQLBoolean,
-      description: "True if there is an unread message by the user.",
+      description: "True if there is an unread message by the Collector(from).",
+      deprecationReason: "Use `unreadByCollector` instead",
       resolve: (conversation) => {
         const memoizedLastMessageId = lastMessageId(conversation)
         const { from_last_viewed_message_id } = conversation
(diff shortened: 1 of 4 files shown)


src/schema/v2/conversation/index.ts
```diff
@@ -329,6 +329,13 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
       type: GraphQLString,
       resolve: ({ from_last_viewed_message_id }) => from_last_viewed_message_id,
     },
+    toLastViewedMessageID: {
+      type: GraphQLString,
+      resolve: ({ from_last_viewed_message_id, ...rest }) => {
+        console.log("### rest", rest)
+        return from_last_viewed_message_id
```
damassi, line 336: This might be pretty confusing long-term. Why is the field name different from the field as returned from the API? Shouldn't it be `fromLastViewedMessageID`? If we're going to change it to the inverse of the data being returned there has to be a really good reason 😄

araujobarret: Apologies 🙇 
I forgot to clean up this new field https://github.com/artsy/metaphysics/pull/4952/commits/1a003df7ec83ebed5a6b325324ff35f503354622 should be okay in this commit 😓

damassi (approved):

src/schema/v2/conversation/index.ts
```diff
@@ -463,7 +469,21 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
...
+      resolve: (conversation) => {
+        const memoizedLastMessageId = lastMessageId(conversation)
+        const { from_last_viewed_message_id } = conversation
+        return (
+          !!from_last_viewed_message_id &&
+          !!memoizedLastMessageId &&
+          from_last_viewed_message_id < memoizedLastMessageId
+        )
+      },
+    },
```
damassi, line 483: #non-blocking suggestion: maybe move this resolver into a shared function and then add a test for that function? Not sure if it makes sense or will ergonomically feel right, just thinking that there's enough shared data between the two that we could unify / test.

lidimayra (approved):

src/schema/v2/conversation/index.ts
araujobarret: totally, check it out please https://github.com/artsy/metaphysics/pull/4952/commits/ecaa29a40857c400a175443d0dc8a2b198bea7dd 
any further requests I can do in follow-up PRs

PR #4966, opened 2023-04-24
feat: support marketable flag for artworksForUser

MounirDhahri:
This PR makes the following changes:

- Support `marketable` arg in order to only return marketable artworks inside `artworksForUser` interface.
- Updates loader used for artworksForUser to be the using filter artworks endpoint:
  - Why? The latter not only supports `marketable` filtering  but also uses ES for filtering and should allow, if we want to, better filtering

| Condition                          | Screenshot                                                                                                                                                             |
|------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| When `marketable` is `true`        | Screenshot 2023-04-24 at 15 26 37 |
| When `marketable` is `false`       | Screenshot 2023-04-24 at 15 26 45 |
| When `marketable` is not specified | Screenshot 2023-04-24 at 15 34 36 |

### DIFF
--- a/src/schema/v2/artworksForUser/artworksForUser.ts
+++ b/src/schema/v2/artworksForUser/artworksForUser.ts
@@ -28,6 +28,7 @@ export const artworksForUser: GraphQLFieldConfig<void, ResolverContext> = {
     userId: { type: GraphQLString },
     version: { type: GraphQLString },
     maxWorksPerArtist: { type: GraphQLInt },
+    marketable: { type: GraphQLBoolean },
   }),
   resolve: async (_root, args: CursorPageable, context) => {
     if (!context.artworksLoader) return
--- a/src/schema/v2/artworksForUser/artworksForUser.ts
+++ b/src/schema/v2/artworksForUser/artworksForUser.ts
@@ -13,6 +13,7 @@ import { ResolverContext } from "types/graphql"
 import { artworkConnection } from "schema/v2/artwork"
 import {
   getBackfillArtworks,
+  getMarketableArtworks,
   getNewForYouArtworks,
   getNewForYouRecs,
 } from "./helpers"
@@ -28,7 +29,7 @@ export const artworksForUser: GraphQLFieldConfig<void, ResolverContext> = {
     userId: { type: GraphQLString },
     version: { type: GraphQLString },
     maxWorksPerArtist: { type: GraphQLInt },
-    marketable: { type: GraphQLBoolean },
+    marketable: { type: GraphQLBoolean, defaultValue: false },
   }),
   resolve: async (_root, args: CursorPageable, context) => {
     if (!context.artworksLoader) return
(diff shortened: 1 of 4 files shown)


olerichter00 (approved):
💯

MounirDhahri:
Fixing the broken test now and will test locally in my sim to make sure we didn't break something

olerichter00 (approved):
👍

jonallured:
I think I understand the intent here - filter out works that are not marketable but I'm not clear on how this plays with pagination nor what the performance implications are.

If the query is for the first 10 but one of those is not marketable will I get 10 back? What if the query is for the next page?

I'm also looking at the overall behavior of this query and unless I'm missing something I think this PR introduces more API calls. Have we considered the performance implications of that?

* make API call to Vortex to get the recs for a user
* make API call to Gravity to turn those rec artwork ids into full artworks
* make another API call to Gravity (when marketable flag exists) to remove any artworks that are not marketable

Maybe it's fine but have we considered another approach that could get us down to just one Gravity API call?

In terms of code design, my feedback is that I'd like to see the resolver know less. Can we pass that marketable flag into the helper function and let it to the conditional logic? Something more like:

```
const newForYouArtworks = await getNewForYouArtworks(
      newForYouArtworkIds,
      args.marketable,
      gravityArgs,
      context
    )
```

My thinking is that the resolver shouldn't know about all these different API calls and should just be at the level of deciding whether backfill is required, etc. What do you think?

Finally: when do we ever want recs that are NOT marketable? Why not filter these things out from the start?

src/schema/v2/artworksForUser/artworksForUser.ts
```diff
@@ -28,6 +29,7 @@ export const artworksForUser: GraphQLFieldConfig<void, ResolverContext> = {
     userId: { type: GraphQLString },
     version: { type: GraphQLString },
     maxWorksPerArtist: { type: GraphQLInt },
+    marketable: { type: GraphQLBoolean, defaultValue: false },
```
dblandin, line 32: **suggestion**: Thoughts on making this nullable or unspecified? It feels like with this default of `false`, any queries not specifying an explicit `marketable` value would not receive any marketable works. Is that the behavior we want for currently executing queries?

src/schema/v2/artworksForUser/artworksForUser.ts
```diff
@@ -50,7 +52,16 @@ export const artworksForUser: GraphQLFieldConfig<void, ResolverContext> = {
       context
     )
 
-    const artworks = [...newForYouArtworks, ...backfillArtworks]
+    const allArtworks = [...newForYouArtworks, ...backfillArtworks]
+
+    // Because we load the data from two different sources that do not support
+    // marketable flag, we need to apply the filtering at the end
+    const artworks = args.marketable
+      ? await getMarketableArtworks(
```
dblandin, line 60: **question**: Bummer that we need to re-fetch the artworks here. Do we have a Jira ticket for supporting a similar `marketable` query param at least on the non-fallback artworks API endpoint in Gravity?

MounirDhahri:
Thanks @dblandin and @jonallured for the thorough review. My bad here, I didn't think this thoroughly and thought we need to only solve the problem for the recs notifications and not in most areas of the client apps. Therefore I dismissed the performance issues and the pagination since we only return a few. 

I will close this PR then given it doesn't make sense to do the work here and more thinking is required to build this.

PR #5003, opened 2023-05-10
Revert "chore: make my collection info node"

MounirDhahri:
Reverts artsy/metaphysics#5000
The id from my collection is not returning properly. Reverting this until we fix it.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -12378,13 +12378,7 @@ type MyCollectionConnection {
 
   # A list of edges.
   edges: [MyCollectionEdge]
-
-  # A globally unique ID.
-  id: ID!
   includesPurchasedArtworks: Boolean!
-
-  # A type-specific ID likely used as a database ID.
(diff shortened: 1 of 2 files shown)


olerichter00:
> The id from my collection is not returning properly. Reverting this until we fix it.

Which id is it?

MounirDhahri:
my collection id is available when I am testing from Insomnia, it's being printed but it's raising an error on my simulator. I reverted this to avoid potentially breaking something in prod.

olerichter00:
Seems like MP returns `null` for `pageInfo` which shouldn't be `null`. Is this after refetching the connection?

MounirDhahri:
As you see, in Insomnia it's working well. which leads me to believe we have an issue with something else that only showed up now
Screenshot 2023-05-10 at 11 43 32

olerichter00:
You could compare the query in Insomnia with the query that Relay generates to see if there is a difference.

PR #5014, opened 2023-05-17
feat: support searching IDV records by name

mdole:
Companion to https://github.com/artsy/gravity/pull/16367


iskounen (approved):
LGTM. I'll assign back to you in case this is really a draft.

PR #5022, opened 2023-05-19
feat: add ability to pass `includesArtworkID` arg to `collectionsConnection`

dimatretyak:
Jira ticket: [FX-4784](https://artsyproduct.atlassian.net/browse/FX-4784)

### Demo
https://github.com/artsy/metaphysics/assets/3513494/9ac28f2d-4dca-4219-bb87-248b0d24d71d

[FX-4784]: https://artsyproduct.atlassian.net/browse/FX-4784?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/me/__tests__/collectionsConnection.test.ts
+++ b/src/schema/v2/me/__tests__/collectionsConnection.test.ts
@@ -32,51 +32,49 @@ describe("collectionsConnection", () => {
     }
   })
 
-  describe("passes correct args to Gravity", () => {
-    it("when `sort` arg is specified", async () => {
-      await runAuthenticatedQuery(query, context)
+  it("passes correct args to Gravity", async () => {
+    await runAuthenticatedQuery(query, context)
 
-      expect(context.collectionsLoader as jest.Mock).toHaveBeenCalledWith({
-        user_id: "user-42",
-        private: true,
-        saves: true,
-        sort: "-created_at",
-        offset: 0,
-        size: 1,
-        total_count: true,
-      })
+    expect(context.collectionsLoader as jest.Mock).toHaveBeenCalledWith({
+      user_id: "user-42",
+      private: true,
+      saves: true,
+      sort: "-created_at",
+      offset: 0,
+      size: 1,
+      total_count: true,
     })
+  })
 
-    it("when `includesArtworkID` arg is specified", async () => {
-      const query = gql`
-        query {
-          me {
-            collectionsConnection(
-              first: 1
-              saves: true
-              includesArtworkID: "123-abc"
-            ) {
-              edges {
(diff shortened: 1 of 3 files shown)


src/schema/v2/me/__tests__/collectionsConnection.test.ts
```diff
@@ -32,17 +32,51 @@ describe("collectionsConnection", () => {
...
+      await runAuthenticatedQuery(query, context)
 
-    expect(context.collectionsLoader as jest.Mock).toHaveBeenCalledWith({
-      user_id: "user-42",
-      private: true,
-      saves: true,
-      sort: "-created_at",
-      offset: 0,
-      size: 1,
-      total_count: true,
```
anandaroop, line 45: **suggestion:** Honestly, I don't think adding a whole new test case here is providing much value here.

We just want to know that this resolved sends the expected payload upstream.

I would leave the `it("passes correct args to Gravity")` test as it was, and only make two small changes: 
- add another `includesArtworkID` on L11 above
- add `artwork_id: "123-abc"` to the expected payload here.

anandaroop (changes requested):
Thanks! 

The change looks good, but I would recommend simplifying the test as suggested below.

src/schema/v2/me/__tests__/collectionsConnection.test.ts
dimatretyak: Done: https://github.com/artsy/metaphysics/pull/5022/commits/663ced4abcccf7358259c2d2991c9595afbb765d

src/schema/v2/me/__tests__/collectionsConnection.test.ts
```diff
@@ -43,6 +48,39 @@ describe("collectionsConnection", () => {
       offset: 0,
       size: 1,
       total_count: true,
+      artwork_id: "123-abc",
+    })
+  })
+
+  it("passes correct args to Gravity when `includesArtworkID` arg is specified", async () => {
+    const query = gql`
```
anandaroop, line 56: We just need to remove this test case now, since it's already covered above 😄

anandaroop (approved):

src/schema/v2/me/__tests__/collectionsConnection.test.ts
dimatretyak: Addressed here: https://github.com/artsy/metaphysics/pull/5022/commits/4bb9e4d4861a1434b82f15da72fd40d6112803cb

PR #5024, opened 2023-05-20
chore(GRO-1671): update meta tags based on the selected tab on the Artist page

pam-:
The type of this PR is: **Chore**

This PR solves [GRO-1671]

### Description

* adds `ArtistPageEnumType` to `ArtistMeta` to make it easier to get the tag's content depending on the Artist page tab

This is 🤝 with https://github.com/artsy/force/pull/12430

[GRO-1671]: https://artsyproduct.atlassian.net/browse/GRO-1671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/artist/meta.ts
+++ b/src/schema/v2/artist/meta.ts
@@ -28,7 +28,7 @@ const ArtistMetaType = new GraphQLObjectType<any, ResolverContext>({
       type: GraphQLString,
       resolve: ({ artist, page }) => {
         const blurb = artist.blurb?.length
-          ? ` ${markdownToText(artist.blurb)}`
+          ? ` ${markdownToText(artist.blurb).slice(0, 70)}`
           : ""
 
         switch (page) {
@@ -39,7 +39,7 @@ const ArtistMetaType = new GraphQLObjectType<any, ResolverContext>({
           case "ARTWORKS":
             return `Discover and purchase ${metaName(
               artist
-            )}’s artworks, available for sale. Browse our selection of paintings, prints, and sculptures by the artist, and find art you love.${blurb}`
+            )}’s artworks, available for sale. Browse our selection of paintings, prints, and sculptures by the artist, and find art you love.`
           case "AUCTION_RESULTS":
             return `Find out about ${metaName(
               artist
--- a/src/schema/v2/artist/meta.ts
+++ b/src/schema/v2/artist/meta.ts
@@ -8,9 +8,48 @@ export const metaName = (artist) => {
   return "Unnamed Artist"
 }
 
+export const formatDescription = (description: string, blurb?: string) => {
+  return truncate(compact([description, blurb]).join("."), 157)
+}
+
 const ArtistMetaType = new GraphQLObjectType<any, ResolverContext>({
   name: "ArtistMeta",
   fields: {
+    artworksDescription: {
+      type: GraphQLString,
+      resolve: (artist) => {
+        const blurb = artist.blurb.length
+          ? markdownToText(artist.blurb)
+          : undefined
+
+        return formatDescription(
+          `Discover and purchase ${metaName(
+            artist
+          )}’s artworks, available for sale. Browse our selection of paintings, prints, and sculptures by the artist, and find art you love.`,
+          blurb
+        )
+      },
+    },
+    artworksTitle: {
+      type: GraphQLString,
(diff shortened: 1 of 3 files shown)


src/schema/v2/artist/meta.ts
```diff
@@ -1,45 +1,91 @@
...
+} from "graphql"
 import { ResolverContext } from "types/graphql"
 
 export const metaName = (artist) => {
   if (artist.name) return stripTags(artist.name)
   return "Unnamed Artist"
 }
 
+export const formatDescription = (description: string, blurb?: string) => {
+  return truncate(compact([description, blurb]).join("."), 157)
```
dzucconi, line 17: I don't think this join is what you want. Though it's not clear to me what we do want here? What types of content are in the description and blurb?

src/schema/v2/artist/meta.ts
```diff
@@ -1,45 +1,91 @@
...
+}
+
+const ArtistTabEnumType = new GraphQLEnumType({
+  name: "ArtistTab",
```
dzucconi, line 21: In general, we want to avoid tying our schema to our UI too closely. And I think "tab" gets a little too close to the UI for comfort. "page" is more future proof of a concept. Maybe "mode"? 🤷 

We should also namespace this under any direct types. So this would be `ArtistMetaPage` or `ArtistMetaMode`, in that case.

cc @mzikherman

src/schema/v2/artist/meta.ts
```diff
@@ -1,45 +1,91 @@
...
       },
     },
   },
 })
 
 const Meta: GraphQLFieldConfig<void, ResolverContext> = {
   type: ArtistMetaType,
-  resolve: (x) => x,
+  args: {
+    tab: { type: ArtistTabEnumType },
```
dzucconi, line 84: This should have a default to avoid breaking existing queries.

src/schema/v2/artist/meta.ts
pam-: we want the bio first sentence after the description as specified in [this doc](https://docs.google.com/document/d/1P7f8LINemT1avr_4xU9sBYroefh5b9h3Gkar3vymtKE/edit) -- i was just recycling what was in there before but you're right! i could just build a new string aka `${description} ${blurb}` and truncate the whole thing. Lmk if you have a smoother way to do dis in mind

src/schema/v2/artist/meta.ts
pam-: easy peasy -- will change to `Page`

src/schema/v2/artist/meta.ts
dzucconi: I mean it's joining with simply a `'.'` (no space). You would want to join with `'. '` The compact & join is preferable in the case of nulls.

But this doc is also specifically asking for the first sentence (which honestly I can see that being pretty weird).

src/schema/v2/artist/meta.ts
```diff
@@ -1,45 +1,86 @@
...
+              `Explore ${metaName(
+                artist
+              )}'s biography, achievements, artworks, auction results, and shows on Artsy. ${blurb}`,
+              157
+            )
+          case "ARTWORKS":
+            return truncate(
+              `Discover and purchase ${metaName(
+                artist
+              )}’s artworks, available for sale. Browse our selection of paintings, prints, and sculptures by the artist, and find art you love. ${blurb}`,
```
dzucconi, line 46: Why I think this should have some specs:

```tsx
`Foo. ${null}`
// => 'Foo. null'
`Foo. ${undefined}`
// => 'Foo. undefined'
`Foo. ${null ?? ''}`
// => 'Foo. ' // Note the trailing space
```

pam-: specced it and changed `undefined` to `""` &  made it so the space only shows up if there's a blurb.

I have some Qs, though:
* Google will automatically truncate the length of the element if it's too long, so given that the blurb is supposed to be short anyway, does it matter if we truncate or not?
* inversely, why include the blurb/bio at all, if it's getting truncated? like, will that really impact SEO?

(I'll raise in #grow)

PR #5031, opened 2023-05-23
feat(artist): sorts insights by rank

dzucconi:
@TMcMeans and I paired through this just now.

This PR ensures we sort the insights in a consistent fashion. I don't think there's a ticket for this explicitly.

Also, now we know which insights are missing and there's two which were not ranked (added to the end). Will make tickets for each of those.


TMcMeans (approved):
we paired on this, so it LGTM as long as all specs pass. The overall next step on artist data is to get these missing data points together!

PR #5044, opened 2023-05-30
feat(NX-3536): Add query for collector profiles search.

jo-rs:
Resolves [NX-3536]

Adds query for collector profiles search.

Screenshot 2023-05-30 at 20 11 22

cc @artsy/negotiate-devs 

[NX-3536]: https://artsyproduct.atlassian.net/browse/NX-3536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -14574,7 +14574,7 @@ type Query {
     last: Int
     partnerID: ID
 
-    # If present, will search by term
+    # Term used for searching collector profiles
     term: String
   ): CollectorProfileTypeConnection
 
@@ -19073,7 +19073,7 @@ type Viewer {
     last: Int
     partnerID: ID
 
-    # If present, will search by term
+    # Term used for searching collector profiles
     term: String
   ): CollectorProfileTypeConnection
 
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -3792,6 +3792,26 @@ type CollectorProfileType {
   userInterests: [UserInterest]!
 }
 
+# A connection to a list of items.
+type CollectorProfileTypeConnection {
+  # A list of edges.
+  edges: [CollectorProfileTypeEdge]
+  pageCursors: PageCursors!
+
+  # Information to aid in pagination.
+  pageInfo: PageInfo!
+  totalCount: Int
+}
+
+# An edge in a connection.
+type CollectorProfileTypeEdge {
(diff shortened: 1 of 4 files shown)


src/schema/v2/CollectorProfile/collectorProfiles.ts
```diff
@@ -0,0 +1,59 @@
...
+    { term, partnerID, ...args },
+    { collectorProfilesLoader }
+  ) => {
+    if (!collectorProfilesLoader)
+      throw new Error(
+        "A X-Access-Token header is required to perform this action."
+      )
+
+    const { page, size, offset } = convertConnectionArgsToGravityArgs(args)
+    const { body, headers } = await collectorProfilesLoader({
```
lidimayra, line 40: Would it make sense to add a check for the required args (partnerId and term) before this?

_schemaV2.graphql
```diff
@@ -14546,6 +14566,18 @@ type Query {
...
 
+  # A list of collector profiles that have sent an inquiry to a partner
+  collectorProfilesConnection(
+    after: String
+    before: String
+    first: Int
+    last: Int
+    partnerID: ID
+
+    # If present, will search by term
```
lidimayra, line 14577: minor/non-blocking: I wonder if the `If present, ...` here might give the idea that would also be possible to perform the query without passing a term (which is not true, in case a partnerId is passed, our endpoint in Gravity [requires it to be passed as well](https://github.com/artsy/gravity/blob/9f45bcd3dc7c4ced064aa36d4c4e3f3be8bd7e8a/app/api/v1/collector_profiles_endpoint.rb#L15))

lidimayra (approved):
Great work! Left two small comments, but LGTM! 🚀

src/schema/v2/CollectorProfile/collectorProfiles.ts
jo-rs: It does, indeed! 
Addressed and tested on [310dc2f](https://github.com/artsy/metaphysics/pull/5044/commits/310dc2f27dcfe25e2753d6afd6a5ddd0606b380f).

lidimayra (approved):

PR #5045, opened 2023-05-30
chore(GRO-1671) Follow up

pam-:
Follow up on https://github.com/artsy/metaphysics/pull/5024 and making `ArtistMeta` type non-nullable.
I thought this was committed to my previous PR but the git hooks kept reverting that change so here it is again!


dzucconi (approved):

dzucconi:
Hm, not sure what's up with the CI failure — try rebasing?

PR #5059, opened 2023-06-02
Only return all active hero units

jonallured:
This purpose of this PR is to troubleshoot why we can't seem to get hero units to show up over on the Force review app for https://github.com/artsy/force/pull/12459 when signed out. There were some theories that it was something to do with different loaders being used so here I'm eliminating that as a possibility by only using the basic hero units loader all the time.

Note that this does break forque but that's ok since we haven't fully cut anything over and are still WIP on this stuff.

/cc @artsy/grow-devs

### DIFF
--- a/src/schema/v2/HeroUnit/heroUnitsConnection.ts
+++ b/src/schema/v2/HeroUnit/heroUnitsConnection.ts
@@ -19,22 +19,7 @@ export const heroUnitsConnection: GraphQLFieldConfig<void, ResolverContext> = {
     },
   }),
   resolve: async (_root, args, context) => {
-    const {
-      authenticatedHeroUnitsLoader,
-      heroUnitsLoader,
-      matchHeroUnitsLoader,
-    } = context
-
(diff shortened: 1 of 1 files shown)


src/schema/v2/HeroUnit/heroUnitsConnection.ts
```diff
@@ -19,22 +19,7 @@ export const heroUnitsConnection: GraphQLFieldConfig<void, ResolverContext> = {
...
   resolve: async (_root, args, context) => {
-    const {
-      authenticatedHeroUnitsLoader,
-      heroUnitsLoader,
-      matchHeroUnitsLoader,
-    } = context
-
-    const { term } = args
-
-    if (term && !matchHeroUnitsLoader)
```
damassi, line 30: FWIW we should make sure to _only_ guard against the unauthenticated loader access up above. If we're missing `term`, explicitly error against that (though that shouldn't be necessary since we can make term non-nullable)

PR #5067, opened 2023-06-07
feat(CX-3724): Collect Multiple Artists Mutation

lordkiz:
Addresses: [CX-3724]

### Description
The added mutation allows for collecting multiple artists

[CX-3724]: https://artsyproduct.atlassian.net/browse/CX-3724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/me/collectArtistsMutation.ts
+++ b/src/schema/v2/me/collectArtistsMutation.ts
@@ -45,6 +45,7 @@ export const collectArtistsMutation = mutationWithClientMutationId<
           return meCreateUserInterestLoader(gravityPayload)
         })
       )
+      return {}
     } catch (err) {
       if ("body" in (err as any)) {
         const e = err as GravityError
--- a/src/schema/v2/me/collectArtistsMutation.ts
+++ b/src/schema/v2/me/collectArtistsMutation.ts
@@ -1,0 +1,62 @@
+import { GraphQLList, GraphQLString } from "graphql"
+import { mutationWithClientMutationId } from "graphql-relay"
+import { GraphQLNonNull } from "graphql"
+import { ResolverContext } from "types/graphql"
+import { meType } from "./index"
+
+interface Input {
+  artistIDs: string[]
+  userId: string
+}
+
+export const collectArtistsMutation = mutationWithClientMutationId<
+  Input,
+  any | null,
+  ResolverContext
+>({
+  name: "CollectArtistsMutation",
+  description: "Collect Multiple Artists",
+  inputFields: {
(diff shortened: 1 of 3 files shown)


src/schema/v2/me/collectArtistsMutation.ts
```diff
@@ -0,0 +1,63 @@
...
+export const collectArtistsMutation = mutationWithClientMutationId<
+  Input,
+  any | null,
+  ResolverContext
+>({
+  name: "CollectArtistsMutation",
+  description: "Collect Multiple Artists",
+  inputFields: {
+    artistIDs: { type: new GraphQLNonNull(new GraphQLList(GraphQLString)) },
+    userId: { type: new GraphQLNonNull(GraphQLString) },
```
MrSltun, line 21: **Question:** can we infer the user id from the session instead of sending it with the request?

src/schema/v2/me/collectArtistsMutation.ts
```diff
@@ -0,0 +1,63 @@
...
+    try {
+      await Promise.all(
+        uniqueArtistIDs.map((artistID) => {
+          const gravityPayload = {
+            interest_id: artistID,
+            user_id: args.userId,
+            interest_type: "Artist",
+            category: "collected_before",
+          }
+          return meCreateUserInterestLoader(gravityPayload)
```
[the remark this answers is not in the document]
MounirDhahri: you can address that on a separate PR in Gravity

src/schema/v2/me/collectArtistsMutation.ts
lordkiz: From session inside a mutation body?
How do you do that?

lordkiz:
@MounirDhahri @MrSltun All your concerns are valid and probably the best way to go.
To address these concerns:
- We first have to provision a service to handle batch creation of user interest on Gravity. The current existing user_interest endpoints do not support that and we shouldn't be making those changes on those.
- We have to decide whether we have to move the batch creation to a sidekiq job or do it synchronously (which will be blocking if we send large lists)
- If moved to a delayed job, then we have to adopt an optimistic approach, because we cannot guarantee that the job won't fail. This is fine, I suppose. However, immediately refreshing Me for updated `collectedArtist` may not return an up to date list of collected artist, hence the optimistic approach
- We have to provide an endpoint for this batch collect artist. The PR you referenced Mounir assumes we will use existing user_interest endpoints for collecting artists.

Based on these, I will close this PR and revisit after we can batch create.

PR #5072, opened 2023-06-08
chore: Add flag to include private hero units GRO-1755

jonallured:
This PR adds some much needed specs around picking which hero units loader to use. Prior to this PR it was not easy enough to understand when which of the three loaders would be used and I wanted to further complicate things so I started by extracting a function and wrapping the existing behavior with tests.

Then I added the complexity - if a flag called `private` is passed then use the authenticated loader and if not use the normal one.

So that's what this PR does - adds a flag to the hero units connection and then improves the resolver function by extracting the business logic to a separate function. That function is now wrapped with tests that help articulate the cases and should clarify when each of the loaders will be used.

This is all in service of changing Force to conditionally send `private` based on whether we want to only show the active hero units or if we want to preview the ones that haven't been marked as active yet. This functionality is something that will help our admins to verify they have their hero units configured correctly before pushing them to active.

Note: this PR can land first and then I will do separate PRs for Force and Forque to begin sending this flag. I've provided a default value so my understanding is that it is not required and will not break anything - do I have that right @dzucconi ?

https://artsyproduct.atlassian.net/browse/GRO-1755

/cc @artsy/grow-devs

### DIFF
--- a/src/schema/v2/HeroUnit/heroUnitsConnection.ts
+++ b/src/schema/v2/HeroUnit/heroUnitsConnection.ts
@@ -1,6 +1,6 @@
 import { connectionWithCursorInfo } from "../fields/pagination"
 import { HeroUnitType } from "./HeroUnitType"
-import { GraphQLString, GraphQLFieldConfig } from "graphql"
+import { GraphQLBoolean, GraphQLString, GraphQLFieldConfig } from "graphql"
 import { ResolverContext } from "types/graphql"
 import { pageable } from "relay-cursor-paging"
 import { convertConnectionArgsToGravityArgs } from "lib/helpers"
@@ -9,8 +9,13 @@ import { paginationResolver } from "schema/v2/fields/pagination"
 export const tokenRequiredMesage =
   "You need to pass a X-Access-Token header to perform this action"
 
+export const invalidArgsMessage =
+  "Invalid arguments - can't have both showAll and term"
+
 export const pickLoader = (args, context) => {
-  const { term } = args
+  const { showAll, term } = args
+
+  if (showAll && term) throw new Error(invalidArgsMessage)
 
   const {
     authenticatedHeroUnitsLoader,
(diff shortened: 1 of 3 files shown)


src/schema/v2/HeroUnit/heroUnitsConnection.ts
```diff
@@ -1,40 +1,59 @@
...
 })
 
 export const heroUnitsConnection: GraphQLFieldConfig<void, ResolverContext> = {
   type: HeroUnitConnection.connectionType,
   args: pageable({
+    showAll: {
+      type: GraphQLBoolean,
+      description: "If true will all hero units.",
+      defaultValue: false,
+    },
```
dzucconi, line 47: I think a better argument for this would be `private`. Like why, as a consumer of this field, would I not want to `showAll`?

jonallured: Ok happy to follow the precedent!

jonallured: Hrm - can you link me to a place where we do this?

```
[tsserver] Identifier expected. 'private' is a reserved word in strict mode. Modules are automatically in strict mode.
```

dzucconi: https://github.com/artsy/metaphysics/blob/cb99c685c09c1aa155419e75d548364d4629e480/src/schema/v2/me/savedArtworks.ts#L36

https://github.com/artsy/metaphysics/blob/cb99c685c09c1aa155419e75d548364d4629e480/src/schema/v2/partner/partnerArtworks.ts#L23

jonallured: ah ha so just never destructure it and instead call args.private - got it 👍

dzucconi (approved):

PR #5081, opened 2023-06-13
feat: Add includePartnersNearIpBasedLocation param to partnersConnection

olerichter00:
## Description

Add includePartnersNearIpBasedLocation param to partnersConnection.

This PR adds the parameter `includePartnersNearIpBasedLocation` to partnersConnection. When set to true, the IP address from context is used to do the geo lookup (introduced in https://github.com/artsy/metaphysics/pull/5055).

I removed the param `ip`. It's not used yet by any clients and this breaking change shouldn't break anything.

--- 

**Related PRs:**

- https://github.com/artsy/metaphysics/pull/5079

### DIFF
--- a/src/lib/locationHelpers.ts
+++ b/src/lib/locationHelpers.ts
@@ -15,7 +15,7 @@ export const getLocationArgs = async (
 ) => {
   let location = near
 
-  if (ip) {
+  if (!location && ip) {
     const {
       body: { data: locationData },
     } = await requestLocationLoader({ ip })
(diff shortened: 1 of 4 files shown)


src/lib/locationHelpers.ts
```diff
@@ -15,7 +15,7 @@ export const getLocationArgs = async (
 ) => {
   let location = near
 
-  if (ip) {
+  if (!location && ip) {
```
olerichter00, line 18: If `near` is passed and `location` already set,   we don't want to do an IP geo lookup.

MounirDhahri (approved):

PR #5087, opened 2023-06-14
chore: remove Algolia stitching and env #migration

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-4819

> **Note**
> https://github.com/artsy/gravity/pull/16457 depends on this

This removes the support for Algolia from Metaphysics.

Previously Eigen clients would query MP for an Algolia app id and a short-lived key in order to issues queries against Algolia indices. This functionality was [removed](https://github.com/artsy/eigen/pull/8161) from Eigen several versions ago.

### Migration

```sh
hokusai staging env unset ALGOLIA_RESTRICT_INDICES
hokusai staging env unset ALGOLIA_SEARCH_API_KEY

# These^ not set in production, so unneeded there
```


src/lib/stitching/gravity/v2/stitching.ts
```diff
@@ -173,11 +173,6 @@ export const gravityStitchingEnvironment = (
         curatedMarketingCollections(size: Int): [MarketingCollection]
       }
 
-      extend type System {
-        algolia: Algolia
-      }
```
anandaroop, line 178: I'm assuming we are fine removing this field outright (as opposed to marking it a deprecated)

damassi: [Wouldn't this](https://github.com/artsy/eigen/blob/a3871266daf9d7d836bc62fb8656195210e22ffb/src/app/Scenes/Search/Search.tsx#L247) break old versions of the app?

anandaroop: Because the Algolia contract ends this week, we can expect those API calls to start falling any moment now too. 

So search functionality in old versions of the app will _definitely_ be broken whether we do this or not, the only question is to what degree this will be graceful. 

I think we confirmed earlier that old versions would just show a blank screen and not an error once Algolia API access is cut off — this change maybe makes it more likely that there will be a hard error screen. 

But, at this point, 4 months out from the cutover, we're talking [essentially 0% of users (iOS)](https://artsy.looker.com/explore/frontend_analytics/eigen_screens?qid=7KaBqF2IikzeaTSuXIYjPa&origin_space=712&toggle=vis) on versions prior to 8.9.0. (I'm not sure how to find the equivalent dash for Android, I'll try to locate...)

I'd favor pruning the schema of (nearly) unused fields, esp if they can't actually return any data. But I'm open to argument on this.

dblandin: That plan sounds good to me @anandaroop. Keeping the schema compatible for a bit longer would retain our expected degraded state for users still on old versions, which is basically that the Top and Artworks pills would work normally but all other pills would be greyed out and inaccessible.

We haven't tested the old app experience with an incompatible schema. It's possible that that would cause a more intrusive error state.

gkartalis (commented):
Looking good! 

Just one thing, the https://github.com/artsy/eigen/pull/8161 PR is adding the feature flag and hiding the old implementation but we def need to clean the old implementation 🧹.

Happy to pair on this if you need anything!

src/lib/stitching/gravity/v2/stitching.ts
gkartalis: @dblandin true, what we tested back then was if algolia was failing silently when metaphysics either returned an empty string or weird invalid key for this.

Also happy to help test out any weird edge cases.

anandaroop:
> but we def need to clean the old implementation 🧹.
> 
> Happy to pair on this if you need anything!
 

Thanks @gkartalis I've got a branch for that as well — that'll come next & I'll reach out if needed.

anandaroop:
Added another commit https://github.com/artsy/metaphysics/pull/5087/commits/cc4457e0c703bce5ce0502314cb2838866c4a3ec that implements the strategy from the earlier comment https://github.com/artsy/metaphysics/pull/5087#discussion_r1231433030

##  Before

before

## After

after

- The API key is no longer generated (and wouldn't work anyway)
- Everything else should match up field for field
- All Algolia fields are marked as deprecated

I believe this should give us the schema compatibility we want in order to safely start pruning Gravity while still supporting older Eigen versions.

anandaroop:
✅  With help from @gkartalis I was able to confirm that the old Algolia-based search implementation + the changes from this PR result in the expected fallback behavior —


algolia

anandaroop:
I'm satisfied with ☝🏽  Going to go ahead and merge this, in order to unblock https://github.com/artsy/gravity/pull/16457

PR #5089, opened 2023-06-15
feat: Add maxDistance param to partnersConnection

olerichter00:
## Description

This PR adds the param `maxDistance` to `partnersConnection` to be able to pass it to Gravity. This is needed to not default to 75 when sorting by distance on the new "Galleries for You" screen in the app.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -13583,6 +13583,9 @@ type PartnerCategory {
     # If true, will only return partners that list artists that the user follows
     includePartnersWithFollowedArtists: Boolean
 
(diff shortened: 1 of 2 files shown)


MounirDhahri (approved):
I believe the default is set in Gravity since you are not adding a default value in MP. I don't have strong preferences here so let's keep it as it is. Only thing that comes to my mind is if it makes sense to add maximum distance for security but I am not sure if that's even a problem

olerichter00:
Yes, the default is set in Gravity. In other places, it's possible to adjust but here the type of the `near` param is a simple string, so we need to add the param separately.

PR #5114, opened 2023-07-05
fix: Pass partners connection arguments correctly

olerichter00:
[Jira Ticket](https://artsyproduct.atlassian.net/browse/CX-3809)

## Description

The `partnerConnection`'s `maxDistance` parameter is currently not passed on to Gravity if the `near` parameter is set. This pull request fixes the issue and adds some tests. 

It also fixed a bug where Boolean parameters aren't passed to Gravity if they are set to `false`.

### DIFF
--- a/src/lib/locationHelpers.ts
+++ b/src/lib/locationHelpers.ts
@@ -8,11 +8,17 @@ interface Location {
 
 const DEFAULT_MAX_DISTANCE_KM = 75
 
-export const getLocationArgs = async (
-  near: Location | undefined,
-  ip: string | undefined,
+export const getLocationArgs = async ({
+  ip,
+  maxDistance,
+  near,
+  requestLocationLoader,
+}: {
+  ip?: string
+  maxDistance?: number
+  near?: Location
   requestLocationLoader: any
-) => {
+}) => {
   let location = near
 
   if (!location && ip) {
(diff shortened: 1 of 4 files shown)


src/lib/locationHelpers.ts
```diff
@@ -32,6 +38,7 @@ export const getLocationArgs = async (
 
   return {
     near: isString(near) ? near : `${location.lat},${location.lng}`,
-    max_distance: location.maxDistance || DEFAULT_MAX_DISTANCE_KM,
+    max_distance:
+      maxDistance || location.maxDistance || DEFAULT_MAX_DISTANCE_KM,
```
olerichter00, line 42: Because `near` is just a simple string in some cases, we need to be able to pass `maxDistance` separately.

src/schema/v2/partner/partners.ts
```diff
@@ -245,7 +245,10 @@ export const PartnersConnection: GraphQLFieldConfig<void, ResolverContext> = {
     }
 
     // Removes null/undefined values from options
-    const cleanedOptions = pickBy(clone(options), identity)
+    const cleanedOptions = pickBy(
+      clone(options),
+      (option) => option !== null && option !== undefined
+    )
```
olerichter00, line 251: Replacing `identity` with `(option) => option !== null && option !== undefined` to not filter out options set to `false`.

MrSltun (approved):
Looks great! ✨ Thank you Ole!!

PR #5115, opened 2023-07-05
feat: add GAINING_FOLLOWERS artist insight

kiryl-zubarau:
This PR adds `GAINING_FOLLOWERS` artist insight.

JIRA - [DIA-26]

[DIA-26]: https://artsyproduct.atlassian.net/browse/DIA-26?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/schema/v2/artist/helpers.ts
```diff
@@ -138,6 +147,19 @@ const splitEntities = (value, delimiter = "|") => {
...
+    artist.follow_count - artist.last_year_follow_count === 0
+  )
+    return 0
+
+  return Math.floor(
+    (artist.last_year_follow_count * 100) /
+      (artist.follow_count - artist.last_year_follow_count)
+  )
+}
+
```
jonallured, line 162: **suggestion [blocking]:** export function and extract more local vars

I think this function should be exported so that it can be tested directly. It'll mean you can move some of the outer loop tests out of the tests for `getArtistInsights` and then can simplify the tests so that it's more clear what's happening here.

I took a stab at how you might want to extract some local vars here but I confess I did this all in the GitHub comment editor so it could be that I've made a typo or this isn't how Prettier would format - sorry about that!

```suggestion
export const computeFollowerGrowth = (artist) => {
  const thisYearCount = artist.follow_count
  const lastYearCount = artist.last_year_follow_count
  const followerGrowthAmount = thisYearCount - lastYearCount
  
  if (thisYearCount === 0 || followerGrowthAmount === 0)
    return 0

  const growthPercentage = Math.floor(
    (thisYearCount * 100) / followerGrowthAmount
  )
  
  return growthPercentage
}

```

So yeah what I'm doing here is trying to put some words around the concepts I see. I don't have a good sense for how much this code will change over time but I think it's worth starting out with a few more details in case we're back in here to update something.

src/schema/v2/artist/__tests__/helpers.test.ts
```diff
@@ -133,4 +133,58 @@ describe("getArtistInsights", () => {
       })
     })
   })
+
+  describe("gaining followers insight", () => {
```
jonallured, line 137: **suggestion:** reduce outer loop tests

If we extract the function that computes the follower growth percentage then I would reduce the number of tests here - maybe just for the happy path?

src/schema/v2/artist/__tests__/helpers.test.ts
```diff
@@ -133,4 +133,58 @@ describe("getArtistInsights", () => {
...
+      }`, () => {
+        const artist = field.artist
+
+        const insights = getArtistInsights(artist)
+        const insight = insights.find((insight) => insight.kind === field.kind)!
+
+        expect(!!insight).toEqual(field.returnsInsight)
+      })
+    })
+  })
```
jonallured, line 189: **suggestion:** Add tests for exported function

Now with fewer outer loop tests and an exported function to compute the growth percent, I'd add a separate `describe` block with more focused tests for that function.

src/schema/v2/artist/helpers.ts
```diff
@@ -92,6 +92,15 @@ export const ARTIST_INSIGHT_MAPPING: Record<
     getEntities: (artist) => artist.vanguard_year && [],
     getLabel: (artist) => `The Artsy Vanguard ${artist.vanguard_year}`,
   },
+  GAINING_FOLLOWERS: {
+    getDescription: (artist) =>
+      `${artistNewFollowersPercentage(
+        artist
+      )}% increase in Artsy followers compared to same time last year.`,
+    getEntities: (artist) =>
+      artistNewFollowersPercentage(artist) >= 20 ? [] : null,
```
jonallured, line 101: **question [non-blocking]:** 

I'm confused about this line - why would we want to sometimes return an empty array and other times return null? And what does the number 20 signify? 🤔 

jonallured (commented):
Hey @kiryl-zubarau, I read through this and noticed an opportunity to refactor a bit and export the function that computes the growth percentage. I think it's better to have that function more directly tested and for a few more local variables to help Future Us read the code. What do you think?

src/schema/v2/artist/helpers.ts
araujobarret: :ceiling-cat: I don't wanna block any work, but shouldn't all this data come from Gravity? I know sometimes we manipulate somehow some data in MP, but I think it makes sense to make it totally dumb, isn't it?

araujobarret: It could be just a virtual field there

kiryl-zubarau: Right, initially I was thinking it will be useful to calculate the value in MP but look like it isn't necessary and can be moved to Gravity.

src/schema/v2/artist/helpers.ts
kiryl-zubarau: Moved this function to the Gravity.

egdbear (approved):
LGTM

PR #5119, opened 2023-07-06
feat: add recent show to insights

egdbear:
Wire recent_show data point
https://artsyproduct.atlassian.net/browse/DIA-23


src/schema/v2/artist/helpers.ts
```diff
@@ -190,3 +195,21 @@ export const getAuctionRecord = async (artist, auctionLotsLoader) => {
 
   return highAuctionRecord
 }
+
+const getRecentShow = (artist): string[] => {
+  // dd/mm/yyy|slug|Group or Solo|Show Title
+  const entities = splitEntities(artist.recent_show)
```
dzucconi, line 201: Can you leave an example artist that actually has this field as non-null? I can't find one, including artists that have current shows in prod.

We're sure this actually conforms to that weird string format? IMO this should have a runtime-codec or something to verify and cast the type.

dzucconi: (Ah, OK this is coming from the `artist_insights_importer`)

dzucconi (approved):

src/schema/v2/artist/helpers.ts
```diff
@@ -190,3 +195,21 @@ export const getAuctionRecord = async (artist, auctionLotsLoader) => {
 
   return highAuctionRecord
 }
+
+const getRecentShow = (artist): string[] => {
```
jonallured, line 199: **suggestion [blocking]:** export function and test directly

Same feedback from me on this - I think these helper functions can be exported and tested directly with some happy path tests on the outer loop.

src/schema/v2/artist/helpers.ts
```diff
@@ -190,3 +195,21 @@ export const getAuctionRecord = async (artist, auctionLotsLoader) => {
...
+  const show = entities[entities.length - 1]
+
+  const showDate = new Date(date)
+  const today = new Date()
+  const threeYearsAgo = new Date()
+
+  threeYearsAgo.setFullYear(today.getFullYear() - 3)
+
+  if (showDate < threeYearsAgo) return []
+  return [show]
```
kiryl-zubarau, line 214: Empty array and array with show both are valid result for [getEntities](https://github.com/artsy/metaphysics/blob/510f1be086baff907ec2b865ddb483ce28ad9ad9/src/schema/v2/artist/helpers.ts#L150-L151), this badge will be shown in both cases. I used `null` [value instead of empty array](https://github.com/artsy/metaphysics/pull/5115#discussion_r1254587168).

kiryl-zubarau (dismissed):
LGTM!

src/schema/v2/artist/helpers.ts
```diff
@@ -97,6 +97,11 @@ export const ARTIST_INSIGHT_MAPPING: Record<
     getEntities: (artist) => artist.critically_acclaimed && [],
     getLabel: () => "Critically Acclaimed",
   },
+  RECENT_CAREER_EVENT: {
+    getDescription: (artist) => artist.recent_show && getRecentShow(artist),
+    getEntities: (artist) => artist.recent_show && [],
```
araujobarret, line 102: is it intentional to return `null` or `[]`?

egdbear: yes

araujobarret (dismissed):

araujobarret:
Don't forget to change the title before merging. Suggestion `feat(DIA-23): add recent show to insights`.

PR #5130, opened 2023-07-10
chore: removing some private mentions from artwork

oxaudo:
This is the references I could find to Artwork `private` field here.

A bit confusing one is the one on `collectionsConnection`. Those are requesting artworks but I believe `private` in those requests refers to `private` collection vs `private` artwork...


olerichter00:
Looks good 👍

This is related to getting rid of artwork's `private` field, right?

PR #5138, opened 2023-07-11
fix(artist): adjusts insights copy

dzucconi:


src/schema/v2/artist/helpers.ts
```diff
@@ -29,8 +29,10 @@ type ArtistInsightKind = typeof ARTIST_INSIGHT_KINDS[number]
 export const ARTIST_INSIGHT_MAPPING: Record<
   ArtistInsightKind,
   {
+    // FIXME: There should be a function that determines visibility.
+    // Currently it uses `getEntities` but not empty array but an existential check.
```
dzucconi, line 33: This wound up being pretty confusing! Types were incorrect because an `any` got through and then worked on that basis. 😞

PR #5142, opened 2023-07-12
chore: better formatted logs in MP

MounirDhahri:
While checking the logs printed in MP, I noticed that the logs printed are pretty hard to read and spent some time to see if I can make them a bit prettier.

**Before**
Screenshot 2023-07-12 at 16 39 20

___
**After**
Screenshot 2023-07-12 at 16 40 54

### DIFF
--- a/src/lib/timer.ts
+++ b/src/lib/timer.ts
@@ -7,7 +7,7 @@ export default (key) => {
   return {
     start: () => {
       time = process.hrtime()
-      info(`Loading: ${key}`)
+      info(`Loading: ${decodeURIComponent(key)}`)
       return time
     },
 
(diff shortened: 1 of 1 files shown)


olerichter00 (approved):
Nice improvement 🌟

mzikherman:
Yea, nice! We can format it even more if needed but this is an improvement.

Screen Shot 2023-07-12 at 6 10 48 PM

Also as an aside I noticed in the screenshot, it is clear we are sending some extraneous parameters to Gravity (which it is ignoring). `excludePurchasedArtworks=false...sortByLastAuctionResultDate=false` 😆

PR #5157, opened 2023-07-19
feat: Add artistIDs param to myCollectionConnection

olerichter00:
Addresses [ONYX-140]

- Related Gravity PR: https://github.com/artsy/gravity/pull/16558

## Description

This PR adds an `artistIDs` param to the connection `me.myCollectionConnection` to filter artworks by artist ID.

[ONYX-140]: https://artsyproduct.atlassian.net/browse/ONYX-140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -9,9 +9,11 @@ import {
   isEmpty,
   isObject,
   isString,
+  mapKeys,
   omit,
   pick,
   reject,
+  snakeCase,
   trim,
   values,
 } from "lodash"
@@ -218,7 +220,7 @@ export const defineCustomLocale = (
 /**
  * Extracts nodes from a GraphQL connection.
  */
-export const extractNodes = <Node extends object, T = Node>(
+export const extractNodes = <Node extends Record<string, unknown>, T = Node>(
   connection:
     | {
         readonly edges?: ReadonlyArray<{
@@ -227,11 +229,11 @@ export const extractNodes = <Node extends object, T = Node>(
       }
     | undefined
     | null,
-  mapper?: (node: Node) => T
+  mapper?: (node?: Node | null) => T
 ): T[] => {
   return (
     connection?.edges
-      ?.map((edge) => (mapper ? (mapper(edge?.node!) as any) : edge?.node!))
+      ?.map((edge) => (mapper ? (mapper(edge?.node) as any) : edge?.node))
       .filter((x) => x != null) ?? []
   )
 }
@@ -248,4 +250,22 @@ export const CatchCollectionNotFoundException = (error) => {
   if (error.statusCode === 404) return emptyConnection
 
   throw error
+}
+
+/**
+ * Converts all object keys to snake case.
+ * @param object — The object to convert.
+ * @return — Returns the object with converted keys.
+ */
+export const snakeCaseKeys = (
+  obj: Record<string, any>
+): Record<string, any> => {
+  return mapKeys(obj, (_, key) => {
+    // Special case for ID to not be converted to I_D
+    if (key.includes("ID")) {
+      return snakeCase(key.replace(/ID/g, "_id"))
+    }
+
+    return snakeCase(key)
+  })
 }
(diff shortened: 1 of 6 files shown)


src/lib/helpers.ts
```diff
@@ -249,3 +251,21 @@ export const CatchCollectionNotFoundException = (error) => {
 
   throw error
 }
+
+/**
+ * Converts all object keys to snake case.
+ * @param object — The object to convert.
+ * @return — Returns the object with converted keys.
+ */
+export const snakeCaseKeys = (
```
olerichter00, line 260: I've added this helper to convert keys to snake case to not have to add them manually and avoid sending the unconverted params when using `...convertConnectionArgsToGravityArgs(options)` ([example](https://github.com/artsy/metaphysics/pull/5157/files#diff-d51c53fbca4329082a76fe9d905e961bd040f965f24579f19217ae5806cf4abbR100)).

src/schema/v2/me/myCollection.ts
```diff
@@ -58,6 +60,11 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
         },
       }),
     },
+    artistIDs: {
+      type: new GraphQLList(GraphQLString),
```
MounirDhahri, line 64: **minor:**
```typescript
      type: new GraphQLNonNull(new GraphQLList(GraphQLString)),
```

src/schema/v2/me/myCollection.ts
```diff
@@ -90,14 +97,13 @@ export const MyCollection: GraphQLFieldConfig<any, ResolverContext> = {
...
-      ? { size: MAX_COLLECTION_SIZE }
+    const optionsAndPaginationArgs = options.sortByLastAuctionResultDate
+      ? { size: MAX_COLLECTION_SIZE, ...options }
       : convertConnectionArgsToGravityArgs(options)
 
     const gravityOptions = {
-      exclude_purchased_artworks: options.excludePurchasedArtworks,
       total_count: true,
-      ...paginationArgs,
+      ...snakeCaseKeys(optionsAndPaginationArgs),
```
MounirDhahri, line 106: **Q:** shouldn't we still send `artist_ids` if it's undefined since we're only looking for its presence [here](https://github.com/artsy/gravity/pull/16558/files#r1268100822). Or are we stripping falsy values somewhere?

src/lib/helpers.ts
```diff
@@ -249,3 +251,21 @@ export const CatchCollectionNotFoundException = (error) => {
...
+  obj: Record<string, any>
+): Record<string, any> => {
+  return mapKeys(obj, (_, key) => {
+    // Special case for ID to not be converted to I_D
+    if (key.includes("ID")) {
```
MounirDhahri, line 265: good that we have no "ID" field on it's own.

MounirDhahri (approved):
LGTM, I left only one small question and it might be an obvious thing that we do in Gravity

src/schema/v2/me/myCollection.ts
olerichter00: If `artist_ids` isn't passed in the GraphQL query, we don't send it to Gravity either. I guess this is the expected behavior. The presence check will fail and the artworks won't be filtered by artistIDs.

MounirDhahri: my bad, i was multi tasking. That's what I meant, it's being sent now even if it's `undefined` right?

src/schema/v2/me/myCollection.ts
dzucconi: I think what you're looking for is `new GraphQLList(new GraphQLNonNull(GraphQLString))`? Meaning: this argument isn't required but if it is you want a real array of strings not an array of strings + nulls?

src/schema/v2/me/myCollection.ts
olerichter00: I tested it without `artistIds` and the param wasn't sent as `undefined`  to Gravity.

src/lib/helpers.ts
olerichter00: Which would then be converted to `id` by this function.

PR #5165, opened 2023-07-21
feat(ONYX-160): add updateUserInterest mutation

MounirDhahri:
Related ticket: [ONYX-160]

This PR adds support for updating user interests.

Screenshot 2023-07-24 at 12 20 13

[ONYX-160]: https://artsyproduct.atlassian.net/browse/ONYX-160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/schema/v2/me/updateUserInterestMutation.ts
```diff
@@ -0,0 +1,97 @@
...
+    { meUpdateUserInterestLoader }
+  ) => {
+    if (!meUpdateUserInterestLoader) {
+      throw new Error("You need to be signed in to perform this action")
+    }
+
+    try {
+      const userInterest: UserInterest = await meUpdateUserInterestLoader?.(
+        id,
+        { private: !!isPrivate }
```
nickskalkin, line 79: **question**: Let's imaging that in future this mutation will support other input fields and `private` field is not sent as an input argument (something like this):

```graphql
mutation {
  updateUserInterest(
      input: { id: "id", someOtherField: "newValue" }
   ) {
    ...
  }
}
```

If I understand this code correctly, then `private` will be sent to gravity in any case as `false`, am I right? It might be not very transparent for developers why `private` suddenly becomes `false`. I think that we should either handle this in the code (if field is not passed - do not send it to gravity) or at least make this field required)

@MounirDhahri

src/schema/v2/me/updateUserInterestMutation.ts
```diff
@@ -0,0 +1,97 @@
...
+      )
+
+      return userInterest
+    } catch (err) {
+      if ("body" in (err as any)) {
```
nickskalkin, line 84: **suggestion**: I would suggest handling errors [this way](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/me/updateCollectionMutation.ts#L84), it simply looks cleaner to me (and it was tested by me recently, so I have more confidence that this works)

src/schema/v2/me/updateUserInterestMutation.ts
```diff
@@ -0,0 +1,97 @@
...
+  private?: boolean
+}
+
+const SuccessType = new GraphQLObjectType<any, ResolverContext>({
+  name: "UpdateUserInterestSuccess",
+  isTypeOf: (data) => data.id,
+  fields: () => ({
+    userInterest: {
+      type: userInterestType,
+      resolve: async (response) => {
```
nickskalkin, line 24: **question**: is there a need for async?

nickskalkin (changes requested):
Looks cool, I left a few small questions

src/schema/v2/me/updateUserInterestMutation.ts
MounirDhahri: good catch 👍

MounirDhahri: will remove the `!!`

src/schema/v2/me/updateUserInterestMutation.ts
MounirDhahri: cool, I will update this. i copied what we do in some other files

src/schema/v2/me/updateUserInterestMutation.ts
MounirDhahri: nope, fixed

nickskalkin (approved):
👍

PR #5168, opened 2023-07-24
feat(AMBER-28): create messages and conversation events query

mdole:
Companion to https://github.com/artsy/impulse/pull/956.

This PR introduces a new `messagesAndConversationEventsConnection` field inside a `conversation`, which fetches messages and events (like an offer being submitted or rejected) together in a single chronologically-ordered connection.

Sample query (note that to run this, you'll need to change the ID to be a conversation you're part of):
```graphql
query {
	me {
		conversation(id: "208433") {
			messagesAndConversationEventsConnection(first: 50) {
				edges {
					node {
						__typename
						... on Message {
							internalID
							isFromUser
							body
						}
						... on ConversationEvent {
							buyerBody
							sellerBody
							eventKey
						}
					}
				}
			}
		}
	}
}
```

Jira ticket: https://artsyproduct.atlassian.net/browse/AMBER-28

### DIFF
--- a/src/schema/v2/conversation/index.ts
+++ b/src/schema/v2/conversation/index.ts
@@ -30,6 +30,10 @@ import { UserType } from "../user"
 import { InquiryRequestType } from "../partner/partnerInquiryRequest"
 import { CollectorProfileType } from "../CollectorProfile/collectorProfile"
 import { ConversationEventType } from "./conversationEvent"
+import {
+  connectionWithCursorInfo,
+  paginationResolver,
+} from "../fields/pagination"
 
 export const BuyerOutcomeTypes = new GraphQLEnumType({
   name: "BuyerOutcomeTypes",
@@ -179,19 +183,6 @@ const MessageOrConversationEventType = new GraphQLUnionType({
       default:
         return null
     }
-  },
-})
-
-export const {
-  connectionType: MessagesAndConversationEventsConnection,
-  edgeType: MessageOrConversationEventEdge,
-} = connectionDefinitions({
-  nodeType: MessageOrConversationEventType,
-  connectionFields: {
-    totalCount: {
-      resolve: ({ total_count }) => total_count,
-      type: GraphQLInt,
-    },
   },
 })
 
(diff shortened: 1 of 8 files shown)


damassi:
@mdole - Sorry for always dinging you on PRs 😅. Mind updating the PR description with an actual description? It's best to avoid external links there as those aren't searchable, and don't provide any context for folks reviewing.

src/schema/v2/conversation/index.ts
```diff
@@ -479,6 +508,61 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
...
+                    conversation_from_address: from_email,
+                    conversation_id: id,
+                  })
+                } else {
+                  return messageOrEvent
+                }
+              }
+            )
+            return {
+              ...connectionFromArraySlice(
```
damassi, line 552: Is there any chance we can use our [`paginationResolver`](https://github.com/artsy/metaphysics/blob/main/docs/pagination.md) helper here?

mdole:
Nothing to apologize for @damassi, that's a fair point! Updated :)

src/schema/v2/conversation/index.ts
mdole: ooo we totally can! did not know that was a thing. definitely cleans up the code a lot. done in https://github.com/artsy/metaphysics/pull/5168/commits/c22c7b724280a135c90ae2ac2ba639d2c7539a50

src/schema/v2/partner/partner.ts
```diff
@@ -646,7 +646,7 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
           )
 
           const gravityArgs = {
-            published: true,
+            private: false,
```
jpotts244, line 649: Is this an intentional change here?

jpotts244 (dismissed):
Looks great - just one Q, will merge in a minute!

src/schema/v2/partner/partner.ts
mdole: oh yeah I meant to leave a comment about that, it was a drive-by - `published` is [not actually a parameter on this Gravity endpoint](https://github.com/artsy/gravity/blob/764e9360a762896c0203477b212f783bab09f5aa/app/api/v1/partners_locations_endpoint.rb#L70), but `private` is.

But looking at this again, I should add a spec for it (and weirdly I think this should be giving a type error due to [this declaration](https://github.com/artsy/metaphysics/blob/3f4b5272944121443f3cc1559acd72ef6faa1895/src/lib/loaders/loaders_without_authentication/gravity.ts#L156), but it doesn't seem to be).

I'll update quick so this is an actual change and not half-assed!

jpotts244: Nice catch on the parameters

jpotts244 (approved):
🚀 ! thanks for this!

PR #5169, opened 2023-07-24
feat: add batch update user interests mutation

MounirDhahri:
This PR resolves [ONYX-152]

The goal here is to add support for updating multiple user interests at once.


olerichter00 (approved):

MounirDhahri:
Let's gooo

PR #5175, opened 2023-07-27
feat(onyx-170): profile search entities return exect href instead of vanity urls

nickskalkin:
# Motivation

`SearchableItem` has a `href` field. In current implementation, for `Profile` type this field always returns `/<profile_id>`, which is a vanity URL. When user is redirected to a such URL, Metaphysics redirects user to a corresponding page based on profile's `owner_type` field. For instance, if `owner_type` is `Fair` - user is redirected to `/fair/<fair_slug>`, if `owner_type` is `PartnerGallery` - user is redirected to `/partner/<partner_slug>`.

## Example

- We search for British museum

```graphql
{
	searchConnection(first: 20, mode: SITE, query: "british museum", aggregations: [TYPE], entities: [PROFILE]) {
			edges {
				node {
					href
				}
			}
  }
```

- In response `href` is "/britishmuseum", which is a `Profile` slug.
- User is redirected to this page and vanity URL resolution redirects user to `/partner/british-museum` page. (MP makes request `partnerLoader` request to correctly resolve partner's slug, note that profile's and partner's slugs are different).

Vanity URL resolution is intentional and we want to keep it, it is useful to have a "short" url's like `artsy.net/britishmuseum`.

But there is no reason why we should make an extra "hop" when searching from app or web search input - it's more performant to redirect a user to a correct link right away.

# Implementation notes

In this pull request I wanted to establish a direct mapping between profiles of different types and URLs corresponding to them. I pulled all profile types from Gravity console like that:

```ruby
> Profile.distinct(:owner_type)
=>
["Fair",
 "FairOrganizer",
 "PartnerAuction",
 "PartnerBrand",
 "PartnerDemo",
 "PartnerGallery",
 "PartnerInstitution",
 "PartnerInstitutionalSeller",
 "PartnerPrivateCollector",
 "PartnerPrivateDealer"]
```

I also referred to existing vanity URL [resolution logic](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/vanityURLEntity.ts#L36). It seems like that all types of profiles redirect to `/partner/<partner_id>` pages (where `<partner_id>` is taken from `profile.owner.id`) with some exceptions:

- `Fair` redirects to a `/fair/<fair_slug>` URL
- `FairOrganizer` also redirects to a `/fair/<fair_slug>` URL, but `<fair_slug>` now is a `default_fair_id` taken from `FairOrganizer`

I've added a new `owner_slug` field. It doesn't exist in Gravity yet, but I've drafted a [PR](https://github.com/artsy/gravity/pull/16592) to add it. We can avoid introducing this field, but then we will have to make an extra request to a `profileLoader` to get `owner` data (like VanityURLEntity does [here](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/vanityURLEntity.ts#L36)). I've added this field to improve performance, but we can discuss it.

### DIFF
--- a/src/schema/v2/SearchableItem/SearchItemRawResponse.ts
+++ b/src/schema/v2/SearchableItem/SearchItemRawResponse.ts
@@ -12,6 +12,7 @@ export type SearchItemRawResponse = {
   location: string
   model: string
   owner_type: string
+  owner_slug: string
   profile_id: string
   published_at: string
   start_at: string
--- a/src/schema/v2/SearchableItem/SearchableItemPresenter.ts
+++ b/src/schema/v2/SearchableItem/SearchableItemPresenter.ts
@@ -42,12 +42,36 @@ export class SearchableItemPresenter {
   }
 
   href(): string {
-    const { href, label, id, profile_id, model } = this.item
+    const {
+      href,
+      label,
+      id,
+      profile_id,
+      model,
+      owner_type,
+      owner_slug,
+    } = this.item
 
     if (href) return href
+
     switch (label) {
       case "Profile":
-        return `/${id}`
+        switch (owner_type) {
+          case "Fair":
+          case "FairOrganizer":
+            return `/fair/${owner_slug}`
+          case "PartnerAuction":
+          case "PartnerBrand":
+          case "PartnerDemo":
+          case "PartnerGallery":
+          case "PartnerInstitution":
+          case "PartnerInstitutionalSeller":
+          case "PartnerPrivateCollector":
+          case "PartnerPrivateDealer":
+            return `/partner/${owner_slug}`
+          default:
+            return `/${id}`
+        }
       case "Fair":
         return `/${profile_id}`
       case "Sale":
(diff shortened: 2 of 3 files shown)


joeyAghion:
Thanks for digging that up... I think _yes_ deferring that responsibility to Gravity makes total sense, especially if there's already the `href` attribute in place to communicate that.

My only question: does any of this require coordination with Eigen? I know it has its own navigation code based on these paths.

nickskalkin:
> My only question: does any of this require coordination with Eigen? I know it has its own navigation code based on these paths.

Good point. I think we should have coordinate with Eigen in case there were things like `navigate("partner/${partner.href}"`, but I cannot find anything like that. Still looking

anandaroop:
Once these `searchConnection` responses change because of the upstream change in https://github.com/artsy/gravity/pull/16592,  Eigen should already be able to do the correct thing with [partner/:id](https://github.com/artsy/eigen/blob/41c2fa73df4b28d99ac25872d9c2e8cc841d4ffd/src/app/routes.ts#L230) and [fair/:id](https://github.com/artsy/eigen/blob/41c2fa73df4b28d99ac25872d9c2e8cc841d4ffd/src/app/routes.ts#L178) urls

PR #5179, opened 2023-07-28
feat(EMI-1286) Handle gravity errors in metaphysics verifyAddress query resolver

fladson:
The current `verifyAddress` error handling approach is not in pair with what we usually have for other queries and mutations, which is preventing us to handle errors correctly client side.

This updates the resolver to be more in pair with the pattern we use across Gravity queries, i.e., using the `...OrError` and `Success` or `Failure` types.

**After**

_successful response_
SCR-20230731-g72

_failed response_
SCR-20230731-g79

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -106,15 +106,6 @@ type addOrderedSetItemSuccess {
   setItem: OrderedSetItem
 }
 
-input AddressInput {
-  addressLine1: String!
-  addressLine2: String
-  city: String
-  country: String!
-  postalCode: String!
-  region: String
-}
-
 type addUserRoleFailure {
   mutationError: GravityMutationError
 }
@@ -15749,7 +15740,7 @@ type Query {
   ): VanityURLEntityType
 
   # Verify a given address.
-  verifyAddress(address: AddressInput!): VerifyAddressType
+  verifyAddress(input: VerifyAddressInput!): VerifyAddressPayload
 
   # A wildcard used to support complex root queries in Relay
   viewer: Viewer
@@ -19284,6 +19275,27 @@ enum VerificationStatuses {
   VERIFIED_WITH_CHANGES
 }
 
+type VerifyAddressFailureType {
+  mutationError: GravityMutationError
+}
+
+input VerifyAddressInput {
+  addressLine1: String!
+  addressLine2: String
+  city: String
+  clientMutationId: String
+  country: String!
+  postalCode: String!
+  region: String
+}
+
+union VerifyAddressMutationType = VerifyAddressFailureType | VerifyAddressType
+
+type VerifyAddressPayload {
+  clientMutationId: String
+  verifyAddressOrError: VerifyAddressMutationType
+}
+
 type VerifyAddressType {
   inputAddress: InputAddressFields!
   suggestedAddresses: [SuggestedAddressFields]!
@@ -20203,7 +20215,7 @@ type Viewer {
   ): VanityURLEntityType
 
   # Verify a given address.
-  verifyAddress(address: AddressInput!): VerifyAddressType
+  verifyAddress(input: VerifyAddressInput!): VerifyAddressPayload
   viewingRoomsConnection(
     after: String
     first: Int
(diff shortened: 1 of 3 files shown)


MrSltun (approved):
LGTM!! We just need to update the tests for the mutation
Also, I'm not sure if it's a good idea to remove types from the schema, as it may cause issues with force (the warning from peril)

erikdstock:
non-blocking: I'd like to understand exchange's philosophy on handling errors better since I can't remember seeing it in any other application. Some [incomplete] context:
- What I think of as the default/convention of top-level `data` and `errors` keys is allowed, even with partial data in the presence of one or more errors, and [graphQL ruby handles this automatically](https://graphql-ruby.org/errors/overview.html)
- There is also precedent for returning errors as a rich type the way we do. @alloy [wrote about it for artsy](https://artsy.github.io/blog/2018/10/19/where-art-thou-my-error/#Make.exceptions.first-class.citizens.of.your.schema) around the time we were developing exchange.
- Another convention is to [return an `errors` field inside your `data`](https://graphql-ruby.org/mutations/mutation_errors#errors-as-data). It looks like we briefly tried this in (https://github.com/artsy/exchange/pull/58 but then switched to a union type inside `orderOrErrors` in artsy/exchange#93 which makes sense because it forces you to handle both types in your response.
- In [artsy/exchange#184](https://github.com/artsy/exchange/pull/184/files#diff-30736993def73ca73dfd3eb2d4aae0d44a8b69adf1fedb73268c5cd992f3db87) we added lots of error codes.
- [error_handler.rb](https://github.com/artsy/exchange/blob/main/app/controllers/concerns/error_handler.rb) to this day still formats some things as root-level errors. It handles `ValidationError` which we raise frequently in the code, which suggests to me that these errors would appear on the top level.
- for order mutations the [possible `OrError`s are very limited](https://github.com/artsy/exchange/blob/eb7363920941750a0ad97e46fd6d8d78c89bbee7/app/graphql/mutations/submit_order.rb#L20), presumably falling back to the top-level errors otherwise.

My understanding form all this is that
- some operations fail for good reasons and we want to return a rich result to the client (`data: _OrError`, payment failed), - some fail for bad reasons and belong in `errors: [{message: '...'}]`. 
- If we are basically getting a bad result from gravity, I _think_ we would want that that to be added to top-level `errors` because we want the client to be resilient to the field being broken - we don't want them to update their smarty credentials or fix a bug in gravity. Partial `data` is still acceptable in the presence of top-level errors, and in this case we could fall back to `UNAVAILABLE`.

However before pressing submit on this commend @fladson pointed me to [this slack thread](https://artsy.slack.com/archives/C02JHHHKP5K/p1690476901140099) (cc @damassi) that suggests we always want to use the nested union error type. All of which is to say i don't know the right approach and this is not a blocking comment.

damassi:
The approach is to use the `onError` pattern, which we can control in more subtle ways if need be. This PR looks good 👍

erikdstock:
Rebased off main and force-pushed back.

iskounen:
> Also, I'm not sure if it's a good idea to remove types from the schema, as it may cause issues with force (the warning from peril)

@MrSltun I believe you were referring to the `AddressInput` type in your previous comment. I think it's safe to remove that type from the schema because it is only used by the unreleased address verification flow.

Regarding the warning, the Force schema is also being updated to reflect this change in the companion PR for this work: artsy/force#12723

_schemaV2.graphql
```diff
@@ -19284,6 +19275,27 @@ enum VerificationStatuses {
   VERIFIED_WITH_CHANGES
 }
 
+type VerifyAddressFailureType {
+  mutationError: GravityMutationError
+}
```
erikdstock, line 19280: This has me wondering - we made this a Query field early on after some discussion back and forth whether it was truly a query or a mutation, but now plan to back these queries up with a model in gravity. While we are changing the schema anyways, should we make this a mutation?

iskounen: I think it would be more straightforward as a mutation since we are performing an operation. A query implies that we are querying a type in our data model, which we aren't doing.

I also think that the `AddressVerificationFlow` component in Force will be greatly simplified if we were requesting a mutation and handling its response.

Having said that, I don't think there is a benefit to making that change today.

iskounen (approved):

_schemaV2.graphql
starsirius: Yeah I agree it makes more sense now to use a mutation for this operation, and agree we can do it in a followup, given it also requires some UI updates.

PR #3864, opened 2022-03-15
feat: add previewSavedSearch query field

dblandin:
Allows a client to preview the formatted labels for an unpersisted alert criteria.

Jira: [FX-3819](https://artsyproduct.atlassian.net/browse/FX-3819)

**Request**

```graphql
{
  previewSavedSearch(attributes: { attributionClass: ["unique"], artistIDs: ["4e934002e340fa0001005336"], acquireable: true }) {
    labels {
      field
      name
      value
    }
  }
}
```

**Response**

```json
{
  "data": {
    "previewSavedSearch": {
      "labels": [
        {
          "field": "artistIDs",
          "name": "Artist",
          "value": "KAWS"
        },
        {
          "field": "attributionClass",
          "name": "Rarity",
          "value": "Unique"
        },
        {
          "field": "acquireable",
          "name": "Ways to Buy",
          "value": "Buy Now"
        }
      ]
    }
  }
}
```

PR #5186, opened 2023-08-01
feat(alerts): add a displayName field for saved search criteria

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-173

This nests a new `displayName` field under the root-level `previewSavedSearch` field.

```graphql
query {
  previewSavedSearch(
    attributes: {
      artistIDs: ["kaws"]
      priceRange: "100-1000"
      additionalGeneIDs: "prints"
      attributionClass: "limited edition"
    }
  ) {
    
    displayName # 👈🏽👈🏽👈🏽

    labels {
      displayValue
    }
  }
}
```

```json
{
  "data": {
    "previewSavedSearch": {

      "displayName": "KAWS — $100–$1,000, Print, Limited edition",  👈🏽👈🏽👈🏽

      "labels": [
        {
          "displayValue": "KAWS"
        },
        {
          "displayValue": "Limited edition"
        },
        {
          "displayValue": "Print"
        },
        {
          "displayValue": "$100–$1,000"
        }
      ]
    }
  }
}
```

- `previewSavedSearch` was added in https://github.com/artsy/metaphysics/pull/3864 so that we could centralize some logic for displaying an unpersisted saved search in human-friendly form

- It [does not appear](https://github.com/search?q=org%3Aartsy%20previewSavedSearch&type=code) to have been wired up yet, though

- Since it is an unauthenticated root-level field that takes search criteria as its input, we saw this as a good place to nest this new `displayName` field

- It follows the logic outlined in the ticket to display these values, in this order, when present: **artist, price, medium, rarity**

---

I wanted to get a sense of what kind of suggested output we would get from real-world alert criteria, so I took the 500 newest SearchCriteria instances in staging Gravity and ran them through this logic (code in this commit https://github.com/artsy/metaphysics/pull/5186/commits/1f2159fdeff6df7105ae9a9cda011c0e5cd10c2a)

<details>
<summary>Previous real world output (outdated now)</summary>
👉🏽 https://docs.google.com/spreadsheets/d/1dLPcKBUGsm1MVmpjeGcHWEy4zmg5700ZmWxM_76ZdiM/edit?usp=sharing

Screenshot 2023-08-01 at 5 25 43 PM

<br/><br/>

You can see from this^ that the output can range from the simple…

```
John Kiki
```
…to the ideal…
```
Tom Blachford, $0–$7,000, Photography, Limited edition
```
…to the very verbose…
```
Pablo Picasso, $250–$1,000, Design/Decorative Art or Drawing or Ephemera or Merchandise or Jewelry or Painting or Photography or Print or Sculpture or Drawing, Collage or other Work on Paper
```

Since these are suggested/default names, this seems fine for a first iteration. And the vast majority of these come out to a reasonable length.
</details>

### DIFF
--- a/src/schema/v2/previewSavedSearch.ts
+++ b/src/schema/v2/previewSavedSearch.ts
@@ -70,6 +70,12 @@ const previewSavedSearchArgs: GraphQLFieldConfigArgumentMap = {
 const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
   name: "PreviewSavedSearch",
   fields: () => ({
+    displayName: {
+      type: new GraphQLNonNull(GraphQLString),
+      resolve: generateDisplayName,
+      description:
+        "A suggestion for a name that describes a set of saved search criteria in a conventional format",
+    },
     labels: {
       type: new GraphQLNonNull(new GraphQLList(SearchCriteriaLabel)),
       resolve: resolveSearchCriteriaLabels,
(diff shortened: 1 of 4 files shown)


src/schema/v2/previewSavedSearch.ts
```diff
@@ -99,3 +109,46 @@ export const PreviewSavedSearchField: GraphQLFieldConfig<
...
+  const suggestedNameElements: string[] = []
+
+  const { artistIDs, priceRange, additionalGeneIDs, attributionClass } = parent
+
+  // artist
+  const artistLabels = (await getArtistLabels(artistIDs, artistLoader)).map(
+    (l) => l.displayValue
+  )
+  if (artistLabels.length) {
+    suggestedNameElements.push(artistLabels.join(" or "))
```
anandaroop, line 128: Not explicitly mentioned in the ticket but we need to handle array-valued inputs to most of these fields. When multiple values are supplied these act as `OR`'d terms, so the naming reflects this.

src/schema/v2/searchCriteriaLabel.ts
```diff
@@ -117,7 +117,7 @@ export const resolveSearchCriteriaLabels = async (
   return labels.flat().filter((x) => x !== undefined) as SearchCriteriaLabel[]
 }
 
-async function getArtistLabels(artistIDs: string[], artistLoader) {
+export async function getArtistLabels(artistIDs: string[], artistLoader) {
```
anandaroop, line 120: Move of the logic we wanted was already defined previously, so I just export these functions for re-use. Maybe some larger code re-org is warranted by this as  well 🤷🏽 ?

anandaroop:
cc @olerichter00 @MounirDhahri (FYI, per our sprint planning notes)

dblandin (commented):
This is looking great!!

I think the implementation here slightly differs from what's described in [ONYX-173](https://artsyproduct.atlassian.net/browse/ONYX-173) and the associated parent ticket [ONYX-169](https://artsyproduct.atlassian.net/browse/ONYX-169) though.

> If there is more than 1 filter applied, we should choose the most relevant filter, and display Artist name + 1 filter. In order of Price, Medium, Unique/Edition, Other

I read that to mean that instead of the following which includes both Medium Type and Rarity:

`Zoya Cherkassky-Nnadi, Painting or Drawing, Collage or other Work on Paper, Unique`

that we'd simplify the display name to:

`Zoya Cherkassky-Nnadi, Painting or Drawing, Collage or other Work on Paper`

If there's no Price, Medium Type, or Rarity attribute specified, we can choose the first other filter to display:

`Zoya Cherkassky-Nnadi, Medium (40 – 100cm)`

and finally, falling back to just the artist name if _no_ filters are applied at all:

`Zoya Cherkassky-Nnadi`

anandaroop:
Based on the recent feedback ([see comment](https://artsyproduct.atlassian.net/browse/ONYX-173?focusedCommentId=58211)) I substantially re-worked this implementation.

Hope I interpreted this correctly 😅 —

- The first (artist) criteria is delimited with ` — ` and all others with `, `
- Artist, price, medium, rarity are still always preferred
- If there is room for other criteria to be displayed they will be (still capped at 3)
- The other criteria are displayed in unspecified priority, though they match the same order as the pills would because I have leaned heavily on the existing `resolveSearchCriteriaLabels` labels to generate the display texts 

Because of the new implementation there is now also a bonus commit (https://github.com/artsy/metaphysics/pull/5186/commits/38f34b8e0eece60bf9ec0c82dd4ee71dce3e8ca2) that implements the earlier suggestion of indicating `+ N more` when there are more criteria than can be displayed. Feel free to drop that if we don't want it.

So, from the specs, here are some sample outputs:


- Basic criteria
  ```
  KAWS — $100–$1,000, Print, Limited edition
  ```

- Array-valued (OR'd) criteria
  ```
  KAWS or Banksy — Painting or Print, Limited edition or Unique
  ```

- When there are other, non-prioritized criteria
  ```
  KAWS — Small (under 40cm), Buy Now or Bid or Inquire or Make Offer, Acrylic
  KAWS — 1990–1999, Yellow
  KAWS — New York, NY, USA, Foo Bar Gallery
  ```

- When there are more criteria than can be displayed
  ```
  KAWS — $100–$1,000, Print, Limited edition + 7 more
  ```

anandaroop:
@dblandin I'll leave this in your hands. 

Fwiw should be safe to merge and iterate since of course this new field isn't in use anywhere.

dblandin (approved):

MounirDhahri:
Nice one

PR #5197, opened 2023-08-09
fix: Disable IP-based location "near you" feature

olerichter00:
## Description
Disabling the "IP-based" location feature for now until we resolve this rate-limiting issue. ([Slack Thread](https://artsy.slack.com/archives/C02BAQ5K7/p1691567528217389)).

This change will affect the "Shows for You" rail on the home screen and the "Galleries for You" screen in the app.

### DIFF
--- a/src/lib/locationHelpers.ts
+++ b/src/lib/locationHelpers.ts
@@ -7,6 +7,7 @@ interface Location {
 }
 
 const DEFAULT_MAX_DISTANCE_KM = 75
+const DISABLE_IP_BASED_LOCATION = true
 
 export const getLocationArgs = async ({
   ip,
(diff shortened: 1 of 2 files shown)


joeyAghion:
This should be configurable, especially since we're planning to toggle it back on in a few days.

olerichter00:
> This should be configurable, especially since we're planning to toggle it back on in a few days.

Very good idea! 👍 Do you think making it configurable with an environment variable would be a good idea? For example `ENABLE_IP_BASED_LOCATION_LOOKUP` or ``ENABLE_IP_BASED_LOCATION_FALLBACK`.

joeyAghion:
I'd suggest `ENABLE_IP_BASED_LOCATION`.

PR #5199, opened 2023-08-09
fix: Prevent sorting partners by distance without provided location param

olerichter00:
## Description

We need to prevent sorting partners by distance when no location is provided in the following query (because Gravity expects the `near` param when sorting by distance). This case can happen when the IP-based location lookup fails ([Slack Thread](https://artsy.slack.com/archives/C02BAQ5K7/p1691567528217389)).

```graphql
{
  partnersConnection(
    maxDistance: 10
    type: GALLERY
    eligibleForListing: true
    excludeFollowedPartners: false
    defaultProfilePublic: false
    sort: DISTANCE
  ) {
    edges {
      node {
        internalID
      }
    }
  }
}
```

Gravity returns the following error when sorting by `location` without providing `near`:

```
 when `near` param is not provided and ip based location lookup fails › returns a list of partners without sending `distance` param
 ```
 
 **Open Question:**

This could also be fixed in Gravity by defaulting to another sorting when `near` is not provided, but I think Metaphysics is the right place to fix this because it's where the IP-based location lookup happens (which can fail).

### DIFF
--- a/src/schema/v2/partner/__tests__/partners.test.js
+++ b/src/schema/v2/partner/__tests__/partners.test.js
@@ -55,4 +55,61 @@ describe("PartnersConnection", () => {
       "5a958e8e7622dd49f4f4176d"
     )
   })
+
+  describe("when `near` param is not provided and ip based location lookup fails", () => {
+    it("returns a list of partners without sending `distance` param", async () => {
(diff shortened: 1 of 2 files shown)


araujobarret (approved):
it looks fine 👍

PR #5206, opened 2023-08-11
chore: Add more fields to artist update mutation DIA-22

jonallured:
In order to support migrating more fields to the Forque artist edit page this PR adds more fields to the input type in the update mutation. Note that I was very confused about why adding the fields to the interfaces wasn't enough so I pulled up the actual `inputFields` too. This makes it even easier to see that when you add a field you have to add it in three places. :hurtrealbad: 

https://artsyproduct.atlassian.net/browse/DIA-22

/cc @artsy/diamond-devs

### DIFF
--- a/src/schema/v2/artist/updateArtistMutation.ts
+++ b/src/schema/v2/artist/updateArtistMutation.ts
@@ -1,4 +1,5 @@
 import {
+  GraphQLList,
   GraphQLNonNull,
   GraphQLObjectType,
   GraphQLString,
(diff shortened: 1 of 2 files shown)


jonallured:
Oh hrm, making alternateNames NonNull is a breaking change:

https://app.circleci.com/pipelines/github/artsy/metaphysics/18768/workflows/4c44be46-d58d-44bd-b969-1cdc8c73ee08/jobs/39649

But but but arrays shouldn't be null... 😭

src/schema/v2/artist/updateArtistMutation.ts
```diff
@@ -14,21 +15,50 @@ import {
...
+  hometown?: string
   id: string
   last?: string
+  location?: string
   middle?: string
+  nationality?: string
+}
+
+const inputFields = {
+  alternateNames: { type: new GraphQLNonNull(new GraphQLList(GraphQLString)) },
```
araujobarret, line 34: With this type in your tests, GraphQL at least expects an empty list, and looks like we're not providing in the tests or am I tripping?

jonallured: Yeah that's right - that's why CI is failing. But what should I do? Should I introduce a breaking change that something is required or remove the non null even though in theory arrays should not be nullable??

dzucconi: If you want to satisfy the Gravity input; then leaving that optional with a `defaultValue` of empty array, would suffice.

jonallured:
> So what you want is slightly different

ah ha - thanks @dzucconi !!

PR #5226, opened 2023-08-23
feat: Add mutations to manage verified representatives DIA-100

jonallured:
This PR starts by correcting a type mistake so that when we work with `VerifiedRepresentative` records from Gravity we ensure that we have their id. That means this is a breaking change but that's ok! We haven't actually shipped our work to users just yet. But yeah, there will need to be an update in Force so that we can get back in sync.

From there I added the create and delete mutations for `VerifiedRepresentative` records so that I can setup Forque to manage them. I think it's all pretty copy/paste-type stuff but lmk if I missed something!

https://artsyproduct.atlassian.net/browse/DIA-100

/cc @artsy/diamond-devs


dzucconi (approved):
Looks good!

jonallured:
Gonna merge and then I have a Force PR to do.

src/schema/v2/verifiedRepresentative/verifiedRepresentative.ts
```diff
@@ -0,0 +1,26 @@
...
+import { PartnerType } from "schema/v2/partner/partner"
+
+export const VerifiedRepresentativeType = new GraphQLObjectType<any, any>({
+  name: "VerifiedRepresentative",
+  interfaces: [NodeInterface],
+  fields: () => {
+    return {
+      ...IDFields,
+      artist: {
+        type: ArtistType,
```
dzucconi, line 13: Can we make these fields non-nullable?

src/schema/v2/verifiedRepresentative/verifiedRepresentative.ts
```diff
@@ -0,0 +1,26 @@
...
+        resolve: ({ artist_id }, {}, { artistLoader }) => {
+          return artistLoader(artist_id)
+        },
+      },
+      partner: {
+        type: PartnerType,
```
dzucconi, line 19: As well

PR #5228, opened 2023-08-24
feat: improve sending of sale end date params

egdbear:
This PR addresses an issue where null values were causing parameter exceptions in the diffusion. It has been observed that when nulls are sent, they are being converted to empty strings "", leading to unexpected behavior. 
To resolve this, we've enhanced the parameter handling logic to exclude sending parameters when they are null. 

cc @artsy/diamond-devs

### DIFF
--- a/src/schema/v2/artist/index.ts
+++ b/src/schema/v2/artist/index.ts
@@ -316,9 +316,11 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
             sizes,
             sort: options.sort,
             state: options.state,
-            sale_start_year: options.saleStartYear,
-            sale_end_year: options.saleEndYear,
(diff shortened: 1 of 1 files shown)


src/schema/v2/artist/index.ts
```diff
@@ -316,9 +316,11 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
             sizes,
             sort: options.sort,
             state: options.state,
-            sale_start_year: options.saleStartYear,
-            sale_end_year: options.saleEndYear,
             allow_unspecified_sale_dates: options.allowUnspecifiedSaleDates,
+            ...(options.saleStartYear && {
```
egdbear, line 320: Obviously, if we notice a repeating pattern, we could create a method for omitting null values.

egdbear:
Merging

PR #5230, opened 2023-08-24
Add not null to verified rep type

jonallured:
Reissue of #5229 this time from origin rather than my fork.

### DIFF
--- a/src/schema/v2/verifiedRepresentative/verifiedRepresentative.ts
+++ b/src/schema/v2/verifiedRepresentative/verifiedRepresentative.ts
@@ -1,4 +1,4 @@
-import { GraphQLObjectType } from "graphql"
+import { GraphQLNonNull, GraphQLObjectType } from "graphql"
 import { IDFields, NodeInterface } from "schema/v2/object_identification"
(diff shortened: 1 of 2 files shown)


egdbear (approved):

src/schema/v2/verifiedRepresentative/verifiedRepresentative.ts
```diff
@@ -10,13 +10,13 @@ export const VerifiedRepresentativeType = new GraphQLObjectType<any, any>({
     return {
       ...IDFields,
       artist: {
-        type: ArtistType,
+        type: new GraphQLNonNull(ArtistType),
```
egdbear, line 13: I believe `ArtistType!` _should_ also work here

jonallured: I don't think that's how it works! I think you do have to use the `GraphQLNonNull` thingy:

Screenshot 2023-08-24 at 10 05 41 AM

egdbear: Yeah, that seems quite unusual 😓 
image

PR #5282, opened 2023-09-13
fix: Change formatting of price high/lows

jpotts244:
Addressing part of this ticket:
https://artsyproduct.atlassian.net/browse/AMBER-191

Convert asterisks to:
$0 if there’s no min specified (e.g. `$0-$1,000` instead of `*-$1000`)
“+” if there’s no max specified (e.g. `$1,000+` instead of `$1000-*`

### DIFF
--- a/src/lib/stitching/gravity/__tests__/stitching.test.ts
+++ b/src/lib/stitching/gravity/__tests__/stitching.test.ts
@@ -1134,7 +1134,7 @@ describe("gravity/stitching", () => {
           {}
         )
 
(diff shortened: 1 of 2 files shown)


mdole (approved):
🙏

oxaudo:
This works! Though `$0 - xx` seems a bit contrived. My other possible suggestions would be `< $xxx` , `below $xxx` and or `$xxx and above` for the other end - might read a bit more natural to partners.

PR #5296, opened 2023-09-18
feat(alerts): add `isEligibleToCreateAlert` to Artwork type

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-363

This moves the responsibility for determining _whether an artwork is eligible to have an alert created from it_ up into Metaphysics, instead of being scattered across client code paths in a divergent or error-prone way, e.g.:

- Force: [ArtworkSidebarCreateArtworkAlert.tsx ](https://github.com/artsy/force/blob/e0118180edc02645e595100177e67453e8acfdfa/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCreateArtworkAlert.tsx#L79C1-L79C1)
- Force: [ArtworkSidebarBiddingClosedMessage.tsx](https://github.com/artsy/force/blob/156b302c328600c704360e4998168c1a72186d44/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarBiddingClosedMessage.tsx#L24)
- Eigen: [ArtworkScreenHeaderCreateAlert.tsx](https://github.com/artsy/eigen/blob/090a826251e0e796dc3904f5f0fa7a35df122cf9/src/app/Scenes/Artwork/Components/ArtworkScreenHeaderCreateAlert.tsx#L22)
- Eigen: [ArtworkAuctionCreateAlertHeader.tsx](https://github.com/artsy/eigen/blob/a68106cb6f6a15fd93f43302cca6b36873f642ef/src/app/Scenes/Artwork/ArtworkAuctionCreateAlertHeader.tsx#L34)
- etc…

Will need to be followed by PRs to update the above paths

#### Some real-world examples…

Screenshot 2023-09-18 at 6 32 48 PM

### DIFF
--- a/src/schema/v2/artwork/utilities.ts
+++ b/src/schema/v2/artwork/utilities.ts
@@ -3,6 +3,7 @@ import qs from "querystring"
 import { Artwork } from "types/runtime/gravity"
 import { parse } from "url"
 import { normalizeImageData } from "../image"
+import mediums from "lib/artworkMediums"
 
 export const isTwoDimensional = ({
   width_cm,
@@ -124,4 +125,20 @@ export const isEligibleForOnPlatformTransaction = ({
   }
 
   return false
+}
+
+export const isEligibleToCreateAlert = (artwork: {
+  artists: { id: string }[]
+  category: string | null
+}) => {
+  const { artists, category } = artwork
+
+  // an alert must be associated with at least one valid artist
+  if (artists.length == 0) return false
+
+  // an alert should be associated with a real medium, and not just "Other"
+  if (!category) return false
+  if (category == mediums.Other.name) return false
+
+  return true
 }
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2255,6 +2255,9 @@ type Artwork implements Node & Searchable & Sellable {
 
   # Artwork is eligible for on-platform transaction
   isEligibleForOnPlatformTransaction: Boolean!
+
+  # Artwork meets minimum metadata criteria to have an alert created from it
+  isEligibleToCreateAlert: Boolean!
   isEmbeddableVideo: Boolean
   isForSale: Boolean
   isHangable: Boolean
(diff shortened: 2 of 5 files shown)


src/schema/v2/artwork/utilities.ts
```diff
@@ -125,3 +126,19 @@ export const isEligibleForOnPlatformTransaction = ({
...
+  category: string | null
+}) => {
+  const { artists, category } = artwork
+
+  // an alert must be associated with at least one valid artist
+  if (artists.length == 0) return false
+
+  // an alert should be associated with a real medium, and not just "Other"
+  if (!category) return false
+  if (category == mediums.Other.name) return false
```
anandaroop, line 141: So far our logic has only tested for the presence of artists (i.e. L137)

This adds an opinionated test for the presence of a valid medium as well (L140–141), on the reasoning that…

- A null medium would likely result in overly broad alerts
- A catch-all "Other" medium would not generate a useful alert, and presumably [not even a valid one](https://github.com/artsy/metaphysics/blob/89e7f5b1821136db52cb0a32bb111f79f8bd9005/src/lib/artworkMediums.ts#L91)

mzikherman (approved):
Nice - makes sense for an MP field!

src/schema/v2/artwork/utilities.ts
nickskalkin: > A null medium would likely result in overly broad alerts

But it is still possible to create and alert without it, from an artist page, for instance. I guess I'm ok with adding such check here, but wondering if this mismatch in validations can cause problems in future

nickskalkin (approved):
awesome 👍

src/schema/v2/artwork/utilities.ts
anandaroop: >  still possible to create and alert without it, from an artist page

Yep, and I expect this flag would not be consulted in the case of setting an alert from the artist artwork grid. This should only govern the "alert-ability" of an individual artwork, where I think these tighter rules make sense.

dblandin (approved):
Awesome 👍🏼

PR #5306, opened 2023-09-21
chore: Remove custom price formatting stitching in favor of gravity formatte…

jpotts244:
…dPriceRange value

Removing price format stitching from MP as we can now rely on Gravity to handle the price range formatting for SearchCriteria
- Supporting Gravity Changes: https://github.com/artsy/gravity/pull/16865
- Supporting Volt V2 Changes: https://github.com/artsy/volt-v2/pull/742

I've removed the use of this from Volt V2 yesterday so I dont think any other clients are using the `lowPriceAmount`/`highPriceAmount` fields


damassi:
🙏 tyy!!

PR #5308, opened 2023-09-22
fix(artist): returns null for bio when partner bio is false and we do not have a blurb

dzucconi:
Closes [DIA-170](https://artsyproduct.atlassian.net/browse/DIA-170)

[DIA-170]: https://artsyproduct.atlassian.net/browse/DIA-170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


egdbear (approved):
👀

PR #5312, opened 2023-09-26
feat: embed the url in the selected insights

egdbear:
It embeds the urls to selected insights.

[Ticket](https://artsyproduct.atlassian.net/jira/software/projects/DIA/boards/146?selectedIssue=DIA-179)


dzucconi (changes requested):
This winds up being a breaking change: if deployed we will see  unformatted HTML show up. Also its just annoying to get HTML when you don't ask for it. It might make sense to add a 'format' argument. We could do it in stages wherein we accept the arg, update clients to request PLAIN, then implement these as Markdown fields to be consistent with the rest of the schema.

egdbear:
Good point @dzucconi

src/schema/v2/artist/insights.ts
```diff
@@ -36,6 +37,13 @@ export const ArtistInsight = new GraphQLObjectType<any, ResolverContext>({
     },
     description: {
       type: GraphQLString,
+      args: {
+        ...markdown().args,
+      },
+      resolve: ({ description }, { format }) => {
+        if (!format) return formatMarkdownValue(description, "plain").trim()
```
egdbear, line 44: @dzucconi we are avoid breaking changes here

dzucconi: You should be able to just use the normal markdown field in it's entirety, no?

src/schema/v2/artist/helpers.ts
```diff
@@ -254,3 +272,7 @@ export const getRecentShow = (artist): string | null => {
 
   return title
 }
+
+export const toMarkDownUrl = (url: string, text: string) => {
+  return `[${text}](${url})`
```
dzucconi, line 277: This is just a confusing helper. Just write Markdown without functions.

src/schema/v2/artist/insights.ts
egdbear: it doesn't have the defaultValue

src/schema/v2/artist/helpers.ts
dzucconi: (Should be fine to delete now)

src/schema/v2/artist/insights.ts
dzucconi, line 44: Would be better to set the `defaultValue` property in the args so that the resolver can be the same.

egdbear: yeah

src/schema/v2/artist/insights.ts
```diff
@@ -36,6 +37,17 @@ export const ArtistInsight = new GraphQLObjectType<any, ResolverContext>({
     },
     description: {
       type: GraphQLString,
+      args: {
+        ...markdown().args,
+        defaultValue: {
+          type: GraphQLString,
+          defaultValue: "plain",
```
dzucconi, line 44: So this isn't specified correctly. You want to merge into the markdown args and provide a `defaultValue` to the `format` arg. It shoudn't be a string but will rather be of type `Format` with a value of `PLAIN`

egdbear: I'm not quite sure if that's possible. Do you have an example?

dzucconi: Something like this would work?

```
--- a/src/schema/v2/artist/insights.ts
+++ b/src/schema/v2/artist/insights.ts
@@ -14,7 +14,8 @@ import {
   getArtistInsights,
   getAuctionRecord,
 } from "./helpers"
-import { markdown, formatMarkdownValue } from "schema/v2/fields/markdown"
+import { formatMarkdownValue } from "schema/v2/fields/markdown"
+import Format, { FORMATS } from "schema/v2/input_fields/format"
 
 export const ArtistInsightKind = new GraphQLEnumType({
   name: "ArtistInsightKind",
@@ -38,14 +39,12 @@ export const ArtistInsight = new GraphQLObjectType<any, ResolverContext>({
     description: {
       type: GraphQLString,
       args: {
-        ...markdown().args,
-        defaultValue: {
-          type: GraphQLString,
-          defaultValue: "plain",
+        format: {
+          ...Format,
+          defaultValue: FORMATS.PLAIN.value,
         },
       },
       resolve: ({ description }, { format }) => {
-        if (!format) return formatMarkdownValue(description, "plain").trim()
         return formatMarkdownValue(description, format).trim()
       },
     },
diff --git a/src/schema/v2/input_fields/format.ts b/src/schema/v2/input_fields/format.ts
index e9df54f1..562234d1 100644
--- a/src/schema/v2/input_fields/format.ts
+++ b/src/schema/v2/input_fields/format.ts
@@ -1,12 +1,12 @@
-import { GraphQLEnumType } from "graphql"
+import { GraphQLEnumType, GraphQLArgumentConfig } from "graphql"
 
-const FORMATS = {
+export const FORMATS = {
   HTML: { value: "html" },
   PLAIN: { value: "plain" },
   MARKDOWN: { value: "markdown" },
 } as const
 
-const Format = {
+export const Format: GraphQLArgumentConfig = {
   type: new GraphQLEnumType({
     name: "Format",
     values: FORMATS,


```

dzucconi (approved):
Cool, looks good!

jonallured (approved):
Nice job working through this one!

egdbear:
merging

PR #5316, opened 2023-09-27
fix: avoid format if description not present

egdbear:
@MounirDhahri noticed that we are experiencing an issue when trying to obtain insights without specifying the kind. This problem was introduced on the staging environment with [this PR](https://github.com/artsy/metaphysics/pull/5312).

This PR adds the ability to skip the "marked" function call if we don't have a description.


araujobarret (approved):
Das war super schnell!

dzucconi (approved):

MounirDhahri:
thanks 🙏

PR #5325, opened 2023-10-04
chore(AMBER-215): remove lastMessageDeliveryID

mdole:
Sister PR to https://github.com/artsy/impulse/pull/987.

Pretty sure this is safe to remove entirely and not just deprecate; it's not referenced in Eigen and AFAIK conversations haven't been touched there for a while. If folks feel it would be safer to mark as deprecated though, can switch to that.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -7091,9 +7091,6 @@ type Conversation implements Node {
     timezone: String
   ): String
 
-  # Delivery id if the user is a recipient of the last message, null otherwise.
-  lastMessageDeliveryID: String
-
   # Impulse id of the last message.
   lastMessageID: String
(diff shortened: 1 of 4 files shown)


damassi (commented):
(Also don't see it in Force or Volt 2)

erikdstock (approved):
[Searching the artsy org](https://github.com/search?q=org%3Aartsy+lastMessageDeliveryID&type=code) i see it referenced in a few places, but only graphQL schemas. I also searched for snake-cased in case a rails app was being clever with it and came up empty. Seems safe, but don't know if we have other norms around this. Approved but defer to others.

joeyAghion (approved):
🤷

PR #5331, opened 2023-10-09
feat: remove the old created year aggregation for auction results

egdbear:
This PR removes the old loader for created dates.

Currently, this query is only used in [Eigen](https://github.com/artsy/eigen/blob/a031acc72df3314723456c9523500cd9dbdb7e18/src/app/Components/Artist/ArtistInsights/ArtistInsightsAuctionResults.tsx#L93). 
We can create a follow-up ticket to replace the Eigen aggregation and then remove `createdYearRange` completely from the schema. @MounirDhahri 

[Metaphysics - the new aggregations](https://github.com/artsy/metaphysics/pull/5255/files)
[Diffusion PR](https://github.com/artsy/diffusion/pull/780)
[DIA-133]

@artsy/diamond-devs 

[DIA-133]: https://artsyproduct.atlassian.net/browse/DIA-133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/auction_result.ts
+++ b/src/schema/v2/auction_result.ts
@@ -32,6 +32,7 @@ import date from "./fields/date"
 import { InternalIDFields, NodeInterface } from "./object_identification"
 // Taken from https://github.com/RubyMoney/money/blob/master/config/currency_iso.json
 import { YearRange } from "./types/yearRange"
+import { GraphQLError } from "graphql"
 
 export const AuctionResultSorts = {
   type: new GraphQLEnumType({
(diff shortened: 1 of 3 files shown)


jonallured (dismissed):
LGTM!! 🎉

MounirDhahri (approved):
Nice! One less endpoint to maintain. Surprising that this didn't fail tests, we should probably also add some tests here.

src/schema/v2/auction_result.ts
```diff
@@ -414,13 +415,41 @@ export const auctionResultConnection = connectionWithCursorInfo({
...
+            }
+          }
+        } catch (error) {
+          throw new GraphQLError(`createdYearRange resolver error: ${error}`)
+        }
+
+        return {
+          startAt: null,
+          endAt: null,
+        }
```
MounirDhahri, line 452: we can maybe move this to line 444, not a blocker though

araujobarret: it's easier to read if we move it there ++

egdbear: good one!

egdbear:
> Nice! One less endpoint to maintain. Surprising that this didn't fail tests, we should probably also add some tests here.

We could add more tests for this query; however, I'm not sure if it makes sense because we already [have the test](https://github.com/artsy/metaphysics/blob/4b9212334643b08707c59647c76d8300f7c47461/src/schema/v2/__tests__/auction_results.test.js#L66) for the loader, and mocking again sounds repetitive. What do you think?

MounirDhahri:
I think a test here would be not to cover the loader mainly but instead to unit test the start/end date response depending on the loader response. The logic is quite straightforward though so we can maybe go away without adding a test here

araujobarret:
I agree with @MounirDhahri , the new flow has at least 4 different ways that would be nice to cover with unit tests.

egdbear:
> I think a test here would be not to cover the loader mainly but instead to unit test the start/end date response depending on the loader response. The logic is quite straightforward though so we can maybe go away without adding a test here

On a second thought this is the safe option!

MounirDhahri (approved):

MounirDhahri:
Awesome, thanks @egdbear

egdbear:
Okay let's merge this!

PR #5332, opened 2023-10-09
fix: remove description for some insights

egdbear:
Remove the descriptions so we use the entities instead.

image

[Slack](https://artsy.slack.com/archives/C05EEBNEF71/p1696869404116629)


jonallured (approved):
LGTM!! 🎉

PR #5347, opened 2023-10-13
fix: incorporate -dev compose template

ovasdi:
Follow-up to https://github.com/artsy/metaphysics/pull/5341

Fixes broken [builds](https://app.circleci.com/pipelines/github/artsy/metaphysics/19542/workflows/f5a03efe-b092-402c-a818-b05997961431/jobs/41167?invite=true#step-106-1555_25) as a result of `docker-compose` template change.

### DIFF
--- a/hokusai/development.yml
+++ b/hokusai/development.yml
@@ -1,7 +1,7 @@
 version: "2"
 services:
   metaphysics:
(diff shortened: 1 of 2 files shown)


mc-jones (approved):
🙏

PR #5351, opened 2023-10-16
[ONYX-454] feat(notifications/activity): expose artworks within PartnerShowOpenedActivity

dblandin:
[ONYX-454](https://artsyproduct.atlassian.net/browse/ONYX-454)

Artworks are currently exposed for `SavedSearchHitActivity` and `ArtworkPublishedActivity` events. While other types do not currently include artwork IDs within the events `object_ids` structure, we can utilize other data loaders to fetch associated artworks.

This also includes a switch to `filterArtworksLoader` which supports more filter params and is a bit more performant than `artworksLoader`, primarily due to our use of Elasticsearch under the hood.

[ONYX-454]: https://artsyproduct.atlassian.net/browse/ONYX-454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/notifications/index.ts
+++ b/src/schema/v2/notifications/index.ts
@@ -99,14 +99,35 @@ export const NotificationType = new GraphQLObjectType<any, ResolverContext>({
     artworksConnection: {
       type: artworkConnection.connectionType,
       args: pageable(),
-      resolve: async ({ object_ids: ids }, args, { artworksLoader }) => {
+      resolve: async (
+        { activity_type, object_ids },
+        args,
+        { filterArtworksLoader }
+      ) => {
+        let loaderArgs = {}
+
+        switch (activity_type) {
+          case "PartnerShowOpenedActivity":
+            loaderArgs = { partner_show_id: object_ids[0] }
+            break
+          default: {
+            loaderArgs = { ids: object_ids }
+            break
+          }
+        }
+
         const { page, size } = convertConnectionArgsToGravityArgs(args)
-        return artworksLoader({ ids }).then((body) => {
-          const totalCount = body.length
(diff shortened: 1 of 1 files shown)


dblandin:
cc/ @olerichter00 @nickskalkin @anandaroop as a conversation starter.

It think the primary risk of this approach is that it'll make `notificationsConnection` resolution slower, especially if the query includes entries of these other types. This does however move to a more performant resolution of artworks, via `filterArtworksConnection` which might offset that concern.

This also intentionally picks out the first object of the activity entry, whether it's a show (or viewing room if this approach is expanded), and only includes artworks from that first object.

anandaroop:
Makes sense, and the tradeoff re: switching but also now using the more performant loader sounds plausible.

Since we can't predict net performance, we can always keep an eye on [this DD entry](https://app.datadoghq.com/apm/resource/metaphysics.graphql/graphql.execute/b8ba750f6c86f6ec?query=env%3Aproduction%20service%3Ametaphysics.graphql%20operation_name%3Agraphql.execute&env=production&topGraphs=latency%3Alatency%2Chits%3Aversion_count%2Cerrors%3Aversion_count%2CbreakdownAs%3Aabsolute&traces=qson%3A%28data%3A%28%29%2Cversion%3A%210%29&start=1697405348645&end=1697491748645&paused=false).

dblandin:
> looks good to me! I don't understand though why this change should make the query slower

I suppose you're right — this shouldn't be any slower than the current `artworksConnection` resolutions for artwork alert or artwork published events.

PR #5358, opened 2023-10-19
feat: enable more queries for the braze

egdbear:
This PR enables the remaining queries from [the ticket](https://artsyproduct.atlassian.net/browse/DIA-206).

```sh
watchedLotConnection
userInterestsConnection
myCollectionConnection
followsAndSaves {
    galleriesConnection
    fairsConnection
    artistsConnection
}
artworkInquiriesConnection
collectionsConnection
lotStanding
auctionResultsByFollowedArtists
myCollectionAuctionResults
```

[Gravity PR](https://github.com/artsy/gravity/pull/16974)
@artsy/diamond-devs

### DIFF
--- a/src/schema/v2/me/index.ts
+++ b/src/schema/v2/me/index.ts
@@ -578,6 +578,15 @@ const MeField: GraphQLFieldConfig<void, ResolverContext> = {
       "showsByFollowedArtists",
       "newWorksFromGalleriesYouFollowConnection",
       "myCollectionInfo",
+      "watchedLotConnection",
+      "userInterestsConnection",
+      "recentlyViewedArtworksConnection",
+      "myCollectionConnection",
+      "artworkInquiriesConnection",
+      "collectionsConnection",
+      "auctionResultsByFollowedArtists",
+      "myCollectionAuctionResults",
+      "newWorksByInterestingArtists",
     ]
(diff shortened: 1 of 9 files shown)


src/schema/v2/me/index.ts
```diff
@@ -578,6 +578,15 @@ const MeField: GraphQLFieldConfig<void, ResolverContext> = {
...
       "myCollectionInfo",
+      "watchedLotConnection",
+      "userInterestsConnection",
+      "recentlyViewedArtworksConnection",
+      "myCollectionConnection",
+      "artworkInquiriesConnection",
+      "collectionsConnection",
+      "auctionResultsByFollowedArtists",
+      "myCollectionAuctionResults",
+      "newWorksByInterestingArtists",
```
damassi, line 589: Can we alphabetize this whole array?

joeyAghion (commented):
I don't see problems, but I probably _wouldn't_ so will defer to the other reviewers.

src/schema/v2/me/index.ts
egdbear: As far as I know, this occurs in a couple of places in metaphysics, and it could be resolved using the general ESLint rule `sort-keys`, perhaps. But definitely, as a code bestpractice/guideline improvement in separate PR, what do you think?

egdbear:
It is Friday and why not to merge this???

PR #5374, opened 2023-10-25
feat: add eslint rule sort-keys

egdbear:
I was motivated to create this PR after reading this [comment](https://github.com/artsy/metaphysics/pull/5358#discussion_r1370290136). To simplify the linting process for everyone, we could introduce the ESLint rule `sort-keys`. We can begin with 'warn' first before implementing it as an 'error' to avoid any blockers.

Screenshot 2023-10-25 at 17 43 13

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -16,6 +16,11 @@
   ],
   "parser": "@typescript-eslint/parser",
   "rules": {
+    "sort-keys": [
+      "warn",
+      "asc",
+      { "caseSensitive": true, "natural": false, "minKeys": 2 }
+    ],
     "import/first": 0,
     "import/named": 0,
     "import/default": 0,
--- a/package.json
+++ b/package.json
@@ -156,7 +156,7 @@
       "yarn prettier --write",
       "git add"
     ],
-    "*.@(js)": [
+    "*.@(js|ts)": [
       "eslint --fix",
       "yarn prettier --write",
       "git add"


damassi:
If sort keys hasn't been updated with an eslint-autofixer, we'll want to use this one! https://www.npmjs.com/package/eslint-plugin-sort-keys-fix. 

One thing i worry about tho is the size of the codebase at this point in time as there will likely be manyyyy warnings, every time one opens a file, and some diff thrashing if the autofix is enabled. 

Not opposed to having ESLint rules for stuff like this, but at the same time it's also one of those things that feels like a good habit to just learn and do (where appropriate, because sometimes a certain non-alphabetical order is more appropriate... sometimes.)

Any chance you could add the auto fix plugin, then run eslint over the whole codebase and open a PR for that, and see what happens with the tests? For things like this, fixing everything up front will probably be worthwhile.

egdbear:
I noticed that eslint is only running for [js files](https://github.com/artsy/metaphysics/blob/main/package.json#L159C5-L159C5). 
Should we keep that still?

egdbear:
>Any chance you could add the auto fix plugin, then run eslint over the whole codebase and open a PR for that, and see what happens with the tests? For things like this, fixing everything up front will probably be worthwhile, similar to adding something like prettier.

Sounds like a plan! 
Also do we want to include `*.ts` files, I see that [`*.ts` files are excluded.](https://github.com/artsy/metaphysics/blob/main/package.json#L159C5-L159C5)

damassi:
Wow, nice find @egdbear - eslint works in an editor (like VSCode, with it enabled), but it def is not running on precommit! We should enable that for sure.

damassi:
Actually @egdbear - now that i'm thinking about it, for a codebase this sensitive where git history is imperative, i think autofix / running it over everything might be to severe... 

Lets update the precommit hook settings in package.json to check for tsx as well -- `(js|ts)` -- and then we can merge this 👍

egdbear:
this should be ok now @damassi

damassi:
Thanks so much for the tooling improvement 🙏

PR #5378, opened 2023-10-26
chore: userRoles.name should never be null

joeyAghion:
It was suggested around [this downstream change](https://github.com/artsy/forque/pull/1064) that `userRoles`' `name`s should never be `null`.

### DIFF
--- a/src/schema/v2/system/userRoles.ts
+++ b/src/schema/v2/system/userRoles.ts
@@ -3,6 +3,7 @@ import {
   GraphQLFieldConfig,
   GraphQLObjectType,
   GraphQLString,
+  GraphQLNonNull,
 } from "graphql"
 import { ResolverContext } from "types/graphql"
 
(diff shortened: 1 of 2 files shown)


src/schema/v2/system/userRoles.ts
```diff
@@ -11,7 +12,7 @@ const UserRoleType = new GraphQLObjectType<any, ResolverContext>({
   description: "Fields corresponding to a given product privilege",
   fields: {
     name: {
-      type: GraphQLString,
+      type: new GraphQLNonNull(GraphQLString),
```
damassi, line 15: Note that if this in anyway _can_ be null or undefined, it will error out at the MP level. So we only mark as non-nullable if we're absolutely certain that there is a value upstream from here. If not sure, better to be safe and leave it alone, and deal with it at the client consumer level.

joeyAghion:
After some discussion we went with [this more radical alternative](https://github.com/artsy/metaphysics/pull/5379).

PR #5383, opened 2023-10-31
feat: add vortex loader for the braze queries

egdbear:
This PR adds a new GraphQL loader in `loaders_without_authentication/vortex`.

Reasoning: `vortexGraphqlLoaderFactory` is _produced_ from `vortexLoader*WithAuthentication*Factory`, which calls the `me` endpoint in the background. Therefore, I aimed to avoid this and add a loader that can be utilised for the Braze-related queries.

It would be amazing to utilise a single GraphQL loader in `loaders_without_authentication` for both Braze-related and user-related queries. However, this may require some additional changes and more thoughts.

I'm also open for suggestions here 😅 

@artsy/diamond-devs

### DIFF
--- a/src/lib/loaders/loaders_without_authentication/vortex.ts
+++ b/src/lib/loaders/loaders_without_authentication/vortex.ts
@@ -1,12 +1,61 @@
 import factories from "../api"
+import config from "config"
+const { VORTEX_API_BASE } = config
+import fetch from "node-fetch"
+import urljoin from "url-join"
+import { GraphQLError } from "graphql"
 
-interface LoaderArgs {
+interface GraphQLArgs {
   query: string
   variables?: any
 }
 
 export default (opts) => {
   const { vortexLoaderWithoutAuthenticationFactory } = factories(opts)
+
+  const vortexGraphqlImpersonationLoader = (opts) => async <T = unknown>({
+    query,
+  }: GraphQLArgs): Promise<Record<string, T>> => {
+    const body = JSON.stringify({
+      query,
+    })
+    const response = await fetch(urljoin(VORTEX_API_BASE, "graphql"), {
+      body,
+      headers: {
+        Authorization: `${opts.appToken}`,
+        "Content-Type": "application/json",
+      },
+      method: "POST",
+    })
+
+    const json = await response.json()
+    const { error, errors: vortexErrors } = json
+    if (error) {
+      console.log(error)
+      throw new Error(`[loaders/vortex.ts]: ${error}`)
+      // If the vortex request failed for some reason, throw its errors.
+    } else if (vortexErrors) {
+      const errors = vortexErrors.reduce((acc, error) => {
+        return acc + " " + error["message"]
+      }, "From vortex service:")
+
+      throw new GraphQLError(`[loaders/vortex.ts]: ${errors}`)
+    } else {
+      return json
+    }
+  }
+
+  const vortexGraphqlLoaderFactory = (appToken) => {
+    return ({ query, variables }: GraphQLArgs) => {
+      return setup(appToken)(
+        "/graphql",
+        { query, variables: JSON.stringify(variables) },
+        {
+          method: "POST",
+        }
+      )
+    }
+  }
 
   return {
     vortexGraphqlLoader: ({ query, variables }: LoaderArgs) => {
--- a/src/lib/loaders/loaders_without_authentication/vortex.ts
+++ b/src/lib/loaders/loaders_without_authentication/vortex.ts
@@ -1,9 +1,4 @@
 import factories from "../api"
-import config from "config"
-const { VORTEX_API_BASE } = config
-import fetch from "node-fetch"
-import urljoin from "url-join"
-import { GraphQLError } from "graphql"
 
 interface GraphQLArgs {
   query: string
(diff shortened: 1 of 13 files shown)


src/lib/loaders/loaders_without_authentication/vortex.ts
```diff
@@ -14,17 +19,52 @@ export default (opts) => {
...
+  const vortexGraphqlImpersonationLoader = (opts) => async <T = unknown>({
+    query,
+  }: GraphQLArgs): Promise<Record<string, T>> => {
+    const body = JSON.stringify({
+      query,
+    })
+    const response = await fetch(urljoin(VORTEX_API_BASE, "graphql"), {
+      body,
+      headers: {
+        Authorization: `${opts.appToken}`,
```
mzikherman, line 31: I'm wondering how this loader gets authorized to hit Vortex. Should this be `Bearer ...` ? And also, what is `opts.appToken` - is that a Vortex token?

egdbear: `opts.appToken` is the `X-Xapp-Token` and in this case it always come as vortex token.

mzikherman: Isn't that a Gravity thing though? Only Gravity knows about roles and verifying a Gravity-issued xapp token. Does Vortex know/respect those?

egdbear: Nope [check this out](https://github.com/artsy/vortex/blob/d742bcd5999dcd934d71176b863bf257b925d2b0/app/graphql/resolvers/base_resolver.rb#L22)

mzikherman: Gotcha

mzikherman: I guess I still don't quite understand how Vortex will verify this token though. The `opts.appToken` is either MP's own xapp-token when the incoming request doesn't specify one, _or_ it's the incoming request's one.

So Braze's xapp token flows through, and is a Gravity xapp token. How does Vortex verify that token? I would think this might still need the `setup(appToken)` part to swap a gravity xapp token to a Vortex token.

(sorry for the many questions!)

egdbear: Got it. Does it make more sense to use `config.VORTEX_TOKEN` instead?

mzikherman: That's a token set in Metaphysics ENV and MP isn't the `email_provider`. Most likely this needs a `setup(appToken)` type callback like the other loader to swap an incoming gravity xapp token for a vortex one.

mzikherman: It also could be that I'm totally not understanding something about Vortex and how it works...I've never used that app before. I would defer to @jonallured or even @joeyAghion on these maybe? Or if we have a little mob session on this to do some knowledge-sharing.

mzikherman: https://github.com/artsy/metaphysics/pull/4030 looks to add some of this vortex functionality (which i honestly don't fully understand), so maybe @jonallured is the person

mzikherman:
Ok, after rebasing https://github.com/artsy/metaphysics/pull/5392 , I think the vortex loader construction will simplify greatly, as you'll simply be able to reuse the existing loader, and likely won't need any changes in that file anymore.

Sorry for the conflicts 😅  but I think this will be a lot cleaner.

src/lib/loaders/loaders_without_authentication/vortex.ts
```diff
@@ -9,7 +9,10 @@ export default (opts) => {
   const { vortexLoaderWithoutAuthenticationFactory } = factories(opts)
 
   return {
-    vortexGraphqlLoader: ({ query, variables }: LoaderArgs) => {
+    vortexGraphqlLoaderWithoutAuthentication: ({
```
mzikherman, line 12: You don't need to rename this, all the loaders are exposed in the context under the `loadersWithAuthentication` and `loadersWithoutAuthentication` objects already, so can pluck both/whichever ones you need. This is in addition to them being available at the root with the same-name collision.

That's in https://github.com/artsy/metaphysics/blob/aefe429bf964cd4ad90252dc98c06817b0ec4c7b/src/lib/loaders/index.ts#L93-L94

src/schema/v2/artworksForUser/helpers.ts
```diff
@@ -7,15 +7,21 @@ export const getNewForYouRecs = async (
   args: CursorPageable,
   context: ResolverContext
 ): Promise<string[]> => {
-  const { vortexGraphqlLoader } = context
+  if (!context.vortexGraphqlLoader) return Promise.resolve([])
```
mzikherman, line 10: I don't follow....why return a promise / vs. `[]` / vs. any early-return?

mzikherman:
I pushed a commit (that is a ❌ btw) - that I think demonstrates a sensible setup for `artworksForUser` and could likely be reused for the other examples.

egdbear:
> I pushed a commit (that is a ❌ btw) - that I think demonstrates a sensible setup for `artworksForUser` and could likely be reused for the other examples.

not sure how this is going to work within the me query.

src/schema/v2/me/artistRecommendations.ts
```diff
@@ -44,14 +55,19 @@ export const ArtistRecommendations: GraphQLFieldConfig<
           }
         }
       `,
-    })()
+    }
+
+    const isAuthenticatedRequest = !!vortexGraphQLAuthenticatedLoader
```
egdbear, line 60: I believe this will consistently fail on non-authenticated requests and the following code will be inaccessible? 🤔 

So something like can work here?

```js    
let vortexResult = null

try {
     vortexResult = await vortexGraphQLAuthenticatedLoader!(query)()
} catch (error) {
      vortexResult = await vortexGraphQLUnauthenticatedLoader(query)()
}
```

src/schema/v2/artworksForUser/helpers.ts
```diff
@@ -33,7 +43,13 @@ export const getNewForYouRecs = async (
           }
         }
       `,
-  })()
+  }
+
+  const isAuthenticatedRequest = !!vortexGraphQLAuthenticatedLoader
+
+  const vortexResult = isAuthenticatedRequest
+    ? await vortexGraphQLAuthenticatedLoader(query)()
```
mzikherman, line 51: So does this work? If a request contains the impersonate headers, it still counts as authenticated, but then this loader will fail
Screen Shot 2023-11-07 at 10 18 42 AM

Notice how it hits a `me` route. Maybe it's better to check `xImpersonateUserId` instead?

egdbear: makes sense!

egdbear:
This should be ready for merge. 
Note that when testing, an `artworkRecommendation` record with the same `user_id` should exist.

src/schema/v2/artworksForUser/helpers.ts
```diff
@@ -33,7 +43,11 @@ export const getNewForYouRecs = async (
           }
         }
       `,
-  })()
+  }
+
+  const vortexResult = xImpersonateUserID
+    ? await vortexGraphQLUnauthenticatedLoader(query)()
+    : await vortexGraphQLAuthenticatedLoader!(query)()
```
mzikherman, line 50: Rather than use `!`, why not follow the TS suggestion and do an exist-y check? If an un-authenticated request that doesn't specify an impersonate user hits this endpoint, it'll try to use this loader, and probably fail with an odd error cause this loader doesn't exist.

Whereas TS is telling you that, so you can specifically do `if (!authenticatedLoader) throw new Error('You must be logged in')` or something like that.

In general we overuse `!` a lot and it should probably only be used rarely.

This isn't a blocker in this case however, b/c while following TS will enable you to show a more sensible error message, we know the request can't succeed if there's no impersonation _and_ there's no access token.

egdbear:
Oh well, this went without squashing the commit 😄

PR #5384, opened 2023-10-31
fix: adding missing color to saved search preview name

MounirDhahri:
This PR solves [APPL-605]

The issue came from the fact that we were missing some colors in the list in MP. I added the missing color as well as made this less defensive to avoid breaking the UI.

https://github.com/artsy/metaphysics/assets/11945712/63af94fe-13f5-4450-ac29-a986b3b5a504

[APPL-605]: https://artsyproduct.atlassian.net/browse/APPL-605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/colors.ts
```diff
@@ -1,138 +1,149 @@
...
+  foregroundColor: string
+  name: string
+  value: string
+}[] = [
+  {
+    backgroundColor: "#7B5927",
+    foregroundColor: "#fff",
+    name: "Brown",
+    value: "brown",
+  },
```
anandaroop, line 94: 👍🏽

anandaroop (approved):
Sounds reasonable, though it's interesting it took this long to surface!

MounirDhahri:
will address your comments @dblandin

dblandin (approved):
Perfect. Thank you!

PR #5385, opened 2023-10-31
feat(searchCriteriaConnection): Add `represented` argument to field

damassi:
Trivial update to wire up the stitched argument `represented` through to GravityQL

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,7 +17,8 @@
   "parser": "@typescript-eslint/parser",
   "rules": {
     "sort-keys": [
-      "warn",
+      // TODO: Reenable this with "warn" after we auto-fix whole codebase
+      "off",
       "asc",
       { "caseSensitive": true, "natural": false, "minKeys": 2 }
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -17,7 +17,8 @@
   "parser": "@typescript-eslint/parser",
   "rules": {
     "sort-keys": [
-      "warn",
+      // TODO: Reenable this with "warn" after we auto-fix whole codebase
+      "off",
```
damassi, line 21: @egdbear - Lets revisit this! The combined issue of the state of MP + eslint + VSCode errors panel makes it so that the number of warnings completely drowns out real issues. Either we suffer the git blame hit by running an auto-fixer over the entire codebase, or we just try to advocate for a culture of sorting (where appropriate). Not sure if we can have both as MP is simply to old / historical / big to make big aesthetic changes to atm.

egdbear: Looks good to me!

PR #5387, opened 2023-11-01
fix: add improvements for recently viewed artwork ids queries

egdbear:
This PR is addressing the issues caused by recent changes for braze impersonations.

### DIFF
--- a/src/schema/v2/me/index.ts
+++ b/src/schema/v2/me/index.ts
@@ -550,7 +550,12 @@ export const meType = new GraphQLObjectType<any, ResolverContext>({
 
 const MeField: GraphQLFieldConfig<void, ResolverContext> = {
   type: meType,
-  resolve: (_root, _options, { userID, meLoader }, info) => {
+  resolve: (
+    _root,
+    _options,
+    { userID, meLoader, xImpersonateUserID },
+    info
+  ) => {
     if (!meLoader) return null
     const fieldsNotRequireLoader = [
       "id",
@@ -574,13 +579,11 @@ const MeField: GraphQLFieldConfig<void, ResolverContext> = {
       "identityVerification",
       "unreadNotificationsCount",
       "unseenNotificationsCount",
-      "similarToRecentlyViewedConnection",
       "showsByFollowedArtists",
       "newWorksFromGalleriesYouFollowConnection",
       "myCollectionInfo",
       "watchedLotConnection",
       "userInterestsConnection",
-      "recentlyViewedArtworksConnection",
       "myCollectionConnection",
       "artworkInquiriesConnection",
       "collectionsConnection",
@@ -588,6 +591,11 @@ const MeField: GraphQLFieldConfig<void, ResolverContext> = {
       "myCollectionAuctionResults",
       "newWorksByInterestingArtists",
     ]
+
+    if (xImpersonateUserID) {
+      return {}
+    }
+
     if (includesFieldsOtherThanSelectionSet(info, fieldsNotRequireLoader)) {
       return meLoader()
     }
--- a/src/schema/v2/me/recentlyViewedArtworks.ts
+++ b/src/schema/v2/me/recentlyViewedArtworks.ts
@@ -15,23 +15,20 @@ export const RecentlyViewedArtworks: GraphQLFieldConfig<
   args: pageable({}),
   description: "A list of the current user’s recently viewed artworks.",
   resolve: async (
-    { recently_viewed_artwork_ids },
+    _me,
     args,
-    { artworksLoader, recentlyViewedArtworkIdsLoader, xImpersonateUserID }
+    { artworksLoader, recentlyViewedArtworkIdsLoader, userID }
   ) => {
+    if (!userID) return null
+
     const { page, size, offset } = convertConnectionArgsToGravityArgs(args)
-
     try {
-      let recentlyViewedIds = recently_viewed_artwork_ids || []
+      const recentlyViewedArtworksBody =
+        (await recentlyViewedArtworkIdsLoader(userID))?.body || []
 
-      if (xImpersonateUserID) {
-        const recentlyViewedArtworksBody = (
-          await recentlyViewedArtworkIdsLoader(xImpersonateUserID)
-        )?.body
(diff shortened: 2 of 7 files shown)


mzikherman (changes requested):
I think this is better, but it doesn't seem to quite work for me. It looks like when used by Eigen, and _not_ requesting any other fields, something is skipped and the data isn't returned.
Screen Shot 2023-11-02 at 11 32 15 AM


Notice that when I add a `bio` selection, for example - the data does resolve.

Screen Shot 2023-11-02 at 11 32 58 AM

It's likely the skipping of the `me` fetch when _just_ `recentlyViewedArtworks` is being requested, which was added back in https://github.com/artsy/metaphysics/pull/5358

So I think this all still needs to be tightened up so that all the queries resolve as expected.

Maybe an alternative is to skip that `me` call when the `xImpersonateUserId` value is passed instead?

egdbear:
> Maybe an alternative is to skip that me call when the xImpersonateUserId value is passed instead?

How about we use this loader consistently, as long as we have the `userID` which is also referred to as the `xImpersonateUserId`?

mzikherman:
I don't think that would work b/c the "by id's" loader/endpoint _only_ allows Braze requests, so how would the non-Braze requests be able to use that loader? It probably only works in development right now because Gravity has that endpoint temporarily relaxed, which will be imminently reverted - https://github.com/artsy/gravity/pull/17016

src/schema/v2/me/index.ts
```diff
@@ -574,20 +579,23 @@ const MeField: GraphQLFieldConfig<void, ResolverContext> = {
       "identityVerification",
       "unreadNotificationsCount",
       "unseenNotificationsCount",
-      "similarToRecentlyViewedConnection",
```
mzikherman, line 577: Because we need to have fetched `me` in order to receive the `recently_viewed_artwork_ids` , we don't want to short-circuit for these fields.

src/schema/v2/me/index.ts
```diff
@@ -574,20 +579,23 @@ const MeField: GraphQLFieldConfig<void, ResolverContext> = {
...
-      "recentlyViewedArtworksConnection",
       "myCollectionConnection",
       "artworkInquiriesConnection",
       "collectionsConnection",
       "auctionResultsByFollowedArtists",
       "myCollectionAuctionResults",
       "newWorksByInterestingArtists",
     ]
+
+    if (xImpersonateUserID) {
```
mzikherman, line 595: But, for the recently/similar fields that we were short-circuiting above, we _do_ want to short-circuit when impersonation is happening, instead of hitting `meLoader` we'll fetch the id's directly in those resolvers.

src/schema/v2/me/recentlyViewedArtworks.ts
```diff
@@ -17,27 +17,35 @@ export const RecentlyViewedArtworks: GraphQLFieldConfig<
...
-      if (!recentlyViewedIds.length) {
-        const recentlyViewedArtworksBody = (
-          await recentlyViewedArtworkIdsLoader(userID)
-        )?.body
-        recentlyViewedIds = recentlyViewedArtworksBody || []
-      }
+      let artworkIDs
 
-      const pageArtworkIDs = recentlyViewedIds.slice(offset, offset + size)
+      // If `recently_viewed_artwork_ids` exists, use those
```
mzikherman, line 34: The approach here is:

  - if `recently_viewed_artwork_ids` exists (even if it's empty), then we already fetched it via `meLoader` and don't need to fetch anything further
  - otherwise, if impersonation is happening, only then can we fetch the id's specifically

The same approach is taken in the `similarToRecentlyViewed` resolver below

mzikherman (approved):

mzikherman:
The spec failure looks kinda weird and unrelated to this PR, although it did fail on CI several times. It passes locally.

I might still merge with the ❌ and see what happens when it's on `main`.

src/schema/v2/me/index.ts
egdbear: Cool makes sense!

egdbear:
> The spec failure looks kinda weird and unrelated to this PR, although it did fail on CI several times. It passes locally.
> 
> I might still merge with the ❌ and see what happens when it's on `main`.

I'm not to able merge it.

image


Can you?

egdbear:
>The spec failure looks kinda weird and unrelated to this PR, although it did fail on CI several times. It passes locally.

It doesn't pass locally if you use the latest LTS `v20.9.0` which has introduced this test failure.

mzikherman:
Got it, @egdbear in that case if you can repro locally try to fix it? I'm not sure why MP is on v20, shouldn't it be on v18?

Re: merging, looks like when it runs and is ❌ the merge button is greyed out, but if you retry, while it's running, the merge button becomes usable.

I'll trust you to take it over the finish line! And then once MP is deployed we can merge and deploy https://github.com/artsy/gravity/pull/17016

egdbear:
> Got it, @egdbear in that case if you can repro locally try to fix it? I'm not sure why MP is on v20, shouldn't it be on v18? We should probably be upgrading node versions specifically/carefully via separate ticket.
> 
> Re: merging, looks like when it runs and is ❌ the merge button is greyed out, but if you retry, while it's running, the merge button becomes usable.
> 
> I'll trust you to take it over the finish line! And then once MP is deployed we can merge and deploy [artsy/gravity#17016](https://github.com/artsy/gravity/pull/17016)

Here is [the fix](https://github.com/artsy/metaphysics/pull/5395)

mzikherman:
Ah cool! I merged that in so can rebase here to get this in.

egdbear:
> Got it, @egdbear in that case if you can repro locally try to fix it? I'm not sure why MP is on v20, shouldn't it be on v18? We should probably be upgrading node versions specifically/carefully via separate ticket.
> 
> Re: merging, looks like when it runs and is ❌ the merge button is greyed out, but if you retry, while it's running, the merge button becomes usable.
> 
> I'll trust you to take it over the finish line! And then once MP is deployed we can merge and deploy [artsy/gravity#17016](https://github.com/artsy/gravity/pull/17016)

It doesn't work for me  :'( 
image

PR #5389, opened 2023-11-01
chore(alerts): drop unsupported colors from alert labels

anandaroop:
We noticed an error condition when trying to render some alerts in Eigen/Force. This led to the addition of a missing filter color in https://github.com/artsy/metaphysics/pull/5384

That caused us to remember that there was _another_ piece of unfinished work in the big color update we implemented last year.

https://github.com/artsy/metaphysics/blob/6fb0305c0cfc26dd099f7376a018f1b0e9c04387/src/lib/colors.ts#L1-L2

**tldr;** We [added](https://github.com/artsy/metaphysics/pull/3762/commits/983cd6b83c6e183d71015f00919a4efd4720efec) these human-readable labels; then we updated the set of filterable colors in Gravity; then we [patched up](https://github.com/artsy/metaphysics/pull/3824/commits/891f25563b3fee64b30ed310c4c90a19a69cd049?diff=split&w=0) this list to handle both the old and new colors; but we never cleaned up support for the old colors in MP. 

This cleanup is basically safe to do now, since — 

<details>
<summary>
I confirmed that all published artworks reference only the new color palette
</summary>

colors
</details>

<details>
<summary>
I confirmed that only 3 SearchCriteria instances that refer to a deprecated color ("gold")
</summary>

<br>

Apparently we missed "gold" in [earlier](https://artsyproduct.atlassian.net/browse/FX-4012) [migrations](https://artsyproduct.atlassian.net/browse/FX-3721). I'd recommend re-mapping it to "yellow" in these three cases.

```rb
deprecated_values = old_values - new_values
# => ["violet", "gold", "darkviolet", "lightgreen", "lightblue", "darkblue", "darkorange", "darkgreen"]

SearchCriteria.where('cardinality(colors) > 0').select{ |sc| (sc.colors & deprecated_values).any? }.count
# => 3

SearchCriteria.where('cardinality(colors) > 0').select{ |sc| (sc.colors & deprecated_values).any? }.pluck(:colors)
# => [["gold"], ["gold", "orange", "purple"], ["blue", "gold", "green", "orange", "pink", "red", "yellow"]]

```
</details>

## Migration

Related to the second collapsible point above —

```rb
SearchCriteria.where('cardinality(colors) > 0').select{ |sc| sc.colors.include? "gold" }.each do |sc|
  pp ["before", sc.colors]
  sc.colors.delete("gold")
  sc.colors << "yellow"
  sc.save
  pp ["after", sc.colors]
end
```


dblandin (approved):
Awesome! Thanks for following up on this 💯

LGTM 👍🏼

MounirDhahri (approved):

MounirDhahri:
Thanks for following up on this

gkartalis (approved):

PR #5416, opened 2023-11-14
feat: introduce mock structure for suggested filters

MounirDhahri:
Resolves [ONYX-515]

This PR brings a mock structure for the upcoming suggested filters work. We are starting here by returning a mock then we will introduce the logic in the upcoming tickets.

Screenshot 2023-11-14 at 09 52 49

[ONYX-515]: https://artsyproduct.atlassian.net/browse/ONYX-515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/previewSavedSearch.ts
+++ b/src/schema/v2/previewSavedSearch.ts
@@ -93,6 +93,14 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
         "URL for a user to view the artwork grid with applied filters matching saved search criteria attributes",
       resolve: resolveHref,
     },
+    suggestedFilters: {
+      description:
+        "Suggested filters for the user to use based on their search criteria",
+      type: new GraphQLList(new GraphQLNonNull(SearchCriteriaLabel)),
+      resolve: (_) => {
+        return mockSuggestedFilters
+      },
+    },
   }),
 })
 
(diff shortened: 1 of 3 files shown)


src/schema/v2/previewSavedSearch.ts
```diff
@@ -93,6 +93,14 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
         "URL for a user to view the artwork grid with applied filters matching saved search criteria attributes",
       resolve: resolveHref,
     },
+    suggestedFilters: {
+      description:
+        "Suggested filters for the user to use based on their search criteria",
+      type: new GraphQLList(new GraphQLNonNull(SearchCriteriaLabel)),
```
MounirDhahri, line 99: note: I am not sure here if we should allow nulls - I will get to this later when we implement the logic and have a better understanding of what we are going to be returning.

```suggestion
      type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(SearchCriteriaLabel))),
```

dblandin: I think we should always be able to return an empty array or non-null suggestions here.

dblandin (dismissed):
LGTM! - I would make this field non-nullable though 👍

src/schema/v2/previewSavedSearch.ts
MounirDhahri: Cool - I will do it now then

PR #5418, opened 2023-11-14
feat(ONYX-520): add rarity as a suggested filter

MounirDhahri:
Resolves [ONYX-520]

This PR adds rarity as a suggested filter. We are getting the most popular rarity for artists based on aggregations. 

**Open questions:**
1. The above logic considers "unique" and "limited edition" as rarities with the same weight. I do however believe that they're not the same and users might be interested in unique works more if both counts are close. I would suggest here to increase the weight of unique artworks to cover for that. Open to other ideas -> Answer: no need for that now

2. Should we consider "Uknown edition" as a valid rarity suggestion? -> Answer: let's remove it

Both questions answered ✅ 

[ONYX-519]: https://artsyproduct.atlassian.net/browse/ONYX-519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ONYX-520]: https://artsyproduct.atlassian.net/browse/ONYX-520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/previewSavedSearch.ts
+++ b/src/schema/v2/previewSavedSearch.ts
@@ -103,12 +103,8 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
       ),
       resolve: async (_root, _args, { filterArtworksLoader }) => {
         const suggestedFiltersByArtist: SearchCriteriaLabel[][] = await Promise.all(
-          _root.artistIDs.map(
-            async (artistSlug) =>
-              await getSuggestedFiltersByArtistSlug(
-                artistSlug,
-                filterArtworksLoader
-              )
+          _root.artistIDs.map((artistSlug) =>
+            getSuggestedFiltersByArtistSlug(artistSlug, filterArtworksLoader)
           )
         )
 
--- a/src/schema/v2/previewSavedSearch.ts
+++ b/src/schema/v2/previewSavedSearch.ts
@@ -8,14 +8,16 @@ import {
   GraphQLObjectType,
   GraphQLString,
 } from "graphql"
+import attributionClasses from "lib/attributionClasses"
+import { snakeCaseKeys } from "lib/helpers"
+import _ from "lodash"
+import { stringify } from "qs"
 import { ResolverContext } from "types/graphql"
 import ArtworkSizes from "./artwork/artworkSizes"
 import {
-  resolveSearchCriteriaLabels,
   SearchCriteriaLabel,
+  resolveSearchCriteriaLabels,
 } from "./searchCriteriaLabel"
-import { snakeCaseKeys } from "lib/helpers"
-import { stringify } from "qs"
 
 const previewSavedSearchArgs: GraphQLFieldConfigArgumentMap = {
   acquireable: {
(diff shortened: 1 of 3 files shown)


src/schema/v2/previewSavedSearch.ts
```diff
@@ -99,13 +101,72 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
...
-      resolve: (_) => {
-        return mockSuggestedFilters
+      resolve: async (_root, _args, { filterArtworksLoader }) => {
+        const suggestedFiltersByArtist: SearchCriteriaLabel[][] = await Promise.all(
+          _root.artistIDs.map(
+            async (artistSlug) =>
+              await getSuggestedFiltersByArtistSlug(
+                artistSlug,
+                filterArtworksLoader
+              )
```
olerichter00, line 111: No need to await the result if it's returned immediately.

```suggestion
           (artistSlug) =>
              getSuggestedFiltersByArtistSlug(
                artistSlug,
                filterArtworksLoader
              )
```

src/schema/v2/previewSavedSearch.ts
```diff
@@ -99,13 +101,72 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
...
+}): string => {
+  return Object.values(aggregation).reduce((acc, curr) => {
+    return curr.count > acc.count ? curr : acc
+  }).name
+}
+
+const getRaritySearchCriteriaLabel = (
+  name: string | undefined
+): SearchCriteriaLabel => {
+  if (!name) {
```
olerichter00, line 135: Is this the case if there are no published artworks for the artist with any rarity associated?

src/schema/v2/previewSavedSearch.ts
```diff
@@ -99,13 +101,72 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
...
+                artistSlug,
+                filterArtworksLoader
+              )
+          )
+        )
+
+        return _.chain(suggestedFiltersByArtist)
+          .flatten()
+          .uniqBy((searchCriteria) => JSON.stringify(searchCriteria))
+          .value()
```
olerichter00, line 118: Magic 🪄

olerichter00 (dismissed):
Looks good to me 🌟

src/schema/v2/previewSavedSearch.ts
MounirDhahri: good point, i am doing it within the function already

src/schema/v2/previewSavedSearch.ts
MounirDhahri: this is actually for the case where aggregations is empty - I can work around it maybe to make sure we don't get here

olerichter00: Not sure if there could be other cases where aggregations are empty, but anyway, it's probably a good idea to return a default in this case 👍

MounirDhahri:
Will merge this - glad to make changes if needed

src/schema/v2/previewSavedSearch.ts
```diff
@@ -99,13 +101,70 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
...
       },
     },
   }),
 })
 
+const getMostPopularField = (aggregation: {
```
anandaroop, line 120: **suggestion:** The "field" naming here might be unintentionally confusing. I'd opt for "most popular value" or "most popular option" instead.

src/schema/v2/previewSavedSearch.ts
```diff
@@ -99,13 +101,70 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
...
+  artistSlug: string,
+  filterArtworksLoader
+): Promise<SearchCriteriaLabel[]> => {
+  const gravityArgs = {
+    published: true,
+    aggregations: ["attribution_class"],
+    artist_id: artistSlug,
+  }
+
+  const { aggregations } = await filterArtworksLoader(gravityArgs)
```
anandaroop, line 153: **suggestion:** In cases where are interested _only_ in the aggregations but _not_ in the artwork results, we can also, specify `size: 0` to reduce the speed/size of the response from gravity. 

Example: https://github.com/artsy/metaphysics/blob/7cdcd8ba40a28bdf39c4dcc310be6f19439634de/src/schema/v2/artists/curatedTrending.ts#L11-L17

anandaroop (commented):
Apologies for the late feedback!

No blockers here, can be addressed as follow-ups…

MounirDhahri:
thanks for the review @anandaroop - I will create a PR with the fixes

src/schema/v2/previewSavedSearch.ts
```diff
@@ -99,13 +101,70 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
       type: new GraphQLNonNull(
         new GraphQLList(new GraphQLNonNull(SearchCriteriaLabel))
       ),
-      resolve: (_) => {
-        return mockSuggestedFilters
+      resolve: async (_root, _args, { filterArtworksLoader }) => {
+        const suggestedFiltersByArtist: SearchCriteriaLabel[][] = await Promise.all(
+          _root.artistIDs.map((artistSlug) =>
+            getSuggestedFiltersByArtistSlug(artistSlug, filterArtworksLoader)
+          )
```
dblandin, line 108: **suggestion**: Instead of iterating over artists and determining suggested filters independently for each, what do you think of executing a single query via the `filterArtworksLoader` to fetch the aggregations we need across both sets of inventory? I'm thinking of something like:

```ts
const gravityArgs = {
  published: true,
  aggregations: ["attribution_class", "medium"],
  artist_ids: artistIDs,
}

const { aggregations } = await filterArtworksLoader(gravityArgs)
```

dblandin (commented):
Looking good! One suggestion for how we fetch the data needed for suggested filters. I don't think we need independent suggestions for each artist associated with the criteria but suggestions for the criteria overall.

PR #5419, opened 2023-11-14
feat(ONYX-519): add medium suggested filter

MounirDhahri:
Resolves [ONYX-519]

This PR adds medium as a suggested filter type. I also realized adds some sorting to make sure filters from the same type are aligned together. 

Screenshot 2023-11-14 at 16 41 42


olerichter00 (dismissed):

dariakoko (dismissed):

src/schema/v2/previewSavedSearch.ts
```diff
@@ -108,27 +109,42 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
...
+  data: { name: string; count: number }
+} => {
+  const fieldValueArray = Object.entries(aggregation).reduce((acc, curr) => {
+    return curr[1].count > acc[1].count ? curr : acc
+  })
+
+  return {
+    value: fieldValueArray[0],
+    data: fieldValueArray[1],
+  }
```
MounirDhahri, line 141: I needed access to the entire object here because the name was no longer enough

src/schema/v2/previewSavedSearch.ts
```diff
@@ -137,30 +153,65 @@ const getRaritySearchCriteriaLabel = (
...
+    return null
+  }
+
+  const mediumData = Object.values(artworkMediums).find((artworkMedium) => {
+    return artworkMedium.mediumFilterGeneSlug === value
+  })
+
+  if (!mediumData || !mediumData.mediumFilterGeneSlug) {
+    return null
+  }
```
MounirDhahri, line 176: this is in case we have a medium in gravity that doesn't match metaphysics

src/schema/v2/previewSavedSearch.ts
```diff
@@ -137,30 +153,65 @@ const getRaritySearchCriteriaLabel = (
...
+
+  return {
+    displayValue: mediumData.name,
+    field: "additionalGeneIDs",
```
MounirDhahri, line 180: we call medium `additionalGeneIdDs` in our client apps therefore the naming here

olerichter00 (approved):

PR #5420, opened 2023-11-14
chore: docker development changes

ovasdi:
Supports https://artsyproduct.atlassian.net/browse/PHIRE-293

This tweaks the `.env.development` to use docker service name when connecting to gravity and also increments the Memcached port exposed on the host to prevent port collision when `gravity-memcached` service is already running as part of gravity local dev stack.

If needed to use `localhost`, those overrides can go into `.env`.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,6 +1,4 @@
-GRAVITY_API_BASE=http://localhost:3000/api/v1
-GRAVITY_GRAPHQL_ENDPOINT=http://localhost:3000/api
-GRAVITY_API_URL=http://localhost:3000
-#GRAVITY_ID=
-#GRAVITY_SECRET=
-MEMCACHED_URL=localhost:11211
+GRAVITY_API_BASE=http://gravity:3000/api/v1
+GRAVITY_GRAPHQL_ENDPOINT=http://gravity:3000/api
+GRAVITY_API_URL=http://gravity:3000
+MEMCACHED_URL=metaphysics-memcached:11211
--- a/.env.development
+++ b/.env.development
@@ -1,4 +1,6 @@
 GRAVITY_API_BASE=http://gravity:3000/api/v1
 GRAVITY_GRAPHQL_ENDPOINT=http://gravity:3000/api
 GRAVITY_API_URL=http://gravity:3000
+#GRAVITY_ID=metaphysics-client-id
+#GRAVITY_SECRET=metaphysics-client-secret
 MEMCACHED_URL=metaphysics-memcached:11211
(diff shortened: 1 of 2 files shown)


hokusai/development.yml
```diff
@@ -4,8 +4,8 @@ services:
 {% include 'templates/docker-compose-service-dev.yml.j2' %}
     command: ["yarn", "start"]
     env_file:
-      - ../.env.development
       - ../.env.shared
+      - ../.env.development
```
ovasdi, line 8: By my understanding we want to make this override `.env*`, correct? But not `.env`

mc-jones: My original thinking is it would be nice if it also overrides `.env` so we always know whatever is in that file will be applied to local development. But I don't think in the repos I've seen so far it would matter too much, as `.env` is pretty empty.

hokusai/development.yml
```diff
@@ -16,4 +16,4 @@ services:
   metaphysics-memcached:
     image: memcached:1.4.34-alpine
     ports:
-      - 11211:11211
+      - 11212:11211
```
mc-jones, line 19: Lets remove the port. There is no need to expose it to the host. And without it we don't have to worry about conflicts with `brew` managed services. 

LMK if I'm forgetting something, though.

.env.development
```diff
@@ -1,6 +1,4 @@
-GRAVITY_API_BASE=http://localhost:3000/api/v1
-GRAVITY_GRAPHQL_ENDPOINT=http://localhost:3000/api
-GRAVITY_API_URL=http://localhost:3000
-#GRAVITY_ID=
-#GRAVITY_SECRET=
```
mc-jones, line 5: Maybe we should leave these in. I'm realizing in situations where they don't want to sync data, we will need to include the values populated by `gravity`'s seed data.

hokusai/development.yml
ovasdi: I vote to keep `.env` as is because it's for personal overrides and that applies to any environment. As you say, it wouldn't matter much since those files tend to be empty but if someone needs to override a development var, they can use `.env`.

hokusai/development.yml
ovasdi: Good call. Done in https://github.com/artsy/metaphysics/pull/5420/commits/0431e7f4166b07921c9a0f4d38a1b6ba493b2dd9

hokusai/development.yml
mc-jones: fair point. Agreed.

.env.development
ovasdi: Another good callout. Added in https://github.com/artsy/metaphysics/pull/5420/commits/a32ffe82aedc9a599508ef4ab767db6062e81149

I assume we want those commented out by default?

mc-jones: Yeah, I think so. Might be worth a note in the readme.

PR #5427, opened 2023-11-16
chore: address review comments on rarity suggested filters

MounirDhahri:
This PR addresses review comments on https://github.com/artsy/metaphysics/pull/5418 

Thanks @anandaroop for the thorough review.

I added a check for the `artistIDs` since we need them for suggestions. Let me know what you think about that

Screenshot 2023-11-16 at 11 02 21

_Making sure I didn't break anything_

### DIFF
--- a/src/schema/v2/previewSavedSearch.ts
+++ b/src/schema/v2/previewSavedSearch.ts
@@ -74,6 +74,8 @@ const previewSavedSearchArgs: GraphQLFieldConfigArgumentMap = {
   },
 }
 
+const SUPPORTED_ATTRIBUTIONS_CLASSES = ["unique", "limited edition"]
+
 const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
   name: "PreviewSavedSearch",
   fields: () => ({
@@ -121,7 +123,7 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
           getMostPopularOption(aggregations["attribution_class"]).value
         )
 
-        if (rarity) {
+        if (rarity && SUPPORTED_ATTRIBUTIONS_CLASSES.includes(rarity.value)) {
           suggestedFilters.push(rarity)
         }
 
@@ -155,12 +157,10 @@ const getMostPopularOption = (aggregation: {
   }
 }
 
-const SUPPORTED_ATTRIBUTIONS_CLASSES = ["unique", "limited edition"]
-
 const getRaritySearchCriteriaLabel = (
   value: string | undefined
 ): SearchCriteriaLabel | null => {
-  if (!value || !SUPPORTED_ATTRIBUTIONS_CLASSES.includes(value)) {
+  if (!value) {
     return null
   }
 
--- a/src/schema/v2/previewSavedSearch.ts
+++ b/src/schema/v2/previewSavedSearch.ts
@@ -102,9 +102,9 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
       type: new GraphQLNonNull(
         new GraphQLList(new GraphQLNonNull(SearchCriteriaLabel))
       ),
-      resolve: async (_root, _args, { filterArtworksLoader }) => {
+      resolve: async ({ artistIDs }, _args, { filterArtworksLoader }) => {
         const suggestedFiltersByArtist: SearchCriteriaLabel[][] = await Promise.all(
-          _root.artistIDs.map((artistSlug) =>
+          artistIDs.map((artistSlug) =>
             getSuggestedFiltersByArtistSlug(artistSlug, filterArtworksLoader)
           )
         )
@@ -125,7 +125,7 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
   }),
 })
 
-const getMostPopularField = (aggregation: {
+const getMostPopularOption = (aggregation: {
   [key: string]: { name: string; count: number }
 }): {
   value: string
@@ -184,6 +184,7 @@ const getSuggestedFiltersByArtistSlug = async (
   filterArtworksLoader
 ): Promise<SearchCriteriaLabel[]> => {
   const gravityArgs = {
+    size: 0,
     published: true,
     aggregations: ["attribution_class", "medium"],
     artist_id: artistSlug,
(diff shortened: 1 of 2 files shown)


src/schema/v2/previewSavedSearch.ts
```diff
@@ -102,9 +102,13 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
...
+      resolve: async ({ artistIDs }, _args, { filterArtworksLoader }) => {
+        if (!artistIDs) {
+          throw new Error("artistIDs are required to get suggested filters")
+        }
+
         const suggestedFiltersByArtist: SearchCriteriaLabel[][] = await Promise.all(
-          _root.artistIDs.map((artistSlug) =>
+          artistIDs.map((artistSlug) =>
             getSuggestedFiltersByArtistSlug(artistSlug, filterArtworksLoader)
           )
```
anandaroop, line 113: I think the same comment applies here as @dblandin [noted already](https://github.com/artsy/metaphysics/pull/5418#discussion_r1395862549), sorry I missed that the first time.

anandaroop (dismissed):
Looks great! 🙏🏽 

Just adding a +1 to another review comment below. Feel free to merge this and follow up, or to go ahead with that last change here as well.

src/schema/v2/previewSavedSearch.ts
MounirDhahri: nice idea - didn't know that's possible. I will do that

MounirDhahri:
@dblandin @anandaroop I was expecting the aggregations response when specifying an array of `artist_slugs` to be an array of aggregations but it was just a single aggregations object with the specified fields.
Screenshot 2023-11-16 at 17 23 33

This seems to be represent the sum for both artists. Is this still what we want to do? The tradeoff will be less suggested filters but that's not so important here since most users won't be setting alerts for two artists anyway.

dblandin:
> This seems to be represent the sum for both artists. Is this still what we want to do?

Yes, exactly. That's what we want to do for this iteration — generate a list of suggested filters based on the combined query, representing the combined inventory.

For example, using [this API request](https://api.artsy.net/api/v1/filter/artworks?size=0&artist_ids[]=4d8b92b34eb68a1b2c0003f4&artist_ids[]=4d8b926a4eb68a1b2c0000ae&aggregations[]=attribution_class&aggregations[]=medium) combining Andy Warhol and Damien Hirst we get:

```json
{
  "hits":[],
  "aggregations":{
    "medium":{
      "prints":{
        "name":"Prints",
        "count":8394
      }
    },
    "attribution_class":{
      "limited edition":{
        "name":"limited edition",
        "count":4732
      }
    }
  }
}
```

so we'd return `Medium Type: Prints` and `Rarity: Limited Edition` as the suggested filters. For the response in your screenshot, I'm assuming that "Open Edition" is the most common rarity and according to our [initial rubric (Slack)](https://artsy.slack.com/archives/C05EQL4R5N0/p1700042617043859?thread_ts=1700041955.767889&cid=C05EQL4R5N0), we'd only return `Medium Type: Prints` as a suggested filter.

src/schema/v2/previewSavedSearch.ts
```diff
@@ -74,6 +74,8 @@ const previewSavedSearchArgs: GraphQLFieldConfigArgumentMap = {
   },
 }
 
+const SUPPORTED_ATTRIBUTIONS_CLASSES = ["unique", "limited edition"]
```
dblandin, line 77: **suggestion**: Maybe `ALLOWED_RARITY_SUGGESTIONS` might be more intention-revealing?

src/schema/v2/previewSavedSearch.ts
```diff
@@ -102,30 +103,45 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
...
-            .groupBy((searchCriteria) => searchCriteria.field)
-            .values()
-            .flatten()
-            .value()
+        if (rarity && SUPPORTED_ATTRIBUTIONS_CLASSES.includes(rarity.value)) {
+          suggestedFilters.push(rarity)
+        }
+
+        const medium = getMediumSearchCriteriaLabel(
+          getMostPopularOption(aggregations["medium"]).value
```
dblandin, line 131: **suggestion**: Given that the aggregations are returned from the API pre-sorted by popularity, we might be able to simplify this further by simply plucking out the first entry if present. It's also possible for this aggregation to be empty so we should guard for that as well.

```ts
const medium = getMediumSearchCriteriaLabel(
  getMostPopularOption(aggregations["medium"])
)

const getMostPopularOption = (aggregation: {
  [key: string]: { name: string; count: number }
}): string | undefined => {
  if (isExisty(aggregation)) {
    const key = Object.keys(aggregation)[0]
    return aggregation[key].value
  }
}
```

dblandin (commented):
Looking good! Left a couple minor comments.

src/schema/v2/previewSavedSearch.ts
MounirDhahri: makes sense. I will make the change

src/schema/v2/previewSavedSearch.ts
MounirDhahri: good idea - do we always sort by popularity in all contexts?

dblandin: I believe so — yes, that's just how Elasticsearch aggregation are returned but it's up to the client whether to simply echo that sort order or do something different like Alphabetical sorting.

src/schema/v2/previewSavedSearch.ts
```diff
@@ -102,49 +103,57 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
...
-            getSuggestedFiltersByArtistSlug(artistSlug, filterArtworksLoader)
-          )
+      resolve: async ({ artistIDs }, _args, { filterArtworksLoader }) => {
+        if (!artistIDs) {
+          throw new Error("artistIDs are required to get suggested filters")
+        }
+
+        const gravityArgs = {
+          size: 0,
+          published: true,
```
dblandin, line 113: **question**: I don't see `published` as a supported query param on the [API endpoint](https://github.com/artsy/gravity/blob/main/app/api/v1/filter_endpoint.rb#L36-L37). Did you find that you needed this? I would expect that, by default, only published artworks are queried for.

dblandin (dismissed):
One minor question but LGTM overall 👍🏼

PR #5428, opened 2023-11-16
feat: mutation to create a partner offer

narikazu:
# Summary

The PR creates a new mutation to create a partner offer by hitting [the gravity RESTful endpoint](https://github.com/artsy/gravity/pull/17047). 

# Ticket
https://artsyproduct.atlassian.net/browse/EMI-1537


_schemaV2.graphql
```diff
@@ -14317,6 +14348,36 @@ type PartnerMeta {
   title: String
 }
 
+type PartnerOffer {
+  artworkId: String
+  createdAt(
```
iskounen, line 14353: **question:** I assumed that this field would be automatically set by ActiveRecord when the `partner_offer` is saved to the database. Is there a reason to include this as a configurable parameter?

leamotta (approved):
LGTM!

_schemaV2.graphql
narikazu: Thank you for reviewing the PR, @iskounen. This type is for the response, not for the input. You can see the request parameters that you can set [here](https://github.com/artsy/metaphysics/pull/5428/files#diff-3a304b8e0fa043b030a2e5b3040547c6e4008ebebed2a9207ba67d27dc18192dR7820-R7825).

PR #5432, opened 2023-11-20
allow passing preferred metric to previewSavedSearch

nickskalkin:
Take a look at this quick demo of the issue:

https://github.com/artsy/metaphysics/assets/3934579/e18c2ead-d541-4fe3-9e50-309e03dcfe34

Currently metaphysics uses only preferred units for displaying pills. I find it weird that when user specifies inches or centimetres explicitly - pils are displayed in other metric units.

I suggest passing `metic` argument to the `previewSavedSearch` query. When metric is not specified - metaphysics fallbacks to the preferred units.

```graphql
query MyQuery {
  previewSavedSearch(attributes: { width: "40-100" }, metric: "in") {
    displayName
    labels {
      displayValue # will be "w: 40–100 in"
    }
  }
}
```

```graphql
query MyQuery {
  previewSavedSearch(attributes: { width: "40-100" }, metric: "cm") {
    displayName
    labels {
      displayValue # will be "w: 102–254 cm"
    }
  }
}
```

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -15902,6 +15902,7 @@ type Query {
   # A previewed saved search
   previewSavedSearch(
     attributes: PreviewSavedSearchAttributes
+    metric: String
   ): PreviewSavedSearch
 
   # Get all price insights for an artist.
@@ -20556,6 +20557,7 @@ type Viewer {
   # A previewed saved search
   previewSavedSearch(
     attributes: PreviewSavedSearchAttributes
+    metric: String
   ): PreviewSavedSearch
 
   # A Profile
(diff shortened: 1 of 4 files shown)


olerichter00 (commented):
Good idea! 👍 

I wonder if we should go even further and update the user's preferred metric in Gravity when the user switches the metric in the filter panel. We do this in the app's "My Collection" artwork form (src/app/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx).

My assumption is that the main reason to change the metric is to adjust the filters to the user's preferred metric. I don't think that there are many people using both metrics.

dblandin:
> I wonder if we should go even further and update the user's preferred metric in Gravity when the user switches the metric in the filter panel.

I was thinking something similar. If we were to update the preferred metric via MP whenever a user switches between (cm) and (in), then the rest of the UI flow should match without us needing to thread this preference in to other API surfaces.

Regardless of our approach here — this is still relevant at a broader level so we should keep chatting with design.

nickskalkin:
Interesting, thanks for feedback both! Going to close this PR for now and re-open if needed. I'll ask Barney what he thinks

PR #5438, opened 2023-11-21
feat(EMI-1585): discount percentage instead of price

leamotta:
https://artsyproduct.atlassian.net/browse/EMI-1585

Replace partner offers fixed value with discount percentages


narikazu (approved):
LGTM

PR #5442, opened 2023-11-27
fix: previewSavedSearch returns artist series slugs instead of IDs

nickskalkin:
Frontends works with `slugs` instead of `IDs`: when a user filters artworks grid by artist series - `slug` is being added to the filters / alert context. `previewSavedSearch` returns `IDs`, and because of this the logic to remove a pill doesn't work.

### DIFF
--- a/src/schema/v2/__tests__/searchCriteriaLabel.test.ts
+++ b/src/schema/v2/__tests__/searchCriteriaLabel.test.ts
@@ -664,16 +664,16 @@ describe("resolveSearchCriteriaLabels", () => {
         .mockReturnValueOnce(
           Promise.resolve({
             artistSeries: {
-              internalID: "abc-123",
               title: "Astroboy",
+              slug: "kaws-astroboy",
             },
           })
         )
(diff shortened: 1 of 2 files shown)


dblandin (changes requested):
One pattern that I see in other label resolvers, which I think is a good one, is to always match the input arg (whether ID or slug) in the output.

For example, if `artistSeriesIDs` is ['slug1', 'slug2'] then those slugs should show up in the output as `value`. Also, if `artistSeriesIDs` is ['id1', 'id2'] then those internal IDs should show up in the output as `value`. That way we'll always match what the frontend currently holds in its alert context and we don't have to decide here which is most "correct" to return:

```
artistSeriesIDs.map(async (id) => {
  ...

  return {
    name: "Artist Series",
    displayValue: title,
    field: "artistSeriesIDs",
    value: id,
  }
}
```

nickskalkin:
@dblandin, thanks, re-implemented following your suggestion!

dblandin (approved):
Nice. LGTM!

PR #5457, opened 2023-12-01
chore: hide same type suggested filter attributes

MounirDhahri:
This PR addresses some feedback from Sam about hiding the suggested filter attributes that are already set when the modal opens. 

https://github.com/artsy/metaphysics/assets/11945712/d226b2ae-b565-4632-be2b-ba3b32b46835


MounirDhahri:
I couldn't test this on my simulator so please don't merge this.

anandaroop (approved):
Nice 👌🏽

dblandin (approved):
Awesome! Really nice to see how easy this is given the field inputs 👍🏼

Let's sync up with Barney as this will yield even more cases where "Suggested Filters" is seemingly empty, with only a "More filters" CTA — something we've raised in the past as a potential concern.

PR #5471, opened 2023-12-13
feat(alerts): include artist series in suggested filters

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-603

This PR mainly adds this —

- `previewSavedSearch.suggestedFilters` now returns artist series suggestions from the artist aggregation (9d8c6a093da85d7571ea3a20e10fbea68e6129da)

-  also allows for `suggestedFilters` to take an optional new argument `source: { type: ARTWORK, id: "abc123" }` (8be328ee721f5fa1e3d2bf979fa22035f5d828e3)

 - when that^ is supplied it derives artist series suggestions from the supplied _artwork_, instead of from artist aggregations (9086cbdb2253d6ff3e8431d13ad35de8aa100d71)

---

In order to tackle that I decided to refactor the `previewSavedSearch` code a bit, as it was growing unwieldy. (This is mostly busywork, feel free to skip and review the main commits highlighted above ☝🏽) —

- extracted `previewSavedSearch.ts` into a separate directory (1794eb7336268c46a1b21e0558326dc7a55c0ff4)
- extracted a bunch of methods into separate files (1f2034d579be29d84eb0ab15a0e1365fc95718f5, 932c514cfc291b895ba859b82ff392dd7f3a25b3, 2b0c7a72b0598ffdaa21830466e1b20d333493b8)

---

### Without the new `source` argument

It sources suggestions from the artist (e.g. Toys, Companions)

without

### With the new `source` argument referring to an artwork

It sources suggestions from the artwork (e.g. SpongeBob)

with


.vscode/settings.json
```diff
@@ -43,6 +43,6 @@
   "typescript.surveys.enabled": false,
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
damassi, line 46: What does this do?

MounirDhahri: this is coming from VSCode latest version https://code.visualstudio.com/updates/v1_85#_code-actions-on-save-and-auto. It's the same as `true`, it should be safe to commit it

Screenshot 2023-12-13 at 10 09 02

src/schema/v2/previewSavedSearch/previewSavedSearch.ts
```diff
@@ -0,0 +1,160 @@
...
+      description:
+        "Suggested filters for the user to use based on their search criteria",
+      type: new GraphQLNonNull(
+        new GraphQLList(new GraphQLNonNull(SearchCriteriaLabel))
+      ),
+      args: {
+        source: {
+          type: AlertSourceType,
+          description: "The context from which the alert originates",
+        },
```
MounirDhahri, line 109: Should we maybe default this to `Artist`?

src/schema/v2/previewSavedSearch/previewSavedSearch.ts
```diff
@@ -0,0 +1,160 @@
...
+  fields: {
+    type: {
+      type: AlertSourceTypeEnum,
+      description: "The type of object from which the alert originates",
+    },
+    id: {
+      type: GraphQLID,
+      description:
+        "The database id of the object from which the alert originates",
+    },
```
MounirDhahri, line 141: **Q:** Would this work with a slug as well?

MounirDhahri: maybe it doesn't matter since we will anyway have the id by the time the UI is ready

MounirDhahri: okey checked the code below and I believe the answer is yes. I think we don't need an extra field here for slug since we do this in most areas.

MounirDhahri (approved):

src/schema/v2/previewSavedSearch/suggestedFilters.ts
```diff
@@ -0,0 +1,198 @@
...
+    artistSeriesOptions = await getArtistSeriesSearchCriteriaLabelsFromArtwork(
+      source,
+      gravityGraphQLLoader
+    )
+  } else {
+    artistSeriesOptions = getArtistSeriesSearchCriteriaLabelsFromArtist(
+      aggregations["artist_series"],
+      MAX_SUGGESTIONS
+    )
+  }
```
dblandin, line 53: **question**: Just to make sure I'm understanding this correctly — when creating an alert from an artwork source we return up to 5 artist series suggestions and when creating an alert from an artist source we return up to 2 artist series suggestion. Is that right?

Also, something that we've discussed but we should document/align on is what precedence these suggestions have over the existing medium+rarity suggestions. I think you made a good point recently that when creating an alert from an artwork source, the artist series suggestions are more relevant as they relate directly to the source artwork and not general inventory. So I might recommend something like this in terms of precedence and limit:

- Artwork source: `<artist series| max 5>, <medium | max 2>, <rarity | max 2>`
- Artist source: `<artist series | max >, <medium | max 2>, <rarity | max 2>`

dblandin (approved):
Most of the diff seems to be refactoring so I zoomed in to last couple commits which look good to me! Left one non-blocking clarifying comment.

.vscode/settings.json
anandaroop: Good catch! I noticed and un-committed this a couple of times but it still snuck through 🙈  

I hadn't looked into the cause yet, thanks @MounirDhahri. Sounds like it's okay to actually leave this in.

src/schema/v2/previewSavedSearch/previewSavedSearch.ts
anandaroop: Based on some recent team conversations I decided to be more explicit about the `id` field expecting an actual id. (It may of course still work with slugs, as is often the case in Gravity)

src/schema/v2/previewSavedSearch/suggestedFilters.ts
anandaroop: Yeah, these numbers are certainly debatable, but good idea to clarify the reasoning here:

- Artist source: up to 2 artist series suggestions, in order to be consistent with the existing pattern for medium and rarity

- Artwork source: up to 5 artist series suggestions, based on (a) the assumption that these will be highly relevant to the work in question, and (b) analysis of existing artist series inventory stats (see [here](https://github.com/artsy/force/pull/13244#discussion_r1419289315) — tldr  95% of works in artist series belong to no more than 2 series;  99.98% belong to no more than 5 series)

And yes, I should have noted that I _prepend_ the artist series suggestions to the list of all suggestions in this PR, based on that same assumption that these will be highly relevant to the artwork in question (though perhaps less so when based on overall artist aggregations.)

src/schema/v2/previewSavedSearch/previewSavedSearch.ts
MounirDhahri: perfect

src/schema/v2/previewSavedSearch/previewSavedSearch.ts
anandaroop: I think this can be left to the client, esp since the default behavior (sourcing the suggestions from the artist aggregation) will always be in effect when the Artwork source is _not_ supplied.

src/schema/v2/previewSavedSearch/suggestedFilters.ts
dblandin: Thanks! I started [a thread here in Slack](https://artsy.slack.com/archives/C05EQL4R5N0/p1702492997351529) to continue aligning on this.

anandaroop:
Going to merge this to be able to make some more FE progress, but happy to make updates in light of the active discussions.

PR #5482, opened 2023-12-20
chore(onyx-503): Clean up unused code paths for TinEye / reverse image search

nickskalkin:
[Ticket](https://artsyproduct.atlassian.net/browse/ONYX-503)

We’ve already retired TinEye functionality from Gravity and clients, so it should be safe to proceed with MP.


anandaroop (approved):
Nice and thorough 🧹 

(And a little [ironic](https://www.notion.so/artsy/7e929ff6c67d472481c9fb79b3a22f9b?v=df2710e734f04d4088d3832dc19364b2&p=4b549462189d42d992e2453d56488a62&pm=s), but still the right call for now I think!)

damassi (commented):
Thank you for the cleanup!

PR #5486, opened 2024-01-03
feat(EMI-1589): expose Artwork#price_listed field as Float

fladson:
https://artsyproduct.atlassian.net/browse/EMI-1589

When an Artwork has the price hidden, the `listPrice` field is not reliable, so we need to expose the actual `price_listed`.

This field will be used for discount calculations, so we need it as Float, for displaying the price, we already have `internalDisplayPrice` doing the formatting.

Example where the artwork has the price hidden:
```
{
  "node": {
    "internalDisplayPrice": "$123",
    "priceListed": 123,
    "listPrice": null
  }
}
```

Note: you may need to set `"shallow": false` for it to appear on certain endpoints.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2458,7 +2458,7 @@ type Artwork implements Node & Searchable & Sellable {
   priceCurrency: String
   priceIncludesTax: Boolean
   priceIncludesTaxDisplay: String
-  priceListed: Money
+  priceListed: Float
 
   # The price paid for the artwork in a user's 'my collection'
   pricePaid: Money
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2458,6 +2458,7 @@ type Artwork implements Node & Searchable & Sellable {
   priceCurrency: String
   priceIncludesTax: Boolean
   priceIncludesTaxDisplay: String
+  priceListed: Money
 
   # The price paid for the artwork in a user's 'my collection'
   pricePaid: Money
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -4269,4 +4269,26 @@ describe("Artwork type", () => {
       })
     })
   })
+
+  describe("#priceListed", () => {
+    const query = `
+      {
+        artwork(id: "richard-prince-untitled-portrait") {
+          priceListed{
+            display
+          }
+        }
+      }
+    `
+
+    it("returns artworks price_listed", () => {
+      artwork.price_listed = 123
+      artwork.price_currency = "USD"
+      return runQuery(query, context).then((data) => {
+        expect(data).toEqual({
+          artwork: { priceListed: { display: "US$123" } },
+        })
+      })
+    })
+  })
 })
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -4274,9 +4274,7 @@ describe("Artwork type", () => {
     const query = `
       {
         artwork(id: "richard-prince-untitled-portrait") {
-          priceListed{
-            display
-          }
+          priceListed
         }
       }
     `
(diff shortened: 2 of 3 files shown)


src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -4269,4 +4269,24 @@ describe("Artwork type", () => {
...
+      {
+        artwork(id: "richard-prince-untitled-portrait") {
+          priceListed
+        }
+      }
+    `
+
+    it("returns artworks price_listed", () => {
+      artwork.price_listed = 123
+      artwork.price_currency = "USD"
```
iskounen, line 4284: It doesn't hurt the test, but is this line necessary?

iskounen (approved):

src/schema/v2/artwork/__tests__/artwork.test.js
fladson: Nope, this was when I had it as `Money`, I will remove it.

MrSltun:
I have a concern about accessing this data with [unauthenticated](https://github.com/artsy/metaphysics/blob/main/src/lib/loaders/loaders_without_authentication/gravity.ts)/[authenticated](https://github.com/artsy/metaphysics/blob/main/src/lib/loaders/loaders_with_authentication/gravity.ts) gravity loaders 🤔

fladson:
> I have a concern about accessing the unlisted price data with [unauthenticated](https://github.com/artsy/metaphysics/blob/main/src/lib/loaders/loaders_without_authentication/gravity.ts)/[authenticated](https://github.com/artsy/metaphysics/blob/main/src/lib/loaders/loaders_with_authentication/gravity.ts) gravity loaders 🤔
> 
> Should we default it to `null` in case the loader is the unauthenticated one?
> 
> Another thing to note as well is that the authenticated gravity loader just sends the tokens to Gravity and returns whatever Gravity returns. Do we have a check for which data gets returned to which users/applications? 👀

@MrSltun thanks for the concern, the way I see this is that we didn't change the field original visibility, but I will check with unauthenticated requests to see what we get back.

_schemaV2.graphql
```diff
@@ -2458,6 +2458,7 @@ type Artwork implements Node & Searchable & Sellable {
   priceCurrency: String
   priceIncludesTax: Boolean
   priceIncludesTaxDisplay: String
+  priceListed: Float
```
starsirius, line 2461: I saw you initially had this as the `Money` type, which would be helpful centralizing the conversion logic in the API. Curious if there is challenge using that.

iskounen: In addition to displaying the price, we also want to use it in a calculation to display the discounted price. Does the `Money` type have a numeric field as well as a text field?

fladson: I didn't have the conversion in mind when switching to Float, I thought the use of Money was more related to formatting and displaying the currency and since we already have a field with the display text ready, I thought that using Money here would be redundant.

The only challenge was to convert the value to cents before using the helpers and that it requires a currency for it work properly.

I haven't tried converting the values to other currencies though.

starsirius: Sorry for the confusion. By conversion, I meant major <-> minor unit conversion. The Gravity `price_listed` in major unit is not recommended, and I'd love to stop propagating that more.

> Does the Money type have a numeric field as well as a text field?

The `Money` type exposes the amount in [minor unit](https://github.com/artsy/metaphysics/blob/84314447e8401678f7a3f9a4991f59818a89990d/src/schema/v2/fields/money.ts#L162), [major unit](https://github.com/artsy/metaphysics/blob/84314447e8401678f7a3f9a4991f59818a89990d/src/schema/v2/fields/money.ts#L177), and formatted string. You can see an example query of the type in [`internationalShippingFee`](https://github.com/artsy/metaphysics/blob/84314447e8401678f7a3f9a4991f59818a89990d/src/schema/v2/artwork/__tests__/artwork.test.js#L2429C11-L2433).

PR #5489, opened 2024-01-04
feat: Add labels fields to Alert type

olerichter00:
## Description

This pull request adds the `labels` field to the `Alert` type in order to display the search criteria pills within the new Alert Notification screen in the app.

Screenshot 2024-01-04 at 12 47 42

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -24,6 +24,11 @@ import { CursorPageable, getPagingParameters } from "relay-cursor-paging"
 import { formatMarkdownValue } from "schema/v2/fields/markdown"
 import { emptyConnection } from "schema/v2/fields/pagination"
 
+// These default values are only necessary due to caching issues in Gravity.
+// Normally Gravity should always send values for these preferences.
+export const DEFAULT_CURRENCY_PREFERENCE = "USD"
+export const DEFAULT_LENGTH_UNIT_PREFERENCE = "in"
+
 const loadNs = performance.now()
 const loadMs = Date.now()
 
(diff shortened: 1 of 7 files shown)


src/lib/helpers.ts
```diff
@@ -24,6 +24,11 @@ import { CursorPageable, getPagingParameters } from "relay-cursor-paging"
 import { formatMarkdownValue } from "schema/v2/fields/markdown"
 import { emptyConnection } from "schema/v2/fields/pagination"
 
+// These default values are only necessary due to caching issues in Gravity.
+// Normally Gravity should always send values for these preferences.
+export const DEFAULT_CURRENCY_PREFERENCE = "USD"
+export const DEFAULT_LENGTH_UNIT_PREFERENCE = "in"
```
olerichter00, line 30: Moving these constants to this helper file because importing them in `alerts.ts` from `me/index.ts` caused issues.

src/schema/v2/previewSavedSearch/searchCriteriaLabel.ts
```diff
@@ -108,7 +108,8 @@ export const resolveSearchCriteriaLabels = async (
     majorPeriods,
     colors,
     partnerIDs,
-  } = parent
+    // Converting all keys to camel case because they can come from Gravity's REST or GraphQL APIs.
+  } = camelCaseKeys(parent) as any
```
olerichter00, line 112: Before, we didn't have to convert the keys because Gravity's stitched GraphQL response already returned them like this. Now, we're not stitching and using the criteria from Gravity's new REST endpoint.

Let me know if you have any good ideas to make this code a bit cleaner. 🧹

anandaroop (dismissed):
LGTM but wondering about the camel logic below…

src/schema/v2/previewSavedSearch/searchCriteriaLabel.ts
olerichter00: Good catch! I've updated the function to correctly convert IDs and added some test cases:

```
export const camelCaseKeys = (
  obj: Record<string, any>
): Record<string, any> => {
  return mapKeys(obj, (_, key) => {
    // Special case for ID to not be converted to I_D
    if (key.includes("_id")) {
      return camelCase(key.replace(/id/g, "ID"))
    }

    return camelCase(key)
  })
}
```

PR #5491, opened 2024-01-04
feat: support returning closed shows and installShots owned by partners

brainbicycle:
Addresses: https://artsyproduct.atlassian.net/browse/PHIRE-551

- adds an authenticated loader for partnerShowImages to allow images accessible to partners to be returned
- adds a param to show() field to support returning inactive/non-displayable shows

use cases needed for folio.


src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -578,6 +578,7 @@ export default (accessToken, userID, opts) => {
       {},
       { headers: true }
     ),
+    partnerShowImagesLoader: gravityLoader((id) => `partner_show/${id}/images`),
```
brainbicycle, line 581: need an authenticated loader to get all partner images

damassi (approved):

PR #5497, opened 2024-01-08
feat(EMI-1589): change Artwork#price_listed field type to Money

fladson:
Changing the type to Money as per @starsirius comment at https://github.com/artsy/metaphysics/pull/5486/files#r1441712833

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -1123,6 +1123,13 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           return price_includes_tax ? "VAT included in price" : null
         },
       },
+      priceListed: {
+        type: Money,
+        resolve: ({ price_listed: price_listed, price_currency: currency }) => {
+          const cents = price_listed * 100
+          return { cents, currency }
+        },
+      },
       taxInfo: TaxInfo,
       artaShippingEnabled: {
         type: GraphQLBoolean,
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -1123,10 +1123,6 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           return price_includes_tax ? "VAT included in price" : null
         },
       },
-      priceListed: {
-        type: GraphQLFloat,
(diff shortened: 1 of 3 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -1124,8 +1124,11 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       priceListed: {
-        type: GraphQLFloat,
-        resolve: ({ price_listed }) => price_listed,
+        type: Money,
+        resolve: ({ price_listed: price_listed, price_currency: currency }) => {
+          const cents = price_listed * 100
+          return { cents, currency }
```
fladson, line 1130: For `display`, the idea is to use the existing `internalDisplayPrice`.

src/schema/v2/artwork/index.ts
```diff
@@ -1124,8 +1124,11 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       priceListed: {
-        type: GraphQLFloat,
-        resolve: ({ price_listed }) => price_listed,
+        type: Money,
+        resolve: ({ price_listed: price_listed, price_currency: currency }) => {
+          const cents = price_listed * 100
```
starsirius, line 1129: We can't simply `* 100` to convert an amount from major unit to minor unit because different currencies have different ratio, e.g. USD is 100 and JPY is 1.

Metaphysics maintains its own ratios for common currencies in [currency_codes.json](https://github.com/artsy/metaphysics/blob/98e7b3b7f6dc00932f9614b87f885417d51a4a08/src/lib/currency_codes.json). We can use it for the conversion based on the price currency. An example conversion can be seen [here](https://github.com/artsy/metaphysics/blob/98e7b3b7f6dc00932f9614b87f885417d51a4a08/src/schema/v2/fields/money.ts#L73).

src/schema/v2/artwork/index.ts
starsirius: 👍  I'm fine to skip implementing `display` now if it's not needed yet.

starsirius (approved):
💪  Thank you for following up!

PR #5519, opened 2024-01-19
fix: preserve sorting when requesting non-shallow partner.artworkConnection properties

erikdstock:
This PR adds the incoming `sort` param to the `partner/:id/artworks/all` query when a query on Partner.artworksConnection includes `shallow: false`.


iskounen (approved):

iskounen:
I found two instances of Volt where `shallow: false` could be used, and neither uses a `sort` param:

- https://github.com/artsy/volt/blob/0eda81c30679fb8d696bef9999a68cbe535c2e59/app/javascript/routes/home/HomePageChecklist/components/ArtworksMissingMetadata.tsx#L201
- https://github.com/artsy/volt/blob/0eda81c30679fb8d696bef9999a68cbe535c2e59/app/javascript/routes/home/HomePageChecklist/HomePageChecklist.tsx#L75

PR #5521, opened 2024-01-22
[Draft]feat: Add subject to Notification

olerichter00:
Resolves [ONYX-657]

## Description

With the new Activity panel design, the title displayed for each notification type also changes. The new title is different from the existing fields `title` and `message`. I have named the new field `subject` for now because it matches the subject line of the emails we send for some of the notification types.

This is a draft for an implementation in Metaphysics, but probably Gravity would be a better place to generate and store the value 🤔 

Screenshot 2024-01-22 at 09 57 28

[ONYX-657]: https://artsyproduct.atlassian.net/browse/ONYX-657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -13615,6 +13615,7 @@ type Notification implements Node {
     # pass `RELATIVE` to display the human-friendly date (e.g. "Today", "Yesterday", "5 days ago")
     format: String
   ): String!
+  subject: String!
   targetHref: String!
   title: String!
 }
--- a/src/schema/v2/notifications/Item/__tests__/index.test.ts
+++ b/src/schema/v2/notifications/Item/__tests__/index.test.ts
@@ -6,6 +6,7 @@ describe("NotificationItem", () => {
     id: "user-notification-id",
     actor_ids: ["actor-id"],
     object_ids: ["object-id1"],
+    objects_count: 1,
   }
   let meNotificationLoader
   let mePartnerOfferLoader = jest.fn()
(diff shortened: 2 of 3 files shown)


mzikherman (commented):
There's no right or wrong way to go about this - this is fine! But somehow I tend to agree with your sentiment that this should go into Gravity. While it is a UI-specific attribute that MP can be good for...it might also benefit from having first-class access to relations (the artist of the saved artwork in the offer activity case), _and_ if there's some overlap between logic to craft subject lines, maybe that can be consolidated into Gravity.

Then, MP would just pass-through and expose that field per usual.

Re: naming - `subject` seems fine _but_ in the context of `objects` and `actors` it almost reads a bit abstract. I think it's fine to use it, but maybe `title` or `headline` may work just as well, and be clearer in context?

olerichter00:
Alright, the implementation in Gravity should be relatively easy. I will just add another for the new title [here](https://github.com/artsy/gravity/blob/f2dd2e778d634f12d7618c8d638b0fd2988d5ec9/app/models/activity_feed/activity.rb#L33) and pass it through MP.

About the naming:
I guess,`title` would be a good name. It's not yet used in Gravity, but already exposed as a field in [MP](https://github.com/artsy/metaphysics/blob/9ea5689eb3476be7e057c7c3a2b8a7591ce30c60/src/schema/v2/notifications/index.ts#L45) (returning the actors). This `title` field won't be needed anymore in the new design but in the meantime, until the new design is released. Would be nice to find a way to re-use this field and not add another one which is basically the new title.

olerichter00:
Closing this PR in favour of https://github.com/artsy/gravity/pull/17284

PR #5527, opened 2024-01-24
feat(EMI-1623): expose page arg for Partner#artworksConnection

fladson:
Pagination of the CMS send offers table only works when navigating client-side (e.g. by clicking on the paginator), but does not work when visiting the URL with params directly (e.g. https://cms-staging.artsy.net/send-offers?page=12 ) or refreshing the page. It seems the page param was ignored and it always fetches the first page.

This will enable pagination using query string and allow volt-v2 to use url params to navigate to the page correctly.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -13982,6 +13982,7 @@ type Partner implements Node {
 
     # Return artworks that are missing priority metadata
     missingPriorityMetadata: Boolean
+    page: Int
 
     # Only return artworks that are partner-offerable
     partnerOfferable: Boolean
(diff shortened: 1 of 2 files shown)


erikdstock:
Question: Do we want to use `page` args, or do we want to use cursor-based pagination with the [`pageable()` fields?](https://github.com/artsy/metaphysics/blob/8b8e99e090d195b0629aa8500b532df8335c9326/src/schema/v2/artwork/index.ts#L1614). Given the outcome of the recent discussion in slack I assume `?page=x` is preferred over `first=10&after=a1s2d3f4`, but wanted to ask.

damassi:
@erikdstock - The page parameter is bi-directional with cursors; ie, we can translate between the two when making requests for pagination data (from UI like the pagination component, being sent up via relay to MP and back down)

PR #5530, opened 2024-01-25
chore(DIA-398): Expose inquiry requests count on collector profile

jo-rs:
Resolves [DIA-398]

Exposes inquiry requests count on the collector profile.

[DIA-398]: https://artsyproduct.atlassian.net/browse/DIA-398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/CollectorProfile/collectorProfile.ts
+++ b/src/schema/v2/CollectorProfile/collectorProfile.ts
@@ -110,6 +110,11 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
     resolve: ({ artwork_inquiry_requests_count }) =>
       artwork_inquiry_requests_count >= 25,
   },
+  inquiryRequestsCount: {
+    type: GraphQLInt,
+    resolve: ({ artwork_inquiry_requests_count }) =>
+      artwork_inquiry_requests_count,
+  },
   isActiveBidder: {
     type: GraphQLBoolean,
     resolve: ({ previously_registered_for_auction }) =>
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -4066,6 +4066,7 @@ type CollectorProfileType {
       reason: "identityVerified is going to be removed, use isIdentityVerified instead"
     )
   initials(length: Int = 3): String
+  inquiryRequestsCount: Int
   institutionalAffiliations: String
   intents: [String]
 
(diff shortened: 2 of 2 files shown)


egdbear (approved):
⭐

src/schema/v2/CollectorProfile/collectorProfile.ts
```diff
@@ -110,6 +110,11 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
     resolve: ({ artwork_inquiry_requests_count }) =>
       artwork_inquiry_requests_count >= 25,
   },
+  inquiryRequestsCount: {
+    type: GraphQLInt,
+    resolve: ({ artwork_inquiry_requests_count }) =>
+      artwork_inquiry_requests_count,
+  },
```
dzucconi, line 117: I think typically we expose a `counts` sub-type. All counts should be non-nullable and fallback to 0.

dzucconi: This is like one of those things I can't emphasize enough: any count has to be non-nullable. It's _so_ annoying to have to do `foo?.bar?.count ?? 0` all the time in our codebases.

jo-rs: Makes sense. Gravity already has a 0 fallback, but it makes sense to foolproof it 👍 
I'll add this in a sec.
Thanks for the feedback! 🙏

dzucconi: If you're in there then I would also mark the rest of those Ints! 🙏

jo-rs: No prob, but do you mean in general or on that file? 🤔 
If it's the file, the only other int I see there atm is the collector level, which...I don't really know what that's about. Should that be non-nullable 🤔

egdbear: This field is [already 0 by default](https://github.com/artsy/gravity/blob/2400f289ea37ffd0e72e0debebfeb834360d3a9b/app/models/domain/collector_profile.rb#L114) in Gravity.
Do you see any potential downstream issues for not having resolving type non-nullable here?

dzucconi: Moving from nullable to non-nullable is not a breaking change. There are no issues.

PR #5531, opened 2024-01-25
fix(DIA-398): Make inquiry requests count non-nullable

jo-rs:
Addresses [this](https://github.com/artsy/metaphysics/pull/5530#discussion_r1466346675) comment.

Makes the inquiry requests count exposed on #5530 non-nullable.


dzucconi:
Thanks for the follow up

dzucconi (approved):

PR #5533, opened 2024-01-25
feat: add formatted artwork location display

rquartararo:
This adds a resolver to the location display field to return a formatted artwork location address.

Screenshot 2024-01-25 at 12 44 42 PM


src/schema/v2/location.ts
```diff
@@ -106,6 +106,10 @@ export const LocationType = new GraphQLObjectType<any, ResolverContext>({
 
     display: {
       type: GraphQLString,
+      resolve: ({ address, city, state, country }) =>
+        `${address && address + ", "}${city && city + ", "}${
+          state && state + ", "
+        }${country && country}`,
```
starsirius, line 112: - For Gravity `Location` that already has its `display`, we can just use it and not break existing behavior.
- For formatting addresses from parts, we can do something similar to [`Location`](https://github.com/artsy/gravity/blob/b448fcf07b6d8871f0767d364fe792b2f5c407a4/app/models/domain/location.rb#L117).

So maybe something like 👇  (untested):

```typescript
resolve: ({ display, address, address_2, city, state, postal_code, country }) => {
  display ||
    compact([address, address_2, city, state, postal_code, country]).join(", ")
}
```

Also, some unit tests would be great!

rquartararo: Thanks! I tried to use `compact` and it wasn't working, but now I see you have to explicitly return it 😅 

I'm not seeing a test file for `location.ts` do you know if this is tested in another file? Otherwise, I will create one.

src/schema/v2/location.ts
```diff
@@ -106,6 +107,22 @@ export const LocationType = new GraphQLObjectType<any, ResolverContext>({
...
+        city,
+        state,
+        postal_code,
+        country,
+      }) => {
+        return (
+          display ||
+          compact([address, address_2, city, state, postal_code, country]).join(
+            ", "
+          )
```
iskounen, line 123: **issue (non-blocking):** I was taught to not place a comma between the state and postal code when addressing letters as a kid, so the resulting address string will look funny to me, but I don't think it matters.

iskounen (dismissed):

iskounen (approved):
I love the test scenarios. LGTM 👍

PR #5544, opened 2024-01-30
fix(EMI-1623): force an empty return if there are no artwork IDs

fladson:
While doing https://github.com/artsy/volt-v2/pull/1141, I noticed that if a page number that is 1 more than the maximum page number is given, it would still return some data when it should return an empty array.

I am not sure why this is happening but forcing a return when the `artworkIds` is empty solved the issue.

I've checked Gravity and it is returning the expected empty array, so there's is something wrong with Metaphysics pagination helpers.

Check slack thread for more context: https://artsy.slack.com/archives/CP9P4KR35/p1706528000848069
Relates to: https://github.com/artsy/volt-v2/pull/1141

### DIFF
--- a/src/schema/v2/partner/partner.ts
+++ b/src/schema/v2/partner/partner.ts
@@ -455,21 +455,14 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
 
           const { body, headers } = await partnerArtworksLoader(id, gravityArgs)
           const totalCount = parseInt(headers["x-total-count"] || "0", 10)
+          const artworkIds = body.map((artwork) => artwork._id)
 
           // Only accept shallow = false argument if requesting user is authorized admin/partner
-          if (args.shallow === false && partnerArtworksAllLoader) {
-            const artworkIds = body.map((artwork) => artwork._id)
-            if (artworkIds.length === 0) {
-              return paginationResolver({
-                totalCount,
-                offset,
-                page,
-                size,
-                body: [],
-                args,
-              })
-            }
-
+          if (
+            args.shallow === false &&
+            artworkIds.length > 0 &&
+            partnerArtworksAllLoader
+          ) {
             const gravityArtworkArgs = {
               artwork_id: artworkIds,
               sort: args.sort,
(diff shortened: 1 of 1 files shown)


src/schema/v2/partner/partner.ts
```diff
@@ -459,10 +459,23 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
+                size,
+                body: [],
+                args,
+              })
+            }
+
             const gravityArtworkArgs = {
               artwork_id: artworkIds,
               sort: args.sort,
+              page: page,
```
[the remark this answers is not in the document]
fladson: That's right, given that we will have artworks ids in the query, this is not needed. I will remove it.

iskounen (approved):

src/schema/v2/partner/partner.ts
```diff
@@ -455,14 +455,19 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
           const { body, headers } = await partnerArtworksLoader(id, gravityArgs)
           const totalCount = parseInt(headers["x-total-count"] || "0", 10)
+          const artworkIds = body.map((artwork) => artwork._id)
 
           // Only accept shallow = false argument if requesting user is authorized admin/partner
-          if (args.shallow === false && partnerArtworksAllLoader) {
-            const artworkIds = body.map((artwork) => artwork._id)
+          if (
+            args.shallow === false &&
+            artworkIds.length > 0 &&
```
starsirius, line 463: This solves the issue for this particular case, but I still think [MP should not drop query params that are empty arrays](https://artsy.slack.com/archives/CP9P4KR35/p1588364384042800). Gravity handles the empty `artwork_id` array param correctly. It'll be nice to try out the [new qs `allowEmptyArrays` option](https://github.com/ljharb/qs/pull/487) and fix this globally. The package hasn't been releasing new versions for 8 months so we might need to pull from github directly.

starsirius: Followup ticket [here](https://artsyproduct.atlassian.net/browse/EMI-1658).

PR #5558, opened 2024-02-05
feat(onyx-718): add partnerOffer field for PartnerOfferCreated notifications

nickskalkin:
Follow-up for https://github.com/artsy/gravity/pull/17314

We need to display price with and without discount in activity panel for partner offer notifications.

### DIFF
--- a/src/schema/v2/partnerOffer.ts
+++ b/src/schema/v2/partnerOffer.ts
@@ -8,8 +8,7 @@ import {
 } from "graphql"
 import { ResolverContext } from "types/graphql"
 import { IDFields } from "./object_identification"
-import { Money } from "./fields/money"
-import currencyCodes from "lib/currency_codes.json"
+import { priceDisplayText } from "lib/moneyHelpers"
 
 export const PartnerOfferType = new GraphQLObjectType<any, ResolverContext>({
   name: "PartnerOffer",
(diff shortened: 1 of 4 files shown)


nickskalkin:
Changing Float to Money..

nickskalkin:
Moving to draft once again. I want to format price message on MP side

src/schema/v2/partnerOffer.ts
```diff
@@ -16,6 +18,9 @@ export const PartnerOfferType = new GraphQLObjectType<any, ResolverContext>({
       type: GraphQLString,
       resolve: ({ artwork_id }) => artwork_id,
     },
+    available: {
+      type: GraphQLBoolean,
+    },
```
anandaroop, line 23: **suggestion (non-blocking):** Minor, but I believe we prefer to rename these upstream booleans to e.g. `isAvailable`

(Examples [here](https://github.com/artsy/metaphysics/blob/836358a9fc9b960ee18ab7ec606ed331bbe488bf/src/schema/v2/me/index.ts#L388-L391) and [here](https://github.com/artsy/metaphysics/blob/836358a9fc9b960ee18ab7ec606ed331bbe488bf/src/schema/v2/gene.ts#L137-L140))

anandaroop (dismissed):
👍🏽

src/schema/v2/partnerOffer.ts
nickskalkin: thanks, fixed!

src/schema/v2/partnerOffer.ts
```diff
@@ -25,6 +31,28 @@ export const PartnerOfferType = new GraphQLObjectType<any, ResolverContext>({
       type: GraphQLString,
       resolve: ({ partner_id }) => partner_id,
     },
+    priceListedMessage: {
+      type: GraphQLString,
```
MounirDhahri, line 35: I see you initially went with this type as a `Money` interface then changed your mind. Is there a specific reason for that?

nickskalkin: @MounirDhahri sure! I've just noticed that Artwork details component uses `saleMessage` field from Metaphysics, so I've decided to stick to the same approach. `saleMessage` is a formatted string 

- saleMessage field resolver in MP: https://github.com/artsy/metaphysics/blob/ec13f424f9547d2961f2d71a6ff1f920a14ebbb9/src/schema/v2/artwork/index.ts#L1501
- place where gravity queries it: https://github.com/artsy/force/blob/main/src/Components/Artwork/Details.tsx#L376

nickskalkin: You think it's worth re-doing it with Money?

PR #5564, opened 2024-02-08
chore: Expose collector profile counts

jo-rs:
Exposes the new `saved_artworks_count`, `followed_artists_count`, `collected_artworks_count`, and `total_bids_count` fields on the collector profile.

### DIFF
--- a/src/schema/v2/CollectorProfile/collectorProfile.ts
+++ b/src/schema/v2/CollectorProfile/collectorProfile.ts
@@ -113,7 +113,23 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
   inquiryRequestsCount: {
     type: new GraphQLNonNull(GraphQLInt),
     resolve: ({ artwork_inquiry_requests_count }) =>
-      artwork_inquiry_requests_count,
+      artwork_inquiry_requests_count || 0,
+  },
+  savedArtworksCount: {
+    type: new GraphQLNonNull(GraphQLInt),
+    resolve: ({ saved_artworks_count }) => saved_artworks_count || 0,
+  },
+  followedArtistsCount: {
+    type: new GraphQLNonNull(GraphQLInt),
+    resolve: ({ followed_artists_count }) => followed_artists_count || 0,
+  },
+  collectedArtworksCount: {
+    type: new GraphQLNonNull(GraphQLInt),
+    resolve: ({ collected_artworks_count }) => collected_artworks_count || 0,
+  },
+  totalBidsCount: {
+    type: new GraphQLNonNull(GraphQLInt),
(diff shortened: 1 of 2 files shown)


src/schema/v2/CollectorProfile/collectorProfile.ts
```diff
@@ -113,7 +113,23 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
   inquiryRequestsCount: {
     type: new GraphQLNonNull(GraphQLInt),
     resolve: ({ artwork_inquiry_requests_count }) =>
-      artwork_inquiry_requests_count,
+      artwork_inquiry_requests_count || 0,
+  },
+  savedArtworksCount: {
+    type: new GraphQLNonNull(GraphQLInt),
+    resolve: ({ saved_artworks_count }) => saved_artworks_count || 0,
```
egdbear, line 120: **suggestion:** I think `saved_artworks_count` (and other added resolving fields) `size` in gravity is returning 0 by default. `|| 0` is not needed here.

jo-rs: I think most, if not all, are safe on the gravity side, but I was still getting `Cannot return null for non-nullable field CollectorProfileType...` before they finished resolving 🤷

jo-rs: I don't know if something changed, but last week `inquiryRequestsCount` was fine, but yesterday it was complaining.

egdbear: It might be related with data-sync? Regardless, I think we might want to confirm this with the test and we can move on:)

dzucconi: Avoid trusting Gravity, fallback to 0

dzucconi (approved):
Personally, I still prefer having a `counts` nested sub-object but it doesn't matter. Let's get this merged.

src/schema/v2/CollectorProfile/collectorProfile.ts
egdbear: I haven't found any examples with fallback on the root level. Why would one avoid trusting Gravity? It should be the source of truth! 😅

egdbear: I feel like there are no many on the root level like this, maybe in resolvers more tough

egdbear: Makes sense, add more defence - wasn't ever the issue 😅 
I will not comment if we add fallbacks to the clients as well 💯

dzucconi: 😄 well the clients will have type information from Metaphysics

PR #5565, opened 2024-02-09
feat: expose purchasable

leamotta:
https://artsyproduct.atlassian.net/browse/EMI-1689

Expose purchasable in order to be used by the saved artworks table in Volt V2


fladson (approved):
I am surprised it wasn't available before.

PR #5571, opened 2024-02-16
fix(collectorprofile): exposes owner field

dzucconi:
This is to get at the `interestsConnection` on the `UserType`.

The existing `userInterests` field on the collector profile only fetches for the logged in user and isn't a connection type anyway.

We'll need to add filtering options to the `interestsConnection` once that's done in Gravity.

### DIFF
--- a/src/schema/v2/CollectorProfile/collectorProfile.ts
+++ b/src/schema/v2/CollectorProfile/collectorProfile.ts
@@ -18,6 +18,7 @@ import Image, { normalizeImageData } from "schema/v2/image"
 import { userInterestType } from "../userInterests"
 import { myLocationType } from "../me/myLocation"
 import initials from "schema/v2/fields/initials"
+import { UserType } from "schema/v2/user"
 
 export const CollectorProfileFields: GraphQLFieldConfigMap<
   any,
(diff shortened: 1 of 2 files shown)


src/schema/v2/CollectorProfile/collectorProfile.ts
```diff
@@ -58,20 +59,25 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
...
     type: new GraphQLNonNull(new GraphQLList(userInterestType)),
     resolve: (_collectorProfile, _args, { meUserInterestsLoader }) => {
       return meUserInterestsLoader?.().then(({ body }) => body)
     },
   },
-
-  // moved InquirerCollectorProfileFields here
   location: { type: myLocationType },
   artsyUserSince: dateFormatter(({ artsy_user_since }) => artsy_user_since),
   ownerID: {
```
araujobarret, line 70: should we deprecate this then?

araujobarret (approved):

src/schema/v2/CollectorProfile/collectorProfile.ts
dzucconi: In theory... Though, I've been finding them useful sometimes where you might want to defer loading of some sub-object to the client.

egdbear (approved):

PR #5570, opened 2024-02-16
feat: Return `me` from notification mutations

olerichter00:
* [ONYX-750](https://artsyproduct.atlassian.net/browse/ONYX-750)

## Description

In order to update the notification read count, this PR updates all Notification mutations to return `me`. This should allow us to update `unreadNotificationsCount` automatically in the Relay store when notifications are marked as read without manipulating the store imperatively ([example](https://github.com/artsy/force/blob/29a650d2548a31325e8d507acdfa12d7c27ca425/src/Components/Notifications/Mutations/markNotificationsAsSeen.ts#L11))

```graphql
mutation {
	markNotificationAsRead(input: { id: "feed_id" }) {
		responseOrError {
			... on MarkNotificationAsReadSuccess {
				me {
					unreadNotificationsCount
				}
				success
			}
		}
	}
}
```

[ONYX-750]: https://artsyproduct.atlassian.net/browse/ONYX-750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5574, opened 2024-02-20
chore: Add eslint recommended rule to eslint config

olerichter00:
## Description

This PR adds `"eslint:recommended",` to [.eslintrc.json](https://github.com/artsy/metaphysics/pull/5574/files#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090f) to add some more rules (e.g. [to avoid committing `debugger`](https://github.com/artsy/metaphysics/pull/5570#discussion_r1494314305)).

### DIFF
--- a/src/lib/graphqlTimeoutMiddleware.ts
+++ b/src/lib/graphqlTimeoutMiddleware.ts
@@ -65,9 +65,9 @@ export const graphqlTimeoutMiddleware = (defaultTimeoutInMS: number) => {
           )
         }, timeoutInMS)
       }),
-      new Promise(async (resolve, reject) => {
+      new Promise((resolve, reject) => {
         try {
-          const result = await middlewareResolver(parent, args, context, info)
+          const result = middlewareResolver(parent, args, context, info)
           resolve(result)
         } catch (error) {
           reject(error)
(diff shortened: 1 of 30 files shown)


damassi (approved):

src/lib/graphqlTimeoutMiddleware.ts
```diff
@@ -65,9 +65,10 @@ export const graphqlTimeoutMiddleware = (defaultTimeoutInMS: number) => {
           )
         }, timeoutInMS)
       }),
-      new Promise((resolve, reject) => {
+      // eslint-disable-next-line no-async-promise-executor
+      new Promise(async (resolve, reject) => {
         try {
-          const result = middlewareResolver(parent, args, context, info)
+          const result = await middlewareResolver(parent, args, context, info)
```
damassi, line 71: Kinda curious about this one... how did it ever work?

olerichter00: I guess you can use an async function as a promise executor. [It's just not encouraged because of error handling, and it shouldn't be necessary because the async function already returns a promise. ](https://eslint.org/docs/latest/rules/no-async-promise-executor)

Probably, the idea here was to be able to use the[ finally block](https://github.com/artsy/metaphysics/blob/f48040d9d8a0c42e6e6aeafca948ab08ec0295d8/src/lib/graphqlTimeoutMiddleware.ts#L75) to clear the timeout 🤔

damassi: Ahh right! That makes sense 👍

PR #5586, opened 2024-02-22
feat: [EMI-1732] expose user partner offer connections on Me

erikdstock:
HOLD - Do not merge
Relates to [EMI-1723]
Completes [EMI-1732]
Depends on https://github.com/artsy/gravity/pull/17374 🔒 
[slack breadcrumbs thinking about schema design](https://artsy.slack.com/archives/C02JHHHKP5K/p1708625247424299) 🔒 
cc @artsy/emerald-devs 

This PR
- adds the `isActive` field from the linked gravity PR to the PartnerOffer type (end_at in the future)
- adds new types and fields (open to revision on these):
  - Type: `PartnerOfferToCollector` - the collector's view of a partner offer. We discussed other names too, like `CollectorOffer`.
  - Field: on `Me.partnerOffersConnection`. Earlier had, but then removed `Artwork.partnerOffersToCollectorConnection`. If we need this later it is trivial to add, but we can also just query off the `me` for now.

_Screenshot shows old field name_
image

[EMI-1723]: https://artsyproduct.atlassian.net/browse/EMI-1723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EMI-1732]: https://artsyproduct.atlassian.net/browse/EMI-1732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/schema/v2/artwork/index.ts
```diff
@@ -1094,31 +1097,57 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+            total_count: true,
+            page,
+            size,
+          }
+
+          if (args.sort) {
+            gravityArgs["sort"] = args.sort
+          }
+          if (args.includeExpired) {
+            gravityArgs["include_expired"] = args.includeExpired
```
starsirius, line 1127: This hasn't been implemented yet in Gravity right?

erikdstock: It has now - artsy/gravtiy#17374

src/schema/v2/partnerOffer.ts
```diff
@@ -18,15 +20,16 @@ export const PartnerOfferType = new GraphQLObjectType<any, ResolverContext>({
...
+    },
     isAvailable: {
       type: GraphQLBoolean,
       resolve: ({ available }) => available,
     },
     createdAt: date(),
     endAt: date(),
-    id: {
-      type: GraphQLString,
-    },
```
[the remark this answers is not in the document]
erikdstock: I could split this fix out, but I don't think we use the `id` field. For now updated the test, will split if we prefer!

src/schema/v2/artwork/index.ts
```diff
@@ -1094,31 +1101,52 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           return partnerLoader(partner.id).catch(() => null)
         },
       },
+      partnerOffersToCollectorConnection: {
```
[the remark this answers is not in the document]
damassi: Yah, better on me! Then we don't need any funny names, since the me namespace captures the context 👍

src/schema/v2/partnerOfferToCollector.ts
```diff
@@ -0,0 +1,73 @@
...
+      resolve: ({ partner_id }) => partner_id,
+    },
+    priceWithDiscountMessage: {
+      type: GraphQLString,
+      resolve: ({ price_with_discount: price, price_currency: currency }) => {
+        if (!price || !currency) {
+          return null
+        }
+
+        const priceCents = price * 100
```
starsirius, line 46: We should avoid harding the major to minor unit conversion here, and instead, use the correct ratio based on  the currency. Here is an [example](https://github.com/artsy/metaphysics/blob/23fae702991340bfa4901b90224e6189da0912c0/src/schema/v2/artwork/index.ts#L1195-L1197) to look up with MP's own json file.

src/schema/v2/partnerOfferToCollector.ts
```diff
@@ -0,0 +1,73 @@
...
+    },
+    isAvailable: {
+      type: GraphQLBoolean,
+      resolve: ({ available }) => available,
+    },
+    partnerId: {
+      type: GraphQLString,
+      resolve: ({ partner_id }) => partner_id,
+    },
+    priceWithDiscountMessage: {
```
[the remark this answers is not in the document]
erikdstock: Ah, this is just copied from `PartnerOffer`. I'll update and add this field as plain `priceWithDiscount`, then follow up on the original type, marking that field deprecated.

erikdstock:
!squashongreen

src/schema/v2/partnerOfferToCollector.ts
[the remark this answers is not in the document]
erikdstock: Updated to add the money type to both this and the base PartnerOffer after the linked gravity PR is merged.

_schemaV2.graphql
```diff
@@ -14819,6 +14838,63 @@ enum PartnerOfferSorts {
...
+    # https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+    timezone: String
+  ): String
+
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID.
+  internalID: ID!
+  isActive: Boolean
```
narikazu, line 14865: question (non-blocking): I'm curious if we have a plan in the future to use the `isActive` field anywhere since [the endpoint](https://github.com/artsy/gravity/pull/17374/files) to return partner offers on me only gives us active partner offers.

narikazu (approved):
LGTM

_schemaV2.graphql
erikdstock: Removed `isActive` since we don't need it right now :+1:

narikazu (approved):
LGTM

PR #5593, opened 2024-02-27
feat(EMI-1710, EMI-1711): add shareableWithPartners to Collections

fladson:
Make `shareable_with_partners` available as a not required param for Collections creation. 

Followup from https://github.com/artsy/gravity/pull/17388

image

**Param not given**
image

### DIFF
--- a/src/schema/v2/me/createCollectionMutation.ts
+++ b/src/schema/v2/me/createCollectionMutation.ts
@@ -3,6 +3,7 @@ import {
   GraphQLUnionType,
   GraphQLNonNull,
   GraphQLString,
+  GraphQLBoolean,
 } from "graphql"
 import { mutationWithClientMutationId } from "graphql-relay"
 import {
@@ -45,6 +46,7 @@ const ResponseOrErrorType = new GraphQLUnionType({
 
 interface InputProps {
   name: string
+  shareableWithPartners: boolean
 }
 
 export const createCollectionMutation = mutationWithClientMutationId<
@@ -56,6 +58,7 @@ export const createCollectionMutation = mutationWithClientMutationId<
   description: "Create a collection",
   inputFields: {
     name: { type: new GraphQLNonNull(GraphQLString) },
+    shareableWithPartners: { type: GraphQLBoolean },
   },
   outputFields: {
     responseOrError: {
@@ -73,6 +76,7 @@ export const createCollectionMutation = mutationWithClientMutationId<
         name: args.name,
         user_id: context.userID,
         saves: true,
+        shareable_with_partners: args.shareableWithPartners,
       })
 
       return response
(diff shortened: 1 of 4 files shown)


src/schema/v2/me/createCollectionMutation.ts
```diff
@@ -56,6 +58,7 @@ export const createCollectionMutation = mutationWithClientMutationId<
   description: "Create a collection",
   inputFields: {
     name: { type: new GraphQLNonNull(GraphQLString) },
+    shareableWithPartners: { type: GraphQLBoolean },
```
starsirius, line 61: Should this also be `new GraphQLNonNull(GraphQLBoolean)` so it's non nullable?

fladson: Do we want to make this a required field? In Gravity it's optional, so I reflected that logic here.

fladson:
This also cover https://artsyproduct.atlassian.net/browse/EMI-1711

src/schema/v2/me/createCollectionMutation.ts
```diff
@@ -73,6 +76,7 @@ export const createCollectionMutation = mutationWithClientMutationId<
         name: args.name,
         user_id: context.userID,
         saves: true,
+        shareable_with_partners: args.shareableWithPartners,
```
starsirius, line 79: The `shareableWithPartners` argument is optional but we always pass it to Gravity for creating a collection. I wonder if there are cases that we'll call Gravity with `shareable_with_partners: null`. For example,

- Client calls the mutation without `shareableWithPartners`. I think this can be addressed by setting a `defaultValue` for this optional argument.
- Client calls the mutation explicitly with `shareableWithPartners: null` (maybe by mistake). A default value might not help here.

Maybe we should should omit passing `shareable_with_partners` to gravity when it's not present?

fladson: I've set a default value in Gravity to handle this.
- If no param is given, it will be `true` by default in Gravity;
- If it is given as `null`, we will raise an error, since it's a boolean field;

starsirius: The [database schema](https://github.com/artsy/gravity/blob/834c48b9a36cae875a2bb4a9d2a3c1bdfe30ecbc/db/schema.rb#L222) does not indicate `null: false`, so _I think_ it's possible to set it to `nil` in Gravity. Or we can specify [`allow_blank: false`](https://github.com/ruby-grape/grape?tab=readme-ov-file#allow_blank) in the [Gravity endpoint](https://github.com/artsy/gravity/blob/834c48b9a36cae875a2bb4a9d2a3c1bdfe30ecbc/app/api/v1/collections_endpoint.rb#L76) to reject empty value.

starsirius (approved):

PR #5594, opened 2024-02-27
feat(DIA-481): wire up CollectorProfile#interestsConnection

araujobarret:
[DIA-481]

Wires up interestsConnection to the new endpoint `PartnerCollectorProfile#userInterests` endpoint on Gravity.

[DIA-481]: https://artsyproduct.atlassian.net/browse/DIA-481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/loaders/loaders_with_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_with_authentication/gravity.ts
@@ -561,6 +561,15 @@ export default (accessToken, userID, opts) => {
       ({ partnerId, userId }) =>
         `partner_collector_profile?partner_id=${partnerId}&user_id=${userId}`
     ),
+    partnerCollectorProfileUserInterestsLoader: gravityLoader<
+      any,
+      { collectorProfileId: string; partnerId: string }
+    >(
+      ({ collectorProfileId, partnerId }) =>
+        `partner_collector_profile/${collectorProfileId}/user_interests?partner_id=${partnerId}`,
+      {},
+      { headers: true }
+    ),
     partnerInquiryRequestLoader: gravityLoader<
       any,
       { partnerId: string; inquiryId: string }
--- a/src/schema/v2/CollectorProfile/collectorProfile.ts
+++ b/src/schema/v2/CollectorProfile/collectorProfile.ts
@@ -15,10 +15,14 @@ import {
 import { ResolverContext } from "types/graphql"
 import Image, { normalizeImageData } from "schema/v2/image"
 
-import { userInterestType } from "../userInterests"
+import { UserInterestConnection, userInterestType } from "../userInterests"
 import { myLocationType } from "../me/myLocation"
 import initials from "schema/v2/fields/initials"
 import { UserType } from "schema/v2/user"
+import { pageable } from "relay-cursor-paging"
+import { convertConnectionArgsToGravityArgs } from "lib/helpers"
+import { createPageCursors } from "../fields/pagination"
+import { connectionFromArraySlice } from "graphql-relay"
 
 export const CollectorProfileFields: GraphQLFieldConfigMap<
   any,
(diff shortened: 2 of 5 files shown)


src/schema/v2/CollectorProfile/collectorProfile.ts
```diff
@@ -65,6 +69,43 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
       return meUserInterestsLoader?.().then(({ body }) => body)
     },
   },
+  interestsConnection: {
+    type: UserInterestConnection,
+    args: pageable({}),
+    resolve: async (
```
egdbear, line 75: **suggestion:** It would be nice to test this resolver

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -561,6 +561,15 @@ export default (accessToken, userID, opts) => {
       ({ partnerId, userId }) =>
         `partner_collector_profile?partner_id=${partnerId}&user_id=${userId}`
     ),
+    partnerCollectorProfileUserInterestsLoader: gravityLoader<
+      any,
+      { collectorProfileId: string; partnerId: string }
+    >(
+      ({ collectorProfileId, partnerId }) =>
+        `partner_collector_profile/${collectorProfileId}/user_interests?partner_id=${partnerId}`,
```
iskounen, line 569: **thought:** I didn't notice this in the Gravity PR, but it feels strange to pass a partner ID to this endpoint. Instead of passing a partner ID, I think Gravity can figure out whether the current user has access to a partner by doing:

```ruby
error!('Forbidden', 403) unless current_user.has_authorization_to?(:update, partner)
```

src/schema/v2/CollectorProfile/collectorProfile.ts
```diff
@@ -65,6 +69,43 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
...
+      { id, partnerId },
+      args,
+      { partnerCollectorProfileUserInterestsLoader }
+    ) => {
+      if (!partnerCollectorProfileUserInterestsLoader || !partnerId) {
+        return null
+      }
+
+      const { page, size, offset } = convertConnectionArgsToGravityArgs(args)
```
iskounen, line 84: **question (non-blocking):** Will we need to sort the results on the client side? If so, we may want to include that in the loader.

src/schema/v2/CollectorProfile/collectorProfile.ts
```diff
@@ -65,6 +69,43 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
...
+      const {
+        body,
+        headers,
+      } = await partnerCollectorProfileUserInterestsLoader(
```
egdbear, line 88: **suggestion:** adding defense here would make more sense to prevent breaks. Maybe try catch or alternative

src/schema/v2/conversation/index.ts
```diff
@@ -322,7 +322,10 @@ export const ConversationType = new GraphQLObjectType<any, ResolverContext>({
           })
 
           return {
-            collectorProfile: data.collector_profile,
+            collectorProfile: {
+              ...data.collector_profile,
+              partnerId: to_id,
```
iskounen, line 327: **question:** 

1. Why is the partner ID included in this field?
2. On line 309 above there is a comment that says "Do not use this field for Partners." Do you know what that is referring to?

egdbear, line 327: I agree that there is a long way from the conversation to reach the interestConnection, and I also agree that this is necessary to make it work.

egdbear (changes requested):

src/lib/loaders/loaders_with_authentication/gravity.ts
araujobarret: Was a decision we took back there when pairing with @jonallured , I remember we discussed about but don't recall exactly why we didn't use the `current_user`.

src/schema/v2/CollectorProfile/collectorProfile.ts
araujobarret: yes, but we never do it in MP, all the paginated loaders we query I only kept the style but can move in this direction if you want to.

src/schema/v2/conversation/index.ts
araujobarret: maybe should we comment this section to make it clear what we use it for?

src/schema/v2/CollectorProfile/collectorProfile.ts
araujobarret: There's no test for the `CollectorProfile` resolver, I'd need to create it from scratch, so on it.

src/schema/v2/CollectorProfile/collectorProfile.ts
egdbear: That's true, but IMO risky, nevermind then!

src/schema/v2/conversation/index.ts
araujobarret: 1. I will add a comment, we need to pass down to the `CollectorProfile` type the `partnerId`, if we remove the need for Gravity to pass the `partner_id` that won't be necessary anymore, maybe it's better going into this direction specially for the Schema organization. I will create a follow-up ticket because that needs to be done on Gravity and then back to MP again.

2. I think the message is wrong, it should be `Do not use this field for Collectors`, can correct it in a sec 👍

src/schema/v2/CollectorProfile/collectorProfile.ts
egdbear: is this comment related to [this point](https://github.com/artsy/metaphysics/pull/5594#discussion_r1504026387)? 😅

araujobarret: yes 🙈

egdbear (approved):

src/lib/loaders/loaders_with_authentication/gravity.ts
jonallured: I might be missing something but should we do both? A user can have access to multiple partners so I would expect for an ID to be passed and then also a check in the endpoint to verify whether the user can see that partner.

iskounen: Good point. I forgot that Gravity will need to fetch the `partner` in that snippet.

PR #5597, opened 2024-02-27
fix(feature-flags): fix feature flag creation responses

damassi:
Noticed that creating feature flags was (kinda) broken in forque (would create, but on create it couldnt find the flag), and we needed to update the api path slightly.

### DIFF
--- a/src/schema/v2/admin/mutations/createFeatureFlagMutation.ts
+++ b/src/schema/v2/admin/mutations/createFeatureFlagMutation.ts
@@ -172,8 +172,7 @@ export const createFeatureFlagMutation = mutationWithClientMutationId<
         await addFeatureFlagVariant(args.name, args.variants)
       }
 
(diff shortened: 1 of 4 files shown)


src/schema/v2/admin/mutations/createFeatureFlagMutation.ts
```diff
@@ -172,8 +172,7 @@ export const createFeatureFlagMutation = mutationWithClientMutationId<
         await addFeatureFlagVariant(args.name, args.variants)
       }
 
-      const featureFlag = await adminFeatureFlagLoader(args.name)
```
damassi, line 175: This surfaced the error, but we don't actually need to perform a fetch here anyways since we don't return a single feature flag on success. Just the full list to refresh the table.

damassi: So technically, changing to `return {}` fixes the bug

PR #5605, opened 2024-03-01
feat: allow querying alerts by search criteria

nickskalkin:
[Context for this PR](https://github.com/artsy/gravity/pull/17398)

Adds support for filtering alerts by search criteria attributes:

```graphql
query X {
  me {
    email
    
    alertsConnection(first: 1, attributes: { artistIDs: ["kaws"], acquireable: true }) {
      edges {
        node {
          displayName
        }
      }
    }
  }
}
```

### DIFF
--- a/src/schema/v2/previewSavedSearch/previewSavedSearch.ts
+++ b/src/schema/v2/previewSavedSearch/previewSavedSearch.ts
@@ -113,7 +113,7 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
   }),
 })
 
-const PreviewSavedSearchAttributesType = new GraphQLInputObjectType({
+export const PreviewSavedSearchAttributesType = new GraphQLInputObjectType({
   name: "PreviewSavedSearchAttributes",
   fields: previewSavedSearchArgs,
 })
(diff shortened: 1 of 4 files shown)


src/schema/v2/previewSavedSearch/previewSavedSearch.ts
```diff
@@ -113,7 +113,7 @@ const PreviewSavedSearchType = new GraphQLObjectType<any, ResolverContext>({
   }),
 })
 
-const PreviewSavedSearchAttributesType = new GraphQLInputObjectType({
+export const PreviewSavedSearchAttributesType = new GraphQLInputObjectType({
```
anandaroop, line 116: **thought (non-blocking):** Wonder if this should have a more generic (non-"preview") name now. But I suppose that would be a breaking change…

anandaroop (approved):

src/schema/v2/previewSavedSearch/previewSavedSearch.ts
nickskalkin: Agree with you, but maybe I can extract values to a constant with a generic name and reference these values here. I'll do it in another PR

PR #5631, opened 2024-03-25
feat(EMI-1727): create updateCollectionsMutation

fladson:
Port the work done at https://github.com/artsy/gravity/pull/17431 to metaphysics.

Example request for reference:
```
  mutation {
    updateMeCollectionsMutation(
      input: {
        attributes: [
          {
            id: ""
            shareableWithPartners: false
          },
          {
            id: ""
            shareableWithPartners: true
          },
        ]
      }
    ) {
      meCollectionsOrErrors {
        ... on UpdateMeCollectionsSuccess {
          collection {
            id
          	shareableWithPartners
            name
            artworksCount
          }
        }
        ... on UpdateMeCollectionsFailure {
          mutationError {
            type
            message
          }
        }
      }
    }
  }
```

Successful request
SCR-20240325-owt

Failed request
SCR-20240325-ox2

@artsy/emerald-devs

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -4108,6 +4108,11 @@ type CollectionsEdge {
   node: Collection
 }
 
+input CollectionsInput {
+  id: String!
+  shareableWithPartners: Boolean!
+}
+
 enum CollectionSorts {
   CREATED_AT_ASC
   CREATED_AT_DESC
@@ -13423,6 +13428,11 @@ type Mutation {
   # Update a collection
   updateCollection(input: updateCollectionInput!): updateCollectionPayload
 
+  # Updates the user's collections in batch.
+  updateCollectionsMutation(
+    input: updateCollectionsMutationInput!
+  ): updateCollectionsMutationPayload
+
   # Updating a collector profile (loyalty applicant status).
   updateCollectorProfile(
     input: UpdateCollectorProfileInput!
@@ -18878,6 +18888,28 @@ type updateCollectionPayload {
 union UpdateCollectionResponseOrError =
     UpdateCollectionFailure
   | UpdateCollectionSuccess
+
+type UpdateCollectionsFailure {
+  mutationError: GravityMutationError
+}
+
+input updateCollectionsMutationInput {
+  attributes: [CollectionsInput!]!
+  clientMutationId: String
+}
+
+type updateCollectionsMutationPayload {
+  clientMutationId: String
+  collectionsOrErrors: [UpdateCollectionsResponseOrError!]!
+}
+
+union UpdateCollectionsResponseOrError =
+    UpdateCollectionsFailure
+  | UpdateCollectionsSuccess
+
+type UpdateCollectionsSuccess {
+  collection: Collection
+}
 
 type UpdateCollectionSuccess {
   collection: Collection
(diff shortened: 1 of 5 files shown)


_schemaV2.graphql
```diff
@@ -13423,6 +13428,11 @@ type Mutation {
   # Update a collection
   updateCollection(input: updateCollectionInput!): updateCollectionPayload
 
+  # Updates the user's collections in batch.
+  updateCollectionsMutation(
```
starsirius, line 13432: Naming suggestion: Given the mutation names are global in the schema, without name spacing, let's prepare ourselves for 2 different mutations:

- update collections for myself (based on the access token): our needs
- update collections for a user (e.g. performed by an admin, with a user id param): future use cases

I can't find a consistent pattern in MP for mutations, but I'd suggest we use name

- `updateMeCollectionsMutation`: update collections for myself (I kind of like `updateMyCollectionsMutation` but apparently "MyCollection" means something else already 😞 .)
- `updateUserCollectionsMutation`: update collections for a user

starsirius: And we can also update the input, payload, response types accordingly.

_schemaV2.graphql
```diff
@@ -4108,6 +4108,11 @@ type CollectionsEdge {
   node: Collection
 }
 
+input CollectionsInput {
```
starsirius, line 4111: Naming suggestion: The plural collection"s" here sounds a bit unintuitive with the singular data inside. Maybe we can call it `UpdateMeCollectionInput` so the mutation below reads a bit nicer:

```graphql
input updateMeCollectionsMutationInput {
  attributes: [UpdateMeCollectionInput!]!
  clientMutationId: String
}
```

(There is an existing [`updateCollectionInput`](https://github.com/artsy/metaphysics/blob/0f0cf799cd780744a9e2ce8ac87d7ca01c832c65/_schemaV2.graphql#L18868-L18874) that requires an id and a name. 🤷  The collection modeling for multiple purposes is kind of making naming tricky.)

src/schema/v2/me/updateCollectionsMutation.ts
```diff
@@ -0,0 +1,105 @@
...
+    })
+
+    try {
+      return await meUpdateCollectionsLoader({
+        attributes: JSON.stringify(gravityPayload),
+      })
+    } catch (error) {
+      const formattedErr = formatGravityError(error)
+      if (formattedErr) {
+        return [{ ...formattedErr, _type: "GravityMutationError" }]
```
starsirius, line 99: The corresponding Gravity batch endpoint returns

- An array of updated collections when succeeded
- A single [_object_](https://github.com/artsy/gravity/blob/fa66526dafd211963f9db0fe677e4518c4e610d4/spec/api/v1/me_collections_endpoint_spec.rb#L115-L116) of an error when failed.

Here in MP we turn the error response type into an array as well. I guess it's fine (or even more flexible in the future a Gravity batch endpoint can return multiple errors).

starsirius (commented):
LGTM. I have a few naming suggestions otherwise looking good! 👍

_schemaV2.graphql
fladson: 👍

PR #5632, opened 2024-03-25
feat(artwork): Add isUnlisted to artwork schema

damassi:
This updates our schema with a new `isUnlisted` field on artwork:

```graphql
{
  artwork(id: "foo") {
    isUnlisted
  }
}
```

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2558,6 +2558,9 @@ type Artwork implements Node & Searchable & Sellable {
   isShareable: Boolean
   isSold: Boolean
   isUnique: Boolean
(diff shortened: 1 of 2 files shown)


joeyAghion:
Any reason we can't stick with the "unlisted" language? I recognize we use "private" a lot internally, but translating the status at this intermediate layer risks making client code harder to trace through back-ends.

damassi:
Oh that's a very good point @joeyAghion - yeah, we can do that, keep things 1:1. Lemme update

lilyfromseattle (approved):
👍

PR #5637, opened 2024-03-29
feat: add priceListedDisplay to Artwork that always returns display for price if it's publicly available)

oxaudo:
This to follow up on discussion from https://github.com/artsy/eigen/pull/10008 as I'm about to add this display to Artwork page with offer.

I also followed the logic of price display from saleMessage as that is what we want I believe to match how the artwork price is displayed on the Artwork page when offer is not present except that we always want to return the price if it's available if the work has a publicly available price.


leamotta (approved):
LGTM!

PR #5640, opened 2024-04-02
fix: alert.artworksConnection shows only for sale artworks

nickskalkin:
There is a mismatch between the artworks user sees in the new "View artworks" screen and when navigating to the artworks grid when clicking "See all matching works" button. This happens because "View artworks" screen uses `alert.artworksConnection` field witch lacks `for_sale` attribute. This PR adds the missing attribute.

Demo of the bug (notice 22 artworks before and 11 after):

<video src="https://github.com/artsy/metaphysics/assets/3934579/bc62a8ec-7262-444e-af67-8557cfcabf8f" />

### DIFF
--- a/src/schema/v2/Alerts/index.ts
+++ b/src/schema/v2/Alerts/index.ts
@@ -250,7 +250,7 @@ export const AlertType = new GraphQLObjectType<
           price_range: args.price_range,
           sizes: args.sizes,
           width: args.width,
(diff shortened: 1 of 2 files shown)


olerichter00 (approved):
👏

tam-kis (approved):

dblandin (approved):
Good catch 👍🏼

PR #5651, opened 2024-04-09
feat(EMI-1790): allow shareableWithPartners to be updated per collection

fladson:
Followup from https://github.com/artsy/gravity/pull/17563

Recently, we shipped the possibility to edit the `shareableWithPartners` through the settings modal, in a batch approach, now we also need to provide this control in the collection edit page for individual and atomic edits.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -19006,7 +19006,7 @@ input updateCollectionInput {
   # The internal ID of the collection
   id: String!
   name: String!
-  shareableWithPartners: Boolean!
+  shareableWithPartners: Boolean
 }
 
 type updateCollectionPayload {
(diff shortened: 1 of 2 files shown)


MrSltun (dismissed):

_schemaV2.graphql
```diff
@@ -19006,6 +19006,7 @@ input updateCollectionInput {
   # The internal ID of the collection
   id: String!
   name: String!
+  shareableWithPartners: Boolean!
```
starsirius, line 19009: I think this param can be optional. Otherwise the current UI that does not pass it (like Eigen) will fail.

MrSltun (approved):

_schemaV2.graphql
fladson: Good call 👍

PR #5653, opened 2024-04-09
feat(onyx-878): Schema for Specialist bios

nickskalkin:
[Jira ticket](https://artsyproduct.atlassian.net/browse/ONYX-878)

We want to update the list of specialists, which exists in two places:

1) for Force it is stored in repo
2) Eigen takes this data from s3 during build step and stores it in app.

We want to make it easier to update the list and following our [playbook for dealing with static content](https://www.notion.so/artsy/Dealing-with-static-content-245c55744c82410293750ed7a44d1021) we've decided to move this data to Metaphysics. Now clients can execute such query to receive a list for specialists:

```graphql
{
  specialistBios {
      name
      firstName
      jobTitle
      specialty
      bio
      email
      
      image {
        imageURL
      }
    }
}
```


olerichter00 (approved):
🌟

src/schema/v2/schema.ts
```diff
@@ -332,6 +333,7 @@ const rootFields = {
   shortcut,
   show: Show,
   showsConnection: Shows,
+  specialistBios: SpecialistBios,
```
mzikherman, line 336: I don't think it should be a root field, should go under `staticContent` -

Screen Shot 2024-04-09 at 10 16 47 AM

anandaroop: Yes, thanks ☝🏽

src/lib/specialistBios/index.ts
```diff
@@ -0,0 +1,49 @@
...
+      },
+      jobTitle: { type: GraphQLString },
+      name: { type: GraphQLString },
+      specialty: { type: GraphQLString },
+    },
+  }
+)
+
+export const SpecialistBios: GraphQLFieldConfig<void, ResolverContext> = {
+  description: "A list of SWA specialists",
```
anandaroop, line 46: **suggestion (non-blocking):** Minor, but I'd spell out the SwA acronym here.

anandaroop (commented):
Looking good!

_schemaV2.graphql
```diff
@@ -18488,6 +18491,16 @@ enum sort {
   DESC
 }
 
+type SpecialistBio {
+  bio: String
+  email: String
+  firstName: String
+  image: Image
```
dblandin, line 18498: **praise**: Really glad we're able to use our Image type here to enable clients to crop/resize as necessary. ⭐

src/lib/specialistBios/data.json
```diff
@@ -0,0 +1,137 @@
...
+    "firstName": "Jessica",
+    "name": "Jessica Backus",
+    "email": "jessica@artsymail.com",
+    "jobTitle": "VP, Collector Sales",
+    "specialty": "collectorServices",
+    "bio": "An early Artsy employee, Jessica has nearly 2 decades of experience in art and tech. Driven by a passion to help people connect with art they love, Jessica has launched multiple strategic initiatives to evolve and grow the art market. She began her career working with clients in Berlin at the gallery Peres Projects and worked at Phillips before joining Artsy. She holds a master’s degree in art history from Hunter and a bachelor’s degree from Columbia University.",
+    "imageUrl": "https://files.artsy.net/images/Jessica-Mobile.webp"
+  },
+  {
+    "firstName": "Christine",
```
dblandin, line 12: **question**: Could `firstName` be computed reliably from `name`? We have some similar logic for [splitting names on the backend](https://github.com/artsy/gravity/blob/a319430dadefbcb7bd850a748d19d6a96a0e092f/app/models/domain/artist.rb#L277-L286) we could borrow from.

dblandin (commented):
Looking good! 🙌🏼

src/schema/v2/schema.ts
nickskalkin: Oh, right. There is `staticContent` field under `root` already though and it does something different: https://github.com/artsy/metaphysics/blob/8239cc57c343db7a861d879ffdc09e5b3647e287/src/schema/v2/static_content.ts#L4

Should I create another namespace field? How would you call it? @anandaroop @mzikherman

nickskalkin: I was hoping we do not use it anywhere an can re-use it, but found queries like this in datadog:

```
staticContent(id: "how-auctions-work-bidding") {
    content(format: HTML)
    id
}
```

src/lib/specialistBios/data.json
nickskalkin: thanks, done!

src/schema/v2/schema.ts
mzikherman: I would probably use this field, make the `id` optional and resolve to `{}` when not passed in, and then add `specialistBios` under there.

A little odd, but I would rather use (and reuse, even if a bit odd) the `staticContent` field vs. root.

anandaroop: Yeah I think that's^ reasonable, not much chance of colliding with the existing `content` and `id`, since the current and future staticContent additions will be properly scoped e.g. `staticContent.specialistBios.fooBar`

src/lib/specialistBios/data.json
```diff
@@ -0,0 +1,122 @@
+[
```
damassi, line 1: `lib/specialistBios` feels like an odd place for this kind of content static content. Can we use the existing `src/data` folder for this? And if we need more fine-grained structure we could create a new folder in there.

src/lib/specialistBios/index.test.ts
```diff
@@ -0,0 +1,42 @@
+import gql from "lib/gql"
```
damassi, line 1: tests go in a `__tests__` folder, as seen every else in MP

src/lib/specialistBios/index.ts
```diff
@@ -0,0 +1,55 @@
+import {
```
damassi, line 1: Why are we storing a slice of the schema in the `lib` folder? The schema goes in `schema`, with the rest of our schema.

src/schema/v2/schema.ts
nickskalkin: makes sense 👍 solved

src/lib/specialistBios/data.json
nickskalkin: thanks, solved!

src/lib/specialistBios/index.test.ts
nickskalkin: ✅

src/lib/specialistBios/index.ts
nickskalkin: ✅

PR #5656, opened 2024-04-09
chore: add new display_artist_bio field to Artwork

JanaeHijaz:
Addresses part 1 of Amber 515, adding this new field to ME to then be surfaced for Private Artworks listings.

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -656,6 +656,30 @@ describe("Artwork type", () => {
     })
   })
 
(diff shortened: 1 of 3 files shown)


damassi (approved):
Looks good! And thanks for the test 👍

src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -656,6 +656,30 @@ describe("Artwork type", () => {
     })
   })
 
+  describe("#displayArtistBio", () => {
```
jpotts244, line 659: 👏🏻 !

PR #5663, opened 2024-04-11
chore: add note to partner offer to collector

leamotta:
In order to support showing the partner offer note in the artwork page for this ticket https://artsyproduct.atlassian.net/browse/EMI-1782


erikdstock (approved):

erikdstock:
LGTM. It's so trivial this is non-blocking, but if there is already a test for this type it wouldn't hurt to add this field to the query so we can test it shows up.

PR #5665, opened 2024-04-12
feat: enable more artist insight fields for update

egdbear:
This PR enables the `criticallyAcclaimed`, `vanguardYear`, and `awards` fields to be used as input for updateArtist mutation.

[Gravity PR](https://github.com/artsy/gravity/pull/17582)

### DIFF
--- a/src/schema/v2/artist/index.ts
+++ b/src/schema/v2/artist/index.ts
@@ -241,7 +241,7 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       awards: {
-        type: GraphQLNonNull(GraphQLString),
+        type: GraphQLString,
         resolve: ({ awards }) => awards || "",
       },
       auctionResultsConnection: {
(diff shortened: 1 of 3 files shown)


src/schema/v2/artist/index.ts
```diff
@@ -240,6 +240,10 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
           )
         },
       },
+      awards: {
+        type: GraphQLNonNull(GraphQLString),
+        resolve: ({ awards }) => awards || "",
+      },
```
iskounen, line 246: Why did you choose make this type `GraphQLNonNull`?

egdbear: I believe it's a safe pattern to fallback to an empty string because this field is of type string in Gravity. We still have some records that are `nil`, so using an empty string gives should us consistency.

Here is the [discussion](https://github.com/artsy/metaphysics/pull/5564#discussion_r1484024204) with @dzucconi - it might help 😅

dzucconi: Actually, this should probably be nullable.

Integers and Booleans should _never_ be nullable because the null value is essentially either `false` or `0`; in this case empty string is useless.

src/schema/v2/artist/index.ts
```diff
@@ -654,8 +658,8 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
       },
       createdAt: date(),
       criticallyAcclaimed: {
-        type: GraphQLBoolean,
-        resolve: ({ critically_acclaimed }) => critically_acclaimed,
+        type: new GraphQLNonNull(GraphQLBoolean),
```
egdbear, line 661: okay fixing this one @dzucconi

iskounen (approved):

PR #5672, opened 2024-04-17
feat: add recent_show for managing artist insights

egdbear:
This PR resolves and enables the `recent_show` field to be used as input for updateArtist mutation.

[Gravity PR](https://github.com/artsy/gravity/pull/17602)

### DIFF
--- a/src/schema/v2/artist/index.ts
+++ b/src/schema/v2/artist/index.ts
@@ -860,6 +860,11 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       related: Related,
+      recentShow: {
+        type: GraphQLString,
+        description: "The most recent show for an artist",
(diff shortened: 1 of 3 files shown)


src/schema/v2/artist/index.ts
```diff
@@ -860,6 +860,11 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
         },
       },
       related: Related,
+      recentShow: {
+        type: GraphQLString,
```
araujobarret, line 864: given no recent show we return "" then this type should be wrapped in a NonNull type, right?

egdbear: Makes sense, I will remove `|| ""` here so we resolve to `null`.

araujobarret (approved):

PR #5673, opened 2024-04-17
feat(onyx-886): exclude disliked artworks from artworksForUser recommendations

nickskalkin:
Initial description of a problem is here: https://github.com/artsy/metaphysics/pull/5658

After releasing ☝️ we discovered a bug - metaphysics was sending a non-authenticated request to `/filter/artworks` endpoint with a new `exclude_disliked_artworks=true` parameter. Earlier, thanks to @joeyAghion's suggestion, I made this parameter `user_only` in Gravity, which makes Gravity fail with `Can not access user-only field` error when it's requested without a user. The fact that Gravity crashes was very fortunate - otherwise I would probably not spot that artworks returned by `filter/artworks` endpoint contain disliked artworks.

This PR reverts a revert (https://github.com/artsy/metaphysics/pull/5668) and also switches to authenticated requests.

### DIFF
--- a/src/schema/v2/artworksForUser/__tests__/helpers.test.ts
+++ b/src/schema/v2/artworksForUser/__tests__/helpers.test.ts
@@ -63,6 +63,11 @@ describe("getNewForYouArtworks", () => {
       context
     )
     expect(artworks.length).toEqual(1)
+    expect(mockArtworksLoader).toBeCalledWith({
+      availability: "for sale",
+      exclude_disliked_artworks: true,
+      ids: artworkIds,
+    })
   })
 })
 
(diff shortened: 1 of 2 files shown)


nickskalkin:
My mistake was that I tested the query only with `onlyAtAuction: false` and thus completely skipped `/filter/artworks`

Now this query works:

```
  artworksForUser(includeBackfill: true, first: 20, onlyAtAuction: true) {
    edges {
      node {
        slug
      }
    }
  }
```

joeyAghion (approved):

mzikherman (approved):

PR #5621, opened 2024-03-13
feat: add non bnmo purchases to CollectorResume buyerActivity

egdbear:
This PR addresses the calculation of non-BNMO purchases and combines them with BNMO purchases from Exchange.

Please find related Gravity PR [here](https://github.com/artsy/gravity/pull/17444).

[DIA-503]

[DIA-503]: https://artsyproduct.atlassian.net/browse/DIA-503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5677, opened 2024-04-18
chore: remove v2 file from codebase

egdbear:
This PR removes v2 file added by mistake from in [this PR](https://github.com/artsy/metaphysics/pull/5621/files).


araujobarret (approved):
🧹

PR #5678, opened 2024-04-18
feat(artist): exposes collected artworks count

dzucconi:
The count in the collection table should be the count of the artworks you've collected.

### DIFF
--- a/src/schema/v2/artist/index.ts
+++ b/src/schema/v2/artist/index.ts
@@ -627,6 +627,24 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
                 return headers["x-total-count"] || 0
               },
             },
+            myCollectedArtworks: {
+              type: new GraphQLNonNull(GraphQLInt),
+              resolve: async ({ _id }, _, { meMyCollectionArtworksLoader }) => {
+                if (!meMyCollectionArtworksLoader) {
+                  return 0
+                }
+
+                const { headers } = await meMyCollectionArtworksLoader({
+                  total_count: true,
+                  artist_ids: [_id],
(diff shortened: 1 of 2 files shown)


src/schema/v2/artist/index.ts
```diff
@@ -627,6 +627,24 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
...
               },
             },
+            myCollectedArtworks: {
+              type: new GraphQLNonNull(GraphQLInt),
+              resolve: async ({ _id }, _, { meMyCollectionArtworksLoader }) => {
+                if (!meMyCollectionArtworksLoader) {
+                  return 0
+                }
+
+                const { headers } = await meMyCollectionArtworksLoader({
```
araujobarret, line 637: how about making it more error-proof if the request fails?

dzucconi: In what way: you wouldn't want to return an erroneous count (0). You want it to error and return the error.

araujobarret: true but the client consuming it will have the entire query request broken, isn't it?

dzucconi: Depends on how one handles the error. An error in the field would null the parent like so: 


Either way: IMO you don't want to swallow the error and return null on the field. You want to surface the error for the client to deal with.

I'd rather the page error than have to include error handling code on every field I display on the front end.

araujobarret: agreed, handling the errors in the client sounds more real IMO than fake nulls.

araujobarret (approved):

PR #5679, opened 2024-04-18
add excludeDislikedArtworks option to artworksForUser

nickskalkin:
This PR adds `excludeDislikedArtworks` option to `artworksForUser`. We have some concerns about performance of non-cached personalised calls to `/filter/artworks` endpoint (to exclude disliked artworks we have to make user-authenticated call to gravity).

When `excludeDislikedArtworks` is `false` - we continue using unauthenticated filter artworks loader - otherwise authenticated one.

New option is `false` by default, and only clients (Force / Eigen) will set it to `true` initially. When (if) we feel confident that `artworksForUser` query can handle Braze's load - we can set this option to true by default.

### DIFF
--- a/src/schema/v2/artworksForUser/helpers.ts
+++ b/src/schema/v2/artworksForUser/helpers.ts
@@ -84,7 +84,11 @@ export const getNewForYouArtworkIDs = async (
 }
 
 export const getNewForYouArtworks = async (
-  { ids, marketable }: { ids: string[]; marketable?: boolean },
+  {
+    ids,
+    marketable,
+    excludeDislikedArtworks = false,
+  }: { ids: string[]; marketable?: boolean; excludeDislikedArtworks?: boolean },
   gravityArgs,
   context: ResolverContext
 ): Promise<any[]> => {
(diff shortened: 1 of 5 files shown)


src/schema/v2/artworksForUser/helpers.ts
```diff
@@ -114,19 +118,30 @@ export const getBackfillArtworks = async (
...
+    },
+    unauthenticatedLoaders: {
+      filterArtworksLoader: filterArtworksUnauthenticatedLoader,
+    },
   } = context
 
+  const filterArtworksLoader =
+    excludeDislikedArtworks === true
+      ? filterArtworksAuthenticatedLoader
+      : filterArtworksUnauthenticatedLoader
```
joeyAghion, line 140: We weren't sure if this was the most accurate way to perform this check or not, but the argument isn't specified by any clients yet, so it seems safe to release.

joeyAghion (approved):

mzikherman (commented):
Nice idea to make this opt-in and default to false!

PR #5682, opened 2024-04-18
fix: time zone-sensitive test failures

joeyAghion:
This works to fix the faily test _right now_ locally and in CI. Who knows what tomorrow brings.

### DIFF
--- a/src/schema/v2/notifications/__tests__/index.test.ts
+++ b/src/schema/v2/notifications/__tests__/index.test.ts
@@ -1,4 +1,5 @@
 import gql from "lib/gql"
+import { DEFAULT_TZ } from "lib/date"
 import { runAuthenticatedQuery } from "schema/v2/test/utils"
 import moment from "moment-timezone"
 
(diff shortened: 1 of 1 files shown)


src/schema/v2/notifications/__tests__/index.test.ts
```diff
@@ -165,7 +166,7 @@ describe("notificationsConnection", () => {
       it("should return `Today` label", async () => {
         const loader = () => {
           return Promise.resolve({
-            feed: [{ ...notificationFeedItem, date: moment() }],
+            feed: [{ ...notificationFeedItem, date: moment().tz(DEFAULT_TZ) }],
```
starsirius, line 169: _I think_ this is optional because `moment()` and `moment().tz(DEFAULT_TZ)` are the same point in time, just represent it in different timezone.

starsirius (approved):
Thank you for fixing the flakiness! 🎉  I think we can additionally improve the coverage by testing the [specified timezone provided by the GQL context](https://github.com/artsy/metaphysics/blob/8d7283fad432ca4226c8696d2c670d82103cc9c7/src/schema/v2/notifications/index.ts#L81), in addition to the `DEFAULT_TZ`.

PR #5690, opened 2024-04-24
chore: update schema for exchange

leamotta:
Follow up for https://github.com/artsy/exchange/pull/2104

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -4752,6 +4752,9 @@ input CommerceCreateOfferOrderWithArtworkInput {
   # submitted orders on same artwork/edition with same quantity, we will return that
   findActiveOrCreate: Boolean = true
 
+  # PartnerOffer Id
+  partnerOfferId: String
+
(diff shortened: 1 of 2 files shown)


MrSltun (changes requested):
I don't think it's a good idea to remove fields from the schema 👀 
Maybe something is out of sync?

leamotta:
@MrSltun those fields are deprecated, they were removed [here](https://github.com/artsy/exchange/pull/2052/files) but the schema was never updated in MP :gull-scream:

MrSltun:
@leamotta Oh boi! This might cause breaking changes, it's safer to mark the fields as `deprecated` rather than removing them

I did a quick search for `locationFilledOnOrder` and `location_filled_on_order`, and it seems that they're not used anywhere, so it might be safe to move forward

MrSltun (approved):

leamotta:
@MrSltun thanks for checking and being fully mindful of this change! 🤗

PR #5693, opened 2024-04-24
feat: Adds partnerProvidedBiography field to Artist

jpotts244:
Adds partnerProvidedBiography field to Artist, originally we wanted to use `biographyBlurb` provided function but it only returns the partner provided bio for `featured` partner artists, see [here](https://github.com/artsy/metaphysics/blob/bf5f37c86646adbbe2c68c464bf69e9add480ad9/src/schema/v2/artist/index.ts#L495)

We need a more flexible solution for our new private artworks experience

```graphql
{
  artist(id: "662664897e9e2b00073fae7d") {
    partnerBiographyBlurb {
      text
    }
    biographyBlurb {
      text
    }
  }
}
```

Screenshot 2024-04-24 at 6 16 26 PM

### DIFF
--- a/src/schema/v2/artist/index.ts
+++ b/src/schema/v2/artist/index.ts
@@ -868,6 +868,38 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
         resolve: ({ id }, options, { partnerArtistsForArtistLoader }) =>
           partnerArtistsForArtistLoader(id, options),
       },
+      partnerBiographyBlurb: {
+        description: "The Partner's provided biography for the artist",
+        args: {
+          ...markdown().args,
+        },
+        type: new GraphQLObjectType<any, ResolverContext>({
+          name: "partnerBiographyBlurb",
+          fields: {
+            text: {
+              type: GraphQLString,
+              resolve: ({ text }) => text,
+            },
+          },
+        }),
+        resolve: async (
+          { id },
+          { format },
+          { partnerArtistsForArtistLoader }
+        ) => {
+          const partnerArtists = await partnerArtistsForArtistLoader(id, {
+            size: 1,
+          })
+
+          if (partnerArtists && partnerArtists.length) {
+            const { biography } = first(partnerArtists) as any
+
+            return {
+              text: formatMarkdownValue(biography, format),
+            }
+          }
+        },
+      },
       duplicates: {
         type: new GraphQLList(Artist.type),
         resolve: ({ id }, _args, { artistDuplicatesLoader }, _info) => {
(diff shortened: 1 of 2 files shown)


damassi (approved):
Thanks for the digging 👍

src/schema/v2/artist/index.ts
```diff
@@ -868,6 +868,38 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
...
+          { id },
+          { format },
+          { partnerArtistsForArtistLoader }
+        ) => {
+          const partnerArtists = await partnerArtistsForArtistLoader(id, {
+            size: 1,
+          })
+
+          if (partnerArtists && partnerArtists.length) {
+            const { biography } = first(partnerArtists) as any
```
lidimayra, line 895: Doesn't this mean we might end up querying the bio defined by one partner and displaying it on an artwork that belongs to another partner?

It looks like we're selecting the first partner artist from the list of all partner artists associated with one Artsy.

Example:
If both `Gagosian` and `A+B Gallery` have `Patrick Hughes` listed as a partner artist and Gagosian was the one to register him first, following the logic here, I'd expect that Gagosian custom bio would be displayed for any Patrick Hughes private artworks A+B Gallery uploaded as well.

lidimayra: I tested the above with `Mike Mandel`.

I have `Mike Mandel` registered as a partner artist for `Robert Mann Gallery` and `Commerce Test Partner`

I registered different versions of custom bio for each gallery
Screenshot 2024-04-25 at 14 09 47

Screenshot 2024-04-25 at 14 10 38

Now even if I access an [artwork](https://staging.artsy.net/artwork/639cad86592ddf000bfbed88) that belongs to `Commerce Test Partner`, I see the bio that was added by `Robert Mann Gallery`
Screenshot 2024-04-25 at 14 12 32

jpotts244: Great catch @lidimayra , I think this is also the behavior on the existing 'biographyBlurb' on artist 😬 I wonder if that is OK

lidimayra: I tested it here, it _seems_ fine (not sure, we might want to investigate deeper)
I also added a custom bio for Mike Mandel on Abakus Project gallery

I see their custom bio working fine on [the artist page](https://staging.artsy.net/partner/abakus-projects/artists/mike-mandel)

And the public artwork page uses [the Artsy provided bio](https://staging.artsy.net/artwork/mike-mandel-stacking-pampers)

PR #5694, opened 2024-04-24
chore: adding some additional specs for new Artist field

jpotts244:
Fast follow Just wanted to test the case when PartnerArtist bio is `null` in case of 💥 


damassi (approved):
Its cool, when defining the schema, it will create the structure for you in terms of slots, so null or non-existent data will come back as nullable, but the fields will still remain there. Thanks for the test follow-up 👍

PR #5702, opened 2024-04-26
feat(onyx-911): consignment submission state info

nickskalkin:
[Jira ticket](https://artsyproduct.atlassian.net/browse/ONYX-897).

From the ticket:

> [Force](https://github.com/artsy/force/blob/e64ac8ce6881c2200aa56cc1d32c72d0cbbdfbb8/src/Apps/MyCollection/Routes/MyCollectionArtwork/Components/MyCollectionArtworkSWASectionSubmitted.tsx) and [Eigen](https://github.com/artsy/eigen/blob/addcd491a72c01cf41458b81929d993a48b21711/src/app/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkSubmissionStatus.tsx) display a colorful description of the submission state, along with some help text. There’s some duplication, and in the case of “approved” submissions, the help text is misleading (“determined that we do not currently have a market for it”).

When pairing with @joeyAghion we agreed that it feel awkward to specify palette colors in Metaphysics - instead we are going to expose raw Convection state (draft, approved, rejected, etc.) based on which Force / Eigen can compute colors.

> and in the case of “approved” submissions, the help text is misleading (“determined that we do not currently have a market for it”

In current state of PR I haven't change the copy - how would you communicate to a user that their submission was approved?

## New schema

```graphql
query X {
  me {
    myCollectionConnection(first: 1, sort: CREATED_AT_DESC) {
      edges {
        node {
          consignmentSubmission {
            state # 'approved'
            stateLabel # 'Evaluation Complete'
            stateHelpMessage # 'Our specialists have reviewed...'
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
+++ b/src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
@@ -14,6 +14,7 @@ const ArtworkConsignmentSubmissionType = new GraphQLObjectType<
       },
       displayText: {
         type: GraphQLString,
+        deprecationReason: "Prefer `stateLabel` field.",
         resolve: (consignmentSubmission) => {
           const state =
             consignmentSubmission.saleState || consignmentSubmission.state
@@ -51,6 +52,41 @@ const ArtworkConsignmentSubmissionType = new GraphQLObjectType<
             consignmentSubmission.saleState || consignmentSubmission.state
 
           return ["submitted", "hold", "open"].includes(state?.toLowerCase())
+        },
+      },
+      state: {
+        type: GraphQLString,
+        description: "Submission state.",
+        resolve: ({ state }) => state.toLowerCase(),
+      },
+      stateLabel: {
+        type: GraphQLString,
+        description: "Submission state label visible to the user.",
+        resolve: ({ state }) => {
+          switch (state.toLowerCase()) {
+            case "approved":
+            case "rejected":
+            case "closed":
+            case "published":
+              return "Evaluation Complete"
+            default:
+              return "In Progress"
+          }
+        },
+      },
+      stateHelpMessage: {
+        type: GraphQLString,
+        description: "More information about the submission state.",
+        resolve: ({ state }) => {
+          switch (state.toLowerCase()) {
+            case "approved":
+            case "rejected":
+            case "closed":
+            case "published":
+              return "Our specialists have reviewed this submission and determined that we do not currently have a market for it."
+            default:
+              return "The artwork is being reviewed or is in the sale process."
+          }
         },
       },
     }
--- a/src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
+++ b/src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
@@ -1,5 +1,25 @@
-import { GraphQLBoolean, GraphQLObjectType, GraphQLString } from "graphql"
+import {
+  GraphQLBoolean,
+  GraphQLEnumType,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLString,
+} from "graphql"
 import { ResolverContext } from "types/graphql"
+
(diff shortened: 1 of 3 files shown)


dblandin:
Looking good! A few high-level comments on the schema:
- Would a level of nesting be useful here? Seeing a prefixed attributes makes me think that we could structure the new attributes under a status/state object.
- "state" vs "status": "state" feels a bit overloaded as the general term here but maybe still useful under `status`?
- "details" vs "help message": sounds like some of the strings are less for help and more to expand upon the current status.

```graphql
{
  artwork(id: "ID") {
    consignmentSubmission {
      displayText # < what we currently have
      status {
        state # -> ConsignmentSubmissionStateEnum / could known values be defined/mapped in an enum?
        label # -> String / "Evaluation Pending"
        details # -> String / "Our specialists are reviewing your work..."
      }
    }
  }
}
```

Feel free to push back or offer alternatives! I just think we can iterate a bit more on the schema here before merging.

src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
```diff
@@ -53,6 +54,41 @@ const ArtworkConsignmentSubmissionType = new GraphQLObjectType<
           return ["submitted", "hold", "open"].includes(state?.toLowerCase())
         },
       },
+      state: {
+        type: GraphQLString,
+        description: "Submission state.",
+        resolve: ({ state }) => state.toLowerCase(),
```
joeyAghion, line 60: Why the `toLowerCase`? I think this doesn't change anything at the moment, and if it did introduce a difference with the back-end in the future, it would get confusing.

Q: Should this be an enum, for better self-documentation?

joeyAghion (dismissed):
Not sure what to advise about nesting in a "status" type or not. I find the current schema clear, and it avoids introducing the state/status ambiguity. The 3 fields all change together, but they're all properties of the submission they're based on.

src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
nickskalkin: Removed `toLowerCase`. There was some logic behind it but I can't remember why I did this now.

joeyAghion: Now I'm confused about why the values are all uppercase... I assumed they'd come from Convection as lowercase, like in the raw data. I think I'm missing some in-between step.

damassi: Note that this happened too on amber with `visibility_level`. There's some strange uppercase / lowercase thing going on that I don't quite understand.

dblandin (dismissed):
I'm fine with the proposed schema — let's roll with it as is 👍🏼

joeyAghion:
I've added a commit that switches to an enum for possible state values.

Note that it duplicates a `State` enum already [available in Convection](https://github.com/artsy/convection/blob/55087cd8e2433de779cae9655cdd9fcdd7415d22/_schema.graphql#L1324) and [stitched into Metaphysics' schema](https://github.com/artsy/metaphysics/blob/a12602f8131c06c77fd12c517d8b062f1d027e76/src/lib/stitching/convection/schema.ts#L31) as `ConsignmentSubmissionStateAggregation`. I tried but didn't see a straightforward way to reuse that.

Open to more feedback about naming/structure.

Q: These states are lowercase strings [in the back-end](https://github.com/artsy/convection/blob/55087cd8e2433de779cae9655cdd9fcdd7415d22/app/models/submission.rb#L26-L32), capitalized [for Convection's Graphql response](https://github.com/artsy/convection/blob/55087cd8e2433de779cae9655cdd9fcdd7415d22/app/graphql/types/state_type.rb#L7), and then this code calls `.toLowerCase()` repeatedly and mostly deals with those lowercased strings. This seems messy to me. Can this be improved to just deal with the capitalized enum keys instead?

mzikherman (approved):

PR #5704, opened 2024-04-29
chore(ONYX-938): edit the list of specialists, remove specialty field

dariakoko:
As requested by the Sales team:
- update the list of specialists
-  remove the speciality field as we are removing the specialists' filter

This data is not yet used by Eigen or Force so we should not worry about removing the speciality field in this PR, it will not cause issues


olerichter00 (dismissed):
Looks good 🌟

PR #5716, opened 2024-05-03
fix: EMI-1823 format partner offer to collector price consistently

erikdstock:
Long [investigation](https://artsy.slack.com/archives/C02JHHHKP5K/p1714752513254499?thread_ts=1714740077.802449&cid=C02JHHHKP5K) (🔒) led to a one-line change. The goal is to match our artwork pricing format on a partner offer price. This matches the implementation we use for artwork.priceListedDisplay [here](https://github.com/artsy/metaphysics/blob/8fb38127ad77a80092d8596335db550a522a7f63/src/schema/v2/artwork/index.ts#L1581-L1599).

[Where this is rendered with artwork.priceListedDisplay](https://github.com/artsy/force/blob/3831de05482023c0150a1a6bacd6ce750bef70e0/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx#L344-L349)

Fixes [EMI-1823]

**Bad**
image

**Fixed**
image


src/schema/v2/fields/money.ts
```diff
@@ -225,7 +226,7 @@ export const resolveMinorAndCurrencyFieldsToMoney = async (
       major,
       cents: minor,
       currency: currencyCode,
-      display: formatMoney(major, symbolFromCurrencyCode(currencyCode), 0),
+      display: priceDisplayText(minor, currencyCode, ""),
```
erikdstock, line 229: note this formatting is more in line with our design guidelines but is specific to 'price display'... which may not be what all prices mean. I may follow up with a more specific attempt to address this.

MrSltun (approved):
Looks good! 🌟 
I have one concern (non-blocking) about having the change in the `money` helper since it might cause visual changes in other places that we didn't intend to change 👀

Maybe we could have the change in the field itself? But this is not blocking from merging, I just wanted to raise the concern

erikdstock:
> one concern (non-blocking) about having the change in the `money` helper since it might cause visual changes in other places that we didn't intend to change

This is only used for our partner offer types right now so should be ok! But I agree. I had also implemented parameterized format field to put on the `Money` type, just decided this would fix the most immediate issue before adding another pattern. Thanks!

PR #5720, opened 2024-05-07
fix(DIA-577): does throw an error when term is ''

araujobarret:
[DIA-577]

Fixes the annoying error we see on volt-v2 when the term is an empty string. That's harmless and it is how it works by default in gravity, it accepts an empty string in the `name_contains` filter field.

image

### DIFF
--- a/src/schema/v2/CollectorProfile/collectorProfiles.ts
+++ b/src/schema/v2/CollectorProfile/collectorProfiles.ts
@@ -31,13 +31,16 @@ export const CollectorProfilesConnection: GraphQLFieldConfig<
     { term, partnerID, ...args },
     { collectorProfilesLoader }
   ) => {
-    if (!collectorProfilesLoader)
+    if (!collectorProfilesLoader) {
       throw new Error(
         "A X-Access-Token header is required to perform this action."
       )
-
-    if (!partnerID || !term) {
-      throw new Error("Arguments `partnerID` and `term` are required.")
+    }
+    if (!partnerID) {
+      throw new Error("Argument `partnerID` is required.")
+    }
+    if (term == undefined) {
+      throw new Error("Argument `term` is required.")
     }
 
     const { page, size, offset } = convertConnectionArgsToGravityArgs(args)
--- a/src/schema/v2/__tests__/collectorProfiles.test.ts
+++ b/src/schema/v2/__tests__/collectorProfiles.test.ts
@@ -57,31 +57,57 @@ describe("collectorProfilesConnection", () => {
     })
   })
 
-  it("returns an error if the partnerID or term arguments are missing", async () => {
+  it("returns an error partnerID argument is missing", async () => {
     const missingArgsQuery = gql`
       {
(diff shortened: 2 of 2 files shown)


src/schema/v2/CollectorProfile/collectorProfiles.ts
```diff
@@ -31,13 +31,16 @@ export const CollectorProfilesConnection: GraphQLFieldConfig<
...
         "A X-Access-Token header is required to perform this action."
       )
-
-    if (!partnerID || !term) {
-      throw new Error("Arguments `partnerID` and `term` are required.")
+    }
+    if (!partnerID) {
+      throw new Error("Argument `partnerID` is required.")
+    }
+    if (term == undefined) {
```
egdbear, line 42: **question:** Why do we send `undefined` to Metaphysics? Can we handle this in the UI instead?

araujobarret: we never actually receive an `undefined`, if the arg is not present then JS considers the var value `undefined`, and we might send `null` from the client too.
This check captures if it's missing or if the client sends `null`, take a look at the `==` and not the `===`.

egdbear: I think we are sending an empty string when this request fails
image
Screenshot 2024-05-08 at 09 45 11


Why are we sending this during the initial render? Shouldn't it be handled on the client itself?

araujobarret: well, that's debatable 😅 
I agree that the clients *could/should* prevent this kind of issue, but that's how the backend is designed to work when receiving `term: ''` and I don't think we should change it right now.
If you strongly feel that we should only do it in volt-v2 I can change it there, or we could even do the combo, the change there and here; I will leave the choice to you 😬

egdbear: I think we shouldn't send anything from the client before user interaction, what do you think?

araujobarret: I can't entirely agree 😂 

Mostly because Relay is mainly declarative, so we can't block the query until the field has the `term` to call it(there are some ways of doing it but it's not something we should be doing for this case)

egdbear: I can't entirely agree either 😅 
I don't think we have the filter drawer presented on the first load. I was thinking if we could prevent that from happening.
Anyway - I'll leave this decision to you!

egdbear (approved):

src/schema/v2/CollectorProfile/collectorProfiles.ts
araujobarret: it's rendered in a portal

image

araujobarret: I will dig more 30-60 minutes into Relay docs to see if I catch something we could use(their doc sucks btw 😞 ) instead of this change here

PR #5721, opened 2024-05-07
chore: remove unused `partnerBiographyBlurb` field from artist schema 

jpotts244:
Removes unused `partnerBiographyBlurb` field from artist schema

It was not used and perhaps we need to devise a different solution for our needs as per the discussion [here](https://github.com/artsy/metaphysics/pull/5693#discussion_r1579434764)

Additionally cleans up some of the specs to make the linter happy

### Question
- Im always nervous about this, no clients should be using this field, is it ok to just delete from the schema? vs deprecating etc

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -1841,9 +1841,6 @@ type Artist implements EntityWithFilterArtworksConnectionInterface & Node & Sear
     # The number of PartnerArtists to return
     size: Int
   ): [PartnerArtist]
-
-  # The Partner's provided biography for the artist
-  partnerBiographyBlurb(format: Format): partnerBiographyBlurb
   partnersConnection(
     after: String
     before: String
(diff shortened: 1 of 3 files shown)


_schemaV2.graphql
```diff
@@ -1841,9 +1841,6 @@ type Artist implements EntityWithFilterArtworksConnectionInterface & Node & Sear
     # The number of PartnerArtists to return
     size: Int
   ): [PartnerArtist]
-
-  # The Partner's provided biography for the artist
-  partnerBiographyBlurb(format: Format): partnerBiographyBlurb
```
damassi, line 1846: Doing a search returns a result in Eigen: https://github.com/search?q=partnerBiographyBlurb&type=code

I think it might be too late to remove this unfortunately, since a request in Eigen for this slice of the schema would break the now-released eigen version. 

So that's basically the flow. Do a global search, see if eigen is referencing it at all. If not, all good to remove. Otherwise we need to force a breaking change via force upgrade on mobile.

jpotts244: Ah nice catch re: eigen
Ok cool - Maybe I'll move to deprecate and follow up after

PR #5728, opened 2024-05-08
chore: enable blurhashes on Eigen

MounirDhahri:
This PR is a follow-up on https://github.com/artsy/metaphysics/pull/5722

Since the SEO issue is related to Web, I figured out why not keep this enabled on App. Please let me know what you think.

### DIFF
--- a/src/schema/v2/image/index.ts
+++ b/src/schema/v2/image/index.ts
@@ -70,10 +70,11 @@ export const ImageType = new GraphQLObjectType<any, ResolverContext>({
           defaultValue: 64,
         },
       },
-      resolve: ({ blurhash, aspect_ratio }, { width }) => {
-        const isBlurhashEnabled = isFeatureFlagEnabled(
-          "diamond_blurhash-enabled-globally"
-        )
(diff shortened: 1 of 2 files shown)


src/schema/v2/image/index.ts
```diff
@@ -70,10 +70,11 @@ export const ImageType = new GraphQLObjectType<any, ResolverContext>({
           defaultValue: 64,
         },
       },
-      resolve: ({ blurhash, aspect_ratio }, { width }) => {
-        const isBlurhashEnabled = isFeatureFlagEnabled(
-          "diamond_blurhash-enabled-globally"
-        )
+      resolve: ({ blurhash, aspect_ratio }, { width }, { userAgent }) => {
+        const isEigen = userAgent?.match("Artsy-Mobile") != null
```
MounirDhahri, line 74: Using the same snippet from here
https://github.com/artsy/force/blob/03270b660fcf2f0bac25a16bde20b4c384b0990d/src/System/Router/buildServerAppContext.ts#L22

araujobarret (approved):
thanks for looking into it! 💟

MrSltun (approved):
Great solution Mounir! 🌟🌟

PR #5729, opened 2024-05-08
feat(EMI-1854): add source to partner offer

leamotta:
https://artsyproduct.atlassian.net/browse/EMI-1854

Support querying for https://github.com/artsy/gravity/pull/17669

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -14872,7 +14872,7 @@ type PartnerOffer implements Node {
     @deprecated(
       reason: "This field is deprecated. Use 'priceWithDiscount' instead."
     )
-  source: String
+  source: SourceEnum
   userIds: [String]
 }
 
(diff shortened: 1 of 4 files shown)


src/schema/v2/partnerOfferToCollector.ts
```diff
@@ -54,6 +54,9 @@ export const PartnerOfferToCollectorType = new GraphQLObjectType<
         )
       },
     },
+    source: {
+      type: GraphQLString,
```
fladson, line 58: Any reason for this not being enum?

leamotta: Good catch!

_schemaV2.graphql
```diff
@@ -18522,6 +18524,11 @@ enum sort {
   DESC
 }
 
+enum SourceEnum {
```
fladson, line 18527: 💅 Since this will be in the schema, I think we can prefix it with the model name, looks like it's the approach we use.

leamotta: Good point!

fladson (approved):

PR #5737, opened 2024-05-14
feat: use artist career highlights for artist insights

egdbear:
This PR introduces the `artist_career_highlights` endpoint used for retrieving artist insights (institutions) for solo, group, venues collected by major institutions.

[Gravity PR](https://github.com/artsy/gravity/pull/17668)

### DIFF
--- a/src/schema/v2/artist/helpers.ts
+++ b/src/schema/v2/artist/helpers.ts
@@ -1,6 +1,6 @@
 import { compact, sortBy } from "lodash"
 import { priceDisplayText } from "lib/moneyHelpers"
-
+import camelCase from "lodash/camelCase"
 const auctionRecordsTrusted = require("lib/auction_records_trusted.json")
   .artists
 
@@ -40,7 +40,7 @@ export const ARTIST_INSIGHT_MAPPING: Record<
 > = {
   SOLO_SHOW: {
     getDescription: () => null,
-    getEntities: (artist) => splitEntities(artist.solo_show_institutions),
+    getEntities: (artist) => artist.soloShowInstitutions,
     getLabel: (_artist, count: number) =>
       `Solo show at ${
         count === 1 ? "a major institution" : `${count} major institutions`
@@ -48,7 +48,7 @@ export const ARTIST_INSIGHT_MAPPING: Record<
   },
   GROUP_SHOW: {
     getDescription: () => null,
-    getEntities: (artist) => splitEntities(artist.group_show_institutions),
+    getEntities: (artist) => artist.groupShowInstitutions,
     getLabel: (_artist, count: number) =>
       `Group show at ${
         count === 1 ? "a major institution" : `${count} major institutions`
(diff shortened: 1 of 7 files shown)


src/schema/v2/artist/helpers.ts
```diff
@@ -237,6 +237,80 @@ export const getAuctionRecord = async (artist, auctionLotsLoader) => {
...
+
+  if (kind.includes("SOLO_SHOW")) {
+    const soloShows = await getArtistCareerHighlights(
+      artistCareerHighlightsLoader,
+      artist._id,
+      "solo"
+    )
+
+    /* eslint-disable-next-line require-atomic-updates */
+    artist.soloShowInstitutions = soloShows
```
araujobarret, line 266: maybe just:
```jsx
artist.soloShowInstitutions = await getArtistCareerHighlights(
      artistCareerHighlightsLoader,
      artist._id,
      "solo"
    )
```
declaring the const is maybe redundant, right?

src/schema/v2/artist/insights.ts
```diff
@@ -86,6 +91,36 @@ export const ArtistInsights: GraphQLFieldConfig<any, ResolverContext> = {
       artist.highAuctionRecord = highAuctionRecord
     }
 
+    if (kind.includes("SOLO_SHOW")) {
```
araujobarret, line 94: I see some repetition of what we have in `enrichWithArtistCareerHighlights`, maybe we could reuse something?

src/schema/v2/artist/helpers.ts
egdbear: yes!

src/schema/v2/artist/insights.ts
egdbear: the initial idea was to re-use `enrichWithArtistCareerHighlights`. thanks for flagging!

src/schema/v2/artist/helpers.ts
```diff
@@ -1,6 +1,6 @@
 import { compact, sortBy } from "lodash"
 import { priceDisplayText } from "lib/moneyHelpers"
-
+import { camelCase } from "lodash"
```
araujobarret, line 3: I think the import should be `import camelCase from 'lodash/camelCase';` to avoid importing the whole lodash(which is quite big)

src/schema/v2/artist/helpers.ts
```diff
@@ -237,6 +237,53 @@ export const getAuctionRecord = async (artist, auctionLotsLoader) => {
...
+  try {
+    const response = await artistCareerHighlightsLoader({
+      artist_id,
+      [type]: true,
+    })
+
+    if (!response) return null
+
+    const highlights = response.map((highlight) => highlight.venue.trim())
+    return highlights
```
araujobarret, line 278: nit suggestion:
```jsx
    return response.map((highlight) => highlight.venue.trim())
```

araujobarret (dismissed):
Looks fine, left 2 small requests 👍

src/schema/v2/artist/helpers.ts
araujobarret: doesn't increase memory consumption in this case?

araujobarret (approved):

PR #5741, opened 2024-05-16
chore: deprecate partnerBiographyBlurb field from artist schema

jpotts244:
Marks unused partnerBiographyBlurb field as deprecated from artist schema

It was not used and perhaps we need to devise a different solution for our needs as per the discussion https://github.com/artsy/metaphysics/pull/5693#discussion_r1579434764

As per [here](https://github.com/artsy/metaphysics/pull/5721/files), eigen stores a reference to this field in certain versions so it is not safe to just directly remove


Mitch-Henson (approved):
Looks good to me!

What happens after this is merged? Will Eigen + Force then remove their references to this field seeing as it is marked as deprecated?

PR #5748, opened 2024-05-22
feat: include totalUnreadCount when a partner requests a list of conversations

iskounen:
This work was motivated by artsy/volt#7560, in which the total count of unread conversations was being fetched separately from the list of conversations. I believe that those should be done in one single query since the Metaphysics schema supports it.

It seems like two separate queries were needed because Metaphysics was returning `null` for `totalUnreadCount` when requested by a partner. It turns out that we were not including the `expand=["total_unread_count"]` param when a partner requests a list of their conversations.

It makes sense to expose the total unread count to partners, so I think that including it in the request params is OK.

### DIFF
--- a/src/schema/v2/conversation/conversations.ts
+++ b/src/schema/v2/conversation/conversations.ts
@@ -97,6 +97,7 @@ const Conversations: GraphQLFieldConfig<
       }
 
       params = {
+        expand,
         deleted: false,
         intercepted: false,
         to_id: args.partnerId,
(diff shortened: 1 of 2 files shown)


iskounen:
@damassi I tagged you because it's strange that this wasn't already included before, so I wanted to see if you could think of a reason why we didn't include it before.

araujobarret (approved):
Gj 💪 
I think why it wasn't here was because we forgot during conversations revamp.

iskounen:
Thanks for checking it out. I'm closing this because it seems like the count of unread conversations isn't reliable for partners, so we need to stick with the current implementation in Volt until that is resolved in Impulse.

PR #5756, opened 2024-05-24
feat(exchange): Extend viewer with commerceOrders exchange field

damassi:
Intersected this in the orders app rebuild in volt. In order to support our modern relay pagination patterns, we need to be able to access root fields from `Viewer`. Normally this is easy to add (just [add the type](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/schema.ts#L351) to `rootFields`, and `Viewer` automatically picks it up -- no work necesary), however with orders we're stitching from exchange, so one has to do the annoying SDL stitching dance.  

cc @artsy/amber-devs

### DIFF
--- a/src/lib/stitching/exchange/v2/stitching.ts
+++ b/src/lib/stitching/exchange/v2/stitching.ts
@@ -335,6 +335,31 @@ export const exchangeStitchingEnvironment = ({
       buyerActivity: CommerceBuyerActivity
     }
 
+    extend type Viewer {
+      commerceOrders(
+        # Returns the elements in the list that come after the specified cursor.
+        after: String
+
+        # Returns the elements in the list that come before the specified cursor.
+        before: String
+        buyerId: String
+        buyerType: String
+
+        # Returns the first _n_ elements from the list.
+        first: Int
+        impulseConversationId: String
+
+        # Returns the last _n_ elements from the list.
+        last: Int
+        mode: CommerceOrderModeEnum
+        sellerId: String
+        sellerType: String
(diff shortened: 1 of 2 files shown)


src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -335,6 +335,31 @@ export const exchangeStitchingEnvironment = ({
       buyerActivity: CommerceBuyerActivity
     }
 
+    extend type Viewer {
+      commerceOrders(
```
damassi, line 339: First: use raw SDL to extend the `Viewer` type

src/lib/stitching/exchange/v2/stitching.ts
```diff
@@ -604,6 +629,25 @@ export const exchangeStitchingEnvironment = ({
           },
         },
       },
+      Viewer: {
+        commerceOrders: {
```
damassi, line 633: Then extend `Viewer` via schema delegation, and point it at the `commerceOrders` field.

(It is impossible to expect anyone anywhere to understand this 😭)

jpotts244: 🙏🏻 🙏🏻 
Bless your kind soul

jpotts244 (approved):

src/lib/stitching/exchange/v2/stitching.ts
damassi: 😅 STITCHING 😓

PR #5778, opened 2024-06-04
chore: Add argument to exclude certain works from artwork recs DIA-652

jonallured:
This PR adds a new argument to the `artworksForUser` query called `excludeArtworkIds` and when it is provided then matching recommendations from vortex are excluded.

<details><summary> Here's some visual proof </summary>
<p>

Screenshot 2024-06-04 at 4 07 37 PM
Screenshot 2024-06-04 at 4 08 02 PM

</p>
</details> 

When I make a query for the recs without the new argument then I get a particular artwork. When I use the new argument and provide that artwork's id then it is skipped. 😎 

https://artsyproduct.atlassian.net/browse/DIA-652

/cc @artsy/amber-devs @isakelaris

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -16020,6 +16020,7 @@ type Query {
   artworksForUser(
     after: String
     before: String
+    excludeArtworkIds: [String] = []
     excludeDislikedArtworks: Boolean = false
     first: Int
     includeBackfill: Boolean!
@@ -21044,6 +21045,7 @@ type Viewer {
   artworksForUser(
     after: String
     before: String
+    excludeArtworkIds: [String] = []
     excludeDislikedArtworks: Boolean = false
     first: Int
     includeBackfill: Boolean!
--- a/src/schema/v2/artworksForUser/__tests__/helpers.test.ts
+++ b/src/schema/v2/artworksForUser/__tests__/helpers.test.ts
@@ -15,7 +15,7 @@ const mockLoaderFactory = (affinities) => {
   return loader
 }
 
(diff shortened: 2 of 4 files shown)


egdbear:
Thanks, Jon, for adding this 👑  
It looks good to me. However, I'm curious if defaulting `maxWorksPerArtist` to 1 would be significant and helpful for the campaign?

cavvia (commented):
Thanks for this! It's not clear from the PR and description if the limit in the `first` arg is applied before or after the exclusion filter. It should be applied after so that we can guarantee that we return the number of artworks requested if they are available for the user.

jonallured:
> It's not clear from the PR and description if the limit in the first arg is applied before or after the exclusion filter.

The way this works is we start by getting recs from Vortex. We pass the `first` argument from MP to Vortex. Then we suppress any artworks that are on the exclude list. Then we request those works from Gravity and other filtering is applied. Once we have the works back from Gravity we fill with backfill if there aren't enough.

joeyAghion:
> We pass the `first` argument from MP to Vortex.... we fill with backfill if there aren't enough.

If I understand correctly, that means we might append backfilled results even when there are more personalized recommendations available. Can we avoid this by requesting `first + excludeArtworkIds.length` results from Vortex?

jonallured:
> Can we avoid this by requesting first + excludeArtworkIds.length results from Vortex?

Brilliant suggestion @joeyAghion you win one internet point today!! 🏆

jonallured:
Alright going to merge!

PR #5782, opened 2024-06-06
feat(onyx-1042): Expose the associated My Collection artwork underneath a Submission

nickskalkin:
[Jira ticket](https://artsyproduct.atlassian.net/browse/ONYX-1042)

To more easily query for artwork metadata stored in My Collection via an associated Submission (in Convection), expose an artwork field under Submission within the GraphQL schema.

Example request:

```graphql
query X {
  submission(id: "250396") {
    myCollectionArtworkID
    
    artwork {
      internalID
      title
    }
  }
}
```

Response:

```json
{
  "data": {
    "submission": {
      "myCollectionArtworkID": "665f27035e8fac000efde37c",
      "artwork": {
        "internalID": "665f27035e8fac000efde37c",
        "title": "No Ball Games (Green) Test - MyC changes"
      }
    }
  }
}
```

### DIFF
--- a/src/lib/stitching/convection/v2/stitching.ts
+++ b/src/lib/stitching/convection/v2/stitching.ts
@@ -10,6 +10,7 @@ export const consignmentStitchingEnvironment = (
   extensionSchema: `
     extend type ConsignmentSubmission {
       artist: Artist
+      artwork: Artwork
     }
 
     extend type ConsignmentOffer {
@@ -29,6 +30,23 @@ export const consignmentStitchingEnvironment = (
             schema: localSchema,
             operation: "query",
             fieldName: "artist",
+            args: {
+              id,
+            },
+            context,
+            info,
+            transforms: convectionSchema.transforms,
+          })
+        },
+      },
+      artwork: {
+        fragment: `fragment SubmissionArtwork on ConsignmentSubmission { myCollectionArtworkID }`,
+        resolve: (parent, _args, context, info) => {
+          const id = parent.myCollectionArtworkID
+          return info.mergeInfo.delegateToSchema({
+            schema: localSchema,
+            operation: "query",
+            fieldName: "artwork",
             args: {
               id,
             },
--- a/src/lib/stitching/convection/v2/stitching.ts
+++ b/src/lib/stitching/convection/v2/stitching.ts
@@ -10,7 +10,7 @@ export const consignmentStitchingEnvironment = (
   extensionSchema: `
     extend type ConsignmentSubmission {
       artist: Artist
-      artwork: Artwork
+      myCollectionArtwork: Artwork
     }
 
     extend type ConsignmentOffer {
@@ -39,7 +39,7 @@ export const consignmentStitchingEnvironment = (
           })
         },
       },
-      artwork: {
+      myCollectionArtwork: {
         fragment: `fragment SubmissionArtwork on ConsignmentSubmission { myCollectionArtworkID }`,
         resolve: (parent, _args, context, info) => {
           const id = parent.myCollectionArtworkID
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -6939,6 +6939,7 @@ type ConsignmentSubmission {
   additionalInfo: String
   artist: Artist
   artistId: String!
+  artwork: Artwork
   assets: [ConsignmentSubmissionCategoryAsset]
   attributionClass: ConsignmentAttributionClass
   authenticityCertificate: Boolean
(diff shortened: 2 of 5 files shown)


src/lib/stitching/convection/v2/stitching.ts
```diff
@@ -38,6 +39,23 @@ export const consignmentStitchingEnvironment = (
...
+            fieldName: "artwork",
+            args: {
+              id,
+            },
+            context,
+            info,
+            transforms: convectionSchema.transforms,
+          })
+        },
+      },
```
joeyAghion, line 58: Help me understand this: I'd expect some reference to Gravity's data-loaders, in order to resolve this artwork. Where is that happening?

joeyAghion (commented):
The `artwork` name might be ambiguous, since a submission can have a "collected" artwork as well as one or more "listed" artworks.

src/lib/stitching/convection/v2/stitching.ts
nickskalkin: In this case we do not query Gravity directly, but rely on the `artwork` field which already exists in GraphQL schema. I don't know how it works under the hood, but by doing `delegateToSchema` and specifying to which schema you delegate (in this case it's `localSchema`, which is kinda... global?) you can get access to it's fields.

nickskalkin: I just found a good article which explains it better: https://the-guild.dev/graphql/stitching/docs/approaches/schema-extensions

Not so many details about stitching here, but it was fun to stumble upon and read https://artsy.github.io/blog/2018/12/11/GraphQL-Stitching/

nickskalkin:
> The artwork name might be ambiguous, since a submission can have a "collected" artwork as well as one or more "listed" artworks.

makes sense, I will call it myCollectionArtwork then to avoid confusion 👍

src/lib/stitching/convection/v2/stitching.ts
olerichter00: This looks wild :) I don't really understand **why** we use `delegateToSchema` instead of a data loader here 🤔 Wouldn't it be easier to use [artworkLoader](https://github.com/artsy/metaphysics/blob/051d02bc2fb655c7187f9aa1a9c4e06475f2339f/src/lib/loaders/loaders_without_authentication/gravity.ts#L57) instead?

MounirDhahri: @olerichter00 if I would have to guess, it's probably because we don't have `ConsignmentSubmission` interface in MP and we are stitching it from convection and as @joeyAghion mentioned above, `delegateToScheme` is the only way for it to tell how to treat queries. I guess this raises the question though of whether or not we want to keep it there.

nickskalkin: @MounirDhahri I think we should at some point switch from using Convection's GraphQL schema to REST (like we did with alerts recently in Gravity), since it's not recommended to do stitching anymore. It will require us to write `ConsignmentSubmission` type definition in MP. But for this PR I decided to not invest much time in it.

nickskalkin: @olerichter00 yes, I wrote above that we should probably at some point invest into de-stitching Convection. I'm just unsure if we should do it now or we can do it later. The change I made doesn't look that bad for me, there is literally the same approach for augmenting Submission type with `artist` a few lines above: https://github.com/artsy/metaphysics/blob/main/src/lib/stitching/convection/v2/stitching.ts#L24

MounirDhahri: I agree with your decision here - if we want to move over stitching here, we need a separate work stream for it.

olerichter00: > In this https://github.com/artsy/README/issues/459, Convection is at least quoted as a successful example.

Ah, sorry, this wasn't meant to be an argument against your solution. 

This quote argues for keeping stitching in Convection because it's working quite well:

> A more complex but arguably successful example is [Convection](https://github.com/artsy/metaphysics/blob/main/src/lib/stitching/convection/v2/stitching.ts). Once we had convection stitched in we were able to almost exclusively work within the Convection repo. There was very little friction and we were able to grow Convection's schema to [a reasonable size](https://github.com/artsy/convection/tree/main/app/graphql/types), encapsulating a good amount of functionality without needing to intercept and augment that much.

nickskalkin: oh, sorry @olerichter00, completely misunderstood you!

joeyAghion (approved):

src/lib/stitching/convection/v2/stitching.ts
olerichter00: > In this https://github.com/artsy/README/issues/459, Convection is at least quoted as a successful example.

Yeah, cutting this quote out of the context makes it pretty unclear and it could go in both directions 😃

PR #5800, opened 2024-06-17
chore: AMBER-724 - inquiry sale message

Mitch-Henson:
Solves [AMBER-724]

With the addition of a new type of sale message (added in https://github.com/artsy/gravity/pull/17819) we want to ensure that Metaphysics honours this and doesn't override it with a hardcoded value.

[AMBER-724]: https://artsyproduct.atlassian.net/browse/AMBER-724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/edition_set.ts
+++ b/src/schema/v2/edition_set.ts
@@ -99,7 +99,7 @@ export const EditionSetType = new GraphQLObjectType<any, ResolverContext>({
     },
     saleMessage: {
       type: GraphQLString,
-      resolve: ({ availability, price, forsale }) => {
+      resolve: ({ availability, price, forsale, sale_message }) => {
         // If it's a supported availability, just return it (capitalized).
         if (includes(EditionSetAvailabilities, availability)) {
           return capitalizeFirstCharacter(availability)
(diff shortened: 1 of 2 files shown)


src/schema/v2/edition_set.ts
```diff
@@ -112,7 +112,7 @@ export const EditionSetType = new GraphQLObjectType<any, ResolverContext>({
 
         // If its for sale (and no price), return 'Available'.
```
lidimayra, line 113: Minor: Should we just get rid of this comment above? It was already wrong before the changes on this PR (it wouldn't return `Available`, it would return `Contact for Price`).

Now it will return `sale_message` and it seems to make sense to just not have a comment here at all.

lidimayra (approved):
Minor non-blocking comment, but LGTM!!! 🚀

src/schema/v2/edition_set.ts
Mitch-Henson: I did not pay attention to that at all, thanks for bringing it up! Removed in [51acdbe](https://github.com/artsy/metaphysics/pull/5800/commits/51acdbea653ce386eb4273cd0404f8f935903ee8)

lidimayra (approved):

PR #5803, opened 2024-06-20
feat(onyx-1071): allow syncing my collection artwork data when creating submission

nickskalkin:
This PR allows to sync artwork data when specifying it's ID as `myCollectionArtworkID` argument to `createConsignmentSubmission` mutation.

Previously, `createConsignmentSubmission` was simply stitched from Convection. This PR adds a resolver for this mutation (keeping the same name). This mutation does two things:

- If  `myCollectionArtworkID` is specified - fetches artwork data using `artworkLoader` and prepares it to be passed to Convection.
- Calls `convectionCreateConsignmentSubmission` mutation (which was renamed from `createConsignmentSubmission`) with artwork data + mutation input arguments (mutation input arguments override artwork data if specified)

Example request (uses this [artwork](https://staging.artsy.net/collector-profile/my-collection/artwork/6654590c43bd0200125ea0b1)):

```graphql
mutation {
  createConsignmentSubmission(input: { artistID: "banksy", myCollectionArtworkID: "6654590c43bd0200125ea0b1" }) {
    consignmentSubmission {
      artistId
      myCollectionArtworkID
      title
      
      medium
      category
      year
      
      attributionClass
      
      editionNumber
      editionSize
      
      height
      width
      depth
      
      dimensionsMetric
      provenance
      
      locationCity
      locationCountry
      locationState
      locationCountryCode
      locationPostalCode
    }
  }
}
```

Response:

```json
{
  "data": {
    "createConsignmentSubmission": {
      "consignmentSubmission": {
        "artistId": "banksy",
        "myCollectionArtworkID": "6654590c43bd0200125ea0b1",
        "title": "Happy Choppers",
        "medium": "Screen print in colours on wove paper",
        "category": "Drawing, Collage or other Work on Paper",
        "year": "2003",
        "attributionClass": "LIMITED_EDITION",
        "editionNumber": "1",
        "editionSize": "2",
        "height": "10",
        "width": "20",
        "depth": "",
        "dimensionsMetric": "in",
        "provenance": "Stolen",
        "locationCity": "Berlin",
        "locationCountry": "Germany",
        "locationState": "Berlin",
        "locationCountryCode": null,
        "locationPostalCode": null
      }
    }
  }
}
```

### DIFF
--- a/src/lib/__tests__/artworkToSubmissionMapping.test.ts
+++ b/src/lib/__tests__/artworkToSubmissionMapping.test.ts
@@ -15,7 +15,7 @@ describe("artworkToSubmissionMapping", () => {
     const artwork = {
       artist: { id: "artist-id" },
       title: "Artwork Title",
-      dates: ["2020"],
+      date: "2020",
       medium: "Medium",
       category: "Photography",
       attribution_class: "Attribution Class",
@@ -32,30 +32,36 @@ describe("artworkToSubmissionMapping", () => {
         countryCode: "US",
         postalCode: "10001",
       },
+      signature: "back",
     }
 
     const submission = artworkToSubmissionMapping(artwork)
 
-    expect(submission).toEqual({
-      artistID: "artist-id",
-      title: "Artwork Title",
-      year: "2020",
-      medium: "Medium",
-      category: "PHOTOGRAPHY",
-      attributionClass: "ATTRIBUTION_CLASS",
-      editionNumber: 1,
-      editionSize: 2,
-      height: "10",
-      width: "10",
-      depth: "10",
-      dimensionsMetric: "in",
-      provenance: "Provenance",
-      locationCity: "City",
-      locationCountry: "Country",
-      locationState: "State",
-      locationCountryCode: "US",
-      locationPostalCode: "10001",
-    })
+    expect(submission).toMatchInlineSnapshot(
+      `
+      Object {
+        "artistID": "artist-id",
+        "attributionClass": "ATTRIBUTION_CLASS",
+        "category": "PHOTOGRAPHY",
+        "depth": "10",
+        "dimensionsMetric": "in",
+        "editionNumber": 1,
+        "editionSize": 2,
+        "height": "10",
+        "locationCity": "City",
+        "locationCountry": "Country",
+        "locationCountryCode": "US",
+        "locationPostalCode": "10001",
+        "locationState": "State",
+        "medium": "Medium",
+        "provenance": "Provenance",
+        "signature": true,
+        "title": "Artwork Title",
+        "width": "10",
+        "year": "2020",
+      }
+    `
+    )
   })
 })
 
--- a/src/lib/__tests__/artworkToSubmissionMapping.test.ts
+++ b/src/lib/__tests__/artworkToSubmissionMapping.test.ts
@@ -1,0 +1,90 @@
+import {
+  artworkToSubmissionCategory,
+  artworkToSubmissionMapping,
+} from "lib/artworkToSubmissionMapping"
+
+describe("artworkToSubmissionMapping", () => {
+  it("doesn't fail with empty artwork", () => {
+    const artwork = {}
+    const submission = artworkToSubmissionMapping(artwork)
+
+    expect(submission).toEqual({})
+  })
+
+  it("correctly maps artwork to submission", () => {
+    const artwork = {
+      artist: { id: "artist-id" },
+      title: "Artwork Title",
+      dates: ["2020"],
+      medium: "Medium",
+      category: "Photography",
+      attribution_class: "Attribution Class",
+      edition_sets: [{ available_editions: [1], edition_size: "2" }],
+      height: "10",
+      width: "10",
+      depth: "10",
+      metric: "in",
+      provenance: "Provenance",
+      collector_location: {
+        city: "City",
+        country: "Country",
+        state: "State",
+        countryCode: "US",
+        postalCode: "10001",
+      },
+    }
+
+    const submission = artworkToSubmissionMapping(artwork)
+
+    expect(submission).toEqual({
+      artistID: "artist-id",
+      title: "Artwork Title",
+      year: "2020",
+      medium: "Medium",
+      category: "PHOTOGRAPHY",
+      attributionClass: "ATTRIBUTION_CLASS",
+      editionNumber: 1,
+      editionSize: 2,
+      height: "10",
+      width: "10",
+      depth: "10",
+      dimensionsMetric: "in",
+      provenance: "Provenance",
+      locationCity: "City",
+      locationCountry: "Country",
+      locationState: "State",
+      locationCountryCode: "US",
+      locationPostalCode: "10001",
+    })
+  })
+})
+
+describe("artworkToSubmissionCategory", () => {
+  it("correctly maps artwork category to submission category", () => {
+    expect(artworkToSubmissionCategory("Architecture")).toEqual("ARCHITECTURE")
+    expect(artworkToSubmissionCategory("Design/Decorative Art")).toEqual(
+      "DESIGN_DECORATIVE_ART"
+    )
+    expect(
+      artworkToSubmissionCategory("Drawing, Collage or other Work on Paper")
+    ).toEqual("DRAWING_COLLAGE_OR_OTHER_WORK_ON_PAPER")
+    expect(
+      artworkToSubmissionCategory("Fashion Design and Wearable Art")
+    ).toEqual("FASHION_DESIGN_AND_WEARABLE_ART")
+    expect(artworkToSubmissionCategory("Installation")).toEqual("INSTALLATION")
+    expect(artworkToSubmissionCategory("Jewelry")).toEqual("JEWELRY")
+    expect(artworkToSubmissionCategory("Mixed Media")).toEqual("MIXED_MEDIA")
+    expect(artworkToSubmissionCategory("Painting")).toEqual("PAINTING")
+    expect(artworkToSubmissionCategory("Performance Art")).toEqual(
+      "PERFORMANCE_ART"
+    )
+    expect(artworkToSubmissionCategory("Photography")).toEqual("PHOTOGRAPHY")
+    expect(artworkToSubmissionCategory("Print")).toEqual("PRINT")
+    expect(artworkToSubmissionCategory("Sculpture")).toEqual("SCULPTURE")
+    expect(artworkToSubmissionCategory("Textile Arts")).toEqual("TEXTILE_ARTS")
+    expect(artworkToSubmissionCategory("Video/Film/Animation")).toEqual(
+      "VIDEO_FILM_ANIMATION"
+    )
+    expect(artworkToSubmissionCategory("Posters")).toEqual("OTHER")
+  })
+})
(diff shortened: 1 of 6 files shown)


src/lib/artworkToSubmissionMapping.ts
```diff
@@ -0,0 +1,70 @@
...
+  "JEWELRY",
+  "MIXED_MEDIA",
+  "PAINTING",
+  "PERFORMANCE_ART",
+  "PHOTOGRAPHY",
+  "PRINT",
+  "SCULPTURE",
+  "TEXTILE_ARTS",
+  "VIDEO_FILM_ANIMATION",
+]
```
dblandin, line 16: This makes me cringe a bit knowing that we already have [many versions of this mapping across our repos](https://github.com/search?q=org%3Aartsy+FASHION_DESIGN_AND_WEARABLE_ART&type=code) 😬 

I hope some day soon we won't need a different set of accepted categories between artworks and consignment submissions. Category (Medium Type) is one of those attributes that I'm hoping we can defer fully to the Artwork model and deprecate its presence in the Submission model entirely. 😅

src/lib/stitching/convection/v2/stitching.ts
```diff
@@ -90,5 +100,45 @@ export const consignmentStitchingEnvironment = (
...
+              }
+
+              const artworkSubmissionData = artworkToSubmissionMapping(artwork)
+
+              // use artwork data to fill in submission, but allow input to override
+              createSubmissionArgs = {
+                ...createSubmissionArgs,
+                input: {
+                  ...artworkSubmissionData,
+                  ...createSubmissionArgs.input,
```
dblandin, line 125: I think as written, if we currently have clients which execute `createConsignmentSubmission` while passing both `myCollectionArtworkID` and other artwork data input arguments, any input argument with a value of `null` will override any fetched value from the artwork.

Might be fine for this iteration but something to double-check on in Force and Eigen.

```typescript
const artworkSubmissionData = { title: "Penicillin G Procaine" }
const input = { title: null }

const merged = {
  ... artworkSubmissionData,
  ...input,
}
// { title: null }
```

dblandin (commented):
Looking good!

src/lib/stitching/convection/v2/stitching.ts
olerichter00: In Eigen currently, all values would be overridden by the form inputs when submitting a MyC artwork, and in Force, we would create the submission from the MyC artwork without sending any additional (artwork-related) data.

I guess overriding fields with `null` values is okay. It would allow clearing fields in the submission form that have been set on the MyC artwork.

src/lib/__tests__/artworkToSubmissionMapping.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        country: "Country",
+        state: "State",
+        countryCode: "US",
+        postalCode: "10001",
+      },
+    }
+
+    const submission = artworkToSubmissionMapping(artwork)
+
+    expect(submission).toEqual({
```
olerichter00, line 39: This is what I often use in these cases to save time by not having to write down the expected values.

```suggestion
    expect(submission).toMatchInlineSnapshot()
```

src/lib/artworkToSubmissionMapping.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const artworkToSubmissionCategory = (category: string) => {
+  if (!category) return
+
+  // replace all spaces, commas, and slashes with underscores, remove double underscores, and uppercase
+  const normalizedArtworkCategory = category
```
olerichter00, line 22: 👍

src/lib/artworkToSubmissionMapping.ts
```diff
@@ -0,0 +1,70 @@
...
+      locationCountryCode: artwork.collector_location?.countryCode,
+      locationPostalCode: artwork.collector_location?.postalCode,
+    })
+      // Remove undefined values
+      .filter(([_key, value]) => value !== undefined)
+      .reduce((obj, [key, value]) => {
+        obj[key] = value
+        return obj
+      }, {})
+  )
```
olerichter00, line 69: We could use [Lodash's `omit`](https://lodash.com/docs/4.17.15#omit) to filter out undefined values.

```suggestion
  const values = {
    artistID: artwork.artist?.id,
    title: artwork.title,
    year: (artwork.dates || [])[0]?.toString(),
    medium: artwork.medium,
    category: artworkToSubmissionCategory(artwork.category),
    attributionClass: artwork.attribution_class
      ?.replace(" ", "_")
      ?.toUpperCase(),
    editionNumber: artwork.edition_sets?.[0]?.available_editions?.[0],
    editionSize: artwork.edition_sets?.[0]?.edition_size
      ? +artwork.edition_sets?.[0]?.edition_size
      : undefined,
    height: artwork.height,
    width: artwork.width,
    depth: artwork.depth,
    dimensionsMetric: artwork.metric,
    provenance: artwork.provenance,
    locationCity: artwork.collector_location?.city,
    locationCountry: artwork.collector_location?.country,
    locationState: artwork.collector_location?.state,
    locationCountryCode: artwork.collector_location?.countryCode,
    locationPostalCode: artwork.collector_location?.postalCode,
  }

  return omit(values, isUndefined)
```

src/lib/stitching/convection/v2/stitching.ts
```diff
@@ -90,5 +100,45 @@ export const consignmentStitchingEnvironment = (
...
+          const myCollectionArtworkID = args.input?.myCollectionArtworkID
+          let createSubmissionArgs = args
+
+          // when myCollectionArtworkID is specified, use artwork data to fill in submission
+          if (myCollectionArtworkID) {
+            if (context.artworkLoader) {
+              const artwork = await context.artworkLoader(myCollectionArtworkID)
+
+              if (!artwork) {
+                throw new Error("Artwork not found")
```
olerichter00, line 115: Maybe this is easier to debug 🤷 

```suggestion
                throw new Error(`My Collection artwork (ID: ${myCollectionArtworkID}) not found.`)
```

olerichter00 (approved):
This looks great 🙌

src/lib/artworkToSubmissionMapping.ts
MounirDhahri: I was under the impression only these are the valid categories we accept
```typescript
{
  OTHER: "OTHER",
  PAINTING: "PAINTING",
  DRAWING_COLLAGE_OR_OTHER_WORK_ON_PAPER: "DRAWING_COLLAGE_OR_OTHER_WORK_ON_PAPER",
  PHOTOGRAPHY: "PHOTOGRAPHY",
  PRINT: "PRINT",
  SCULPTURE: "SCULPTURE",
}
```

src/lib/artworkToSubmissionMapping.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const artworkToSubmissionCategory = (category: string) => {
+  if (!category) return
+
+  // replace all spaces, commas, and slashes with underscores, remove double underscores, and uppercase
+  const normalizedArtworkCategory = category
+    .replace(/ /g, "_")
+    .replace(/,/g, "_")
+    .replace(/\//g, "_")
+    .replace(/\_\_/g, "_")
```
MounirDhahri, line 26: would this `.replace(/\W/g, "")` work as well?

src/lib/artworkToSubmissionMapping.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  if (ValidSubmissionCategories.includes(normalizedArtworkCategory)) {
+    return normalizedArtworkCategory
+  }
+
+  return "OTHER"
+}
+
+// artwork is an object taken from the Gravity API (artworkLoader)
+export const artworkToSubmissionMapping = (artwork) => {
```
MounirDhahri, line 37: I created a helper here [getInitialSubmissionFormValuesFromArtwork](https://github.com/artsy/eigen/blob/d87a2641244c4c65f853fa47a75796f7915a3975/src/app/Scenes/SellWithArtsy/ArtworkForm/Utils/getInitialSubmissionValuesFromArtwork.ts#L12) for some mapping

src/lib/artworkToSubmissionMapping.ts
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    Object.entries({
+      artistID: artwork.artist?.id,
+      title: artwork.title,
+      year: (artwork.dates || [])[0]?.toString(),
```
MounirDhahri, line 42: I am using `artwork.date` in the fronted - not sure if that's the best way

src/lib/stitching/convection/v2/stitching.ts
```diff
@@ -90,5 +100,45 @@ export const consignmentStitchingEnvironment = (
...
+              }
+
+              const artworkSubmissionData = artworkToSubmissionMapping(artwork)
```
MounirDhahri, line 118: We need to also specify that the source  is `MY_COLLLECTION` when the artwork comes from my collection.

MounirDhahri (commented):
Nice! looks neat. I aded just a few small comments but overall looks good to me.

src/lib/artworkToSubmissionMapping.ts
dblandin: Yeah – and that does align with available artwork mediums defined [here](https://github.com/artsy/metaphysics/blob/9ccad9c469d57897140a20e8bec62af4ff3d3ff9/src/lib/artworkMediums.ts#L4) in Metaphysics.

I think there's some friction between this MP-backed list, which I'd prefer to use in the UI, and the enum defined in Convection. I'm thinking soon we'll be able to defer to the MyC artwork for this medium type and we can simply stop using the [Convection-backed enum](https://github.com/artsy/convection/blob/45f7d5984bc26ad90a9ed8ef92bb28f60facccfd/app/graphql/types/category_type.rb#L4).

src/lib/stitching/convection/v2/stitching.ts
olerichter00: Good idea! I wonder though whether we should set it in the resolver if myCollectionArtworkId is present or if the client should pass it explicitly as an argument 🤔

MounirDhahri: I feel like if we are creating an artwork using a my collection artwork id, we can confidently tell this artwork source is my collection.

src/lib/artworkToSubmissionMapping.ts
nickskalkin: thanks @MounirDhahri. The logic is more or less the same, but the difference is that your code is backed by GraphQL response, and our code uses raw gravity response (artworkLoader). I've checked if our code misses any fields that your code has and I added missing `signature` field.

nickskalkin:
@olerichter00 I've addressed code review requests and this PR can be merged. Let's test on staging carefully after it is merged

src/lib/artworkToSubmissionMapping.ts
MounirDhahri: perfect - thanks

MounirDhahri (approved):

olerichter00:
> @olerichter00 I've addressed code review requests and this PR can be merged. Let's test on staging carefully after it is merged

@nickskalkin The PR in Force to create a new submission from a MyC artwork is ready for review as well: https://github.com/artsy/force/pull/13998

MounirDhahri:
@olerichter00 @nickskalkin I would also suggest to test things on app since we already send the id there as well

nickskalkin:
I will merge it tomorrow morning, don't want to merge today and get it deployed on prod without me knowing

PR #5804, opened 2024-06-20
fix: trailing comma in inquiries with questions

iskounen:
This PR prevents a trailing comma from appearing in conversation messages in CMS when questions are attached to an inquiry.

Screenshot 2024-06-20 at 9 15 45 AM

### DIFF
--- a/src/schema/v2/partner/partnerInquiryRequest.ts
+++ b/src/schema/v2/partner/partnerInquiryRequest.ts
@@ -40,7 +40,7 @@ export const InquiryRequestType = new GraphQLObjectType<any, ResolverContext>({
               : `• ${question?.question}`
           )
         })
-        if (message) lines.unshift([message, "\n"].join())
+        if (message) lines.unshift(`${message}\n`)
         return lines.join("\n")
       },
     },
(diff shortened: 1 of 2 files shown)


lidimayra (approved):
LGTM!!

Mitch-Henson (approved):
Huh, interesting. I wonder if there was a reason for this originally? Not sure if _Chesterton’s Fence_ applies here, but seems straightforward enough.

src/schema/v2/partner/partnerInquiryRequest.ts
```diff
@@ -40,7 +40,7 @@ export const InquiryRequestType = new GraphQLObjectType<any, ResolverContext>({
               : `• ${question?.question}`
           )
         })
-        if (message) lines.unshift([message, "\n"].join())
```
jonallured, line 43: My only theory is that `join` behaves differently in JS than it does in Ruby. In Ruby this would have just added a new line to the `message` string but in JS it inserts that comma. Maybe this was just a mistake? 🤷

lidimayra: Agreed!

And for a bit of context: the `formattedFirstMessage` didn't even exist in convos before the revamp (it was added as a workaround to include inquiry questions, which used to come directly from the email message in convos v1).

I understand its only purpose today is to be used to present those messages containing the inquiry question.
And looking at the examples on CMS convos on staging today for conversations containing inquiry questions, the comma really seems weird (much better without it)

iskounen: That's right. The `join` method adds a comma between elements in the array, unless you specify another string to use (as we do with `join("\n")` on the next line).

I think that this was just an oversight.

PR #5808, opened 2024-06-24
feat(EMI-1924): return isActive for partner offers

leamotta:
Part of https://artsyproduct.atlassian.net/browse/EMI-1924

Return isActive so clients can easily know if an offer is active or not and can render banners or messaging accordingly.
Basically reverting this revert https://github.com/artsy/metaphysics/pull/5586/commits/7aff9df4157c4ed48d1351261fce67f17c654748


narikazu (approved):
LGTM

PR #5823, opened 2024-07-03
chore: change type for swa specialist fields

dariakoko:
Changing the types of fields from `TYPE` to `new GraphQLNonNull(TYPE)` to simplify implementation on the frontend.


nickskalkin (approved):
makes sense 👍

PR #5830, opened 2024-07-05
Chore: Keeps gravity.graphql up to latest schema

jpotts244:
Chore: Keeps gravity.graphql up to latest schema to remove Location from being queried on Artwork type

Building off of @mzikherman's work here: https://github.com/artsy/gravity/pull/17883

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -12059,6 +12059,8 @@ type MarketingCollection {
   category: String!
   createdAt: ISO8601DateTime!
   credit: String
+
+  # HTML description, derived from the markdown description
   description: String
   descriptionMarkdown: String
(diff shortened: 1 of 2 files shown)


_schemaV2.graphql
```diff
@@ -12059,6 +12059,8 @@ type MarketingCollection {
   category: String!
   createdAt: ISO8601DateTime!
   credit: String
+
+  # HTML description, derived from the markdown description
   description: String
```
jpotts244, line 12064: Most of these are just "note" metadata updates 

The only real details is the removal of `PartnerLocation` on Artwork Type!

lidimayra (approved):

PR #5838, opened 2024-07-12
chore: upgrade to Graphql 15 and Graphql-tools 5 and related dependencies

MounirDhahri:
## Description
This PR is the first of potentially a bunch of PRs that has as a goal to bring metaphysics dependencies close to the latest in order to take advantage of the new APIs, security fixes and performance improvements that we released over the past 6/7 years.

There are a few dependencies here that were problematic:
- graphql-tools
- graphql
- graphql-relay
The above deps were problematic because we had a bunch of patches on top of them that are sometimes not applicable anymore: Either because they are not needed anymore 🙌 or because the maintainers changed how things work. This is why I decided to go just one version above and open this PR at the first commit that *looks* stable, before continuing. 

## What comes next
- **Short term**
  - Figure out how can we merge this safely? can we release this for a small percentage of our users first?

- **Mid/Long term**
  - express-graphql has been deprecated - the version that I upgraded to is new than what we had, but it's still 4 years old. The library grew into `graphql-http` [here](https://github.com/graphql/graphql-http) and I think we should switch to that as well.
  - Migrate to graphql-tools v6. It's also a major upgrade because now they're taking advantage of the mono repo design
	- There was no migration guide from v4 to v5. But there is a guide from v5 to v6.
	- Link: https://the-guild.dev/graphql/tools/docs/migration/migration-from-tools
  - Migrate to next versions of graphql-tools

## Notes for reviewers:
- The first commits are prefixed with ✅: that's to indicate they're functional commits that the project is still running fine at (tests and ts probably not)
- Some peer dependencies needed to be also upgraded - I followed their Changelog to make upgrades to make sure we don't miss any breaking changes.
- Enums were a pain to figure out, the reason behind that is that graphql-tools v4 had an issue in the implementation, we patched it, but we introduced a new pattern that didn't match how the repo maintainers wanted enum values and defaultValues to look like. I removed our patch in favour of following the docs and updated our enum usages everywhere.
- The stitching API changed, but not much, which is great for us.

##  How to review this PR?
1. As mentioned earlier, all commits are valid so you can start by that.
2. (**Recommended**) Guided tour by clicking 👉👉 [[here](https://github.com/artsy/metaphysics/pull/5838/commits/ee44e5312d3d99fb06189d5266caff154d448c9c#r1689334284)]


patches/graphql-relay+0.7.0.patch
```diff
@@ -0,0 +1,188 @@
+diff --git a/node_modules/graphql-relay/connection/arrayConnection.d.ts b/node_modules/graphql-relay/connection/arrayConnection.d.ts
```
MounirDhahri, line 1: Changes made after upgrading to react-relay 0.7.0. They're the same as before but some files moved here and there and I had to adapt to that. Luckily, most changes here are typescript and I think this might be the easier patch that we can get rid of.

package.json
```diff
@@ -60,16 +60,16 @@
...
     "google-libphonenumber": "^3.2.25",
-    "graphql": "14.5.4",
+    "graphql": "15.8.0",
     "graphql-depth-limit": "1.1.0",
-    "graphql-middleware": "1.2.6",
+    "graphql-middleware": "4.0.3",
     "graphql-parse-resolve-info": "^4.12.3",
-    "graphql-relay": "0.5.4",
-    "graphql-tools": "4.0.5",
+    "graphql-relay": "0.7.0",
```
MounirDhahri, line 71: I started initially by upgrading react-relay - that was a straightforward update because not much changed.
We had a patch for it that was not applicable anymore though. I replaced it with a new one and made sure to make the changes that in the new files. [See changes in patch 🔗 ](https://github.com/artsy/metaphysics/pull/5838/commits/ee44e5312d3d99fb06189d5266caff154d448c9c#r1689331684) 

▶️ **Next**

package.json
```diff
@@ -64,12 +64,12 @@
     "express-ipfilter": "0.3.1",
     "express-rate-limit": "3.2.0",
     "google-libphonenumber": "^3.2.25",
-    "graphql": "14.7.0",
+    "graphql": "15.8.0",
```
MounirDhahri, line 67: the troubles started here when I wanted to bump into `grapqhl` 15. The first `graphql-tools` version that allowed that, was 4.0.8. 
Unfortunately, that version brought many issues along with it from broken enums logic, different file structure, removed helpers. You can see me working on making adjustments all over the place now to satisfy the new API.
You can see me working on that here in this commit and this one https://github.com/artsy/metaphysics/pull/5838/commits/bfc66a7c4c47e1acf237876635b8898c343aff91

I also had to make adjustments to the peer dependencies later in the funnel in order to remove the resolution I made. This is the commit for it https://github.com/artsy/metaphysics/pull/5838/commits/faf692583f1b7e34434fb2076d0d9bd759b01dc3

▶️ **Next** [Fix peer dependencies](https://github.com/artsy/metaphysics/pull/5838/commits/77b611d5e1d976313b8fd99fe52c357ddf553837)

patches/graphql-tools+4.0.8.patch
```diff
@@ -14,18 +14,20 @@ index 5c18b05..1aaa8af 100644
...
          values.forEach(function (value) {
              newValues_1[value.name] = {
+-                value: value.value,
 +                // Simply pass-through the name of the enum entry such that the enum type in the original schema will
 +                // receive that and will be able to do its work as if it were never proxied.
-                 value: value.value,
++                value: value.name,
                  deprecationReason: value.deprecationReason,
                  description: value.description,
+                 astNode: value.astNode,
```
MounirDhahri: back to the enums drama - here I tried to bring things as they were (meaning `value = value.name`)

This was annoying because it took me more time.

Spoiler: I got rid of this patch entire eventually 🥇 


▶️ **Next** [Upgrade express-graphl](https://github.com/artsy/metaphysics/pull/5838/commits/77b611d5e1d976313b8fd99fe52c357ddf553837)

package.json
```diff
@@ -69,7 +69,7 @@
     "graphql-middleware": "4.0.3",
     "graphql-parse-resolve-info": "^4.12.3",
     "graphql-relay": "0.7.0",
-    "graphql-tools": "4.0.8",
+    "graphql-tools": "5",
```
MounirDhahri, line 72: I started to see the light at the end of the tunnel here. I figured out why not make this upgrade here to avoid all the previous enums drama since the community mentions that they fixed that in v5 and see what's breaking.
- I figured out there were some API changes around stitching, which I've made in this commit that you can scroll until end of the file to see them.
- Obviously, I had to removed graphql-tools patch next and updated the enums to be defined as described in the docs.

And that's it 🎉 
The next commits fix issues that are present in this commit:
[Issue with FeatureFlagSortEnums] https://github.com/artsy/metaphysics/pull/5838/commits/2a7c2978c5030a19c042ce264f4cc07c3ecc2f57
[Forgot to transform Date from exchange] https://github.com/artsy/metaphysics/pull/5838/commits/da9cc71b160ba6218c81fb0f6653a9ae5a4d7bf3
[**Fix broken stitching changes**] https://github.com/artsy/metaphysics/pull/5838/commits/38f51fdd10c12492d4000473e49317f7cb499a38 this one requires some attention - [**Go to issue**]

src/lib/stitching/vortex/__tests__/pricingContext.test.ts
```diff
@@ -138,9 +138,9 @@ describe("PricingContext type", () => {
     expect(JSON.parse(mockFetch.mock.calls[0][1].body).variables)
       .toMatchInlineSnapshot(`
       Object {
-        "_v1_artistId": "artist-id",
-        "_v2_category": "DRAWING_COLLAGE_OTHER_WORK_ON_PAPER",
-        "_v3_sizeScore": 225,
+        "_v0_artistId": "artist-id",
```
MounirDhahri, line 141: Q1: this seems to me as an implementation detail and not important in practice. I didn't want to apply a patch to update it - what do you think?

src/lib/stitching/vortex/__tests__/rankedStats.test.ts
```diff
@@ -33,11 +33,7 @@ describe("AnalyticsRankedStats type", () => {
     query($id: String!) {
       partner(id: $id) {
         analytics {
-          topArtworks: rankedStats(
-            first: 1
-            objectType: ARTWORK
-            period: ONE_YEAR
-          ) {
+          rankedStats(first: 1, objectType: ARTWORK, period: ONE_YEAR) {
```
MounirDhahri, line 36: **Situation 2**
something seems broken with aliases inside tests - I spent a few weeks working on it and didn't figure out the root cause of it. It might be something small but I can't figure it out. Also I tried this in production and the query is working well.

Screenshot 2024-07-23 at 17 39 54

🔙 [Go back](https://github.com/artsy/metaphysics/pull/5838/commits/38f51fdd10c12492d4000473e49317f7cb499a38#r1689374081)

.vscode/launch.json
```diff
@@ -9,7 +9,6 @@
       "address": "localhost",
       "stopOnEntry": false,
       "runtimeExecutable": null,
-      "runtimeVersion": "8.4.0",
```
MounirDhahri, line 12: tests were running with an old engine

joeyAghion:
Epic work and walk-through!

I see there are some schema changes, but they appear to just be ordering changes. If there aren't breaking changes, it might be revealing to run Integrity tests against a local Force+MP stack using this branch. (Also, obviously we can lean on staging for QA but that's _late_ in the pipeline depending on the types of problems you anticipate.)

joeyAghion:
Yes, we can build an alternate "canary" image, and add an alternate deployment, service, and ingress definition with a `canary-weight` setting like 10% (see [this branch for an example](https://github.com/artsy/gravity/compare/main...prod-es6-canary)).

But that's even later in the pipeline, usually we'd only reserve that for something we suspect will affect scalability or performance in subtle ways, because that requires production. What sort of problems do you anticipate?

damassi:
Future PRs that touch stitching are _not_ type-safe. We write in SDL, fragments are expressed in strings. All of these things = sensitivity for upgrades. If we could segment traffic and find a way to monitor to ensure health, that would be reassuring. (Of course, integrity first tho.) Just thinking that next steps in this MP upgrade effort could get quite hairy.

MounirDhahri:
Thanks all ❤️ 

> About integrity

I tested things in app and everything seems to be working fine - nice idea to try out integrity as well. I will do it.

About directing 10% of traffic to a "canary" image - thanks for pointing that out @joeyAghion and @damassi - I didn't know that was possible from within MP, that will be the next step after testing things out thoroughly.

anandaroop (approved):
Awesome! 👏🏽 

I also wanted some more confidence that there were no breaking changes in the schema, so I diffed the results of an introspection query  — this branch vs staging. And they appear to be effectively the same 👍🏽 Thank you for this painstaking work!

<details>
<summary>introspection diff</summary>

```graphql
# ran this against local and staging, then diffed
{
  __schema {
    queryType {
      fields {
        name
        args {
          name
        }
        type {
          name
        }
      }
    }
  }
}
```

A couple of trivial diffs, due to a change in alphabetization:
- `artistSeries` now  precedes `artists` instead of follows
-  `marketPriceInsights` now  precedes `marketingCategories` instead of follows
```diff
224a219,236
>           },
>           {
>             "name": "artistSeries",
>             "args": [{ "name": "id" }],
>             "type": { "name": "ArtistSeries" }
>           },
>           {
>             "name": "artistSeriesConnection",
>             "args": [
>               { "name": "after" },
>               { "name": "artistID" },
>               { "name": "artworkID" },
>               { "name": "before" },
>               { "name": "excludeIDs" },
>               { "name": "first" },
>               { "name": "last" }
>             ],
>             "type": { "name": "ArtistSeriesConnection" }
253,270d264
<           },
<           {
<             "name": "artistSeries",
<             "args": [{ "name": "id" }],
<             "type": { "name": "ArtistSeries" }
<           },
<           {
<             "name": "artistSeriesConnection",
<             "args": [
<               { "name": "after" },
<               { "name": "artistID" },
<               { "name": "artworkID" },
<               { "name": "before" },
<               { "name": "excludeIDs" },
<               { "name": "first" },
<               { "name": "last" }
<             ],
<             "type": { "name": "ArtistSeriesConnection" }
734a729,733
>           },
>           {
>             "name": "marketPriceInsights",
>             "args": [{ "name": "artistId" }, { "name": "medium" }],
>             "type": { "name": "MarketPriceInsights" }
758,762d756
<             "name": "marketPriceInsights",
<             "args": [{ "name": "artistId" }, { "name": "medium" }],
<             "type": { "name": "MarketPriceInsights" }
<           },
<           {
```
</details>

damassi (dismissed):

_schemaV2.graphql
```diff
@@ -35,102 +46,52 @@ type AccountRequest {
   notes: String
 }
 
-# Autogenerated input type of AddAssetsToConsignmentSubmission
-input AddAssetsToConsignmentSubmissionInput {
+# Autogenerated input type of AddAssetToConsignmentSubmission
+input AddAssetToConsignmentSubmissionInput {
```
mzikherman, line 50: The large diff in the `_schema` file worries me a bit. 'Noise' like comments or whitespace, etc is of course fine, but I wonder if there are 'real' changes (here, `Assets` -> `Asset`).

I don't think this file is actually used by Metaphysics at runtime - but it's used from clients, and so can help represent or call out issues if there are unexpected changes.

So in general, we should carefully look at updates that happen here.

mzikherman: (although is this a noisy git-diff / re-ordering issue? I see `Asset` version going to `Assets` below). I guess it would be nice to verify there's no 'real' diff here.

src/lib/stitching/exchange/transformers/replaceCommerceDateTimeType.ts
```diff
@@ -1,116 +1,35 @@
...
-          (isWrappingType(type) &&
-            type.ofType &&
-            type.ofType.name === "CommerceDateTime"))
-      ) {
-        return dateField.type
-      }
-      return type
-    })
+  if (
+    ["CommerceDateTime", "CommerceDate"].includes(fieldConfig.type.toString())
```
mzikherman, line 24: all of this Exchange-related stitching setup 😿

src/lib/stitching/vortex/__tests__/pricingContext.test.ts
mzikherman: I'd agree, internal-ish argument names the lib uses for stitching queries doesn't seem like something that would affect anything if the mechanism changes.

src/schema/v2/me/createUserInterestsMutation.ts
```diff
@@ -82,15 +82,21 @@ export const createUserInterestsMutation = mutationWithClientMutationId<
       },
     },
   },
-  mutateAndGetPayload: async (args, { meCreateUserInterestLoader }) => {
-    if (!meCreateUserInterestLoader) {
+  mutateAndGetPayload: async (
+    args,
+    { meCreateUserInterestLoader, meLoader }
+  ) => {
+    if (!meCreateUserInterestLoader || !meLoader) {
```
mzikherman, line 89: What's the `meLoader` needed for?

mzikherman (dismissed):
Epic epic epic.

Definitely down to get this merged! We can have a deploy-block for a bit (announce it in #dev), see what Integrity says, try to all do a bunch of staging QA, and so forth.

Canary deploy seems nice - but at the same time, a rollback is a rollback, someone needs to manually dial the canary back to 0 in that case, if there's an issue. Presumably we'd see metrics/errors/issues quickly if there was something really up (and could then rollback), so I'm even in favor of doing a regular deploy + close monitoring afterwards.

_schemaV2.graphql
mzikherman: Ah I see your comment @anandaroop  - and nice thinking on making a live introspection query and diff'ing those, vs. this somewhat-artificial generated file that seems to be subject to a lot of noise (which isn't even used by MP itself).

That totally gives me more confidence the large diff here is in fact just generative noise + git-diffing.

anandaroop: The diff you flagged makes me realize I didn't check _mutations_ though, lemme do that for good measure too…

mc-jones (dismissed):
Nothing to add that hasn't been said. But 👏 👏 👏 .

_schemaV2.graphql
anandaroop: With mutations, a lot more noise in the diff due to alphabetization, but with the help of https://novicelab.org/jsonabc/ — those also appear to be the same ☑️

olerichter00 (approved):
👏 👏 👏

MounirDhahri:
Thanks everyone for reviewing

I finally managed to run Integrity on local force pointing to this branch from MP and everything came green 🟢 

Screenshot 2024-07-30 at 12 40 14

I will merge this PR and keep monitoring MP.

PR #5843, opened 2024-07-16
chore(EMI-1959): load collectorSignals for artwork

erikdstock:
This pr adds a `CollectorSignals` field to artworks along with a loader on both artwork and filterArtworksConnection fields to query the backing data for these objects.

The backing data is not optimized yet and can take some time. These fields are therefore feature flagged. Later we'll need to pay attention to our query complexity here as the field could be accessed from list endpoints, using cached and batch loaders where possible to avoid needing to denormalize new fields on the backend.

Some more detailed description of how we resolve auctions fields [in this comment.](https://github.com/artsy/metaphysics/pull/5843#discussion_r1683516760)


src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,92 @@
...
+interface EnrichedSignals {
+  bidCount?: number
+  lotWatcherCount?: number
+  partnerOffer?: { endAt: string }
+}
+
+export const enrichArtworkWithCollectorSignals = async (artwork, ctx) => {
+  const enrichedSignals: EnrichedSignals = {}
+  artwork.collectorSignals = enrichedSignals
+
```
erikdstock, line 41: This function is used in the `artwork` and `filterArtworksConnection` resolvers to populate a `collectorSignals` object. I did not make efforts to batch requests yet, and some requests might be to the same endpoints that another field would request, but it works:
<details><summary>Single artwork (auction lot) signals query</summary>
<p>

image
</p>
</details> 

<details><summary>Single artwork acquireable (Partner offer) signals query</summary>
<p>
image
</p>
</details> 

<details><summary>Connection query - acquirable (partner offers)</summary>
<p>

image
</details> 
</p>

<details><summary>Connection query - auction lots</summary>
<p>
image
</p>
</details>

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,95 @@
...
+  if (inSale) {
+    const activeAuctions = await activeAuctionsForArtwork(artwork, ctx)
+    const activeAuction = activeAuctions[0]
+
+    if (activeAuction) {
+      const saleArtworkRequest = ctx.saleArtworkLoader({
+        saleId: activeAuction.id, // _id?
+        saleArtworkId: artworkId,
+      })
+      const [activeSaleArtwork] = await Promise.all([saleArtworkRequest])
```
erikdstock, line 56: This logic is a combination of the resolvers for [Artwork.isBiddable](https://github.com/artsy/metaphysics/blob/50c94eb6a3f1ceb1a8223041bbb15b6b5fdf7fbd/src/schema/v2/artwork/index.ts#L745-L761) and [Artwork.saleArtwork](https://github.com/artsy/metaphysics/blob/50c94eb6a3f1ceb1a8223041bbb15b6b5fdf7fbd/src/schema/v2/artwork/index.ts#L1508-L1528). Some concern here that we might add unnecessary extra fetches at this layer but i'm not sure how that works.

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,95 @@
...
+      const partnerOffer = await ctx
+        .mePartnerOffersLoader({
+          artwork_id: artworkId,
+          sort: "-created_at",
+          size: 1,
+        })
+        .then(({ body }) => body[0])
+      enrichedSignals.partnerOffer = partnerOffer
+    }
+  }
```
erikdstock, line 80: This is an alternative to the hook created on the client in https://github.com/artsy/force/pull/14132

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,95 @@
...
+    const { salesLoader } = ctx
+    const sales = await salesLoader({
+      id: artwork.sale_ids,
+      is_auction: true,
+      live: true,
+    })
+    return sales
+  }
+  return []
+}
```
erikdstock, line 95: Code organization could be improved, this is just POC.

src/schema/v2/artwork/collectorSignals.ts
mzikherman, line 56: Prob don't need `Promise.all` if it's just one promise

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,95 @@
...
+    // Handle signals for non-auction acquirable artworks
+  } else if (artwork.acquireable) {
+    if (ctx.mePartnerOffersLoader) {
+      const partnerOffer = await ctx
+        .mePartnerOffersLoader({
+          artwork_id: artworkId,
+          sort: "-created_at",
+          size: 1,
+        })
+        .then(({ body }) => body[0])
```
mzikherman, line 77: Shouldn't use any chaining over `async/await` stuff

src/schema/v2/artwork/index.ts
```diff
@@ -1986,6 +1991,12 @@ export const artworkResolver = async (_source, args, context, resolveInfo) => {
     resolveInfo
   )
 
+  const hasRequestedCollectorSignals = isFieldRequested(
```
mzikherman, line 1994: nice!

src/schema/v2/artwork/index.ts
```diff
@@ -1986,6 +1991,12 @@ export const artworkResolver = async (_source, args, context, resolveInfo) => {
...
+    "collectorSignals",
+    resolveInfo
+  )
```
erikdstock, line 1997: Followed nearby pattern here. I did not do the same for `filterArtworksConnection` yet - just marked it temp in the comments while naively always requesting collector signals.

erikdstock (commented):
Interested in feedback so far as this seems to be working!

src/schema/v2/artwork/collectorSignals.ts
damassi, line 95: (I know this is still wip, but you asked for feedback in slack :)

Thinking: A few of the patterns in here feel a bit idiosyncratic compared to the rest of MP, but maybe i'm not understanding exactly why a lot of this is necessary. Would you mind kinda breaking down in the PR description what the flow is architecture-wise in a bit more detail?

src/schema/v2/artwork/collectorSignals.ts
erikdstock: Correct. I would remove that if we get the 👍 to continue.

src/schema/v2/artwork/collectorSignals.ts
erikdstock: For sure - the auction signals logic was the hardest part. I tried to use the [saleArtworksAllLoader](https://github.com/artsy/metaphysics/blob/613ac7be7c685131905184beaee37134ee678759/src/lib/loaders/loaders_without_authentication/gravity.ts#L223) here but ran into some auth issues. Was hoping i could request `/sale_artworks` with `artwork_id` and `biddable_sale: true`. So what i am doing instead is

1. testing if the artwork has `sale_ids`, otherwise skipping
2. querying for those sale ids + `is_auction` + `live`, which follows the logic in [Artwork.biddable](https://github.com/artsy/metaphysics/blob/50c94eb6a3f1ceb1a8223041bbb15b6b5fdf7fbd/src/schema/v2/artwork/index.ts#L745-L761).
3. Then up above, below where this function is called, if we got a biddable auction, fetch the sale artwork for this sale/artwork combination [as we do in Artwork.saleArtwork](https://github.com/artsy/metaphysics/blob/50c94eb6a3f1ceb1a8223041bbb15b6b5fdf7fbd/src/schema/v2/artwork/index.ts#L1508-L1528)). If we get that back then we know it is visible to the user and we can use that for bid count.
4. For lot watcher count I re-used the `artwork.recent_saves_count` though we might want to get more specific in the future.

There are two things I could do to try and clean this up:
- make that first approach work, maybe i was missing some minor detail
- Skip 2 up above

src/schema/v2/filterArtworksConnection.ts
```diff
@@ -575,6 +572,9 @@ const filterArtworksConnectionTypeFactory = (
         )
       }
 
+      // temp
+      hits.forEach((artwork) => enrichArtworkWithCollectorSignals(artwork, ctx))
```
mzikherman, line 576: This feels problematic - a call to `enrich` for every artwork (often up to 30 for a full artwork grid) - each of which might fan out to _multiple_ fetches - this would block the initial request for a page featuring a grid (artist page, collect/collection page).

Depending on the approach taken - since eventually collector signals _would_ be asked for on every artwork grid/rail - it might be nice to move everything about collector signals to the client - maybe a bespoke endpoint that returns all the collector signals for a list of artwork id's which can be queried + rendered on the client - or perhaps queried all at once.

As-is this probably needs an `await` or something here as the `enrich` is async.

erikdstock: Oh you are right. that might explain some more empty results. I just wanted to see it working as inefficiently as possible. I'll change this to a `await Promise.all(hits.map(...)`.

Regarding doing it on the client vs here, that's what I am wondering too. We have a hook there.

src/schema/v2/artwork/collectorSignals.ts
damassi: This makes sense! Ty. Might be good to drop it in the PR description 👍

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,95 @@
...
+          enrichedSignals.lotWatcherCount = artwork.recent_saves_count
+        }
+        if (activeSaleArtwork.bidder_positions_count) {
+          enrichedSignals.bidCount = activeSaleArtwork.bidder_positions_count
+        }
+      }
+    }
+
+    // Handle signals for non-auction acquirable artworks
+  } else if (artwork.acquireable) {
```
starsirius, line 69: [`acquireable`](https://github.com/artsy/gravity/blob/b8777718374d6af17f651503aa5bb81f7b1b156e/app/models/concerns/for_sale.rb#L249-L251) excludes artworks that are not BNMO (e.g. MOI artworks). I think we want [`purchasable`](https://github.com/artsy/gravity/blob/b8777718374d6af17f651503aa5bb81f7b1b156e/app/models/concerns/for_sale.rb#L254-L256) instead here.

src/schema/v2/artwork/index.ts
starsirius: TIL!

src/schema/v2/filterArtworksConnection.ts
starsirius: Looking at Datadog ([example](https://app.datadoghq.com/apm/trace/4639396345898945969?colorBy=service&env=production&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=4234249432811012679&spanViewType=metadata&timeHint=1721380110116.001&view=spans)), it seems the API call to Gravity filter endpoint is one of the main bottle neck:

<details><summary>Screenshot</summary>
<p>

Screenshot 2024-07-19 at 9 34 04 AM

</p>
</details> 

Additional requests that got fanned out afterwards are surprisingly parallelized and not so bad:

<details><summary>Screenshot</summary>
<p>

Screenshot 2024-07-19 at 9 34 38 AM

</p>
</details> 

I totally agree that moving them to client side can help with unblocking rendering (and we'll need to raise it to design for the right placeholder patterns for this case), but I also feel the exact data fetched here (like bid count etc.) have good opportunities to be optimized (e.g. cached or even persisted).

To help us learn and determine the right next steps, maybe we can put this behind a feature flag (following the [blur hash pattern](https://github.com/artsy/metaphysics/blob/900d031a5df9e7ec3ef89412039c30d62c8f8c2f/src/schema/v2/image/index.ts#L79)) and we can observe the performance with real/staging data?

In the mean time, I'm going to give design a heads-up and get quick reaction re rendering the signals on the client-side.

mzikherman: Definitely agree - anything iterable/behind a feature flag is a good idea and 👍 from me.

Also yup good call out - client-side rendering does have that subtle design consideration of needing to likely support the loading state.

That chart is cool! Makes sense the ES call that front-loads any fanning out of enriching artwork calls is the slowest (and the parallelization of multiple smaller cacheable requests might be _more_ performant than one bigger batched endpoint that returns all signals for a list of artworks, for instance).

This further enriching behind a feature flag definitely makes this totally mergeable.

src/lib/featureFlags.ts
```diff
@@ -8,13 +8,17 @@ const { UNLEASH_API, UNLEASH_APP_NAME, UNLEASH_SERVER_KEY } = config
...
+  "emerald_signals-partner-offers",
+  "emerald_signals-auction-improvements",
+] as const
 
 type FeatureFlag = typeof FEATURE_FLAGS_LIST[number]
 
 let unleashClient: Unleash
 
-export function initilizeFeatureFlags() {
+export function initializeFeatureFlags() {
```
damassi, line 21: I do not know how to spell 🤦

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,121 @@
+import { GraphQLFieldConfig } from "graphql"
+import { GraphQLInt, GraphQLObjectType } from "graphql"
+import { ResolverContext } from "types/graphql"
+import { PartnerOfferToCollectorType } from "../partnerOfferToCollector"
+import { isFeatureFlagEnabled } from "lib/featureFlags"
+
+export const collectorSignals: GraphQLFieldConfig<any, ResolverContext> = {
```
damassi, line 7: `CollectorSignals`

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,121 @@
...
+  },
+}
+
+interface EnrichedSignals {
+  bidCount?: number
+  lotWatcherCount?: number
+  partnerOffer?: { endAt: string }
+}
+
+export const enrichArtworkWithCollectorSignals = async (artwork, ctx) => {
```
damassi, line 39: Can we break this into its own file?

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,121 @@
...
+  const enrichedSignals: EnrichedSignals = {}
+  artwork.collectorSignals = enrichedSignals
```
damassi, line 41: The order of operations here is a bit confusing and relies on mutation in place. Can we move this assignment til after all of the operations on `enrichedSignals` is complete?

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  const artworkId = artwork._id
+
+  const inSale = artwork.sale_ids?.length > 0
```
damassi, line 45: `isInSale`

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,121 @@
...
+      },
+      ctx
+    )
+
+    if (activeSaleArtwork) {
+      if (artwork.recent_saves_count) {
+        enrichedSignals.lotWatcherCount = artwork.recent_saves_count
+      }
+      if (activeSaleArtwork.bidder_positions_count) {
+        enrichedSignals.bidCount = activeSaleArtwork.bidder_positions_count
```
damassi, line 74: Can we keep these field names 1:1? Ie, if the field coming back from gravity is `recent_saves_count`, the field should be `recentSavesCount`. Ditto with `bidderPositionsCount`

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,121 @@
...
+        artwork_id: artworkId,
+        sort: "-created_at",
+        size: 1,
+      })
+
+      const partnerOffer = partnerOffers.body[0]
+
+      enrichedSignals.partnerOffer = partnerOffer
+    }
+  }
```
damassi, line 93:
```tsx
  artwork.collectorSignals = enrichedSignals
}
```

Better, however, would be to return `enrichedSignals` and avoid mutating arguments passed into functions:

```tsx
const collectorSignals = getCollectorSignals(artwork)
artwork.collectorSignals = collectorSignals 
```

This also allows us to simplify naming and future dev comprehension ('what does "enrich" mean?'); now we just get and assign.

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,121 @@
...
+  return artwork
+}
+
+const getActiveSaleArtwork = async ({ artworkId, saleIds }, ctx) => {
+  if (saleIds?.length > 0) {
+    const { salesLoader } = ctx
+    const sales = await salesLoader({
```
damassi, line 100: #superminor - ctx.salesLoader, like we're doing below

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,121 @@
...
+      is_auction: true,
+      live: true,
+    })
+
+    const activeAuction = sales[0]
+    if (!activeAuction) {
+      return null
+    }
+
+    if (activeAuction) {
```
damassi, line 111: No need for this condition, as we're exiting right above

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -0,0 +1,121 @@
...
+      const saleArtwork = await ctx.saleArtworkLoader({
+        saleId: activeAuction.id,
+        saleArtworkId: artworkId,
+      })
+
+      return saleArtwork
+    }
+  }
+  return null
```
damassi, line 120: We should exit early, right at the top of the function, and flatten things out.

src/schema/v2/artwork/index.ts
```diff
@@ -1998,6 +2008,14 @@ export const artworkResolver = async (_source, args, context, resolveInfo) => {
     return enrichedArtworks?.[0]
   }
 
+  if (hasRequestedCollectorSignals && artwork) {
+    const enrichedArtwork = await enrichArtworkWithCollectorSignals(
```
damassi, line 2012:
```tsx
if (!(hasRequestedCollectorSignals && artwork)) {
  return 
}

artwork.collectorSignals = await getCollectorSignals(artwork, context) 
```

src/schema/v2/filterArtworksConnection.ts
```diff
@@ -575,39 +573,50 @@ const filterArtworksConnectionTypeFactory = (
...
-          arrayLength: Math.min(
-            aggregations.total.value,
-            totalPages * gravityOptions.size
-          ),
-          sliceStart: gravityOptions.offset,
-        }
-      )
+      const collectorSignalsRequests = hasRequestedCollectorSignals
+        ? hits.map((artwork) => enrichArtworkWithCollectorSignals(artwork, ctx))
+        : [Promise.resolve()]
```
damassi, line 583: Can we exit early here instead of supplying a faux resolve? 

```tsx
if (!hasRequestedCollectorSignals) {
  return null
}
```

src/schema/v2/filterArtworksConnection.ts
```diff
@@ -575,39 +573,50 @@ const filterArtworksConnectionTypeFactory = (
...
 
-      connection.pageInfo.endCursor = pageToCursor(
-        gravityOptions.page + 1,
-        gravityOptions.size
-      )
+      return Promise.all(collectorSignalsRequests).then(() => {
```
damassi, line 585: The way we're implicitly relying on buried mutation here will get confusing in the future for those who didn't write the code, and lead to more complexity. Like, why is there no result being returned from `then`? Thinking... 

Instead, similar to elsewhere, we should stick to inputs and outputs, and keep it simple. If we're firing off a bunch of parallel requests to loaders via this collected array of promise invocations, we should be evaluating the _return results_ of said calls:

```tsx
if (!hasRequestedCollectorSignals) {
  return null
}

const collectorSignalsRequests = hits.map(artwork => getCollectorSignals(artwork, context))

try {
  const collectorSignalsData = await Promise.all(collectorSignalsRequests)
  
  // Do something with the results 
  const hitsWithCollectorSignals = hits.map((artwork, index) => {
    artwork.collectorSignals = collectorSignalsData[index]
    return artwork
  })
  
  const connection = connectionFromArraySlice(hitsWithCollectorSignals, ...)
  
  // etc
}
```

This keeps things more functional and linear, with less hidden operations. (Also, forgive if i'm misunderstanding some of the implementation somewhere!)

erikdstock: OK. I tried to split the difference on this one without shuffling thing around too much but I think I got it - the promises do return the artwork and we handle them with `Promise.all(requests).then(artworks => ...` - so the promises we wait for are returning the loaded artworks rather than just saying "I'm still promising here...". LMK if this is ok, hopefully it's otherwise good to merge.

src/lib/__tests__/isFieldRequested.test.ts
```diff
@@ -0,0 +1,109 @@
...
+jest.mock("graphql-parse-resolve-info", () => ({
+  parseResolveInfo: jest.fn(),
+  simplifyParsedResolveInfoFragmentWithType: jest.fn(),
+}))
+
+describe("isFieldRequested", () => {
+  const mockResolveInfo: GraphQLResolveInfo = {} as any // Mocked, as the actual value is not used due to mocking of dependencies
+
+  beforeEach(() => {
+    ;(parseResolveInfo as jest.Mock).mockReturnValue({})
```
damassi, line 17: We should define these as variables (for all related functions):
```
const mockParseResolveInfo = parseResolveInfo as jest.Mock
```

src/lib/fillers/__tests__/fetchCollectorSignals.test.ts
```diff
@@ -0,0 +1,136 @@
...
+const artwork = {
+  _id: "artwork1",
+  sale_ids: ["sale1"],
+  purchasable: true,
+  recent_saves_count: 10,
+}
+
+beforeEach(() => {
+  jest.clearAllMocks()
+})
```
damassi, line 29: Can we move these above definitions into the `describe` block and out of the global scope?

src/lib/fillers/fetchCollectorSignals.ts
```diff
@@ -0,0 +1,118 @@
+import { isFeatureFlagEnabled } from "lib/featureFlags"
```
damassi, line 1: is the folder name `fillers` correct? or is it supposed to be `filters`?

src/lib/fillers/fetchCollectorSignals.ts
```diff
@@ -0,0 +1,118 @@
...
+
+interface EnrichedSignals {
```
damassi, line 3:
```tsx
interface CollectorSignal {
  bidCount?: number
  lotWatcherCount?: number
  partnerOffer?: { endAt: string }
}
```

damassi (dismissed):
Updates look beautiful @erikdstock - thanks for working through those change requests!

src/schema/v2/filterArtworksConnection.ts
```diff
@@ -575,39 +573,57 @@ const filterArtworksConnectionTypeFactory = (
...
+            return fetchCollectorSignals(artwork, ctx).then(
+              (collectorSignals) => {
+                artwork.collectorSignals = collectorSignals
+                return artwork
+              }
+            )
+          })
+        : hits
+
+      return Promise.all(collectorSignalsRequests).then((artworks) => {
```
damassi, line 592: #non-blocking (but nice 😉 ): can we use async await here vs chained promises? 

```
try {
  const artworks = await Promise.all(collectorSignalRequests)
  ...
} catch (error) {
  // handle / return error 
}

Comment (mzikherman) [reply src/schema/v2/filterArtworksConnection.ts:592]: I actually want to caution against zealous `catch`. In MP, without any `catch` - you get pretty desirable behavior. There is Sentry reporting, the error shows up in `errors` in the response and the corresponding field is `null`, and the error in the response includes relevant information to help debug.

There are a lot of examples of `catch` that simply log 'Error fetching .... ' -> not actually providing any information, and the resulting field is usually still `null`, thus `catch` does nothing there. Or `catch` and throwing your own error there - more or less the same thing.

Basically you _shouldn't_ need a `catch` in most cases.

File: src/lib/fillers/fetchCollectorSignals.ts (hunk 25 above)
Comment (erikdstock) [reply src/lib/fillers/fetchCollectorSignals.ts:1]: So I put this in the folder next to its neighbor at the point it is called - [enrichArtworkWithPriceInsights()](https://github.com/artsy/metaphysics/blob/537f2d7155c41b775240dc11e33056ed75502ce0/src/lib/fillers/enrichArtworksWithPriceInsights.ts#L8-L26). These are the only two files in that folder.

However it doesn't really enrich anything any more and is really more of a complex loader. I could put it in the context - like [creating all the basic loaders](https://github.com/artsy/metaphysics/blob/a755fd94e5f1afd87cccce1e2d05f4cae2ce8940/src/lib/loaders/index.ts#L73-L96) and then constructing this last one from it.

As a first step I'm just moving it into lib/loaders. If we want to make this a first-class member of the context I can do that as well.

File: src/schema/v2/filterArtworksConnection.ts (hunk 27 above)
Comment (damassi) [reply src/schema/v2/filterArtworksConnection.ts:592]: Yah, i mean it (should) be understood that you have to return the error! but i know what you mean. Still should avoid chained vanilla promises. We can let it error out without try / catch.

Comment (erikdstock) [reply src/schema/v2/filterArtworksConnection.ts]: 👍 Updated the whole resolver to use async/await _except_ this little `hits.map` because it gave me a [probably incorrect] linter error `Possible race condition: `artwork.collectorSignals` might be reassigned based on an outdated value of `artwork.collectorSignals`.eslintrequire-atomic-updates`
image

So that part being quite small I just left as a chained promise.

File: src/lib/isFieldRequested.ts [hunk 28]
```diff
@@ -0,0 +1,46 @@
...
+  parseResolveInfo,
+  ResolveTree,
+  simplifyParsedResolveInfoFragmentWithType,
+} from "graphql-parse-resolve-info"
+
+export const isFieldRequested = (
+  field: string,
+  resolveInfo: GraphQLResolveInfo
+) => {
+  let parsedResolveInfoFragment
```
erikdstock, line 12: file was previously called `ifFieldRequested`. I made it accept dot notation for nested fields (not foolproof - like if you query by `node`!) so that I could use it in the artworksConnection resolver.

src/lib/loaders/collectorSignalsLoader.ts
```diff
@@ -0,0 +1,121 @@
...
+
+        partnerOffer = partnerOffers.body[0]
+      } catch (error) {
+        console.error(
+          "collectorSignalsLoader: Error fetching partner offers",
+          error
+        )
+      }
+    }
+  }
```
erikdstock: @mzikherman - Regarding your comment https://github.com/artsy/metaphysics/pull/5843#discussion_r1687206253:

> I actually want to caution against zealous catch. In MP, without any catch - you get pretty desirable behavior. There is Sentry reporting, the error shows up in errors in the response and the corresponding field is null, and the error in the response includes relevant information to help debug.
> 
> There are a lot of examples of catch that simply log 'Error fetching .... ' -> not actually providing any information, and the resulting field is usually still null, thus catch does nothing there. Or catch and throwing your own error there - more or less the same thing.
> 
> Basically you shouldn't need a catch in most cases.

I took this comment to be in reference to this field & so removed most catches in this file in 78b2a6c7d43f60d1d3c4679a31e0a6cf5f0d660e, but I wonder if this could cause entire artwork connections to be null out as it is part of the initial resolver. If I misinterpreted let me know and I'll put them back.

mzikherman: Try it and see! Put a `throw` in there and see what happens to the field/resolutions. This kind-of a thing, where you don't want the whole connection to `null` out (just the `collectorSignals` part of the `node`) _is_ totally something where you may want to `catch`.

src/schema/v2/filterArtworksConnection.ts
```diff
@@ -568,47 +566,65 @@ const filterArtworksConnectionTypeFactory = (
...
+        ? hits.map((artwork) => {
+            return collectorSignalsLoader(artwork, ctx).then(
+              (collectorSignals) => {
+                artwork.collectorSignals = collectorSignals
+
+                return artwork
+              }
+            )
+          })
+        : hits
```
damassi, line 592: I _think_ this will error out if `hasRequestedCollectorSignals` is false, since `hits` isn't an array of promises at that point.

(Otherwise all good re yr comment / this approach)

erikdstock: I wondered that too because it's awkward to work with, especially before switching to using `async` - but Promise.all does handle non-promises:
```
> Promise.all([1, 2, 3, Promise.resolve(4)])
  .then((results) => results.forEach((r, i) => console.log(i,r)))
> Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 612,
  [Symbol(trigger_async_id_symbol)]: 604
}

0 1
1 2
2 3
3 4
```

src/lib/loaders/collectorSignalsLoader.ts
erikdstock: Confirmed this breaks the entire parent resolution, so put it back. I think this is good to merge 👊

src/schema/v2/filterArtworksConnection.ts
damassi: Thats amazing. TIL.

PR #5844, opened 2024-07-17
fix(onyx-1110): fixes artwork.consignmentSubmission.internalID is always null

nickskalkin:
ArtworkConsignmentSubmission’s internalID was always `null` but now returns the submission's `id`.


olerichter00 (approved):
`rrrrr` 🌟

PR #5848, opened 2024-07-22
feat: add prototype schema for new home view sections

anandaroop:
A first increment towards the goal of a dynamic Art Advisor home view. 

Onyx has been leading a brainstorm and PoC schema for a flexible Home View, which can provide a list of content sections and their recommended components. 

See: [tech plan](https://www.notion.so/artsy/Art-Advisor-Dynamic-Home-View-3615106ef393433a9f16dcf13b2414f2) and [early share-out](https://artsy.slack.com/archives/C07ANEV7RNV/p1721410245056939)

This PR reflects this early, provisional thinking. (We expect this to evolve, perhaps drastically, as the requirements become clearer.)

The high-level goals are:

- the server controls the selection and ordering of such sections (based on personalization/curation/etc)
- the client knows enough to be able to passively render them using standardized components

The section content is provide via a new root field called `homeView`, which serves as a sandbox for us to iterate on the new schema without affecting / being affected by any existing home screen functionality.

A skeletal version of this, with schema from this PR being consumed by Eigen:

skel

Co-authored-by: @nickskalkin

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2288,7 +2288,7 @@ enum ArtistSorts {
 }
 
 # An artists rail section in the home view
-type ArtistsRailSection implements GenericHomeViewSection & Node {
+type ArtistsRailHomeViewSection implements GenericHomeViewSection & Node {
   # The component that is prescribed for this section
   component: HomeViewComponent
 
@@ -3151,7 +3151,7 @@ enum ArtworkSorts {
 }
 
 # An artwork rail section in the home view
-type ArtworksRailSection implements GenericHomeViewSection & Node {
+type ArtworksRailHomeViewSection implements GenericHomeViewSection & Node {
   # The component that is prescribed for this section
   component: HomeViewComponent
 
@@ -11358,7 +11358,7 @@ enum HomeViewComponentType {
   ARTWORKS_RAIL
 }
 
-union HomeViewSection = ArtistsRailSection | ArtworksRailSection
+union HomeViewSection = ArtistsRailHomeViewSection | ArtworksRailHomeViewSection
 
 # A connection to a list of items.
 type HomeViewSectionConnection {
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2288,9 +2288,15 @@ enum ArtistSorts {
 }
 
 # An artists rail section in the home view
-type ArtistsRailSection implements GenericSection {
+type ArtistsRailSection implements GenericSection & Node {
   # The component that is prescribed for this section
   component: Component
+
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
 
   # A stable identifier for this section
   key: String!
@@ -3145,9 +3151,15 @@ enum ArtworkSorts {
 }
 
 # An artwork rail section in the home view
-type ArtworksRailSection implements GenericSection {
+type ArtworksRailSection implements GenericSection & Node {
   # The component that is prescribed for this section
   component: Component
+
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
 
   # A stable identifier for this section
   key: String!
@@ -10974,6 +10986,12 @@ interface GenericSection {
   # The component that is prescribed for this section
   component: Component
 
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
+
   # A stable identifier for this section
   key: String!
 
@@ -11334,6 +11352,12 @@ type HomePageSalesModule {
 type HomeView {
   # A list of sections on the home view
   sections: [Section!]!
+  sectionsConnection(
+    after: String
+    before: String
+    first: Int
+    last: Int
+  ): SectionConnection
 }
 
 type IdentityVerification {
@@ -18509,6 +18533,26 @@ enum SecondFactorKind {
 union SecondFactorOrErrorsUnion = AppSecondFactor | Errors | SmsSecondFactor
 
 union Section = ArtistsRailSection | ArtworksRailSection
+
+# A connection to a list of items.
+type SectionConnection {
+  # A list of edges.
+  edges: [SectionEdge]
+  pageCursors: PageCursors!
+
+  # Information to aid in pagination.
+  pageInfo: PageInfo!
+  totalCount: Int
+}
+
+# An edge in a connection.
+type SectionEdge {
+  # A cursor for use in pagination
+  cursor: String!
+
+  # The item at the end of the edge
+  node: Section
+}
 
 # A piece that can be sold
 interface Sellable {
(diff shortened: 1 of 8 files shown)


_schemaV2.graphql
```diff
@@ -2287,6 +2287,24 @@ enum ArtistSorts {
   TRENDING_DESC
 }
 
+# An artists rail section in the home view
+type ArtistsRailSection implements GenericHomeViewSection & Node {
```
nickskalkin, line 2291: Wondering if these sections also should be prefixed with `HomeView`

nickskalkin (dismissed):
👍

src/schema/v2/homeView/HomeViewComponent.ts
```diff
@@ -0,0 +1,31 @@
+import { GraphQLEnumType, GraphQLObjectType } from "graphql"
+
+// known standard component types
+
+const HomeViewComponentTypeType = new GraphQLEnumType({
+  name: "HomeViewComponentType",
+  description: "Known component types for use in home view",
+  values: {
+    ARTWORKS_RAIL: {
+      value: "artworks_rail",
```
damassi, line 10: We should avoid snake case

src/schema/v2/homeView/HomeViewSection.ts
```diff
@@ -0,0 +1,89 @@
...
+    ...InternalIDFields,
+    key: {
+      type: GraphQLNonNull(GraphQLString),
+      description: "A stable identifier for this section",
+    },
+    title: {
+      type: GraphQLNonNull(GraphQLString),
+      description: "A display title for this section",
+    },
+    component: {
```
damassi, line 27: maybe something more generic like `section` would be better? (Thinking its best to keep our data layer as generic as possible, vs encoding react idioms)

src/schema/v2/homeView/HomeViewSection.ts
```diff
@@ -0,0 +1,89 @@
...
+      type: HomeViewComponent,
+      description: "The component that is prescribed for this section",
+    },
+  },
+  isTypeOf: (value) => {
+    return value.component.type === "artworks_rail"
```
damassi, line 56: `artworksRail`

src/schema/v2/homeView/HomeViewSection.ts
```diff
@@ -0,0 +1,89 @@
...
+    },
+    component: {
+      type: HomeViewComponent,
+      description: "The component that is prescribed for this section",
+    },
+  },
+  isTypeOf: (value) => {
+    return value.component.type === "artists_rail"
+  },
+})
```
damassi, line 82: It might be a little soon given this is a POC, but Rather than duplicating these types we could create a factory function that returns the value based on configuration values:

```tsx
const createSection = ({ name, description, isTypeOf }) => {
  return new GraphQLObjectType<any, ResolverContext>({
    name,
    description,
    interfaces: [GenericHomeViewSectionInterface, NodeInterface],
    fields: {
      ...InternalIDFields,
      key: {
        type: GraphQLNonNull(GraphQLString),
        description: "A stable identifier for this section",
      },
      title: {
        type: GraphQLNonNull(GraphQLString),
        description: "A display title for this section",
      },
      component: {
        type: HomeViewComponent,
        description: "The component that is prescribed for this section",
      },
    },
    isTypeOf,
  })
}

const ArtistsRailSection = createSection({
  name: "ArtistsRailSection",
  description: "An artists rail section in the home view",
  isTypeOf: (value) => value.component.type === "artistsRail"
})
```

_schemaV2.graphql
```diff
@@ -2287,6 +2287,24 @@ enum ArtistSorts {
...
+  component: HomeViewComponent
+
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
+
+  # A stable identifier for this section
+  key: String!
```
iskounen, line 2302: My first thought was "why do we need a key when we have an ID?" and then I realized that this was more like a slug. Maybe this should be `slug` instead?

_schemaV2.graphql
```diff
@@ -10930,6 +10966,24 @@ type GeneMeta {
   description: String!
 }
 
+# Abstract interface shared by every kind of home view section
+interface GenericHomeViewSection {
```
iskounen, line 10970: It would be cool to have a `weight` field on this type that can tell us how to position a section relative to its siblings inside of the home page (or other containers we come up with).

We could then use this `weight` to sort and paginate the sections returned by a `HomeViewSectionConnection`.

src/schema/v2/homeView/HomeViewSection.ts
iskounen: +1, it's unfortunate that React has monopolized the word "component" in my mind, but I totally assumed that this pointed to a specific type of React component that should be rendered by the client.

iskounen: +1, it's unfortunate that React has monopolized the word "component" in my mind, but I totally assumed that this pointed to a specific type of React component that should be rendered by the client.

src/schema/v2/homeView/HomeViewComponent.ts
anandaroop: Fixed in 397eb2849

_schemaV2.graphql
anandaroop: Fixed in 593d9dabc

src/schema/v2/homeView/HomeViewSection.ts
anandaroop: Fixed in bccd7cac4

anandaroop (commented):
Thanks for the comments. I've replied or addressed below…

src/schema/v2/homeView/HomeViewComponent.ts
damassi: Agreed its all over the place, and fixing stuff like this en-masse is delicate... Just thinking of JS norms, and small nudges forward in that way (similar to how we've evolved to name files in `camelCaseFormat.ts`.

damassi (approved):
Looks good! 👍 👍

anandaroop:
Going for it — nothing precious about this, let's feel free to iterate 🚀

_schemaV2.graphql
iskounen: 👍 on having the server return sections in the desired order.

Regarding the scenario you were thinking about: I envisioned clients always requesting sections ordered by `weight`, so the scenario you described shouldn't happen. However, having queries for sections always depend on a weight would be super limiting for us, so I think your idea of doing it server-side is more interesting.

_schemaV2.graphql
```diff
@@ -11279,6 +11333,53 @@ type HomePageSalesModule {
...
+    before: String
+    first: Int
+    last: Int
+  ): HomeViewSectionConnection
+}
+
+# A component specification
+type HomeViewComponent {
+  # Which component type to render
+  type: HomeViewComponentType
```
araujobarret, line 11349: (late here sorry 🙇 ): Is there any reason we're following a pattern different than GraphQL __typename union here?
My question is more about creating a different deviation for a type's `type` rather than using the standards we have.

olerichter00: cc @MounirDhahri

PR #5851, opened 2024-07-24
feat: add submission copy to GraphQL schema

dariakoko:
As discussed in [this thread](https://artsy.slack.com/archives/D032AFMTJES/p1721828639408619), adding submission copies to MP
Addressed [ONYX-1135](https://artsyproduct.atlassian.net/browse/ONYX-1135) 

[ONYX-1135]: https://artsyproduct.atlassian.net/browse/ONYX-1135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
+++ b/src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
@@ -86,13 +86,60 @@ const ArtworkConsignmentSubmissionType = new GraphQLObjectType<
         description: "Submission state label visible to the user.",
         resolve: ({ state }) => {
           switch (state.toLowerCase()) {
-            case "approved":
-            case "rejected":
+            case "draft":
+            case "hold":
             case "closed":
-            case "published":
-              return "Evaluation Complete"
+              return null
+            case "approved":
+              return "Approved"
+            case "rejected":
+              return "Submission Unsuccessful"
             default:
               return "In Progress"
+          }
+        },
+      },
+      stateLabelColor: {
+        type: GraphQLString,
+        resolve: ({ state }) => {
+          switch (state.toLowerCase()) {
+            case "rejected":
+              return "black60"
+            default:
+              return "black100"
+          }
+        },
+      },
+      actionLabel: {
+        type: GraphQLString,
+        description:
+          "Action label asks the user to poseed with the submission.",
+        resolve: ({ state }) => {
+          switch (state.toLowerCase()) {
+            case "draft":
+              return "Complete Submission"
+            case "approved":
+              return "Complete Listing"
+            default:
+              return null
+          }
+        },
+      },
+      buttonLable: {
+        type: GraphQLString,
+        description: "Button label visible to the user.",
+        resolve: ({ state }) => {
+          switch (state.toLowerCase()) {
+            case "draft":
+              return "Complete Submission"
+            case "approved":
+              return "Add Additional Information"
+            case "submitted":
+            case "published":
+              /* case "RESUBMITTED": */
+              return "Edit Submission"
+            default:
+              return null
           }
         },
       },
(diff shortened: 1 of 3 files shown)


olerichter00 (approved):
Looks great! 🌟 

It would be great if we could also get the copy for listed submissions ([this case](https://github.com/artsy/force/blob/742930690d41d7119cd9848cd9b6917e6c4f2f69/src/Apps/MyCollection/Routes/MyCollectionArtwork/Components/MyCollectionArtworkSWASubmissionStatus.tsx#L226)). In Force I've done this by looking at the artwork's `listedArtworks` which should include at least one (published?) artwork in case the submission is already listed.

src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
```diff
@@ -86,28 +86,84 @@ const ArtworkConsignmentSubmissionType = new GraphQLObjectType<
...
+        type: GraphQLString,
+        resolve: ({ state }) => {
+          switch (state.toLowerCase()) {
+            case "rejected":
+              return "black60"
+            default:
+              return "black100"
+          }
+        },
+      },
```
olerichter00, line 112: What is the `stateLabelColor` for? Is it the button? Maybe for the button we could define the color like this:

```suggestion
      buttonVariant: {
        type: GraphQLString,
        resolve: ({ state }) => {
          switch (state.toLowerCase()) {
            case "approved":
            case "draft":
              return "primaryBlack"
            default:
              return "secondaryBlack"
          }
        },
      },
```

src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
```diff
@@ -86,28 +86,84 @@ const ArtworkConsignmentSubmissionType = new GraphQLObjectType<
...
+              return "Add Additional Information"
+            case "submitted":
+            case "published":
+              /* case "RESUBMITTED": */
+              return "Edit Submission"
+            default:
+              return null
+          }
+        },
+      },
```
olerichter00, line 145: Maybe we could define the color for the button like this:

```suggestion
      },
      buttonVariant: {
        type: GraphQLString,
        resolve: ({ state }) => {
          switch (state.toLowerCase()) {
            case "approved":
            case "draft":
              return "primaryBlack"
            default:
              return "secondaryBlack"
          }
        },
      },
```

src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
```diff
@@ -86,28 +86,84 @@ const ArtworkConsignmentSubmissionType = new GraphQLObjectType<
...
+            case "draft":
+              return "Complete Submission"
+            case "approved":
+              return "Complete Listing"
+            default:
+              return null
+          }
+        },
+      },
+      buttonLable: {
```
olerichter00, line 128:
```suggestion
      buttonLabel: {
```

src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
dariakoko: It's the color for `stateLabel`, it's black100 for all statuses and only for Submission Unsuccessful (rejected state) it's black60
Not sure if it's an overkill or not

dariakoko:
Sure, we can add the copy for listed state as well @olerichter00 
Do we need to add another state in order to do that? I see we have `open: "Submission in progress"`, that looks like it could be the same as `listed`. Or, maybe I misused `published` state?

dariakoko:
> Sure, we can add the copy for listed state as well @olerichter00 Do we need to add another state in order to do that? I see we have `open: "Submission in progress"`, that looks like it could be the same as `listed`. Or, maybe I misused `published` state?

I see [here](https://www.notion.so/artsy/Submission-Status-MyC-Artwork-7593508a62df447290deeeee34116899?pvs=4) that this is a new one. Solved ✅

nickskalkin (approved):
👍

PR #5852, opened 2024-07-24
fix(EMI-1959): put collectorSignalsLoader in collectorSignals field resolver

erikdstock:
**FIX**
Completes [EMI-1959]

Hopefullly this should simplify and fix some code for loading collector signals. Use a traditional resolver method instead of trying to pre-fetch them based on the field being requested.

[EMI-1959]: https://artsyproduct.atlassian.net/browse/EMI-1959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -22,9 +23,14 @@ export const CollectorSignals: GraphQLFieldConfig<any, ResolverContext> = {
...
+  description: "Collector signals on artwork",
 
-  resolve: async (artwork, {}, _ctx) => {
-    return artwork.collectorSignals
+  resolve: async (artwork, {}, ctx) => {
+    try {
+      const collectorSignals = await collectorSignalsLoader(artwork, ctx)
+      return collectorSignals
+    } catch (e) {
+      console.error("Error in CollectorSignals resolver: ", e)
```
mzikherman, line 33: Is there a reason to `catch` and log, vs. not catching?

erikdstock: When I didn't catch and just threw an error in the resolver I got an error instead of data in my query.

mzikherman: Screen Shot 2024-07-24 at 5 11 52 PM

You shouldn't `catch` - (per my comment yesterday).

Let the standard error catching happen. It will resolve `collectorSignals: null`, and everything will be fine. Don't `catch` unless you really _need_ to.

erikdstock: Ah, you are totally right. I knew about `data`/`errors`, should have checked closer.

erikdstock:
Pushed a few more changes to clean up complexity that was no longer necessary
- combined file for signals loader
- consolidated tests
- removed now-unnecessary catches since we aren't loading things in the general artwork/connection resolvers any more

PR #5856, opened 2024-07-25
chore: Enable address type filter on partner locations query

lidimayra:
I just realised I forgot to add this when writing [AMBER-828], but we need to be able to filter partner locations by address type so we can use this query in Volt (the one currently consumed by orders filters only partner locations with a `Business` address type)

```graphql
{
    id
    locationsConnection(first: 5, private: true, addressType: BUSINESS) {
      edges {
        node {
          address
          city
        }
      }
    }
  }
}
```

Ref for the same filter in Volt [here](https://github.com/artsy/volt/blob/cefe12e3145bd76dc6f094b1dc92d84cb5bf2459/app/models/partner.rb#L234)

@artsy/amber-devs 

[AMBER-828]: https://artsyproduct.atlassian.net/browse/AMBER-828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -108,6 +108,12 @@ type addOrderedSetItemSuccess {
   setItem: OrderedSetItem
 }
 
+enum addressType {
+  BUSINESS
+  OTHER
+  TEMPORARY
+}
+
 type addUserRoleFailure {
   mutationError: GravityMutationError
 }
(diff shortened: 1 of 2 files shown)


jonallured (approved):

damassi:
thank you @lidimayra !

PR #5860, opened 2024-07-25
test: add base tests for marketing collections

egdbear:
This PR adds the base tests for marketing collections.


mzikherman (approved):
Awesome! This is what I was expecting - flipping the flag to be enabled in a test. This ensures we have test coverage when we flip the flag for real.

What I've seen in the past is:
  - no tests around a new flag (and by default flag is off in tests)
  - flag switched on in prod -> essentially we have untested code!
  - ...time passes...
  - we decide to go back and clean up the flag -> now by default flag is on, and only then realize we need to update tests, which is of course a little bit late in the process

Nicely done.

PR #5864, opened 2024-07-26
POC: Maybe an RFC: Checkout flow in schema

erikdstock:
This is a partly-implemented concept of how we could create a `Checkout` type in metaphysics that could drive UIs in both Web and Apps. It's only a proof-of-concept - I used an ES class but that isn't necessary.

The type mainly exposes a few token properties plus a dynamic list of stages which would represent individual forms or interactions in a checkout (they can have saved values (be complete) or not, and they can be hypothetical future steps (`knownRequired=false`). Haven't put much thought into mutations, how this would work with things like failed payment (is it a stage?), etc. The idea is to make it easier to build interfaces for checkout and easier to support future order types.

Also, stages are not well-typed. mostly just strings of json where each step could have its own kind of `savedValues` and `options`.

This could be an approach to investigate the next time we consider bigger checkout work.

### DIFF
--- a/src/schema/v2/checkout.ts
+++ b/src/schema/v2/checkout.ts
@@ -1,0 +1,413 @@
+import {
+  GraphQLFieldConfig,
+  GraphQLString,
+  GraphQLObjectType,
+  graphql,
+  GraphQLBoolean,
+  GraphQLList,
+} from "graphql"
+import gql from "lib/gql"
+import { ResolverContext } from "types/graphql"
+import { schema } from "schema/v2"
+
+const CheckoutStageType: GraphQLObjectType = new GraphQLObjectType({
+  name: "CheckoutStage",
+  fields: {
+    stage: {
+      type: GraphQLString,
+      description: "The stage of the checkout",
+      resolve: (stage) => {
+        return stage.stage
+      },
+    },
+    knownRequired: {
+      type: GraphQLBoolean,
+      description:
+        "Indicates if the stage is known to be required based on current order data",
+      resolve: (stage) => {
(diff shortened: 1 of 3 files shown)


src/schema/v2/checkout.ts
```diff
@@ -0,0 +1,413 @@
...
+        return JSON.stringify(stage.savedValues)
+      },
+    },
+    options: {
+      type: GraphQLString,
+      description: "The options for the stage",
+      resolve: (stage) => {
+        return JSON.stringify(stage.options)
+      },
+    },
```
erikdstock, line 44: These values could have better types based on what stage it is, and checkout form fields could match them.

src/schema/v2/checkout.ts
```diff
@@ -0,0 +1,413 @@
...
+        return checkout.order.internalID
+      },
+    },
+    helloOrder: {
+      type: GraphQLString,
+      description: "Hello to the order",
+      resolve: (checkout) => {
+        return "Hello " + checkout.order?.state
+      },
+    },
```
erikdstock, line 64: There might be a better way to do this

erikdstock:
Added a concept of required actions like a failed payment
image

erikdstock:
I'm gonna close this for now since we don't have immediate plans to work on it. If we come back to it I wondered if different checkout types would be useful for orders in different states, like `...on PendingOrder { with form values }` vs `...on ApprovedOrder { with shipping details }`.

Given how easily this went when working off of existing force data I think it would be a good approach to take ... some day.

PR #5865, opened 2024-07-27
feat: add artworks connection to ArtworksRailHomeViewSection

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-1142

Co-authored-by: @nickskalkin 
Co-authored-by: @MounirDhahri 

## Description

This furthers the home view schema spike by confirming our ability to resolve the actual artwork data that is required by any section of type `ArtworksRailHomeViewSection`

First we evolve the logic for serving sections:

- 7638abddf54a8eea2f34f38dfbef46fe1df2a052 — adds some fake/placeholder logic for determining a list of sections for a given user, thus getting away from the static array of sections we were serving previously.

Then we get to the real business of this PR:

- 9f7f4f39e2f859b334604b03f61de28bc57da97a — adds an `artworksConnection` field to the `ArtworksRailHomeViewSection` section type, which knows how to resolve its artworks, depending on what kind of artworks rail section it is

- 4eda3207b50ebff371c99e61f71bd5cdb1033b8f — for one particular kind of `ArtworksRailHomeViewSection` — the one with `key: NEW_WORKS_FOR_YOU` — we then teach it how to source artworks from the same resolver that is used by the `artworksForUser` root field

With this logic in place, the request/response now looks like…

### Example query

```graphql
{
  homeView {
    sectionsConnection(first: 4) {
      edges {
        node {
          ... on GenericHomeViewSection {
            key                               #  👈🏽 fields
            title                             #  👈🏽 that are
            component {                       #  👈🏽 common
              type                            #  👈🏽 to all 
            }                                 #  👈🏽 section types
          }        

            artworksConnection(first: 3) {    #  👈🏽 fields
              edges {                         #  👈🏽 that are
                node {                        #  👈🏽 specific
                  slug                        #  👈🏽 to 
                }                             #  👈🏽 sections
              }                               #  👈🏽 of type
            }                                 #  👈🏽 ArtworksRailHomeViewSection
          }
        }
      }
    }
  }
}
```

### Example response

```jsonc
{
  "data": {
    "homeView": {
      "sectionsConnection": {
        "edges": [
          {
            "node": {
              "key": "NEW_WORKS_FOR_YOU",     // 👈🏽 based on this key…
              "title": "New works for you",
              "component": {
                "type": "ARTWORKS_RAIL"
              },
              "artworksConnection": {         // 👈🏽 …we correctly resolve these artworks
                "edges": [
                  {
                    "node": {
                      "slug": "ed-ruscha-that-was-then-this-is-now-ii"
                    }
                  },
                  {
                    "node": {
                      "slug": "josef-albers-i-s-f-60"
                    }
                  },
                  {
                    "node": {
                      "slug": "susumu-kamijo-the-beach-lovers-86"
                    }
                  }
                ]
              }
            }
          },
          {
            "node": {
              "key": "AUCTION_LOTS_FOR_YOU",  // 👈🏽 TODO
              "title": "Auction lots for you",
              "component": {
                "type": "ARTWORKS_RAIL"
              },
              "artworksConnection": {
                "edges": [
                  {
                    "node": {
                      "slug": "TODO-auction_lots_for_you"
                    }
                  }
                ]
              }
            }
          },
          {
            "node": {
              "key": "SUGGESTED_ARTISTS",
              "title": "Suggested artists for you",
              "component": {
                "type": "ARTISTS_RAIL"
              }
            }
          },
          {
            "node": {
              "key": "RECENTLY_VIEWED_ARTWORKS", // 👈🏽 TODO
              "title": "Recently viewed works",
              "component": {
                "type": "ARTWORKS_RAIL"
              },
              "artworksConnection": {
                "edges": [
                  {
                    "node": {
                      "slug": "TODO-recently_viewed_artworks"
                    }
                  }
                ]
              }
            }
          }
        ]
      }
    }
  }
}
```


_schemaV2.graphql
```diff
@@ -3168,6 +3168,13 @@ enum ArtworkSorts {
 
 # An artwork rail section in the home view
 type ArtworksRailHomeViewSection implements GenericHomeViewSection & Node {
+  artworksConnection(
+    after: String
+    before: String
+    first: Int
+    last: Int
+  ): ArtworkConnection
```
anandaroop, line 3176: We discussed the merits of

-  having a simple `artworksConnection` here that takes **basic connection arguments only** as shown here, and letting each home view artworks resolver be more opinionated about what [ultimate args](https://github.com/artsy/metaphysics/pull/5865/files#diff-294b344a66a8ab3968329ebdf23ee095eaa754562be805266f228cd2ac4be3dfR18-R28) and ultimate [resolver should](https://github.com/artsy/metaphysics/pull/5865/files#diff-294b344a66a8ab3968329ebdf23ee095eaa754562be805266f228cd2ac4be3dfR30-R35) be used
    - works well for the paradigm where the _server is the authority_ about what objects should be shown in home view

-versus-

- trying to accommodate many other types of args that could be used by the various kinds of `ArtworksRailHomeViewSection` — e.g. NEW_WORKS_FOR_YOU, RECENTLY_VIEWED_ARTWORKS, etc etc
  - would _allow the client to influence_ what objects should be shown in home view

-versus-

- actually providing multiple, separate artwork fields for different section types
  - can't remember the advantage of this, but I think we agreed it would gunk up the schema

src/schema/v2/homeView/HomeViewSection.ts
```diff
@@ -45,6 +48,13 @@ const ArtworksRailHomeViewSectionType = new GraphQLObjectType<
   interfaces: [GenericHomeViewSectionInterface, NodeInterface],
   fields: {
     ...standardSectionFields,
+
+    artworksConnection: {
+      type: artworkConnection.connectionType,
+      args: pageable({}),
+      resolve: (parent, ...rest) =>
+        ARTWORK_RESOLVERS[parent.key](parent, ...rest),
```
anandaroop, line 56: This bit of runtime polymorphism allows the artworksConnection to decide on-the-fly [which resolver to use](https://github.com/artsy/metaphysics/pull/5865/files#diff-294b344a66a8ab3968329ebdf23ee095eaa754562be805266f228cd2ac4be3dfR88-R92), based on what kind of `key` the parent section has — `NEW_WORKS_FOR_YOU` vs `RECENTLY_VIEWED_ARTWORKS` etc etc.

This is probably the "cleverest" bit in this PR — how do we feel about it?

src/schema/v2/homeView/artworkResolvers.ts
```diff
@@ -0,0 +1,92 @@
...
+    // formerly specified client-side
+    maxWorksPerArtist: 3,
+    includeBackfill: true,
+    first: args.first,
+    version: "C",
+    excludeDislikedArtworks: true,
+    excludeArtworkIds: [],
+
+    ...args,
+  }
```
anandaroop, line 28: The big paradigm shift here is that the _server_ becomes authority for deciding what artworks should be selected for any given artworks rail.

This this kind of filtering logic, which previously was specified in the client…

https://github.com/artsy/eigen/blob/323364090372beec0d2a0bf4555795e7336c4def/src/app/Components/WorksForYouArtworks.tsx#L69-L73

…will now move up to the server, i.e. L20–24 above.

anandaroop: L25 is a hacky way of solving the requirement that certain args be specified with default values. I believe @nickskalkin suggested  a helper (e.g. something like `extractDefaultsFromField` that could introspect over a graphql field config and pull out [any default values that have been defined](https://github.com/artsy/metaphysics/blob/d160f6629aeeecf4089cd652a0c433b87af061be/src/schema/v2/artworksForUser/artworksForUser.ts#L34-L45))

src/schema/v2/homeView/getSectionsForUser.ts
```diff
@@ -0,0 +1,87 @@
...
+  const { meLoader } = context.authenticatedLoaders
+
+  if (!meLoader) throw new Error("You must be signed in to see this content.")
+
+  const me = await meLoader()
+
+  let sections: HomeViewSection[] = []
+
+  if (me.type === "Admin") {
+    sections = [
```
anandaroop, line 18: As the comment above loudly notes, this is just some silly placeholder logic for varying the list of sections based on _some_ user attribute. This area of course needs much more definition.

src/schema/v2/homeView/index.ts
```diff
@@ -16,11 +16,12 @@ const SectionsConnectionType = connectionWithCursorInfo({
 const SectionConnection: GraphQLFieldConfig<any, ResolverContext> = {
   type: SectionsConnectionType,
   args: pageable({}),
-  resolve: async (_parent, args, _context, _info) => {
+  resolve: async (_parent, args, context, _info) => {
     const { page, size, offset } = convertConnectionArgsToGravityArgs(args)
 
-    const totalCount = STUB_SECTIONS.length
-    const data = STUB_SECTIONS.slice(offset, offset + size)
+    const sections = await getSectionsForUser(context)
```
anandaroop, line 22: Slight evolution of this placeholder section data: from a stubbed array identical for everyone → to a slightly varying array based on the user's attributes.

src/schema/v2/homeView/artworkResolvers.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  const result = await artworksForUser.resolve!(
+    parent,
+    finalArgs,
+    context,
+    info
+  )
```
anandaroop, line 35: And once we've assembled the necessary args, we lean on existing resolver logic, rather than rebuilding from scratch. 

`artworksForUser` was instructive in that the resolver encapsulates [quite a bit of logic](https://github.com/artsy/metaphysics/blob/d160f6629aeeecf4089cd652a0c433b87af061be/src/schema/v2/artworksForUser/artworksForUser.ts#L47-L90) — would be bad to try and recreate that!

Do we foresee any risks with this approach, i.e. drift between how we invoke these resolvers here, vs how they were "originally" invoked elsewhere?

MounirDhahri:
As we agreed on this one - let's merge it with the stub data which will be useful for testing things out. We can remove it later

PR #5874, opened 2024-08-01
feat: add homeView > section query field

dblandin:
**Example**

```graphql
{
  homeView {
    section(id: "home-view-section-auction-lots-for-you") {
      __typename
      ... on GenericHomeViewSection {
        component {
          title
        }
      }
    }
  }
}
```

### DIFF
--- a/src/schema/v2/homeView/getSectionsForUser.ts
+++ b/src/schema/v2/homeView/getSectionsForUser.ts
@@ -1,10 +1,11 @@
 import { ResolverContext } from "types/graphql"
 import {
-  AuctionLotsForYouResolver,
-  NewWorksForYouResolver,
-  RecentlyViewedArtworksResolver,
-} from "./artworkResolvers"
-import { GraphQLFieldResolver } from "graphql"
+  AuctionLotsForYou,
+  HomeViewSection,
+  NewWorksForYou,
+  RecentlyViewedArtworks,
+  SuggestedArtists,
+} from "./sections"
 
 export async function getSectionsForUser(
   context: ResolverContext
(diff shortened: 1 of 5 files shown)


src/schema/v2/homeView/getSectionsForUser.ts
```diff
@@ -22,65 +23,19 @@ export async function getSectionsForUser(
...
+      RecentlyViewedArtworks,
     ]
   }
 
   return sections
 }
-
-// stub sections
-
-type HomeViewSection = {
```
MounirDhahri, line 44: nice 👏

MounirDhahri (approved):

src/schema/v2/homeView/index.ts
```diff
@@ -34,13 +40,35 @@ const SectionConnection: GraphQLFieldConfig<any, ResolverContext> = {
...
+      description: "The ID of the section",
+      type: new GraphQLNonNull(GraphQLString),
+    },
+  },
+  resolve: (_root, { id }, context) => {
+    const { meLoader } = context.authenticatedLoaders
+
+    if (!meLoader) throw new Error("You must be signed in to see this content.")
+
+    if (id.length === 0) {
```
MounirDhahri, line 57:
```suggestion
    if (!id) {
```

MounirDhahri:
I will merge this to avoid copy pasting the GraphQL schema and to also in order to merge Nikita's Eigen PR

MounirDhahri:
There is nothing wrong with this PR but something to keep in mind - whenever we're going to make a breaking change, we will break the push schema step in Eigen since we run `yarn relay` in CI. It should be fine though because I am updating Eigen currently with the new schema

Screenshot 2024-08-01 at 15 29 59

src/schema/v2/homeView/index.ts
dblandin: The argument is already enforced by the schema to be non-null. I think this check (which I borrowed from a similar resolver) guards against an empty string.

dblandin:
> It should be fine though because I am updating Eigen currently with the new schema

Yep, good to know. I was (and still am) assuming that we can follow-up quickly with [this PR](https://github.com/artsy/eigen/pull/10524) to resolve the breaking changes. I'll touch base in that PR.

PR #5881, opened 2024-08-02
chore: make some homeView field types non-nullable

dblandin:
Reduces the amount of null-checking needed by the client and we expect all of these types to consistently present data.


olerichter00 (approved):
That's a great improvement! Thanks 🙏

MounirDhahri (approved):
Nice

PR #5889, opened 2024-08-06
feat(EMI-2015): Auction deadline collector signals

erikdstock:
This PR completes [EMI-2015] by adding auction deadline values to the CollectorSignals type.

Only future dates are included for the date values. I have an alternative idea for the interface, but this essentially collects data that we normally check from multiple different sources - it's mostly up to clients to choose which to prioritize based on presence.

- [x] Needs tests

[EMI-2015]: https://artsyproduct.atlassian.net/browse/EMI-2015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/artwork/collectorSignals.ts
+++ b/src/schema/v2/artwork/collectorSignals.ts
@@ -1,8 +1,25 @@
-import { GraphQLFieldConfig } from "graphql"
+import { GraphQLBoolean, GraphQLFieldConfig, GraphQLString } from "graphql"
 import { GraphQLInt, GraphQLObjectType } from "graphql"
 import { ResolverContext } from "types/graphql"
 import { PartnerOfferToCollectorType } from "../partnerOfferToCollector"
 import { isFeatureFlagEnabled } from "lib/featureFlags"
+import { some } from "lodash"
+import { isFieldRequested } from "lib/isFieldRequested"
+
+const NOW = new Date()
+
+interface ActiveLotData {
+  saleArtwork: {
+    bidder_positions_count: number
+    extended_bidding_end_at?: string
+    end_at?: string
+  }
+  sale: {
+    registration_ends_at: string
+    live_start_at: string
+    ended_at?: string
+  }
+}
 
 export const CollectorSignals: GraphQLFieldConfig<any, ResolverContext> = {
   type: new GraphQLObjectType({
@@ -11,11 +28,32 @@ export const CollectorSignals: GraphQLFieldConfig<any, ResolverContext> = {
     fields: {
       bidCount: {
         type: GraphQLInt,
-        description: "Lot bid count",
+        description: "Bid count on lots open for bidding",
       },
       lotWatcherCount: {
         type: GraphQLInt,
-        description: "Lot watcher count",
+        description: "Lot watcher count on lots open for bidding",
+      },
+      registrationEndsAt: {
+        type: GraphQLString,
+        description: "Pending auction registration end time",
+      },
+      lotClosesAt: {
+        type: GraphQLString,
+        description: "Pending auction lot end time for bidding",
+      },
+      onlineBiddingExtended: {
+        type: GraphQLBoolean,
+        description:
+          "Auction lot bidding period extended due to last-minute bids",
+      },
+      liveStartAt: {
+        type: GraphQLString,
+        description: "Auction live bidding start time",
+      },
+      liveBiddingStarted: {
+        type: GraphQLBoolean,
+        description: "Live bidding has started on this lot's auction",
       },
       partnerOffer: {
         type: PartnerOfferToCollectorType,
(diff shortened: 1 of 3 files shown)


src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -61,20 +106,50 @@ const collectorSignalsLoader = async (
...
+        }
+      }
+
+      const extendedBiddingEndAt = saleArtwork.extended_bidding_end_at
+      if (extendedBiddingEndAt && extendedBiddingEndAt.length > 0) {
+        onlineBiddingExtended = true
+        lotClosesAt = extendedBiddingEndAt
+      } else {
+        onlineBiddingExtended = false
+        lotClosesAt = saleArtwork.end_at
```
erikdstock, line 152: lots of hidden logic and assumptions in here:
- `sale.liveStartAt` is not on sale artwork, and implies the non-existence of `end_at`
- `saleArtwork.extendedBiddingEndAt` implies online bidding was extended and `end_at` should be ignored
- `ifFutureDate()` is a filtering helper that returns a date object if it is (present and) in the future, otherwise null.
- At the top of this section everything is wrapped in an additional check for `!sale.ended_at` to ignore artworks with an ended sale, though [technically we should not get these](https://github.com/artsy/metaphysics/pull/5889/files?diff=split&w=1#diff-dc52d603a602c819aef005d759243d63b275674ea6954b70a204fec0f49d29b5R195) - in the case below we query for sales with the `live: true` param, refers to the auction's active bidding period, not live auction integration.

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -61,20 +106,50 @@ const collectorSignalsLoader = async (
...
+    const activeLotData = await getActiveSaleArtwork(
       {
         artworkId,
         saleIds: artwork.sale_ids,
       },
       ctx
     )
 
-    if (activeSaleArtwork) {
+    const activeAuctionLot = activeLotData && !activeLotData.sale.ended_at
```
starsirius, line 117: Question: Do we still need to check `ended_at` if we are already fetching from _active_ sale above?

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -61,20 +106,50 @@ const collectorSignalsLoader = async (
...
       }
-      if (activeSaleArtwork.bidder_positions_count) {
-        bidCount = activeSaleArtwork.bidder_positions_count
+
+      if (saleArtwork.bidder_positions_count) {
+        bidCount = saleArtwork.bidder_positions_count
+      }
+
+      const futureRegistrationEndTime = ifFutureDate(sale.registration_ends_at)
+      if (futureRegistrationEndTime) {
```
starsirius, line 130: #non-blocking: `ifFutureDate` returns a date or `null`, and I guess you can just do assignment directly? Or I wonder if it's more clear to have it return a boolean like `isFuture`. (Not sure what's our current recommendation about date/time libraries, but there could already be helper for this.)

src/schema/v2/artwork/collectorSignals.ts
starsirius: Oh I see your [comment below](https://github.com/artsy/metaphysics/pull/5889/files#r1707107448)! 👍

src/schema/v2/artwork/collectorSignals.ts
starsirius: Maybe we can make the assumption and condition clear in the GraphQL field description?

src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -61,20 +106,50 @@ const collectorSignalsLoader = async (
...
+        if (futureLiveStartTime) {
+          liveBiddingStarted = false
+          liveStartAt = futureLiveStartTime
+        } else {
+          liveBiddingStarted = true
+        }
+      }
+
+      const extendedBiddingEndAt = saleArtwork.extended_bidding_end_at
+      if (extendedBiddingEndAt && extendedBiddingEndAt.length > 0) {
```
starsirius, line 147: Question: what does `extendedBiddingEndAt.length > 0` do?

MrSltun (commented):
Looks good! 🌟
I have only one concern, and I could be wrong about it, but if we call `collectorSignals`, would the whole function run? even if we only request `partnerOffer` for example?

erikdstock:
Talking with @starsirius today we are thinking this might need some adjustments due to new caching in force - returning boolean fields where it would be better for a client to compare a time value with the current real time. We might need to revisit this for other possibly-cached values like bid/watcher counts as well. Will raise in slack.

erikdstock:
> if we call `collectorSignals`, would the whole function run? even if we only request `partnerOffer` for example?

Good point - we have an `isFieldRequested` helper that traverses the query ast to check for specific fields. i can add that, possibly as a followup.

src/schema/v2/artwork/collectorSignals.ts
erikdstock: My hope here was to have some filtering/conditional logic rolled into one - if you pass a date + it is in the future, you know you can work with it from there forwards without needing to say `if (something?.some_date && isFuture(something.some_date)) { const futureDate = something.some_date ; /* ... */ }`.

Updated this function to be a little more explicit about what it does.

src/schema/v2/artwork/collectorSignals.ts
erikdstock: Removed, not needed.

MrSltun (approved):
Looks great!! 🌟

starsirius (approved):
👍  @mzikherman 's comment about breaking the single resolver to field-specific resolvers seems worth exploring. Don't think it's blocking and we can create followup ticket for the refactor!

PR #5898, opened 2024-08-07
feat: add galleries near you partner section

MounirDhahri:
This PR resolves [ONYX-1178]

This PR adds a new galleries near you section. 

Screenshot 2024-08-30 at 11 11 39

**How it will look like in Eigen 👇 **

https://github.com/user-attachments/assets/bd26c192-ad8d-45fa-b802-0aa85e1497fd

[ONYX-1178]: https://artsyproduct.atlassian.net/browse/ONYX-1178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/homeView/HomeViewSection.ts
+++ b/src/schema/v2/homeView/HomeViewSection.ts
@@ -1,7 +1,6 @@
 import {
   GraphQLFieldConfigMap,
   GraphQLInterfaceType,
-  GraphQLNonNull,
   GraphQLObjectType,
   GraphQLUnionType,
 } from "graphql"
@@ -17,7 +16,6 @@ import { heroUnitsConnection } from "../HeroUnit/heroUnitsConnection"
 import { MarketingCollectionType } from "../marketingCollections"
 import { NotificationsConnection } from "../notifications"
 import { InternalIDFields, NodeInterface } from "../object_identification"
-import { PartnersConnection } from "../partner/partners"
 import { SalesConnectionField } from "../sales"
 import { HomeViewComponent } from "./HomeViewComponent"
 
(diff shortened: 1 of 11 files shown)


olerichter00 (dismissed):
Looks great! 🌟

src/schema/v2/homeView/HomeViewSection.ts
```diff
@@ -237,8 +237,19 @@ const SalesRailHomeViewSectionType = new GraphQLObjectType<
   },
 })
 
-// the Section union type of all concrete sections
+export const GalleriesHomeViewSectionType = new GraphQLObjectType<
```
anandaroop, line 240: **thought (non-blocking):** 🤦🏽 sorry for this late feedback. I think I was expecting to see a rail of galleries here, but now I realize of course this is just a big hero image that navigates to a predefined route. 

This seems like the kind of section we _could_ handle generically (HeroImageLink or some such). 

I don't think it's a requirement to do that now, since this is just a one-off in the current home feed design. But should we? A little hard to say without seeing more complete designs. If this pattern is recurring it would make sense to treat it generically. Also that would free up the name GalleriesHomeViewSection for something that really _was_ a list of galleries.

Like I said, nothing urgent, since we are still free to make breaking changes till M2.

joeyAghion: Could we cover >1 section by making this generic? Like, could it reuse the hero section, with different parameters? If so, seems like a win.

Either way, I imagine we _will_ need a general-purpose type of section for static content. So sooner or later...

anandaroop: Followed up here finally, and worth some discussion: https://github.com/artsy/metaphysics/pull/6116/

PR #5899, opened 2024-08-07
fix: Put a cap on amount of collector profile returned from the connection…

jpotts244:
…, allow for pagination args to be passed in from client

Reworks `collectorProfilesConnection` under `Partner` type to limit the about of collector profiles we return to the client 

We were facing some performance issues on really large datasets
Screenshot 2024-08-07 at 9 41 00 AM

---

Downstream Gravity would return this exception
Screenshot 2024-08-07 at 10 51 19 AM

---

Limits to 20 and exposes paginations params so they can be passed downstream
and adjusts `totalCount` to be a count of the full data set

Screenshot 2024-08-07 at 10 42 11 AM

### DIFF
--- a/src/schema/v2/Alerts/index.ts
+++ b/src/schema/v2/Alerts/index.ts
@@ -518,8 +518,11 @@ export const PartnerAlertsEdgeFields = {
   collectorProfilesConnection: {
     type: PartnerCollectorProfilesConnectionType,
     args: pageable({
-      totalCount: {
-        type: GraphQLBoolean,
+      page: {
+        type: GraphQLInt,
+      },
+      size: {
+        type: GraphQLInt,
       },
     }),
     resolve: async (parent, args, { partnerCollectorProfilesLoader }) => {
@@ -533,6 +536,9 @@ export const PartnerAlertsEdgeFields = {
           "partnerId or userIds is undefined in the parent object"
         )
       }
+
+      // Make API call to fetch the first X collector profile records
+      const slicedUserIds = parent.user_ids.slice(0, 20)
 
       type GravityArgs = {
         page: number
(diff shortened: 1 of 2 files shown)


src/schema/v2/Alerts/index.ts
```diff
@@ -518,8 +518,11 @@ export const PartnerAlertsEdgeFields = {
   collectorProfilesConnection: {
     type: PartnerCollectorProfilesConnectionType,
     args: pageable({
-      totalCount: {
```
jpotts244, line 521: This doesnt need to be accessible to clients, we hardcode it below to `true`

src/schema/v2/Alerts/index.ts
```diff
@@ -534,6 +537,9 @@ export const PartnerAlertsEdgeFields = {
         )
       }
 
+      // Make API call to fetch the first X collector profile records
+      const slicedUserIds = parent.user_ids.slice(0, 20)
```
damassi, line 541: A more graphql/connection-y way to do this would be to rely on `first`, vs size (which is a foreign concept in connection-based cursoring). 

So something like: `first: args.first ?? 10` would do it

jpotts244: Good idea ^^

src/schema/v2/Alerts/index.ts
```diff
@@ -534,6 +539,10 @@ export const PartnerAlertsEdgeFields = {
         )
       }
 
+      // Make API call to fetch the first X collector profile records
+      const first = args.first ?? DEFAULT_COLLECTOR_PROFILES_BATCH_SIZE
+      const slicedUserIds = parent.user_ids.slice(0, first)
+
```
jpotts244, line 545: Digging on the side here to see if I can pass `first:` into the gravity call

slicing what gets passed in the `user_ids` field in the API needs to be capped as well

damassi: Yah was thinking there might be some weaving involved...

damassi (approved):

jonallured (approved):

PR #5888, opened 2024-08-06
feat: partners support featured keywords for curated browsing

joeyAghion:
Like, https://github.com/artsy/metaphysics/pull/5878, but for partners. Depends on https://github.com/artsy/gravity/pull/18014 (but can be merged whenever).

PR #5905, opened 2024-08-09
feat: expose featuredKeywords on sales

joeyAghion:
Like https://github.com/artsy/metaphysics/pull/5878 and https://github.com/artsy/metaphysics/pull/5888, but for sales. Depends on https://github.com/artsy/gravity/pull/18025 but can safely be deployed whenever.


ovasdi (approved):
LGTM

PR #5906, opened 2024-08-12
feat(onyx): add hero units to new home view schema

nickskalkin:
[Jira ticket](https://artsyproduct.atlassian.net/browse/ONYX-1187)

Adds `HeroUnitsHomeViewSection`:

```graphql
query X {
  homeView {
    sectionsConnection(first: 10) {
      edges {
        node {
          ... on HeroUnitsHomeViewSection {
            heroUnitsConnection(first: 10, private: false) {
              edges {
                node {
                  title
                }
              }
            }
          }
        }
      }
    }
  }
}

```

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2298,7 +2298,7 @@ type ArtistsRailHomeViewSection implements GenericHomeViewSection & Node {
   ): ArtistConnection!
 
   # The component that is prescribed for this section
-  component: HomeViewComponent!
+  component: HomeViewComponent
 
   # A globally unique ID.
   id: ID!
@@ -3168,7 +3168,7 @@ type ArtworksRailHomeViewSection implements GenericHomeViewSection & Node {
   ): ArtworkConnection!
 
   # The component that is prescribed for this section
-  component: HomeViewComponent!
+  component: HomeViewComponent
 
   # A globally unique ID.
   id: ID!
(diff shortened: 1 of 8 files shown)


_schemaV2.graphql
```diff
@@ -10620,6 +10620,30 @@ type HeroUnitLink {
...
+    after: String
+    before: String
+    first: Int
+    last: Int
+
+    # If true will include inactive hero units.
+    private: Boolean = false
+
+    # If present will search by term.
+    term: String
```
olerichter00, line 10637: Shouldn't both be set in Metaphysics instead of being sent by the client)

src/schema/v2/homeView/sections.ts
```diff
@@ -11,13 +11,14 @@ import {
...
 
 export type HomeViewSection = {
   id: string
   type: string
-  component: {
-    title: string
-  }
+  component?: {
+    title?: string | null
+  } | null
```
olerichter00, line 21: Not directly related to this PR: I really don't get this component field. It only includes the title which some sections have and others don't have. 🤔 Shouldn't the title be part of the specific section type and only be included if the section has a title?

olerichter00: With the new section introduced in this PR we have to make the title nullable for all sections which makes the client implementation more complicated.

_schemaV2.graphql
nickskalkin: Agree, will do

src/schema/v2/homeView/sections.ts
nickskalkin: I think we don't really need it for now, our assumption was that in future components will be more customizable and it would be nice to scope component-level customization

olerichter00: Yeah, it might be useful in the futuer.

olerichter00 (approved):

PR #5912, opened 2024-08-14
feat(onyx-1191): add 'curator's picks: emerging' home view section

nickskalkin:
I am trying to migrate "Curator's picks: emerging" rail to the new schema (which is [MarketingCollectionRail in Eigen](https://github.com/artsy/eigen/blob/main/src/app/Scenes/Home/Components/MarketingCollectionRail.tsx)). There is something different about this rail - it is not "hardcoded", i.e. component info, such as as title, subtitle, background image, etc. is stored in Gravity (accessible via `homePage.heroUnits(platform: MOBILE)` MP query). Current `homeView` implementation does't allow fetching component-level details from backend, this is what I am trying to achieve in this PR.

The implementation feels a bit weird to me and I would like to hear your feedback about it. The biggest change is that sections are now can be not only `HomeViewSection`s, but also callable functions (see `featuredCollection("curators-picks-emerging-app", "curators-picks-emerging")` in `getSectionsForUser` function). I don't know how else to achieve this "dynamic" behaviour.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -9809,6 +9809,25 @@ enum FeatureSorts {
   NAME_DESC
 }
 
+# A featured collection section in the home view
+type FeaturedCollectionHomeViewSection implements GenericHomeViewSection & Node {
+  artworksConnection(
+    after: String
+    before: String
+    first: Int
+    last: Int
+  ): ArtworkConnection!
+
+  # The component that is prescribed for this section
+  component: HomeViewComponent
+
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
+}
+
 # An illustrated link chosen to highlight a Gene from a given GeneFamily
 type FeaturedGeneLink {
   href: String!
@@ -10913,6 +10932,12 @@ type HomeView {
 
 # A component specification
 type HomeViewComponent {
+  # A background color for this section
+  backgroundColor: String
+
+  # A display subtitle for this section
+  subtitle: String
+
   # A display title for this section
   title: String
 }
@@ -10920,6 +10945,7 @@ type HomeViewComponent {
 union HomeViewSection =
     ArtistsRailHomeViewSection
   | ArtworksRailHomeViewSection
+  | FeaturedCollectionHomeViewSection
   | HeroUnitsHomeViewSection
 
 # A connection to a list of items.
--- a/src/schema/v2/homeView/getSectionsForUser.ts
+++ b/src/schema/v2/homeView/getSectionsForUser.ts
@@ -1,8 +1,9 @@
 import { ResolverContext } from "types/graphql"
 import {
   AuctionLotsForYou,
+  featuredCollection,
   HeroUnits,
-  HomeViewSection,
+  HomeViewSectionOrResolver,
   NewWorksForYou,
   NewWorksFromGalleriesYouFollow,
   RecentlyViewedArtworks,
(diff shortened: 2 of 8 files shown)


src/schema/v2/homeView/getSectionsForUser.ts
```diff
@@ -23,10 +24,14 @@ export async function getSectionsForUser(
 
   const me = await meLoader()
 
-  let sections: HomeViewSection[] = []
+  let sections: HomeViewSectionOrResolver[] = []
 
   if (me.type === "Admin") {
     sections = [
+      featuredCollection(
+        "curators-picks-emerging-app",
```
nickskalkin, line 32: Here I pass heroUnit ID and marketingCollection ID. Eigen hardcodes marketingColection ID, then fetches all heroUnits and picks only the one with corresponding marketing collection ID. I can imaging that in future we can have another heroUnit corresponding to another marketing collection. It will be easy to support by adding another `featuredCollection(heroUnitID, marketingCollectionID)`.

src/schema/v2/homeView/sections.ts
```diff
@@ -92,7 +103,29 @@ export const HeroUnits: HomeViewSection = {
   resolver: HeroUnitsResolver,
 }
 
-const sections: HomeViewSection[] = [
+export const featuredCollection = (
```
nickskalkin, line 106: maybe `featuredCollectionSectionBuilder` is a better name for this outer function

anandaroop:
My initial reaction is that this component just wants to be an `ArtworksRailHomeViewSection`, but with extra component attributes such as you've added (bgcolor, subtitle). 




That would allow us to continue treating this content generically, rather than having a new dedicated section type for it. 

Is there something which would forbid us from using that approach?

anandaroop:
I was trying to think of another approach that would that would enable the use of the existing `ArtworksRailHomeViewSection`, and I sketched out something here:

https://github.com/artsy/metaphysics/compare/main...anandaroop/onyx-1190-curators-picks-emerging

This borrows a few things from your approach, but aims to contain the complexity within the section's component definition.

Maybe we can discuss tomorrow.

nickskalkin:
@anandaroop thanks a lot for your effort! Let's pair today?

> My initial reaction is that this component just wants to be an ArtworksRailHomeViewSection, but with extra component attributes such as you've added (bgcolor, subtitle).

Mine was too, but at the same time I am hesitant to "pollute" artworks rails schema with fields that we may not want to have in future, but I am not strongly opinionated.

olerichter00:
> My initial reaction is that this component just wants to be an ArtworksRailHomeViewSection, but with extra component attributes such as you've added (bgcolor, subtitle).

Looking at the two rails (the normal artwork rail and the "Curators' Picks: Emerging" rail), I think, they look quite different, and it might also be a good option to create a new (generic) section type and component for it. Maybe something like `CollectionArtworksRailHomeViewSection`. This way, we would not need to pollute the artwork rail section schema with new fields that are only relevant for this new section, and the frontend implementation would also be simpler by creating two different components instead of having one that tries to implement both versions.

_schemaV2.graphql
```diff
@@ -10913,13 +10932,20 @@ type HomeView {
 
 # A component specification
 type HomeViewComponent {
+  # A background color for this section
+  backgroundColor: String
+
+  # A display subtitle for this section
+  subtitle: String
```
olerichter00, line 10939: What do you think about moving these new fields to the new section type, as they are only relevant for this section? I suppose it might be confusing if they're also available for all other sections.

anandaroop:
([Slack thread](https://artsy.slack.com/archives/C05EQL4R5N0/p1723746300946949) for some further discussion)

nickskalkin:
Closing in favour of https://github.com/artsy/metaphysics/pull/5916

PR #5915, opened 2024-08-16
feat: Allow alertsConnection to accept alert id filter

jpotts244:
Adds support for single alert with partner facing metadata so we can support a show single alert page in CMS

[Wires in new Gravity endpoint](https://github.com/artsy/gravity/pull/18073)

This connection should support:
```
{
    slug
    name
    alertsConnection(first: 10) {
      totalCount
      edges {
        id
        collectorProfilesConnection(first: 10) {
          totalCount
          edges {
            node {
              name
              internalID
            }
          }
        }
        node {
          internalID
          attributionClass
        }
      }
    }
  }
}
```

but also a

```
{
    slug
    name
    alertsConnection(id: "im-a-regular-old-single-record") {
      totalCount
      edges {
        id
        collectorProfilesConnection(first: 10) {
          totalCount
          edges {
            node {
              name
              internalID
            }
          }
        }
        node {
          internalID
          attributionClass
        }
      }
    }
  }
}
```

### DIFF
--- a/src/lib/loaders/loaders_with_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_with_authentication/gravity.ts
@@ -659,12 +659,12 @@ export default (accessToken, userID, opts) => {
       {},
       { headers: true }
     ),
-    partnerSearchCriteriaLoader: gravityLoader(
+    partnerSearchCriteriasLoader: gravityLoader(
       (id) => `/partner/${id}/partner_search_criterias`,
       {},
       { headers: true }
     ),
-    partnerSearchCriteriaSingleLoader: gravityLoader<
+    partnerSearchCriteriaLoader: gravityLoader<
       any,
       { partner_id: string; id: string }
     >(
--- a/src/lib/loaders/loaders_with_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_with_authentication/gravity.ts
@@ -669,7 +669,7 @@ export default (accessToken, userID, opts) => {
       { partner_id: string; id: string }
     >(
       ({ partner_id, id }) =>
-        `/partner/${partner_id}/partner_search_criterias/${id}`,
+        `/partner/${partner_id}/partner_search_criteria/${id}`,
       {},
       { headers: true }
     ),
--- a/src/lib/loaders/loaders_with_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_with_authentication/gravity.ts
@@ -664,18 +664,17 @@ export default (accessToken, userID, opts) => {
       {},
       { headers: true }
     ),
-    partnerSearchCriteriaHitsLoader: gravityLoader(
-      (id) => `partner/${id}/partner_search_criteria_hits`,
-      {},
-      { headers: true }
-    ),
-    // TODO: rename
-    singlePartnerSearchCriteriaLoader: gravityLoader<
+    partnerSearchCriteriaSingleLoader: gravityLoader<
       any,
       { partner_id: string; id: string }
     >(
       ({ partner_id, id }) =>
         `/partner/${partner_id}/partner_search_criterias/${id}`,
+      {},
+      { headers: true }
+    ),
+    partnerSearchCriteriaHitsLoader: gravityLoader(
+      (id) => `partner/${id}/partner_search_criteria_hits`,
       {},
       { headers: true }
     ),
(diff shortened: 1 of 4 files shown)


src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -663,6 +663,16 @@ export default (accessToken, userID, opts) => {
       {},
       { headers: true }
     ),
+    // TODO: rename
+    singlePartnerSearchCriteriaLoader: gravityLoader<
```
mzikherman, line 667: I wonder if the Gravity implementation might add this support on to the existing endpoint (returning `hits: [single-instance]`) and so would then re-use the existing loader.

jpotts244: Good call and question!

Thinking, I do like the rest-full-ness of a separate singular endpoint but up for debate

damassi: Thinking too that moving this to gravity might better support future roll ups for all matching alerts (for a given artwork), a la demand 1 and the issue we're seeing with the new tab

jpotts244: Leaning in that direction as well, was just thinking we want to be able to handle more "fancy" pagination if a user has a lot of collectors we need to paginate through and maybe thats better to handle at the alert level!

Ex:
Screenshot 2024-08-22 at 2 04 22 PM

src/schema/v2/partner/__tests__/partner.test.js
```diff
@@ -1634,6 +1634,57 @@ describe("Partner type", () => {
...
+            id: "3f980bbe-7b9b-4fa9-beb1-69d13e94fb0c",
+            price_range: "1-2",
+          },
+        },
+      }
+
+      const query = gql`
+        {
+          partner(id: "catty-partner") {
+            alertsConnection(
```
jpotts244, line 1653: Ok pretty close, but still need to pass in a `first` argument
I think we want to be like: 'if id is present as an arg, dont worry about needing a first'

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -664,6 +664,15 @@ export default (accessToken, userID, opts) => {
       {},
       { headers: true }
     ),
+    partnerSearchCriteriaSingleLoader: gravityLoader<
```
damassi, line 667: (non-blocking): maybe just `partnerSearchCriteriaLoader` (implies singular)

src/schema/v2/partner/partner.ts
```diff
@@ -346,8 +349,21 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
+          { partnerSearchCriteriaSingleLoader, partnerSearchCriteriaLoader }
+        ) => {
+          if (
+            !partnerSearchCriteriaLoader ||
+            !partnerSearchCriteriaSingleLoader
+          )
+            return null
+
+          if (args.id && !args.first) {
+            args.first = 1
```
jpotts244, line 364: Sets `first: 1` behind the scenes if an id argument is present. This makes sure the pagination resolver doesnt 💥 


Let me know if this is ethical? I guess we could return a response without using the `paginationResolver` if were expecting a singular alert pack here from the connection

src/lib/loaders/loaders_with_authentication/gravity.ts
jpotts244: that works, let me rename it's plural partner to `partnerSearchCriteriasLoader`

damassi (dismissed):
One tiny non-blocking naming thought, but otherwise looks great 👍 

It would be nice to have something in the future where we could do 
```
searchCriteria(id: "foo") {
  // fields 
}
```
but if this is the sort of shape that's needed here, all good

src/schema/v2/partner/partner.ts
damassi: Yah, this was kinda implied by my comment -- wiring up an entirely new field for handling singular items. That said, there are spots all over MP where we do something like this, so seems ok.

mzikherman: I actually don't think we want an entirely new field for handling singular items.

Right now - we have this connection type in our schema, where the node is an `Alert`, and the edge contains useful data.

How would a singular field fit into the schema? You might need to create a _new_ type that combines the `Alert` and some other edge data...essentially you'd need to encode the _join_ between partner/alert/criteria 'stuff' - which is a bit hodge-podge and tied to the current backend implementation, yada yada.

For these kinds of situations, I think it's better to not create any singular type, and use the connection of edges (even when you want a single specific edge).

damassi (approved):

PR #5916, opened 2024-08-16
feat(onyx-1191): add "Curators' Picks: Emerging" home view section

nickskalkin:
Related: https://github.com/artsy/metaphysics/pull/5912

Adds "Curators' Picks: Emerging" section to our new schema. See a discussion why we came up with some of implementation decisions [in Slack](https://artsy.slack.com/archives/C05EQL4R5N0/p1723746300946949).

Example request:

```graphql
query {
  homeView {
    section(id: "home-view-section-curators-picks-emerging") {
        component {
          title
          description
          backgroundImageURL
          href
        }

          edges {
            node {
              title
            }
          }
        }
      }
    }
  }
}

```

And response:

```json
{
  "data": {
    "homeView": {
      "section": {
        "component": {
          "title": "Curators’ Picks: Emerging",
          "description": "The best works by rising talents on Artsy, available now.",
          "backgroundImageURL": "https://d32dm0rphc51dk.cloudfront.net/8hMW7lYWoZZXO3Kv2u8EaA/untouched-jpg.jpg",
          "href": "/collection/curators-picks-emerging"
        },
        "artworksConnection": {
          "edges": [
            {
              "node": {
                "title": "Landscapes (series VIII) #5"
              }
            }
          ]
        }
      }
    }
  }
}
```

### DIFF
--- a/src/schema/v2/homeView/sections.ts
+++ b/src/schema/v2/homeView/sections.ts
@@ -14,15 +14,19 @@ import {
 } from "./artistResolvers"
 import { HeroUnitsResolver } from "./heroUnitsResolver"
 
-type MaybeResolved<T> = T | ((context: ResolverContext) => Promise<T>)
+type MaybeResolved<T> =
+  | T
+  | ((context: ResolverContext, args: any) => Promise<T>)
 
 export type HomeViewSection = {
   id: string
   type: string
   component?: {
     title?: MaybeResolved<string>
+    type?: MaybeResolved<string>
     description?: MaybeResolved<string>
-    backgroundColor?: MaybeResolved<string>
+    backgroundImageURL?: MaybeResolved<string>
+    href?: MaybeResolved<string>
   }
   resolver?: GraphQLFieldResolver<any, ResolverContext>
 }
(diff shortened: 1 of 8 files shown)


src/schema/v2/homeView/sections.ts
```diff
@@ -14,15 +14,19 @@ import {
...
+type MaybeResolved<T> =
+  | T
+  | ((context: ResolverContext, args: any) => Promise<T>)
 
 export type HomeViewSection = {
   id: string
   type: string
   component?: {
     title?: MaybeResolved<string>
+    type?: MaybeResolved<string>
```
anandaroop, line 26: **thought (non-blocking):** In my spike I had this as just `string` instead of `MaybeResolved<string>`. 

Specifying preferred components seems to me a concern of MP's new home view schema, rather than something which is likely to be fetched from _upstream_.  

That's just my best guess, so I'm not insisting — but I might be in favor of simplifying this to just `string`. I admit knowing this distinction ahead of time when introducing a new field might be challenging. (Seems fine for the other fields introduced here.)

anandaroop (approved):
Just one thought, otherwise LGTM

src/schema/v2/homeView/sections.ts
nickskalkin: Totally makes sense, I did it on autopilot - it's very unlikely `type` will ever be not a regular hardcoded string

nickskalkin: I will address in follow-up PR if you don't mind just for my convenience

PR #5917, opened 2024-08-19
feat(updateCollectorProfile): expose CollectorProfileType field 

araujobarret:
This old mutation returns a default `UpdateCollectorProfilePayload` which doesn't fit anywhere using it with relay. 
The changes here expose `collectorProfile` with the type supported by our schema when it comes to spreading fragments. Nothing new here, we have this pattern in other mutations such as `updateMyUserProfile`, and could extend it to something like `collectorProfileOrError` but it seemed too much for this small exposition, open to feedback 😄

### DIFF
--- a/src/schema/v2/CollectorProfile/__tests__/updateCollectorProfileWithID.test.ts
+++ b/src/schema/v2/CollectorProfile/__tests__/updateCollectorProfileWithID.test.ts
@@ -1,6 +1,6 @@
 /* eslint-disable promise/always-return */
 import gql from "lib/gql"
-import { runAuthenticatedQuery, runQuery } from "schema/v2/test/utils"
+import { runAuthenticatedQuery } from "schema/v2/test/utils"
 
 describe("UpdateCollectorProfileWithID", () => {
   it("calls the expected loader with correctly formatted params", async () => {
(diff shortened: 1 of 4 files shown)


src/schema/v2/me/update_collector_profile.ts
```diff
@@ -37,7 +40,13 @@ export default mutationWithClientMutationId<any, any, ResolverContext>({
       type: GraphQLString,
     },
   },
-  outputFields: CollectorProfileFields,
+  outputFields: {
+    ...CollectorProfileFields,
+    collectorProfile: {
+      type: CollectorProfileType,
+      resolve: (collectorProfile) => collectorProfile,
+    },
```
egdbear, line 48: **Questions:**
- What is the reason behind not using the existing `CollectorProfileType` for the `outputFields`?
- I also noticed that this isn't using the correct `typeOrError` along with the success/error type (e.g., [PartnerOfferType](https://github.com/artsy/metaphysics/blob/afe483696f9be14dba60e4e0ef8980ca78f91e79/src/schema/v2/createPartnerOfferMutation.ts#L70)). Implementing this in the PR shouldn't be a heavy lift, and it would help maintain consistency across our mutations.

It doesn’t feel right to me to use `...collectorProfileFields` and name the `collectorProfile` as `node` here. Can you give more context?

araujobarret: > What is the reason behind not using the existing CollectorProfileType for the outputFields?

I think it's going to break Force since it uses it, but I see only a [clientMutationId](https://github.com/artsy/force/blob/main/src/Components/Inquiry/Hooks/useUpdateCollectorProfile.ts#L32) being used, so I think it's a no harm issue

>I also noticed that this isn't using the correct typeOrError along with the success/error type (e.g., [PartnerOfferType](https://github.com/artsy/metaphysics/blob/afe483696f9be14dba60e4e0ef8980ca78f91e79/src/schema/v2/createPartnerOfferMutation.ts#L70)). Implementing this in the PR shouldn't be a heavy lift, and it would help maintain consistency across our mutations

Agree, it seemed too much since I'm improving something unrelated to the work we are doing in Eigen, but once I introduce it there it's going to be stitched to an app version, making improvements later hard to be done. So, on it 🫡

araujobarret: done in https://github.com/artsy/metaphysics/pull/5917/commits/455f322ce6817945b581404ca99613f7e465fa05

egdbear: :chef-kiss:

egdbear: Great to see removing the amount of duplicated lines in the schema!

image

egdbear (approved):

src/schema/v2/CollectorProfile/__tests__/updateCollectorProfileWithID.test.ts
```diff
@@ -83,30 +83,29 @@ describe("UpdateCollectorProfileWithID", () => {
   })
 
   it("throws error when data loader is missing", async () => {
-    const mutation = `
+    const mutation = gql`
       mutation {
-        updateCollectorProfile(input: { professionalBuyer: true, loyaltyApplicant: true, selfReportedPurchases: "trust me i buy art" }) {
```
araujobarret, line 88: idkw this test case was testing another mutation 🤔

PR #5923, opened 2024-08-20
fix(homeview): Fix type error

damassi:
Looks like a type-error creeped through CI from the homeview work; fixing.


nickskalkin:
thanks!

PR #5926, opened 2024-08-20
feat(alerts): Add partnerAlertHitsConnection

damassi:
Addresses https://artsyproduct.atlassian.net/browse/AMBER-914

Adds a new connection to the partners alerts eco system:

```graphql
{
  partner(id: "cypress-test-partner-for-automated-testing-purposes") {
    partnerAlertHitsConnection(first: 10) {
      totalCount
      edges {
        
        internalID
        artwork {
          internalID
          title
        }
        userIDs
        createdAt
        node {
          additionalGeneIDs
        }
      }
    }
  }
}
```

cc @artsy/amber-devs

### DIFF
--- a/src/schema/v2/partner/__tests__/partner.test.js
+++ b/src/schema/v2/partner/__tests__/partner.test.js
@@ -14,7 +14,6 @@ describe("Partner type", () => {
       type: "Gallery",
       has_full_profile: true,
       inquireable: false,
-      profile_banner_display: true,
       distinguish_represented_artists: true,
       profile_banner_display: "Artworks",
       claimed: true,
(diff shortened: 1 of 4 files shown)


src/schema/v2/partner/partner.ts
```diff
@@ -367,6 +386,58 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
           })
         },
       },
+      partnerAlertsHitsConnection: {
```
jonallured, line 389: nit: I would put the plural only on Hit so what about `partnerAlertHitsConnection`?

src/schema/v2/partner/partner.ts
```diff
@@ -367,6 +386,58 @@ export const PartnerType = new GraphQLObjectType<any, ResolverContext>({
...
+          if (!partnerSearchCriteriaHitsLoader) return null
+          const { page, size, offset } = convertConnectionArgsToGravityArgs(
+            args
+          )
+
+          type GravityArgs = {
+            page: number
+            size: number
+            total_count: boolean
+            artist_id?: string
```
jpotts244, line 413: I dont think the API is accepting `artist_id` args at the time nor in the way we are persisting in the [schema](https://github.com/artsy/gravity/blob/2049ddd10a5ab8d964e1bec5644b8bc5737d463b/db/schema.rb#L670-L678)

Maybe we burn that bridge when we need to support additional args?

damassi: ahh copy 🍝 ; fixing

src/schema/v2/partner/__tests__/partner.test.js
```diff
@@ -1560,6 +1560,79 @@ describe("Partner type", () => {
...
+      `
+      const partnerSearchCriteriaHitsLoader = () => Promise.resolve(response)
+
+      const data = await runAuthenticatedQuery(query, {
+        ...context,
+        partnerSearchCriteriaHitsLoader,
+      })
+      expect(data).toEqual({
+        partner: {
+          partnerAlertsHitsConnection: {
```
jpotts244, line 1614:
```suggestion
          partnerAlertHitsConnection: {
```

PR #5936, opened 2024-08-22
feat: add latest activity rail

MounirDhahri:
This PR resolves [ONYX-1221]

This PR adds the latest activity rail as a new section to the home view. 

~~**Approach explanation**~~
~~Here I am only adding the component title without data - The latter will be used for the section title and in the frontend, the plan is to show/hide the latest activity component based on the presence of the section in the response.~~

It turned out it wasn't hard to migration `notificationsConnection` so I just migrated that eventually

**Eigen Implementation**: https://github.com/artsy/eigen/pull/10648

Screenshot 2024-08-22 at 15 54 54

[ONYX-1221]: https://artsyproduct.atlassian.net/browse/ONYX-1221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/schema/v2/me/index.ts
```diff
@@ -142,7 +142,7 @@ export const LengthUnitPreference = new GraphQLEnumType({
 export const meType = new GraphQLObjectType<any, ResolverContext>({
   name: "Me",
   interfaces: [NodeInterface],
-  fields: {
+  fields: () => ({
```
MounirDhahri, line 145: This was needed to avoid the circular dependency as specified here https://github.com/graphql/graphql-js/issues/373#issuecomment-215913280

nickskalkin: hm, not sure how it could cause this issue, but glad that you found solution

src/schema/v2/homeView/activityResolvers.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/*
+ * Resolvers for home view activity sections
+ */
+
+export const LatestActivityResolver: GraphQLFieldResolver<
+  any,
+  ResolverContext
+> = async (parent, args, context, info) => {
+  return await NotificationsConnection.resolve!(parent, args, context, info)
```
nickskalkin, line 13: 💅

nickskalkin (approved):
awesome!

src/schema/v2/me/index.ts
MounirDhahri: it was a pain to figure out 😄

src/schema/v2/homeView/__tests__/HomeViewSection.test.ts
```diff
@@ -478,4 +478,44 @@ describe("HomeViewSection", () => {
...
+
+      expect(data).toMatchInlineSnapshot(`
+        Object {
+          "homeView": Object {
+            "section": Object {
+              "__typename": "ActivityRailHomeViewSection",
+              "component": Object {
+                "title": "Latest Activity",
+              },
+            },
```
anandaroop, line 515: **question (non-blocking):** Since you eventually updated the data resolution (👏🏽) do want to reflect that in the spec as well?

anandaroop (approved):
LGTM just one q below…

dblandin (commented):
Awesome! 💯 

As Roop mentioned, updating the spec to reflect the data resolution would be nice 👍🏼

src/schema/v2/homeView/__tests__/HomeViewSection.test.ts
MounirDhahri: makes sense - I will update the specs

PR #5944, opened 2024-08-26
feat: add add auction results home view section

MounirDhahri:
This PR resolves [ONYX-1234]

This PR adds a new home view section for auction results. It also exposes the latest auction results to client apps.

**Query on Insomnia**

Screenshot 2024-08-26 at 16 16 01

___

**New section on Eigen**

[ONYX-1234]: https://artsyproduct.atlassian.net/browse/ONYX-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -3503,6 +3503,25 @@ type AuctionResultsByArtistsEdge {
   node: AuctionResultsByArtists
 }
 
+# An auction results rail section in the home view
+type AuctionResultsRailHomeViewSection implements GenericHomeViewSection & Node {
+  auctionResultsConnection(
+    after: String
+    before: String
+    first: Int
+    last: Int
+  ): AuctionResultConnection
+
+  # The component that is prescribed for this section
+  component: HomeViewComponent
+
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
+}
+
 enum AuctionResultsState {
   ALL
   PAST
@@ -11059,6 +11078,7 @@ union HomeViewSection =
   | ArticlesRailHomeViewSection
   | ArtistsRailHomeViewSection
   | ArtworksRailHomeViewSection
+  | AuctionResultsRailHomeViewSection
   | FairsRailHomeViewSection
   | HeroUnitsHomeViewSection
   | MarketingCollectionsRailHomeViewSection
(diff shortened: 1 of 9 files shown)


olerichter00 (approved):
🌟

src/schema/v2/homeView/HomeViewComponent.ts
```diff
@@ -99,5 +136,9 @@ export const HomeViewComponent = new GraphQLObjectType({
         }
       },
     },
+    behaviors: {
```
MounirDhahri, line 139: Following the agreed on schema here https://www.notion.so/artsy/Art-Advisor-Dynamic-Home-View-3615106ef393433a9f16dcf13b2414f2#682168e6cf9f47b9ad3cfbe1400a7005

_schemaV2.graphql
```diff
@@ -11054,11 +11076,25 @@ enum HomeViewComponentBackgroundImageURLVersion {
...
+  viewAll: HomeViewComponentBehaviorsViewAll
+}
+
+type HomeViewComponentBehaviorsViewAll {
+  # Text for the CTA of the view all button
+  buttonText: String!
+
+  # href of the view all button
+  href: String!
+}
```
anandaroop, line 11090: **questions (non-blocking):** Nice 👏🏽 

Some further questions:

 - Is this meant to handle both of the current (or future) kinds of placements for the view-all UI? Or would we distinguish between them in the response?
    


 - This clearly sets out the behavior where we want to go to a predefined route. But what about the other case: where we don't have a route but instead just expand to a full-screen view of the data. Would that be indicated by the absence of `href`? Or should it be signaled by the presence of some other attribute, with some more explicit gql modeling?

(No need to answer in this PR — and a reminder that we can introduce breaking changes till Milestone 2)

src/schema/v2/homeView/HomeViewComponent.ts
anandaroop: Seems like PR ticket might cover https://artsyproduct.atlassian.net/browse/ONYX-1157 as well. 

With that in mind, I left some related questions in my other comment.

anandaroop (approved):
Very nice. 

Some Qs about the new behavior, but doesn't have to be blocking.

nickskalkin (dismissed):
awesome!

PR #5945, opened 2024-08-26
feat(EMI-2017): Expose increased interest collector signal

erikdstock:
Completes [EMI-2017]

This PR exposes the new `artwork.increased_interest_signal` field through the metaphysics `artwork.collectorSignals` field. In addition it makes the `artwork.collectorSignals` field explicitly null if the artwork is not subject to collector signals.

[EMI-2017]: https://artsyproduct.atlassian.net/browse/EMI-2017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/artwork/collectorSignals.ts
+++ b/src/schema/v2/artwork/collectorSignals.ts
@@ -166,7 +166,7 @@ export const CollectorSignals: GraphQLFieldConfig<any, ResolverContext> = {
       increasedInterest: {
         type: new GraphQLNonNull(GraphQLBoolean),
         description: "Increased interest in the artwork",
-        resolve: (artwork, {}) => {
+        resolve: (artwork) => {
           return !!artwork.increased_interest_signal
         },
       },
--- a/src/schema/v2/artwork/collectorSignals.ts
+++ b/src/schema/v2/artwork/collectorSignals.ts
@@ -166,11 +166,8 @@ export const CollectorSignals: GraphQLFieldConfig<any, ResolverContext> = {
       increasedInterest: {
         type: new GraphQLNonNull(GraphQLBoolean),
         description: "Increased interest in the artwork",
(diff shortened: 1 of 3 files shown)


src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -163,10 +163,20 @@ export const CollectorSignals: GraphQLFieldConfig<any, ResolverContext> = {
           return activePartnerOffers?.[0]
         },
       },
+      increasedInterest: {
+        type: new GraphQLNonNull(GraphQLBoolean),
+        description: "Increased interest in the artwork",
+        resolve: (artwork, {}) => {
```
damassi, line 169: `{}`

erikdstock:
> Will we be adding a check to not return the increase interest signal if there is a partner offer?

I'm not sure if this would make it any easier for clients - in one case only the ideal data is there, in another extra data is there but in either we have to say `collectorSignals.partnerOffer || collectorSignals.increasedInterest`. 

We could change the type here from `increasedInterest: Boolean` to `label?: 'increased_interest' | x | y | z`. If we do that I might make the whole collector signal type more rigid to narrow presence of partner offer to that specific label, but to me this feels like a lot of extra complexity compared to just saying `collectorSignals.partnerOffer || collectorSignals.increasedInterest || collectorSignals.curatorsPick` on the client.

damassi:
@erikdstock, @rquartararo  - re 
> collectorSignals.partnerOffer || collectorSignals.increasedInterest || collectorSignals.curatorsPick on the client

The problem with this approach is that it doesn't stand up through time, compared to computing this from within MP. If client logic can live in MP, it _should_ live in MP. And if there are issues with this, we fix the issues!

PR #5959, opened 2024-08-29
feat(EMI-1948): merchant accounts loader

leamotta:
https://artsyproduct.atlassian.net/browse/EMI-1948

Add a simple loader to get the first active Stripe account to connect in the frontend at checkout

Screenshot 2024-08-29 at 14 14 25


src/schema/v2/partner/__tests__/partnerMerchantAccountsConnection.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      {
+        partner(id: "partnerID") {
+          merchantAccountsConnection(first: 1) {
+            edges {
+              node {
+                externalId
+              }
+            }
+          }
+        }
```
MrSltun, line 44: **Question:** are we going to get multiple accounts? since in the PR description it says we'll always return the first one

In case we'll always return one account, my suggestion, for simplicity, would be to return the external id instead of an array

MrSltun (commented):
Looks good! 🌟 
I just have a question/suggestion, but it's non-blocking :)

starsirius (commented):
Nice and straightforward! Let's add some tests and should be ready to go!

MrSltun (approved):
🌟🌟🌟

PR #5962, opened 2024-08-30
chore: better section names

MounirDhahri:
This PR resolves [ONYX-1242]

This PR comes as a follow-up on https://artsy.slack.com/archives/C05EQL4R5N0/p1724837376493849

The goal here is to remove the rail mention from the type name since that is already described by `section.component`. I noticed as well that we have no type-safety in place so I added that for a better DX

[ONYX-1242]: https://artsyproduct.atlassian.net/browse/ONYX-1242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nickskalkin (commented):
All good 👍  but please descriptions

anandaroop (commented):
As Nikita noted, I think there was some search-and-replace glitches here.

I went ahead and suggested changes to all the field descriptions since this consistency was already [on my mind](https://artsy.slack.com/archives/C05EQL4R5N0/p1724962528244299?thread_ts=1724962504.954069&cid=C05EQL4R5N0) anyway.

src/schema/v2/homeView/HomeViewSection.ts
```diff
@@ -29,39 +29,50 @@ const standardSectionFields: GraphQLFieldConfigMap<any, ResolverContext> = {
...
+  HeroUnitsHomeViewSection: "HeroUnitsHomeViewSection",
+  FairsHomeViewSection: "FairsHomeViewSection",
+  ArticlesHomeViewSection: "ArticlesHomeViewSection",
+  MarketingCollectionsHomeViewSection: "MarketingCollectionsHomeViewSection",
+  ShowsHomeViewSection: "ShowsHomeViewSection",
+  ViewingRoomsHomeViewSection: "ViewingRoomsHomeViewSection",
+  ActivityHomeViewSection: "ActivityHomeViewSection",
+  AuctionResultsHomeViewSection: "AuctionResultsHomeViewSection",
+  SalesHomeViewSection: "SalesHomeViewSection",
+} as const
```
anandaroop, line 45: Thanks for this also 🙏🏽

_schemaV2.graphql
```diff
@@ -46,8 +46,8 @@ type AccountRequest {
   notes: String
 }
 
-# An rail to show a list of user activity
-type ActivityRailHomeViewSection implements GenericHomeViewSection & Node {
+# An  to show a list of user activity
+type ActivityHomeViewSection implements GenericHomeViewSection & Node {
```
damassi, line 50: #emphatically-suggested-but-non-blocking 😄 Since all of these types are very specific to the home view, and we want to be able to easily scan for homeview types in the type graph (as well as at a glance in code) the best way to name multiple variants of a category is to stick to a common prefix:

```
HomeViewActivitySection
HomeViewArticlesSection
HomeViewFairsSection
etc
```

And then we're not hunting all around graphiql and our schema for things that could be automatically grouped together. 

See the line numbers to get an idea of my point, which will start to show up at scale in our code and elsewhere (including file-names):

Screenshot 2024-09-03 at 10 48 18 PM

src/schema/v2/homeView/HomeViewSection.ts
```diff
@@ -29,39 +29,50 @@ const standardSectionFields: GraphQLFieldConfigMap<any, ResolverContext> = {
...
+  ArtistsHomeViewSection: "ArtistsHomeViewSection",
+  HeroUnitsHomeViewSection: "HeroUnitsHomeViewSection",
+  FairsHomeViewSection: "FairsHomeViewSection",
+  ArticlesHomeViewSection: "ArticlesHomeViewSection",
+  MarketingCollectionsHomeViewSection: "MarketingCollectionsHomeViewSection",
+  ShowsHomeViewSection: "ShowsHomeViewSection",
+  ViewingRoomsHomeViewSection: "ViewingRoomsHomeViewSection",
+  ActivityHomeViewSection: "ActivityHomeViewSection",
+  AuctionResultsHomeViewSection: "AuctionResultsHomeViewSection",
+  SalesHomeViewSection: "SalesHomeViewSection",
```
damassi, line 44:
```tsx
export const HomeViewSectionTypeNames = {
  HomeViewActivitySection: "HomeViewActivitySection",
  HomeViewArticlesSection: "HomeViewArticlesSection",
  HomeViewArtistsSection: "HomeViewArtistsSection",
  HomeViewArtworksSection: "HomeViewArtworksSection",
  HomeViewAuctionResultsSection: "HomeViewAuctionResultsSection",
  HomeViewFairsSection: "HomeViewFairsSection",
  HomeViewGenericSection: "HomeViewGenericSection",
  HomeViewHeroUnitsSection: "HomeViewHeroUnitsSection",
  HomeViewMarketingCollectionsSection: "HomeViewMarketingCollectionsSection",
  HomeViewSalesSection: "HomeViewSalesSection",
  HomeViewShowsSection: "HomeViewShowsSection",
  HomeViewViewingRoomsSection: "HomeViewViewingRoomsSection",
}
```

(Alphabetized as well) 

Also to the point -- With the introduction of application regions (ie, "HomeSections") we're introducing the potential for _other_ application regions to be represented as types/code/file-names. If we don't prefix the region (the common category), finding things will be much more difficult at a glance.

src/schema/v2/homeView/HomeViewSection.ts
```diff
@@ -242,17 +242,17 @@ const SalesRailHomeViewSectionType = new GraphQLObjectType<
...
+    FairsHomeViewSectionType,
     HeroUnitsHomeViewSectionType,
-    MarketingCollectionsRailHomeViewSectionType,
-    ShowsRailHomeViewSectionType,
-    ViewingRoomsRailHomeViewSectionType,
-    SalesRailHomeViewSectionType,
+    MarketingCollectionsHomeViewSectionType,
+    ShowsHomeViewSectionType,
+    ViewingRoomsHomeViewSectionType,
+    SalesHomeViewSectionType,
```
damassi, line 255: Anytime we've got a big list of things, we should make sure they're all alpha

_schemaV2.graphql
araujobarret: + here, plus with the alphabetical sorting makes super easier to trace

_schemaV2.graphql
```diff
@@ -11124,6 +11064,126 @@ type HomeViewSectionEdge {
...
+  # The component that is prescribed for this section
+  component: HomeViewComponent
+
+  # A globally unique ID.
+  id: ID!
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
+}
+
```
damassi, line 11186: All grouped together now 👌

_schemaV2.graphql
damassi: Thanks for the fix! Resolving.

damassi (approved):

_schemaV2.graphql
anandaroop: Made the change in https://github.com/artsy/metaphysics/pull/5962/commits/2c5513da77f6f57cdf0c61af2ef877882dc6be4a

Only difference is I went with e.g. `HomeViewSectionActivity` instead of `HomeViewActivitySection` — while the latter reads more naturally, the former actually does a better job with colocating and skimmability, which was the stated goal.

damassi: totally 👌

anandaroop:
Posted about the friction [over here](https://artsy.slack.com/archives/C02BAQ5K7/p1725997635385819), and going to merge now.

PR #5963, opened 2024-08-30
feat(onyx-1157): add href and button texts to all home view sections that require it

nickskalkin:
I've added hrefs and button texts to all sections following @MounirDhahri 's approach from this PR: https://github.com/artsy/metaphysics/pull/5944

Let's debate if we want to implement the interface somehow else. @anandaroop you've left a few comments here: https://github.com/artsy/metaphysics/pull/5944#discussion_r1731598695

> Is this meant to handle both of the current (or future) kinds of placements for the view-all UI? Or would we distinguish between them in the response?

Currently both title section and buttons lead to the same page. I am not sure if it will be different in future, but if there will be a need to distinguish them, will it be possible to do it by expanding current schema and not changing it a lot? Something like this maybe:

```graphql
behaviors {
  sectionTitle {
    href
  }

  viewAll {
    href
    buttonText
  }
}
```

> This clearly sets out the behavior where we want to go to a predefined route. But what about the other case: where we don't have a route but instead just expand to a full-screen view of the data. Would that be indicated by the absence of href? Or should it be signaled by the presence of some other attribute, with some more explicit gql modeling?

Let's fantasize. Do you have a suggestion how to depict it? I've also noticed that some horizontal rails (Recommended Artists) have infinite pagination (not infinite technically, limited by 20 artists). Should this behavior also be a part of `behaviors`? 

Regarding naming of the field. When I was thinking about it, this is what I had in mind. Rail components are mostly react native flat lists, which have a ListFooterComponent property. We use it to configure "Browse All" buttons (see [ArtworksRail example](https://github.com/artsy/eigen/blob/main/src/app/Components/ArtworkRail/ArtworkRail.tsx#L50)). Lists can be horizontal or vertical, when horizontal - Browse All button moves to the right, when vertical (we don't have them on home page currently, but I saw something similar in draft designs of a new home feed) - to the bottom. I was thinking that maybe component definition should be more down to earth, something like this:

```graphql
component {
  title
  vertical # false by default

  footer { # optional
    button { # optional
      href
      text
    }

    pagination # optional, configure whether component has pagination and maybe how to paginate (max number of entities, etc.)
  }
}
```

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11099,7 +11099,7 @@ type HomeViewComponentBehaviorsViewAll {
   buttonText: String
 
   # href of the view all button
-  href: String!
+  href: String
 }
 
 union HomeViewSection =
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11096,7 +11096,7 @@ type HomeViewComponentBehaviors {
 
 type HomeViewComponentBehaviorsViewAll {
   # Text for the CTA of the view all button
-  buttonText: String!
+  buttonText: String
 
   # href of the view all button
   href: String!
--- a/src/schema/v2/homeView/HomeViewComponent.ts
+++ b/src/schema/v2/homeView/HomeViewComponent.ts
@@ -9,7 +9,7 @@ import { ResolverContext } from "types/graphql"
 export type HomeViewComponentBehaviors = {
   viewAll?: {
     href: string
-    buttonText: string
+    buttonText?: string
   }
 }
 const HomeViewComponentBehaviors = new GraphQLObjectType<
@@ -27,7 +27,7 @@ const HomeViewComponentBehaviors = new GraphQLObjectType<
             description: "href of the view all button",
           },
           buttonText: {
-            type: new GraphQLNonNull(GraphQLString),
+            type: GraphQLString,
             description: "Text for the CTA of the view all button",
           },
         },
--- a/src/schema/v2/homeView/HomeViewComponent.ts
+++ b/src/schema/v2/homeView/HomeViewComponent.ts
@@ -1,14 +1,9 @@
-import {
-  GraphQLEnumType,
-  GraphQLNonNull,
-  GraphQLObjectType,
-  GraphQLString,
-} from "graphql"
+import { GraphQLEnumType, GraphQLObjectType, GraphQLString } from "graphql"
 import { ResolverContext } from "types/graphql"
 
 export type HomeViewComponentBehaviors = {
   viewAll?: {
-    href: string
+    href?: string
     buttonText?: string
   }
 }
@@ -23,7 +18,7 @@ const HomeViewComponentBehaviors = new GraphQLObjectType<
         name: "HomeViewComponentBehaviorsViewAll",
         fields: {
           href: {
-            type: new GraphQLNonNull(GraphQLString),
+            type: GraphQLString,
             description: "href of the view all button",
           },
           buttonText: {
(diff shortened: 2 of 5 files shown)


dblandin:
The viewAll component response can take a few different froms, which I think cover our short-term needs without the need for extra properties:

**Query**

```graphql
{
  homeView {
    sectionsConnection(first: 10) {
      edges {
        node {
          component {
            title
            
            behaviors {
              viewAll {
                href
              }
            }
          }
        }
      }
    }
  }
}
```

**response 1: `null` viewAll**

- signals that no viewAll behavior is enabled
- No chevron next to the section title
- No "Browse all" button at the end of the rail


```json
{
  "homeView": {
    "sectionsConnection": {
      "edges": [
        "node": {
          "component": {
            "title": "New Works for You",
            "behaviors": {
              "viewAll": null
            }
          }
        }
      ]
    }
  }
}
```

**response 2: viewAll present but `null` href**

- signals that clients should enable default viewAll behavior, by navigating to a section-specific screen that knows how to query+render objects in a full width view.
- Section title/header is tappable. Chevron (or other indicator) is rendered. When tapped, client internally navigates to a default viewAll route for that section. Like `/home-view/section/home-view-section-new-works-for-you`.


```json
{
  "homeView": {
    "sectionsConnection": {
      "edges": [
        "node": {
          "component": {
            "title": "New Works for You",
            "behaviors": {
              "viewAll": {
                "href": null
              }
            }
          }
        }
      ]
    }
  }
}
```

**response 3: viewAll present and href value provided**

- Client renders that some UI bits including the chevron and "Browse all" button but instead of navigating to a default internal route, we navigate to the `href` value provided by the server.

```json
{
  "homeView": {
    "sectionsConnection": {
      "edges": [
        "node": {
          "component": {
            "title": "Curators' Picks",
            "behaviors": {
              "viewAll": {
                "href": "/collection/curators-picks-emerging-artists"
              }
            }
          }
        }
      ]
    }
  }
}
```

dblandin:
A couple other stray comments:

- Do we need to supply a "buttonText" property? If viewAll behavior is enabled, I imagine the client would know what kind of objects to expect (artworks/artists/auctions) to be able to render type-specific copy in the button without that being server-driven.
- I'm not super clear on the horizontal vs vertical list options — is there an example of this in Figma?

nickskalkin:
@dblandin Thanks for the detailed answer! A few comments from me:

Response 2 (href is `null`) - I think that when we want to redirect a user to a section-specific screen - Metaphysics could specify href of such section (`/home-view/section/home-view-section-new-works-for-you`). This way your **Response 2** becomes **Response 3** and we save some `if` statements on clients side.

> Do we need to supply a "buttonText" property?

Can't think of a case when we 100% would need `buttonText` at the moment. It gives us more flexibility, but we can add it in future. Do you think it makes API heavier without a reason?

> I'm not super clear on the horizontal vs vertical list options — is there an example of this in Figma?

I couldn't find a Figma link, but take a look a [demo day recording](https://drive.google.com/drive/folders/1loO103fcarfImsczAOlDZzd9fzUhxxbA) (go to 25:20). Jimi mentions that that some sections might become vertical, if I understood him correctly.

nickskalkin:
@dblandin I made viewAll.href optional. Not removing viewAll.buttonText now, because it will break Eigen. Once this PR is merged I will follow-up in Eigen and remove buttonText from here

dblandin (approved):
👍🏼

anandaroop (approved):
Thanks for hashing this out today 🙏🏽

PR #5974, opened 2024-09-04
feat(onyx-1217): add withTimeout to home view section resolvers

nickskalkin:
[Jira ticket](https://artsyproduct.atlassian.net/browse/ONYX-1217) 🔒 

In order to keep the home view responsive, we don’t want to block the home view  response in the case where an individual section is taking too long to resolve its data. We achieve this by wrapping each section's resolver in `withTimeout` helper, which throws an error when exceeding the timeout.

Timeouts are configurable by modifying environment variables, such as `HOME_VIEW_NEWS_TIMEOUT_MS`, `HOME_VIEW_HERO_UNITS_TIMEOUT_MS`, etc. Initially I've set default option to 2000 (2 seconds, which seemed generous to me), but testing on staging showed that it is not enough for some sections. I've increased timeout threshold to 4 seconds by default.


src/lib/loaders/helpers.ts
```diff
@@ -9,7 +9,7 @@ const MAX_STEPS = 1
  */
 export const withTimeout = async (loader: Promise<any>, timeoutMs: number) => {
   try {
-    let data: unknown
+    let data
```
nickskalkin, line 12: When using `withTimeout` in conjunction with `connectionFromArray`, like this:

```
const result = await withTimeout(resolver(...), timeout)

return connectionFromArray(result, args)
```

I was getting this complaint from typescript:

```
Argument of type 'unknown' is not assignable to parameter of type 'readonly unknown[]'.
```

Maybe this is a smarter way to get rid of this error message, let me know

dblandin (commented):
Would it be more maintainable/discoverable to consolidate these timeout configurations within our existing [registry file](https://github.com/artsy/metaphysics/blob/6d077b93db331866a002f19a39cafde0307cb5d7/src/schema/v2/homeView/sections.ts#L4)?

Thinking of something like the following:

```typescript
export const SimilarToRecentlyViewedArtworks: HomeViewSection = {
  // ...snip...
  resolver: withTimeout(SimilarToRecentlyViewedArtworksResolver, config.DEFAULT_HOMEVIEW_SECTION_TIMEOUT),
}

export const CuratorsPicksEmerging: HomeViewSection = {
  // ...snip...
  resolver: withTimeout(CuratorsPicksEmergingArtworksResolver, config.DEFAULT_HOMEVIEW_SECTION_TIMEOUT),
}

export const RecentlyViewedArtworks: HomeViewSection = {
  // ...snip...
  // Example of a specific section choosing a non-default timeout value
  resolver: withTimeout(RecentlyViewedArtworksResolver, 4000),
}
```

Also wondering if a single new environment variable could be shared as the default timeout value while giving specific sections the ability to choose a non-default value if necessary. But ideally, all sections would be able to resolve under a shared threshold.

araujobarret (commented):
Not sure if I follow how that would be in place given the main pagination query on the home feed, can you elaborate more how would it work?

I see you added to the individual level of the resolvers but this mechanism in my mind doesn't work with pagination because on the Relay level you're waiting for the next cursor before being able to paginate more to other cursors.

nickskalkin:
@dblandin I like your idea to move `withTimeout` to sections! Do you think it looks fine like this?

```typescript
export const SimilarToRecentlyViewedArtworks: HomeViewSection = {
  id: "home-view-section-similar-to-recently-viewed-artworks",
  type: "ArtworksRailHomeViewSection",
  resolver: async (parent, args, context, info) => {
    return await withTimeout(
      SimilarToRecentlyViewedArtworksResolver(parent, args, context, info),
      config.HOME_VIEW_SIMILAR_TO_RECENTLY_VIEWED_ARTWORKS_TIMEOUT_MS
    )
  },
}
```

> Also wondering if a single new environment variable could be shared as the default timeout value while giving specific sections the ability to choose a non-default value if necessary. But ideally, all sections would be able to resolve under a shared threshold.

What if I extract this `4000` fallback into a generic environment variable (`HOME_VIEW_RESOLVERS_TIMEOUT_MS`)? This will allow us to be as flexible as we want to - configure timeout for all resolvers at once, but allowing individual resolvers to overwrite the generic setting.

nickskalkin:
Attaching a link to a discussion that @araujobarret started about pagination issue - [slack](https://artsy.slack.com/archives/C05EQL4R5N0/p1725449275651999)

dblandin:
> I like your idea to move withTimeout to sections! Do you think it looks fine like this?

That certainly feels better to me, co-located in the same file, but it still introduces quite a bit of duplication invoking the withTimeout function and passing in the timeout value.

What do you think of this? Seems to work fine locally:

```typescript
const withHomeViewTimeout = (resolver: GraphQLFieldResolver<any, ResolverContext>, timeout: number = config.HOME_VIEW_RESOLVER_TIMEOUT_MS): GraphQLFieldResolver<any, ResolverContext> => {
  return async (parent, args, context, info) => {
    return await withTimeout(resolver(parent, args, context, info), timeout)
  }
}

export const SimilarToRecentlyViewedArtworks: HomeViewSection = {
  // ...snip...
  resolver: withHomeViewTimeout(SimilarToRecentlyViewedArtworksResolver),
}
```

anandaroop:
Nice, I like how this is shaping up per the discussion above.

One thing to note is that when the timeout kicks in the response _continues_ to contain the requested section, but with an `null` connection and a corresponding error key:



This seems fine to me. We haven't fully spec'd out what should happen on the client side in this circumstance, but this might give us the flexibility to partially render the section with something like "Failed to load, try again 🔄" or something like that (if we want).

anandaroop:
(@araujobarret I don't know if that helps — the section _is_ still present, so pagination would remain consistent in this scenario)

nickskalkin:
@dblandin PR looks much better now thanks to you 🙌

dblandin (approved):
Nice! LGTM 👍🏼

araujobarret (approved):

PR #5982, opened 2024-09-06
fix: correct title for Recently viewed home view section

nickskalkin:
correct title for Recently viewed home view section


anandaroop (approved):
👍🏽

PR #5984, opened 2024-09-09
chore: update artwork rails href

MounirDhahri:
This PR updates the home artwork section with a new href that will be valid for the new generic Artwork screens.

The route is open for debate.
Initially, I named the route `home-view/section/sectionID` in Eigen. However that led to more complication around telling which type of section it: Artists? artworks? auctions. For each of those section we have a different placeholder. So there was a need to make 2 requests, one to tell what type of section it is. and one to query for the data. This can be avoided by being explicit about the route URL. I am interested to hear what you think about it?  

https://github.com/user-attachments/assets/1670b06a-2f6b-4568-94ab-c2950d11ed23

Related Eigen PR: https://github.com/artsy/eigen/pull/10546

### DIFF
--- a/src/schema/v2/homeView/__tests__/HomeViewSection.test.ts
+++ b/src/schema/v2/homeView/__tests__/HomeViewSection.test.ts
@@ -38,7 +38,7 @@ describe("HomeViewSection", () => {
           "component": Object {
             "behaviors": Object {
               "viewAll": Object {
-                "href": "/similar-to-recently-viewed",
+                "href": "/home-view/section/artworks/home-view-section-similar-to-recently-viewed-artworks",
               },
             },
             "title": "Similar to Works You’ve Viewed",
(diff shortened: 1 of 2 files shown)


dblandin (commented):
As discussed [here](https://github.com/artsy/metaphysics/pull/5963), if we're intending to signal to the client that it should use a default HomeView/section viewAll behavior, we should pass a `null` or undefined href.

Any `href` values specified at this level should trigger non-default navigation to other screens/pages and should also reference stable cross-platform paths.

MounirDhahri:
Here I am actually not using the default behaviour of going to `home-view/section/sectionID`. Instead here I am navigating directly to `home-view/section/artworks/sectionID`. This is because I wanted to avoid making a different request to get the type of the section and render the right placeholder.

PR #5997, opened 2024-09-12
feat(EMI-2052): avoid returning lotWatcherCount for closed lots

fladson:
https://artsyproduct.atlassian.net/browse/EMI-2052

~Return `null` instead of `0` for the `artwork.collectorSignals.auction.lotWatcherCount` if the lot is closed for bidding.
I got the lot closed logic from the current `lotClosesAt` field, so any feedback is welcome on that.~

Return no auction data if the lot is closed for bidding.

@artsy/emerald-devs

### DIFF
--- a/src/schema/v2/artwork/collectorSignals.ts
+++ b/src/schema/v2/artwork/collectorSignals.ts
@@ -52,7 +52,11 @@ const AuctionCollectorSignals: GraphQLFieldConfig<any, ResolverContext> = {
       (activeLotData.saleArtwork.extended_bidding_end_at ||
         activeLotData.saleArtwork.end_at ||
         activeLotData.sale.ended_at)
-    const isBiddingClosed = lotClosesAt && new Date(lotClosesAt) <= new Date()
+
+    // If the lot is closed for bidding, return null
+    if (lotClosesAt && new Date(lotClosesAt) <= new Date()) {
+      return null
+    }
 
     // Resolve all associated auctions data in one object
     return {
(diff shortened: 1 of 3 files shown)


src/schema/v2/artwork/collectorSignals.ts
```diff
@@ -47,11 +47,20 @@ const AuctionCollectorSignals: GraphQLFieldConfig<any, ResolverContext> = {
...
+        activeLotData.sale.ended_at)
+    const isBiddingClosed = lotClosesAt && new Date(lotClosesAt) <= new Date()
+
     // Resolve all associated auctions data in one object
     return {
       artwork,
       saleArtwork: activeLotData.saleArtwork,
       sale: activeLotData.sale,
+      lotClosesAt,
+      isBiddingClosed,
```
erikdstock, line 63: Rather than loading these values up here, I think the preferred pattern is for this to go in the resolver below. We can use `sale` and `saleArtwork` down there, included because this parent resolver loads them here.

However,
- if the `isBiddingClosed` property means that the entire `auction` field could be null, we could do that here
- if we want to re-use the `lotClosesAt` logic down below, we can extract a function that takes the same sale/saleArtwork args for both.

erikdstock (commented):
Makes sense - I think our preferred pattern is to organize the field resolution in the individual fields so made some suggestions based on that, but it's possible this is so trivial that doing it this way is fine. cc @damassi in case you have opinions on it.

erikdstock (approved):

PR #6007, opened 2024-09-13
chore(homeView): switch from union to interface

dblandin:
This allows us to query for generally support attributes across all sections without needed a named fragment like `... on HomeViewSectionGeneric`. This does switch the connection from exposing a Union to an Interface, but all else seems to work exactly the same.

I don't see much of a downside here. There was some [recent discussion in Slack](https://artsy.slack.com/archives/C05EQL4R5N0/p1726078732679849?thread_ts=1726069270.194679) regarding the `matchConnection` field using a union type instead of an interface implemented for `searchConnection` but I think the real problems with the `searchConnection` field stems more from [the query introspection and this error](https://github.com/artsy/metaphysics/blob/7bd5e7a988f761185a2310b3132eca9533750ca0/src/schema/v2/search/SearchResolver.ts#L64-L68) rather than the implementation of an interface-backed connection.

## Example

```graphql
{
  homeView {
    sectionsConnection(first: 5) {
      __typename
      internalID
      component {
        title
      }
      contextModule
    }
  }
}
```

### DIFF
--- a/src/schema/v2/homeView/HomeViewSection.ts
+++ b/src/schema/v2/homeView/HomeViewSection.ts
@@ -5,7 +5,6 @@ import {
   GraphQLNonNull,
   GraphQLObjectType,
   GraphQLString,
-  GraphQLUnionType,
 } from "graphql"
 import { pageable } from "relay-cursor-paging"
 import { ResolverContext } from "types/graphql"
@@ -60,10 +59,13 @@ export const HomeViewSectionTypeNames = {
   HomeViewSectionViewingRooms: "HomeViewSectionViewingRooms",
 } as const
 
-const HomeViewGenericSectionInterface = new GraphQLInterfaceType({
+export const HomeViewGenericSectionInterface = new GraphQLInterfaceType({
   name: HomeViewSectionTypeNames.HomeViewSectionGeneric,
   description: "Abstract interface shared by every kind of home view section",
   fields: standardSectionFields,
+  resolveType: (value) => {
+    return value.type
+  },
 })
 
 // concrete sections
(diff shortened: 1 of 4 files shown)


src/schema/v2/homeView/HomeViewSection.ts
```diff
@@ -60,10 +59,13 @@ export const HomeViewSectionTypeNames = {
...
 } as const
 
-const HomeViewGenericSectionInterface = new GraphQLInterfaceType({
+export const HomeViewGenericSectionInterface = new GraphQLInterfaceType({
   name: HomeViewSectionTypeNames.HomeViewSectionGeneric,
   description: "Abstract interface shared by every kind of home view section",
   fields: standardSectionFields,
+  resolveType: (value) => {
+    return value.type
+  },
```
anandaroop, line 68: Ahh nice, I actually didn't realize interfaces also took `resolveType`  

(I think we've got dozens of these for unions and [only one](https://github.com/artsy/metaphysics/blame/7bd5e7a988f761185a2310b3132eca9533750ca0/src/schema/v2/artwork/artworkContextGrids/index.ts#L40-L63) for interfaces!)

I think that's what makes this possible — love it.

anandaroop (approved):
Nice!

And confirmed locally that, despite the schema change, this is a no-op over in Eigen — that's your understanding as well right @dblandin?

Thanks for sorting this out 🙏🏽

anandaroop:
> this is a no-op over in Eigen

Yep, all good in TestFlight too

src/schema/v2/homeView/index.ts
```diff
@@ -11,12 +11,12 @@ import {
...
   paginationResolver,
 } from "../fields/pagination"
-import { HomeViewSectionType } from "./HomeViewSection"
+import { HomeViewGenericSectionInterface } from "./HomeViewSection"
 import { getSectionsForUser } from "./getSectionsForUser"
 import { registry } from "./sections"
 
 const SectionsConnectionType = connectionWithCursorInfo({
-  nodeType: HomeViewSectionType,
+  nodeType: HomeViewGenericSectionInterface,
```
MounirDhahri, line 19: Shouldn't this be the union?

dblandin: No, this is an intentional change from Union > Interface.

PR #6013, opened 2024-09-16
chore(analytics): add ownerType value to home view viewAll signals

dblandin:
Add `ownerType` to viewAll behaviors with an explicit `href` provided so we can reliably pass this value through to analytics event payloads.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10912,7 +10912,7 @@ type HomeViewComponent {
   description: String
 
   # A screen to navigate to when this component is clicked
-  href: String
+  href: String @deprecated(reason: "Use `behaviors.viewAll.href` instead")
 
   # A display title for this section
   title: String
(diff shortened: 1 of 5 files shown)


_schemaV2.graphql
```diff
@@ -10912,7 +10912,7 @@ type HomeViewComponent {
   description: String
 
   # A screen to navigate to when this component is clicked
-  href: String
+  href: String @deprecated(reason: "Use `behaviors.viewAll.href` instead")
```
anandaroop, line 10915: **thought (non-blocking):** Thanks! I think we could go one step further and remove this, once we've verified Eigen doesn't currently use it. Might as well clean up cruft while we're still [allowed to make breaking changes](https://www.notion.so/artsy/Art-Advisor-HomeView-Rollout-7b674fb1371c4d3e864fd08111f56644?pvs=4#48ca0c764af445cb955a04a8a8cc71fc) — but that can be a separate PR.

dblandin: Yeah, I'm pretty sure we don't reference it any more but I'll make sure and follow-up with the removal in a separate PR.

PR #6015, opened 2024-09-16
fix(WIP): Do not allow setting MyC artwork condition to Null in mutations

olerichter00:
Related to https://github.com/artsy/eigen/pull/10785

## Description

With this, we don't allow setting MyC artwork's `condition` to `null` in mutations because it's [not allowed in Gravity](https://github.com/artsy/gravity/blob/2049ddd10a5ab8d964e1bec5644b8bc5737d463b/app/models/concerns/my_collection_artwork.rb#L4) and we're getting an error message that `null` is not allowed when we try to. 

This is fix is related to https://github.com/artsy/eigen/pull/10785.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -13937,7 +13937,7 @@ input MyCollectionCreateArtworkInput {
 
   # The given location of the user as structured data
   collectorLocation: EditableLocation
-  condition: ArtworkConditionEnumType
+  condition: ArtworkConditionEnumType!
   conditionDescription: String
   confidentialNotes: String
   costCurrencyCode: String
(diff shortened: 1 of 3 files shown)


nickskalkin:
I read [this](https://github.com/artsy/gravity/blob/2049ddd10a5ab8d964e1bec5644b8bc5737d463b/app/models/concerns/my_collection_artwork.rb#L44) as condition can be `nil`. 

At least I can do this in console:

```ruby
w = Artwork.find("667d740df781c9070ddcc63d")
w.condition = nil
w.save!
```

Could it be that there is an issue with how metaphysics sends null condition (when I pass `condition: null` in mutation payload - MP fires such request to gravity `"artwork/667d740df781c9070ddcc63d?condition="`)? Or something is wrong on gravity's endpoint side?

olerichter00:
@nickskalkin, yeah, probably it's not about setting `condition` to `null`, but how the `null` is passed from the client through MP to Gravity. Maybe it is cast to a string or somewhere in between.

olerichter00:
@nickskalkin, yeah, probably it's not about setting condition to null, but how the null is passed from the client through MP to Gravity. Maybe it is cast to a string or somewhere in between.

PR #6017, opened 2024-09-16
ci(cloudflare): Add automated deployments

damassi:
This has yet to be updated with the exact script, but this PR supplies the infrastructure, pointed at our ci-test-worker namespace. 

Opening to get the CircleCI bits right. 

cc @artsy/diamond-devs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,11 +119,25 @@ workflows:
             - push-staging-image
             - deploy-staging
 
+      - node/run:
+          name: deploy-cloudflare-workers-staging
+          context: cloudflare
+          <<: *only_main
+          pkg-manager: yarn
+          yarn-run: deploy-cloudflare-workers:staging
+
       # release
       - horizon/block:
           <<: *only_release
           context: horizon
           project_id: 18
+
+      - node/run:
+          name: deploy-cloudflare-workers
+          context: cloudflare
+          <<: *only_release
+          pkg-manager: yarn
+          yarn-run: deploy-cloudflare-workers
 
       - hokusai/deploy-production:
           <<: *only_release
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .env.staging
 .env.shared
 .DS_Store
+.wrangler
 build
 node_modules
 dump.rdb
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -119,6 +127,13 @@ workflows:
             - push-staging-image
             - deploy-staging
 
+      - node/run:
+          name: deploy-cloudflare-workers
+          context: cloudflare
```
damassi, line 132: Added new context

.circleci/config.yml
```diff
@@ -25,6 +25,14 @@ only_release: &only_release
     branches:
       only: release
 
+only_staging_or_release: &only_staging_or_release
+  context: hokusai
```
joeyAghion, line 29: Can this be removed?

scripts/deploy-cloudflare-workers.sh
```diff
@@ -0,0 +1,25 @@
...
+# Staging deploys
+if git diff --name-only origin/main | grep '^workers/' > /dev/null; then
+  echo "[deploy-cloudflare-workers] Change detected. Deploying to staging..."
+
+  ENV="staging"
+  DEPLOY=true
+
+# Production Deploys
+elif [[ "$CIRCLE_BRANCH" == "release" ]]; then
+  if git diff --name-only HEAD~1 | grep '^workers/' > /dev/null; then
```
joeyAghion, line 12: I don't think we can guarantee there's a diff in the most recent commit. As an alternative, is it OK to always deploy the worker?

scripts/deploy-cloudflare-workers.sh
```diff
@@ -0,0 +1,25 @@
...
+    DEPLOY=true
+  fi
+else
+  echo "[deploy-cloudflare-workers] No change detected. Skipping deployment."
+fi
+
+# Only deploy if a change was detected
+if [ "$DEPLOY" = true ]; then
+  yarn wrangler deploy --config=workers/test/ciTestWorker.toml --env $ENV
+fi
```
joeyAghion, line 25: I think it would simplify things to call this with an explicit `staging` or `production` argument from `.circleci/config.yml`.

scripts/deploy-cloudflare-workers.sh
damassi: It is ok to deploy every time! I just wanted to avoid it because of the scary intersection. But its likely fine.

PR #6019, opened 2024-09-17
ci(cloudflare): Add mpcache CloudFlare worker deploy

damassi:
Addresses [DIA-849]

Adds our new mpcache cloudflare worker into the repo, and adds support for CI deploys. 

cc @artsy/diamond-devs 

[DIA-849]: https://artsyproduct.atlassian.net/browse/DIA-849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -15,8 +15,8 @@
     "build:remote-schemas": "cp src/data/*.graphql build/src/data/",
     "build": "yarn build:lib && yarn build:index && yarn build:fixtures && yarn build:remote-schemas",
     "ci": "yarn test",
-    "deploy-cloudflare-workers:staging": "yarn wrangler deploy --config=workers/test/ciTestWorker.toml --env staging",
-    "deploy-cloudflare-workers": "yarn wrangler deploy --config=workers/test/ciTestWorker.toml --env staging",
+    "deploy-cloudflare-workers:staging": "yarn wrangler deploy --config=workers/caching/mpcache.toml --env staging",
+    "deploy-cloudflare-workers": "yarn wrangler deploy --config=workers/caching/mpcache.toml --env production",
```
damassi, line 19: This is obviously very minimal, but we'll cross this bridge when a second worker is added.

workers/caching/mpcache.ts
```diff
@@ -0,0 +1,116 @@
...
+      const hashBuffer = await crypto.subtle.digest("SHA-256", msgBuffer)
+
+      // convert bytes to hex string
+      return [...new Uint8Array(hashBuffer)]
+        .map((b) => b.toString(16).padStart(2, "0"))
+        .join("")
+    }
+
+    // Construct a GET request with digest appended to path, to serve as cache key.
+    async function getCacheKey(request) {
```
araujobarret, line 23: can't we move the smaller functions outside? it'd make it better to read the `default` block

araujobarret (approved):
Looks good, once the logs are removed I think we can merge it

.circleci/config.yml
```diff
@@ -120,9 +120,9 @@ workflows:
             - deploy-staging
 
       - node/run:
+          <<: *only_main
```
joeyAghion, line 123: This deploys the worker even before tests pass. I think it should instead be after everything (like `push-schema-changes`). In practice, when we make application-level changes coupled to worker-level changes, I'm not _sure_ which should deploy first. (We should probably avoid that.)

joeyAghion (commented):
Gosh now I'm embarrassed of that code 🙈

.circleci/config.yml
damassi: Updated!

damassi:
@araujobarret - these logs all match code currently deployed on CloudFlare; i just moved that over here. So ideally we don't change anything for now and refactors can take place in future prs, once our pipelines are perfectly in sync with current production code.

workers/caching/mpcache.toml
```diff
@@ -0,0 +1,8 @@
+main = './mpcache.ts'
+compatibility_date = "2024-09-16"
+
+[env.staging]
+name = "mpcache-staging"
+
+[env.production]
+name = "mpcache"
```
joeyAghion, line 8: Is 2 separate (and differently named) workers the "typical" way to separate QA and production environments?

joeyAghion (approved):
I love the code-review this has _already_ inspired. I think organizing the functions, making the worker less verbose, and maybe dropping the `log`s are all good enhancements to make after this (once there's a safe way to iterate)! For this PR's purpose, I think it's acceptable to focus on the release working.

workers/caching/mpcache.toml
damassi: Yup, and can target via the `--env` flag

damassi: All kinds of various settings can be embedded under here, so its highly configurable per environment.

PR #6024, opened 2024-09-17
chore: Remove legacy demand queries AMBER-955

jonallured:
This PR removes two queries from MP:

* artist.alertsConnection
* partner.alertsSummaryArtistsConnection

They are no longer used now that we have shipped the new version of the Demand page in CMS!! 🎉 

Note: I'm leaving this marked as DONT-MERGE and will return to it on Monday in an abundance of caution.

https://artsyproduct.atlassian.net/browse/AMBER-955

/cc @artsy/amber-devs


_schemaV2.graphql
```diff
@@ -14504,19 +14463,6 @@ type Partner implements Node {
     page: Int
     size: Int
   ): PartnerAlertsConnection
-  alertsSummaryArtistsConnection(
```
jpotts244, line 14507: Sweet with the removal of this connection, I think we can also remove these fields from the `AlertType`

https://github.com/artsy/metaphysics/blob/f0fe20e0c37fc2baedd0e8d63f9ac135724f9c6c/src/schema/v2/Alerts/index.ts#L392-L404

_schemaV2.graphql
```diff
@@ -317,7 +317,6 @@ type Alert {
   ): String!
```
jpotts244, line 317: The removal of fields and connection types here looks ✅ but I always get a little nervous of the MP magic.
But to chat out loud, the removal of this stuff here is OK since Volt was the only client using these fields and Volt not longer is using them

src/schema/v2/Alerts/index.ts
```diff
@@ -388,20 +388,6 @@ export const AlertType = new GraphQLObjectType<
...
-      // - `alertsConnection` under `Artist`
-      // - `alertsSummaryArtistsConnection` under `Partner`.
-      totalUserSearchCriteriaCount: {
-        type: GraphQLInt,
-        resolve: ({ count_30d }) => count_30d,
-      },
-      hasRecentlyEnabledUserSearchCriteria: {
-        type: GraphQLBoolean,
-        resolve: ({ count_7d }) => count_7d && count_7d > 0,
-      },
```
jpotts244, line 404: I think we can also remove the `count_30/7` that these guys are using as well

Ex: https://github.com/artsy/metaphysics/pull/6024/files#diff-87dc75f67d185f47ad8c6c3666c7cae7cd3f9ce3759d9145c386e58919ddef7eR166-R168

jpotts244: and if we follow that back can probably remove these Edge types on Alert as well!

```
const AlertsEdgeFields = {
  counts: {
    type: new GraphQLObjectType({
      name: "AlertsCounts",
      fields: {
        totalUserSearchCriteriaCount: {
          type: GraphQLInt,
          resolve: ({ count_30d }) => count_30d,
        },
      },
    }),
    resolve: (x) => x,
  },
  isRecentlyEnabled: {
    type: GraphQLBoolean,
    resolve: ({ count_7d }) => count_7d > 0,
  },
}
```

jpotts244: Update: we are pulling edge field retirement into separate ticket of work
https://artsyproduct.atlassian.net/browse/AMBER-1041

PR #6033, opened 2024-09-19
[HOLD] chore(workers): rename workers to something more self-describing

damassi:
This updates our worker code with the new renames: `metaphysics-cdn`. Deploys point at `-staging`. 

cc @artsy/amber-devs


damassi:
Closing as this is already happening in another branch!

PR #6034, opened 2024-09-19
chore: rename CDN endpoints and workers to metaphysics-cdn*

joeyAghion:
This introduces more natural, consistent names for the CDN-backed Metaphysics endpoints. Once deployed, new `metaphysics-cdn` and `metaphysics-cdn-staging` names will be available.

We'll need to preserve the `-alt` ingress definitions and DNS until that traffic dies down (which could be a while). The `mpcache*` workers can be cleaned up post-release, routing the old names to the new workers instead.

### Migration

I've _already_ completed the following:
* updated existing firewall rules to add new hostnames
* created 2 new workers with identical code and env vars
* defined routes from the new names to the new workers

Post-release, we can:
* route old names to new workers and clean up old workers

Much later, we can:
* retire `*-alt` ingress definitions and DNS


hokusai/production.yml
```diff
@@ -159,7 +159,17 @@ spec:
...
+    http:
+      paths:
+      - path: /
+        pathType: Prefix
+        backend:
+          service:
+            name: metaphysics-web-internal
+            port:
+              name: mp-http
+  - host: metaphysics-production-alt.artsy.net # preserve this legacy worker hostname until traffic fully migrates away
```
damassi, line 172: 🙏

damassi (approved):
Makes much more sense now 👍

PR #6038, opened 2024-09-20
feat: unstitch user devices

egdbear:
This PR is the first step in unstitching devices from GravQL, following up on [Gravity PR](https://github.com/artsy/gravity/pull/18166), related to enabling device retrieval for trusted apps and admins. 

The query has so far only been used in [Pulse](https://github.com/artsy/pulse/blob/cf6a49d42e060b6d838fcecf83655dc2a799d8a8/app/operations/metaphysics.rb#L109), and this update should not introduce any breaking changes.

The code is behind the environment flag `USE_UNSTITCHED_USER_DEVICES`, which is currently set to false on both staging and production. Once the Gravity PR is merged, we can proceed, and the staging environment variable can be updated to true.

For more details, check out the [#unstitching](https://artsy.slack.com/archives/C07NAMFES5Q) channel on Slack.

:future-friday-unstitching:


damassi (approved):
Yayyy

mzikherman (approved):

PR #5893, opened 2024-08-07
docs: add docs for creating a new home view section

MounirDhahri:
This PR is still in progress - reviews are welcome though 👋 

This PR defines some guidelines on how to add a new home view section.

What's missing:
- [ ] Some polishing to content (better english, better links)
- [ ] Link example PR
- [ ] Add KS video link

PR #6043, opened 2024-09-24
docs(homeView): add documentation and an example home view section

anandaroop:
In the course of preparing the KS session I have been:

- annotating the TS types in order to document them and enable intellisense
- creating an `ExampleSection` that can explain all the the various options in greater depth

I'll make use of this in the KS regardless, but…

👉🏽 I'm seeking opinions on whether this is worth **committing**.

I can see it both ways:

**PRO**: Thorough documentation that hopefully facilitates other teams self-serving as they work within the new home view conventions.

**CON**: The added and ongoing burden of keeping everything in sync — the TS doc comments ⇔ GraphQL field descriptions ⇔ the annotated `ExampleSection` file.

Thoughts?


damassi:
The code for all of this home view stuff is very clean and self-describing! I'd vote nay for docs like this, unless its like a best-practices-like doc for how to add new home view sections (through working code that one could copy paste to get started).

src/schema/v2/homeView/HomeViewComponent.ts
```diff
@@ -2,13 +2,21 @@ import { OwnerType } from "@artsy/cohesion"
...
+  /** Represents the behavior of the View All button */
   viewAll?: {
+    /** Route to navigate to, or null for default view-all behavior */
     href?: string | null
+
+    /** Text for the CTA of the View All button */
     buttonText?: string
+
+    /** TODO: confirm whether this correct or HomeViewSection.ownerType is correct */
     ownerType?: OwnerType
```
dblandin, line 16: @anandaroop Yeah, I can see how this is a bit confusing. `viewAll.href` + `viewAll.ownerType` are used together to navigate to an arbitrary screen _and_ give the client awareness of what _type_ of screen it represents for an analytics payload.

If `viewAll.href` is `null`, we can fallback to the client to implement default behavior, which could be to render a section-specific viewAll screen, in which case it seemed appropriate to specify `ownerType` on the top-level section configuration instead of requiring clients to drill down to `component > behaviors > viewAll > ownerType` in the query.

anandaroop:
Thanks — and yes I'm inclined now to _not_ commit this, but boil down a simpler version of it into a markdown doc, including copy-paste-able section example.

dblandin:
I agree that it might not be necessary to add inline code comments to the main supporting files but I feel like the committed _ExampleSection.ts file might be a useful artifact for teams to easily copy/paste into new section definitions.

I've done that a number of times in Force with the [ExampleApp](https://github.com/artsy/force/blob/ffe4f84eb485095fe1cb54a6380129df095c98cc/src/Apps/Example/ExampleApp.tsx).

As you [commented](https://github.com/artsy/metaphysics/pull/6043#issuecomment-2371371452), this file could be slimmed down and inlined into a markdown doc instead, which I'd support also 👍🏼

anandaroop:
Ok this is where I landed: 

- removed the inline TS comments

- kept the` _ExampleSection.ts` file
  -  I like that it actually gets type-checked and thus will do a better job of avoiding drift than copy-paste from a markdown doc 
  -  (And reminder to self that there already are some WIP markdown docs over at https://github.com/artsy/metaphysics/pull/5893)

anandaroop:
I'll go ahead and merge this, we can feel free to modify of course.

MounirDhahri:
cc @araujobarret this is a great starting point as well

PR #6049, opened 2024-09-25
chore: rename legacy zone to default

dblandin:
We initially created support for the "zone" concept thinking that we would switch from a "legacy" set of sections to a completely different set of sections, possibly backed by multiple zones.

That need seems less clear and we're now planning more incremental changes, often backed by feature flags or A/B testing setup, to a default set of sections.

This commit removes the label "legacy" from this set of sections in favor of a more neutral "default" name.


anandaroop (approved):
Ha, I almost did this yesterday but was too busy fussing on the KS.

I'll merge this and rebase my live-coding branches now 👌🏽

PR #6053, opened 2024-09-25
feat(homeView): define an optional displayability check per section

anandaroop:
Here we add the ability for a section to define its own displayability check that can govern whether the section should be displayed, considering any arbitrary data or condition _beyond_ the standard ones (auth, feature flags) that we've already defined

(See https://github.com/artsy/metaphysics/commit/d9562a59558fd710653ceb7cfce446dfe3cb9a38  for a less noisy view)

### DIFF
--- a/src/schema/v2/homeView/helpers/isSectionDisplayable.ts
+++ b/src/schema/v2/homeView/helpers/isSectionDisplayable.ts
@@ -27,5 +27,10 @@ export function isSectionDisplayable(
     })
   }
 
+  // section's display pre-check
+  if (typeof section.shouldBeDisplayed === "function") {
+    isDisplayable = isDisplayable && section?.shouldBeDisplayed(context)
+  }
+
   return isDisplayable
 }
--- a/src/schema/v2/homeView/sections/index.ts
+++ b/src/schema/v2/homeView/sections/index.ts
@@ -43,6 +43,7 @@ export type HomeViewSection = {
   }
   ownerType?: OwnerType
   requiresAuthentication: boolean
+  shouldBeDisplayed?: (context: ResolverContext) => boolean
   resolver?: GraphQLFieldResolver<any, ResolverContext>
   type: keyof typeof HomeViewSectionTypeNames
 }
--- a/src/schema/v2/homeView/helpers/__tests__/isSectionDisplayable.test.ts
+++ b/src/schema/v2/homeView/helpers/__tests__/isSectionDisplayable.test.ts
@@ -74,4 +74,51 @@ describe("isSectionDisplayable", () => {
       ).toBe(false)
     })
   })
+
+  describe("with a section's own displayability check", () => {
+    it("returns true if the section does NOT define a displayability check", () => {
+      const section: Partial<HomeViewSection> = {
+        requiresAuthentication: false,
+        shouldBeDisplayed: undefined,
+      }
+      const context: Partial<ResolverContext> = { userID: "42" }
+
+      expect(
+        isSectionDisplayable(
+          section as HomeViewSection,
+          context as ResolverContext
+        )
+      ).toBe(true)
+    })
+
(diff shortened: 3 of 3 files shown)


src/schema/v2/homeView/helpers/isSectionDisplayable.ts
```diff
@@ -27,5 +27,10 @@ export function isSectionDisplayable(
     })
   }
 
+  // section's display pre-check
+  if (typeof section.shouldBeDisplayed === "function") {
+    isDisplayable = isDisplayable && section?.shouldBeDisplayed(context)
+  }
```
anandaroop, line 33: This comes after the other checks (auth, feature flags) because this one is potentially the most expensive, and we'd prefer to avoid computing it unless necessary

src/schema/v2/homeView/helpers/isSectionDisplayable.ts
```diff
@@ -27,5 +27,10 @@ export function isSectionDisplayable(
     })
   }
 
+  // section's display pre-check
+  if (typeof section.shouldBeDisplayed === "function") {
+    isDisplayable = isDisplayable && section?.shouldBeDisplayed(context)
```
dblandin, line 32: **question**: Is the `typeof` check necessary, given that the `shouldBeDisplayed` property is already typed as a function or undefined?

```suggestion
  if (section.shouldBeDisplayed) {
    isDisplayable = isDisplayable && section.shouldBeDisplayed(context)
```

src/schema/v2/homeView/sections/index.ts
```diff
@@ -43,6 +43,7 @@ export type HomeViewSection = {
   }
   ownerType?: OwnerType
   requiresAuthentication: boolean
+  shouldBeDisplayed?: (context: ResolverContext) => boolean
```
dblandin, line 46: Should we anticipate this function, if defined, to sometimes involve some async logic? Maybe we can handle that conditionally if the return value is a `Promise` and allow teams to decide between sync or async conditions?

```suggestion
  shouldBeDisplayed?: (context: ResolverContext) => boolean | Promise<boolean>
```

dblandin (commented):
Nice! Looking good 👍🏼

src/schema/v2/homeView/helpers/isSectionDisplayable.ts
anandaroop: I went back and forth on this but settled on the current version because `if (section.shouldBeDisplayed)` looked to me too much like `shouldBeDisplayed` was  just a simple boolean flag, rather than a function to be optionally executed. Just a minor readability concern for future-us, but I don't feel _too_ strongly about it.

src/schema/v2/homeView/sections/index.ts
anandaroop: Ah, good idea 👍🏽 (though I'll prob merge this and update after KS in the interest of time)

anandaroop: Actually I took a quick stab, but adding async behavior needed some more careful consideration when computing the list of sections from the default zone. Let's def come back to this.

src/schema/v2/homeView/helpers/isSectionDisplayable.ts
```diff
@@ -27,5 +27,10 @@ export function isSectionDisplayable(
     })
   }
 
+  // section's display pre-check
+  if (typeof section.shouldBeDisplayed === "function") {
```
damassi, line 31: `isFunction(section.shouldBeDisplayed)`

src/schema/v2/homeView/helpers/isSectionDisplayable.ts
damassi: might as well use lodash

PR #6055, opened 2024-09-25
chore: drop type/schema enforcement for home view section contextModule

dblandin:
This will allow devs to optionally omit contextModule (as well as other ownerType value current nullable) from the concrete section defintion while development is ongoing.

Sections shipped to production users should have these values specified and that usually requires some alignment with data to determine values and adding those values in our analytics schema package: [artsy/cohesion][].

[artsy/cohesion]: https://github.com/artsy/cohesion

cc/ @joeyAghion


anandaroop (approved):
👍🏽 

Sort of points to the need for a pre-launch checklist for creating new sections. Probably not a bad idea, in any case.

PR #6059, opened 2024-09-26
fix: allow headers dynamically in response to CORS preflight requests

joeyAghion:
This incorporates Metaphysics' old behavior, of reflecting `Access-Control-Request-Headers` dynamically, to the worker's implementation. I referenced the `cors` module's [source](https://github.com/expressjs/cors/blob/791983ebc0407115bc8ae8e64830d440da995938/lib/index.js) and docs to understand the rules and quirks.

### DIFF
--- a/workers/caching/metaphysics-cdn.ts
+++ b/workers/caching/metaphysics-cdn.ts
@@ -2,8 +2,8 @@
 
 export default {
   async fetch(request, env, ctx) {
-    const cacheKeyGeneration = env.CACHE_KEY_GENERATION || "v1"
-    const defaultMaxAge = env.DEFAULT_MAX_AGE || 30
+    const cacheKeyGeneration = env.CACHE_KEY_GENERATION ?? "v1"
+    const defaultMaxAge = env.DEFAULT_MAX_AGE ?? 30
 
     // Returns a SHA-256 digest of provided string.
     async function sha256(message) {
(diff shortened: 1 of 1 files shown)


workers/caching/metaphysics-cdn.ts
```diff
@@ -64,18 +64,22 @@ export default {
...
           headers: {
-            "Access-Control-Allow-Headers":
-              "cache-control,content-type,user-agent,x-access-token,x-original-session-id,x-timezone,x-user-id",
+            "Access-Control-Allow-Headers": allowedHeaders,
             "Access-Control-Allow-Methods": "GET,POST,OPTIONS",
             "Access-Control-Allow-Origin": "*",
             "Access-Control-Max-Age": "600",
-            Vary: "Origin",
+            "Content-Length": "0",
+            Vary: "Origin,Access-Control-Request-Headers",
```
joeyAghion, line 82: Since we're reflecting the requested headers in the response, browsers should only reuse the response for matching values.

workers/caching/metaphysics-cdn.ts
```diff
@@ -64,18 +64,22 @@ export default {
...
           status: 204,
           headers: {
-            "Access-Control-Allow-Headers":
-              "cache-control,content-type,user-agent,x-access-token,x-original-session-id,x-timezone,x-user-id",
+            "Access-Control-Allow-Headers": allowedHeaders,
             "Access-Control-Allow-Methods": "GET,POST,OPTIONS",
             "Access-Control-Allow-Origin": "*",
             "Access-Control-Max-Age": "600",
-            Vary: "Origin",
+            "Content-Length": "0",
```
joeyAghion, line 81: Apparently works around an old Safari bug (that we have not encountered).

PR #6070, opened 2024-09-30
feat: add discover marketing collections to homeview

egdbear:
Discover Marketing Collection Component

This PR introduces the new **Discover Marketing Collection** component to the homeView:

[Ticket (figma url in the ticket)](https://artsyproduct.atlassian.net/browse/DIA-861)

Component Preview

image

Query

Query Screenshot

Response

image

The placement, content collections, component style, and wording are still being refined and may change as we continue to complete the feature.

### DIFF
--- a/src/schema/v2/homeView/HomeViewSection.ts
+++ b/src/schema/v2/homeView/HomeViewSection.ts
@@ -21,6 +21,7 @@ import { InternalIDFields, NodeInterface } from "../object_identification"
 import { SalesConnectionField } from "../sales"
 import { HomeViewComponent } from "./HomeViewComponent"
 import { toGlobalId } from "graphql-relay"
+import { FeaturedLinkConnectionType } from "../FeaturedLink/featuredLink"
 
 // section interface
 
@@ -56,6 +57,8 @@ export const HomeViewSectionTypeNames = {
   HomeViewSectionArtists: "HomeViewSectionArtists",
   HomeViewSectionArtworks: "HomeViewSectionArtworks",
   HomeViewSectionAuctionResults: "HomeViewSectionAuctionResults",
+  HomeViewSectionDiscoverMarketingCollections:
+    "HomeViewSectionDiscoverMarketingCollections",
   HomeViewSectionFairs: "HomeViewSectionFairs",
   HomeViewSectionGalleries: "HomeViewSectionGalleries",
   HomeViewSectionGeneric: "HomeViewSectionGeneric",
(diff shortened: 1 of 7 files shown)


iskounen (dismissed):

src/schema/v2/homeView/HomeViewSection.ts
```diff
@@ -274,6 +277,24 @@ export const HomeViewGalleriesSectionType = new GraphQLObjectType<
...
+export const HomeViewDiscoverMarketingCollectionType = new GraphQLObjectType<
+  any,
+  ResolverContext
+>({
+  name: HomeViewSectionTypeNames.HomeViewSectionDiscoverMarketingCollections,
+  description: "A section containing a list of curated marketing collections",
+  interfaces: [HomeViewGenericSectionInterface, NodeInterface],
+  fields: {
+    ...standardSectionFields,
+    linksConnection: {
```
anandaroop, line 289: **thought:** I would like to encourage us to think in terms of the reusability of generic section types.

Here we have a section whose standard presentation in the UI will be a set of pills/chips. Each chip represents a link to _something_. 

- Does that something need to be declared as a _marketing collection_ here, as opposed to a more generic group of links or featured links?

- Does the section need to be coupled to the _discover_ language, which seems to me an incidental (and possibly changeable) detail of the current UI?

It makes sense to me that the section declaration is over in `sections/DiscoverMarketingCollections.ts` adopts these more specific details, because that _is a specific instance_ of this section type.

But the language here in the schema strikes me as too specific, too coupled to the language of this particular **Discover Something New** section.

Instead I might envision something like this new GraphQL _section type_:

```graphql
type HomeViewSectionFeaturedLinks {
  # etc…
}
```

and this new _section instance_:


```js
export const DiscoverMarketingCollections: HomeViewSection = {
   id: "home-view-section-discover-marketing-collections",

   type: HomeViewSectionTypeNames.HomeViewSectionFeaturedLinks,
  // etc...
}
```

cc @artsy/onyx-devs for your thought as well

anandaroop (commented):
Sorry for the late review, but I'm leaving some more opinionated naming feedback to try and convey the difference I see between section _types_ and section _instances_.

src/schema/v2/homeView/HomeViewSection.ts
anandaroop: (Noting also an alternative approach sketched out in #6087)

PR #6071, opened 2024-09-30
fix: add totalCount to trending artists section

nickskalkin:
Add missing `totalCount` to `home-view-section-trending-artists` section. Because it is missing - this section is not displayed at all (because of this [condition](https://github.com/artsy/eigen/blob/main/src/app/Scenes/HomeView/Sections/HomeViewSectionArtists.tsx#L72) in Eigen).

```graphql
query {
  homeView {
    section(id: "home-view-section-trending-artists") {
      ... on HomeViewSectionArtists {
        artistsConnection(first: 0) {
          totalCount
        }
      }
    }
  }
}
```

```json
{
  "data": {
    "homeView": {
      "section": {
        "artistsConnection": {
          "totalCount": 12
        }
      }
    }
  }
}
```

There are other sections that do not return totalCount. Although they do not result to a component not being displayed - I think it's worth fixing all of them. I will create a ticket.

### DIFF
--- a/src/schema/v2/homeView/sections/TrendingArtists.ts
+++ b/src/schema/v2/homeView/sections/TrendingArtists.ts
@@ -16,6 +16,9 @@ export const TrendingArtists: HomeViewSection = {
 
   resolver: withHomeViewTimeout(async (_parent, args, context, _info) => {
     const artistRecords = await getCuratedArtists(context)
-    return connectionFromArray(artistRecords, args)
+    return {
+      totalCount: artistRecords.length,
+      ...connectionFromArray(artistRecords, args),
+    }
   }),
 }


dariakoko (dismissed):
Thank you!

src/schema/v2/homeView/sections/TrendingArtists.ts
```diff
@@ -16,6 +16,9 @@ export const TrendingArtists: HomeViewSection = {
 
   resolver: withHomeViewTimeout(async (_parent, args, context, _info) => {
     const artistRecords = await getCuratedArtists(context)
-    return connectionFromArray(artistRecords, args)
+    return {
+      totalCount: artistRecords.length,
+      ...connectionFromArray(artistRecords, args),
+    }
```
anandaroop, line 22: 🙏🏽

anandaroop (approved):
Thank you!

> There are other sections that do not return totalCount. Although they do not result to a component not being displayed - I think it's worth fixing all of them. I will create a ticket. 

Definitely worth fixing these consistently in MP. 

**question (non-blocking):** I also wonder if Eigen is handling these early returns consistently — is it always based on `totalCount` attribute or on the presence of an actual non-empty array of data?

nickskalkin:
@anandaroop 

>  I also wonder if Eigen is handling these early returns consistently — is it always based on totalCount attribute or on the presence of an actual non-empty array of data?

I've seen both patterns, which one do you think is preferable?

dariakoko:
@nickskalkin I will jump in with questionsss. 
Maybe you know why we started using the totalCount for these scenarios?

Unless we are querying for the data conditionally, I'm not sure it is beneficial to rely on totalCount when deciding to display of not display the data. In fact, totalCount could be inconsistent with the actual data returned. For example, the backend might return a totalCount greater than zero, but the data array could still be empty due to a processing error, pagination issue, etc. 
Even if we look at the Eigen code with the issue you are fixing, we don't actually need the totalCount at all  - it just adds an extra level of complexity and another vulnerability. Instead, we could rely on [the code on line 76](https://github.com/artsy/eigen/blob/main/src/app/Scenes/HomeView/Sections/HomeViewSectionArtists.tsx#L76)
```
const artists = extractNodes(section.artistsConnection)

if (!artists || artists.length === 0) {
   return null 
}
```
 or extract the length of the data array like this: 
```
if (!artistsConnection?.edges || artistsConnection.edges.length === 0) {
   return null 
}
```

nickskalkin:
yes, agree with you @dariakoko, your approach seems more robust to me. I don't know why we use `totalCount` for such checks 🤷

PR #6073, opened 2024-09-30
chore: Retire custom AlertsEdgeFields

jpotts244:
Retire custom `AlertsEdgeFields` as they are no longer in use

The usage has been retired in Volt and continued to be retired in gravity

Retired gravity code:
- https://github.com/artsy/gravity/pull/18180
- https://github.com/artsy/gravity/pull/18177
- https://github.com/artsy/gravity/pull/18176

### DIFF
--- a/src/schema/v2/Alerts/index.ts
+++ b/src/schema/v2/Alerts/index.ts
@@ -392,25 +392,6 @@ export const AlertType = new GraphQLObjectType<
   },
 })
 
-const AlertsEdgeFields = {
-  counts: {
-    type: new GraphQLObjectType({
-      name: "AlertsCounts",
-      fields: {
-        totalUserSearchCriteriaCount: {
-          type: GraphQLInt,
-          resolve: ({ count_30d }) => count_30d,
-        },
-      },
-    }),
-    resolve: (x) => x,
-  },
-  isRecentlyEnabled: {
-    type: GraphQLBoolean,
-    resolve: ({ count_7d }) => count_7d > 0,
-  },
-}
-
 export const AlertsConnectionSortEnum = new GraphQLEnumType({
   name: "AlertsConnectionSortEnum",
   values: {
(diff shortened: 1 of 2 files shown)


src/schema/v2/Alerts/index.ts
```diff
@@ -437,7 +418,6 @@ export const resolveAlertFromJSON = (alert) => {
 export const AlertsConnectionType = connectionWithCursorInfo({
   name: "Alert",
   nodeType: AlertType,
-  edgeFields: AlertsEdgeFields,
```
jpotts244, line 440: ChatGPT has assured me that the removal of custom edge fields will just cause us to use defaulting graphql edge fields,
The connection type will still have edges, and each edge will have node and cursor, which is usually sufficient for most use cases.


Looking for ways to test this

jonallured (approved):
🙏

src/schema/v2/Alerts/index.ts
damassi: Yah, this is safe as this will just populate the edge area of the connection; nodes should still be returned. Buttt It might be worth running `yarn sync-schema:local` in volt against a running local MP and then `yarn relay` and `yarn type-check` just to see if any of this edge stuff is being used anywhere. Feels like this will break in consumers (unless _that_ code is already retired, in which case I am just rambling!)

jpotts244: Good idea, my fear is that Force seems to be using this AlertsConnection for the `me { alertsConnection....}` BUT does not use these specific fields

damassi: Cool yah, as long as things type-check we should be good 👍

jpotts244: Thanks Chris!

I've run this MP schema change against local force
```
yarn sync-schema-local
yarn relay
yarn type-check
```

All looks good! Will merge after rebasing with latest `main`

PR #6087, opened 2024-10-02
feat: add HomeViewSectionCards section type

dblandin:
https://artsyproduct.atlassian.net/browse/ONYX-1348

## Description

POC for a more generic section type discussed [here in Slack](https://artsy.slack.com/archives/C07ANEV7RNV/p1727872616918969).

This PR introduces two concrete section instances:

- "Explore by Category"
- "Discovery Something New"

### Explore By Category

**Request**

```graphql
{
  homeView {
    section(id: "home-view-section-explore-by-category") {
      __typename
      component {
        title
      }

      ... on HomeViewSectionCards {
        cardsConnection(first: 10) {
          edges {
            node {
              entityID
              entityType
              image {
                cropped(height: 400, width: 340) {
                  src
                }
              }
            }
          }
        }
      }
    }
  }
}
```

**Response**

```json
{
  "data": {
    "homeView": {
      "section": {
        "__typename": "HomeViewSectionCards",
        "component": {
          "title": "Explore by Category"
        },
        "cardsConnection": {
          "edges": [
            {
              "node": {
                "title": "Medium",
                "entityID": "Medium",
                "entityType": "MarketingCollectionCategory",
                "image": {
                  "cropped": {
                    "src": "https://d7hftxdivxxvm.cloudfront.net?height=400&quality=80&resize_to=fill&src=https%3A%2F%2Fplacehold.co%2F340x400%2FEEE%2F31343C%3Ffont%3Dopen-sans%26text%3DMedium&width=340"
                  }
                }
              }
            },
            {
              "node": {
                "title": "Movement",
                "entityID": "Movement",
                "entityType": "MarketingCollectionCategory",
                "image": {
                  "cropped": {
                    "src": "https://d7hftxdivxxvm.cloudfront.net?height=400&quality=80&resize_to=fill&src=https%3A%2F%2Fplacehold.co%2F340x400%2FEEE%2F31343C%3Ffont%3Dopen-sans%26text%3DMovement&width=340"
                  }
                }
              }
            },
            {
              "node": {
                "title": "Color",
                "entityID": "Collect by Color",
                "entityType": "MarketingCollectionCategory",
                "image": {
                  "cropped": {
                    "src": "https://d7hftxdivxxvm.cloudfront.net?height=400&quality=80&resize_to=fill&src=https%3A%2F%2Fplacehold.co%2F340x400%2FEEE%2F31343C%3Ffont%3Dopen-sans%26text%3DColor&width=340"
                  }
                }
              }
            },
            {
              "node": {
                "title": "Size",
                "entityID": "Collect by Size",
                "entityType": "MarketingCollectionCategory",
                "image": {
                  "cropped": {
                    "src": "https://d7hftxdivxxvm.cloudfront.net?height=400&quality=80&resize_to=fill&src=https%3A%2F%2Fplacehold.co%2F340x400%2FEEE%2F31343C%3Ffont%3Dopen-sans%26text%3DSize&width=340"
                  }
                }
              }
            },
            {
              "node": {
                "title": "Price",
                "entityID": "Collect by Price",
                "entityType": "MarketingCollectionCategory",
                "image": {
                  "cropped": {
                    "src": "https://d7hftxdivxxvm.cloudfront.net?height=400&quality=80&resize_to=fill&src=https%3A%2F%2Fplacehold.co%2F340x400%2FEEE%2F31343C%3Ffont%3Dopen-sans%26text%3DPrice&width=340"
                  }
                }
              }
            },
            {
              "node": {
                "title": "Gallery",
                "entityID": "Gallery",
                "entityType": "MarketingCollectionCategory",
                "image": {
                  "cropped": {
                    "src": "https://d7hftxdivxxvm.cloudfront.net?height=400&quality=80&resize_to=fill&src=https%3A%2F%2Fplacehold.co%2F340x400%2FEEE%2F31343C%3Ffont%3Dopen-sans%26text%3DGallery&width=340"
                  }
                }
              }
            }
          ]
        }
      }
    }
  },
  "extensions": {
    "requests": {},
    "requestID": "f7203e10-80eb-11ef-bc28-67b0991b13df"
  }
}
```

### Discover Something New

**Request**

```graphql
{
  homeView {
    section(id: "home-view-section-discover-something-new") {
      __typename
      component {
        title
      }

      ... on HomeViewSectionCards {
        cardsConnection(first: 10) {
          edges {
            node {
              title
              entityID
              entityType
              href
              image {
                cropped(height: 100, width: 100) {
                  src
                }
              }
            }
          }
        }
      }
    }
  }
}
```

**Response**

```json
{
  "data": {
    "homeView": {
      "section": {
        "__typename": "HomeViewSectionCards",
        "component": {
          "title": "Discover Something New"
        },
        "cardsConnection": {
          "edges": [
            {
              "node": {
                "title": "Figuration",
                "entityID": "368f7dac-9c0a-43ac-8f93-663b5a1531bc",
                "entityType": "MarketingCollection",
                "href": "/collection/figurative-art",
                "image": null
              }
            },
            {
              "node": {
                "title": "New from Leading Galleries",
                "entityID": "e79ec3dd-0412-40ae-a3d8-d1dd0ba77ac0",
                "entityType": "MarketingCollection",
                "href": "/collection/new-from-leading-galleries",
                "image": null
              }
            },
            {
              "node": {
                "title": "Painting",
                "entityID": "42c2945a-b0eb-49f3-8e85-29d66739a613",
                "entityType": "MarketingCollection",
                "href": "/collection/painting",
                "image": null
              }
            },
            {
              "node": {
                "title": "Prints",
                "entityID": "bf1c3a08-9dae-4ec4-99d6-71a6b2ba0989",
                "entityType": "MarketingCollection",
                "href": "/collection/prints",
                "image": null
              }
            },
            {
              "node": {
                "title": "Street Art",
                "entityID": "87c8feb4-05a3-4e76-9d5f-aad427536bde",
                "entityType": "MarketingCollection",
                "href": "/collection/street-art",
                "image": {
                  "cropped": {
                    "src": "https://d7hftxdivxxvm.cloudfront.net?height=100&quality=80&resize_to=fill&src=https%3A%2F%2Fd32dm0rphc51dk.cloudfront.net%2FKmJZWb8ZAhKGv3mi7jT95w%2Fnormalized.jpg&width=100"
                  }
                }
              }
            },
            {
              "node": {
                "title": "Art for Small Spaces",
                "entityID": "5dd94e6b-fe26-41de-a139-01ddb916d61e",
                "entityType": "MarketingCollection",
                "href": "/collection/art-for-small-spaces",
                "image": null
              }
            },
            {
              "node": {
                "title": "Cool-Toned Artworks",
                "entityID": "edac5956-a5e8-4eac-af1b-4d97398985db",
                "entityType": "MarketingCollection",
                "href": "/collection/cool-toned-artworks",
                "image": null
              }
            }
          ]
        }
      }
    }
  }
}
```


src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
```diff
@@ -0,0 +1,46 @@
...
+  contextModule: "" as ContextModule, // TODO: fill in with real value
+  type: HomeViewSectionTypeNames.HomeViewSectionCards,
+  component: {
+    title: "Discover Something New",
+  },
+  requiresAuthentication: false,
+  resolver: async (_parent, args, context, _info) => {
+    const { body } = await context.marketingCollectionsLoader({
+      slugs: marketingCollectionSlugs,
+    })
```
dblandin, line 30: **walkthrough**: Since marketing collections are modeled in a first class way in our backend, I experimented with using a data loader to fetch collection details. This should be easily cacheable as the list of slugs is hardcoded above.

src/schema/v2/homeView/sections/ExploreByCategory.ts
```diff
@@ -0,0 +1,63 @@
...
+        return {
+          ...category,
+          entityID: category.id,
+          entityType: "MarketingCollectionCategory",
+          image_url: new URL(
+            `/340x400/EEE/31343C?${new URLSearchParams({
+              font: "open-sans",
+              text: category.title,
+            })}`,
+            "https://placehold.co"
```
dblandin, line 55: **walkthrough**: We'll obviously want to replace this with actual images but found this service that at least lets you embed some reasonable text into the placeholder image.

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
```diff
@@ -0,0 +1,46 @@
...
+]
+
+export const DiscoverSomethingNew: HomeViewSection = {
+  id: "home-view-section-discover-something-new",
+  featureFlag: "diamond_home-view-marketing-collection-categories",
+  contextModule: "" as ContextModule, // TODO: fill in with real value
+  type: HomeViewSectionTypeNames.HomeViewSectionCards,
+  component: {
+    title: "Discover Something New",
+  },
```
anandaroop, line 25: **thought (non-blocking):** At the moment, both of these will get be rendered by some new `HomeViewSectionCards` component over in Eigen. But one of them will want to be rendered by a different/custom component type.

E.g. something along the lines of…

```ts
export const DiscoverSomethingNew: HomeViewSection
  id: "home-view-section-discover-something-new",

  type: HomeViewSectionTypeNames.HomeViewSectionCards,

  component: {  // ← default visual treatment
     title: "Discover Something New", 
   },

  // […]
}
```

and 


```ts
export const ExploreByCategory: HomeViewSection = {
   id: "home-view-section-explore-by-category",

   type: HomeViewSectionTypeNames.HomeViewSectionCards,

   component: {
     title: "Explore by Category",
     type: "HomeViewSectionCardChips"  // ← custom visual treatment
   }

  // […]
}
```

Is that how you're picturing this as well?

anandaroop (commented):
🤩 Yes I really like this direction — identifying the common attributes shared by these navigational components, but allowing for some differing presentational options.

This feels general enough without being too general (and I even noted another place below where we _might_ be able to make use of this abstraction).

Thanks for sketching this out.

src/schema/v2/homeView/sections/ExploreByCategory.ts
```diff
@@ -0,0 +1,63 @@
...
+    title: "Medium",
+  },
+  Movement: {
+    id: "Movement",
+    title: "Movement",
+  },
+  "Collect by Color": {
+    id: "Collect by Color",
+    title: "Color",
+  },
```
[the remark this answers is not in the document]
dblandin: So far I'm thinking that the Eigen component can prefer `href` if it's provided and otherwise fallback to `entityType` + `entityID` values to determine how to navigate internally.

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
dblandin: Yeah, we'll need to pick a default visual treatment and an alternates can be have a defined `component > type` value which can be interpreted by the client/Eigen to trigger that alternate presentation.

src/schema/v2/homeView/HomeViewCard.ts
```diff
@@ -0,0 +1,40 @@
...
+  GraphQLString,
+  GraphQLObjectType,
+  GraphQLFieldConfig,
+  GraphQLNonNull,
+} from "graphql"
+import { ResolverContext } from "types/graphql"
+import { connectionWithCursorInfo } from "../fields/pagination"
+
+export const HomeViewCardType = new GraphQLObjectType<any, ResolverContext>({
+  name: "HomeViewCard",
```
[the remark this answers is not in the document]
dblandin: Yeah, that's a good thought. We could potentially define a singular-version of the section type defined here which exposes a single "card".

But since the viewAll behavior for the "Galleries Near You" section navigates to a simple list of gallery partners, I wonder if instead we could define an alternate styling of the generic Gallery section type. Maybe the default styling of such a section renders the list of galleries in a rail.

```graphql
{
  homeView {
    section(id: "home-view-section-galleries-near-you") {
      __typename // "HomeViewSectionGalleries"
      component {
        title // "Galleries Near You"
        type // "TeaserCard"
        description // "Follow these local galleries for updates on artists you love."
        backgroundImageURL // "https://files.artsy.net/images/galleries_for_you.webp"
        behaviors {
          viewAll {
            buttonText // "Explore"
          }
        }
      }
    }
  }
}
```

_Note that I think the `backgroundImageURL` string attribute above could easily be upgraded to a `backgroundImage` field backed by our more robust Image type._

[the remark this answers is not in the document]
anandaroop: Yeah in that^ case I'd be in favor of reclaiming the `HomeViewSectionGalleries` name for an _actual_ connection of galleries, which is not at all what that section/component does now. 

> Note that I think the backgroundImageURL string attribute above could easily be upgraded to a backgroundImage field backed by our more robust Image type.

Good idea — placeholder ticket: [ONYX-1339](https://artsyproduct.atlassian.net/browse/ONYX-1339)

src/schema/v2/homeView/sections/ExploreByCategory.ts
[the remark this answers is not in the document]
araujobarret: `id` might mislead here, `id` generally represents the resource ID, I'd rather call it something different

src/schema/v2/homeView/sections/ExploreByCategory.ts
```diff
@@ -0,0 +1,63 @@
...
+              font: "open-sans",
+              text: category.title,
+            })}`,
+            "https://placehold.co"
+          ),
+        }
+      }
+    )
+
+    return connectionFromArray(cards, args)
```
araujobarret, line 61: this is not a connection why do we need to fake it?

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
araujobarret: We can define this one as type Chips if possible bc it's the section which should be using it

src/schema/v2/homeView/HomeViewCard.ts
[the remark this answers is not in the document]
egdbear: Agree on this, this could be re-used on Galleries components as singular Card component.

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
```diff
@@ -0,0 +1,46 @@
...
+  requiresAuthentication: false,
+  resolver: async (_parent, args, context, _info) => {
```
egdbear, line 27: **suggestion (blocking)**: We're keeping this data static for now because we're still figuring out the final content and whether the curation team will manage it. I’d recommend keeping the data as is for now and introducing the network call later.

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
```diff
@@ -0,0 +1,46 @@
...
+    const { body } = await context.marketingCollectionsLoader({
+      slugs: marketingCollectionSlugs,
+    })
+
+    const cards = body.map((marketingCollection) => {
+      return {
+        href: `/collection/${marketingCollection.slug}`,
+        entityID: marketingCollection.id,
```
egdbear, line 35: **thought (non-blocking):** Is entityID and entityType the right naming here? Do we need to be more specific? I’m still not quite sure, as this has come up a few times in different discussions.

src/schema/v2/homeView/sections/index.ts
```diff
@@ -56,19 +58,21 @@ const sections: HomeViewSection[] = [
   Auctions,
   CuratorsPicksEmerging,
   DiscoverMarketingCollections,
+  DiscoverSomethingNew,
+  ExploreByCategory,
+  ExploreByMarketingCollectionCategories,
```
egdbear, line 63: **question:** I haven't seen this order of elements in figma, is this in recent designs?

araujobarret:
About the connections, I think we can move on faking the connections for categories but if we add more static types in the future it would be nice to set it up accordingly to be consistent.

So the only issue right now is the `imageURL` is returning the placeholders for the Explore By, testing locally I can't see `image#url` and `image#imageURL` 

image

egdbear (commented):
Thanks for adding this PR! It's always helpful to see the idea in the code rather than just in Slack, as it makes things easier to understand.

Suggestion: If we're starting to introduce more generic types for navigational components, it might be worth reviewing all the components to ensure the generic types can be reused as much as possible.

src/schema/v2/homeView/sections/index.ts
araujobarret: [Explore by](https://www.figma.com/board/kgm28u8rZLh3EfznOuOYSe/Homefeed-Versions?node-id=1368-2111&t=4MLOZEmkTZCkGOfN-4) should be after Hero units and Discover should be after [Recently Viewed](https://www.figma.com/board/kgm28u8rZLh3EfznOuOYSe/Homefeed-Versions?node-id=1368-2085&t=4MLOZEmkTZCkGOfN-4) isn't it?

dblandin:
I'll try to play back and respond to a few items of feedback so far:

**Image GraphQL Type, Gemini on-the-fly resizing, and image data**

I'm hearing some feedback to suggest that the Image support modeled here might be lacking capabilities or somehow broken, but I don't think that's the case.

We have an [Image GraphQL type defined here](https://github.com/artsy/metaphysics/blob/0416dcd2f1a2309bb318f90d546bd13a23f9ebd2/src/schema/v2/image/index.ts#L51) in Metaphysics.

Image data can be as simple as `{ image_url: "https://example.com/image.jpg" }` to support on-the-fly (and cached) resizing and/or cropping via Gemini+Cloudfront.

There are plenty of additional fields on the Image type which support an image processed by Gemini and backed by Gravity, but for simple rendering needs, including resizing, simply using `resized` and/or `cropped` fields should be sufficient for clients.

For resizing/cropping processes, the original height/width of the image is not referenced at all, nor is it required in the source image data.

For example, given the following source image data:

```
{
  image_url: "https://example.com/image.jpg",
  original_width: 720
  original_height: 720
}
```

if I request a resize of 360x 360, I get the following URL from Metaphysics:

```
https://d7hftxdivxxvm.cloudfront.net?height=360&quality=80&resize_to=fit&src=https%3A%2F%2Fexample.com%2Fimage.jpg&width=360
```

Note that the original height/width value of 720 is nowhere in the URL.

Cloudfront returns a cached version if found, otherwise we request the resize from Gemini which downloads the source image and applies the resizing and finally returns the new image.

**GraphQL Connection vs GraphQLList**

When clients benefit from pagination, you can use a GraphQL Connection type to offer cursor-based pagination control to clients when exposing a list of items.

Data exposed via GraphQL Connections can be backed by data loaders and/or static content. Either way of backing connection is valid and can be paginated by clients. Just because a specific GraphQL Connection is backed by an static list of items, doesn't make it fake or somehow _not_ a real connection. The connection is simply the interface clients use to request and paginate data.

For this UI, it seems like we might want to allow clients to limit the number of items requested even if the source list grows in size. If we want to render this list in another context (like a "view all" screen), we may find ourselves wanting to paginate this list of categories. If we promote this concept of marketing collection category to be modeled in a first class way in Gravity, these categories might then be fetched via the API. And a GraphQL connection can easily be modified to be backed by data loaders if that happens. We have much less flexibility with the GraphQLList type.

Section types in the Home View architecture can choose to expose whatever clients need but for groupings of items displayed in UIs that might benefit from pagination, it seems reasonable to default to connection types, even if the source data is hardcoded and static initially.

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
dblandin: For a section type that renders a list of cards, the default component in Eigen to support the type could defer to an `href` attribute to handle navigation on tap. By also threading through the entity ID/type, clients also have the option of adding this metadata to analytic calls. For example, we often fire an event that includes what the intended destination screen is.

For example, we could have an event with the following structure:

```
{
  action: ActionType.tappedCardGroup,
  context_module: "discoverSomethingNew,
  context_screen_owner_type: OwnerType.home,
  destination_screen_owner_type "MarketingCollection",
  destination_screen_owner_id: "d78e9a17-ccf6-4104-b4e9-95c18f6412df", 
  horizontal_slide_position: index,
  module_height: "double",
  type: "thumbnail",
}
```

You can see other examples of analytics payloads [here in Eigen](https://github.com/artsy/eigen/blob/ea625b401a7bd8f00a7466dc7d326da3fa5c7aee/src/app/Scenes/HomeView/useHomeViewTracking.ts).

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
dblandin: Do we have many examples of when the desired display text differs from the text we already have on the backend?

Looking at [the static data](https://github.com/artsy/metaphysics/blob/0416dcd2f1a2309bb318f90d546bd13a23f9ebd2/src/schema/v2/homeView/sections/DiscoverMarketingCollections.ts#L23-L79), it looks like we're often mirroring the same values that already exist and are easily editable on the backend.

src/schema/v2/homeView/HomeViewSection.ts
```diff
@@ -349,12 +351,31 @@ export const HomeViewDiscoverMarketingCollectionType = new GraphQLObjectType<
...
+  any,
+  ResolverContext
+>({
+  name: HomeViewSectionTypeNames.HomeViewSectionCards,
+  description: "A section containing a list of navigation cards",
+  interfaces: [HomeViewGenericSectionInterface, NodeInterface],
+  fields: {
+    ...standardSectionFields,
+    cardsConnection: {
+      type: HomeViewCardConnectionType,
```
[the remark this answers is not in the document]
dblandin: I expanded on the connection choice [here](https://github.com/artsy/metaphysics/pull/6087#issuecomment-2396691636). The section instances proposed in this PR are meant to replace the existing section instances.

src/schema/v2/homeView/sections/ExploreByCategory.ts
dblandin: Not sure what you mean by faking it, but I expanded on the connection choice [here](https://github.com/artsy/metaphysics/pull/6087#issuecomment-2396691636).

src/schema/v2/homeView/HomeViewCard.ts
```diff
@@ -0,0 +1,40 @@
...
+  fields: {
+    href: { type: GraphQLString },
+    entityType: { type: GraphQLString },
+    entityID: { type: GraphQLString },
+    image: {
+      type: Image.type,
```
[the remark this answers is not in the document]
dblandin: We don't need to define or expose original height or width values. We still gain the ability to easily resize and/or crop by simply using this GraphQL type and providing a source image URL. I've expanded on this [here](https://github.com/artsy/metaphysics/pull/6087#issuecomment-2396691636).

src/schema/v2/homeView/sections/ExploreByCategory.ts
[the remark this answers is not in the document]
dblandin: We're still using "Collect by Color" (and others) as an identifier though. Very similar to how we reference [attribution classes](https://github.com/artsy/metaphysics/blob/a9f38eab4651cba47ba7fb5deeacc0ef67ad2fcf/src/lib/attributionClasses.ts#L4) and [artwork medium types](https://github.com/artsy/metaphysics/blob/0416dcd2f1a2309bb318f90d546bd13a23f9ebd2/src/lib/artworkMediums.ts), both of which lack a 1-1 mapping with a backend resource type.

I'm open to suggestions but I think using `id` would be consistent. I consider it broadly fits as an identifier but not specifically an internal database record identifier.

araujobarret:
Thanks for covering the questions and concerns, I thought it was nice to have the `original_width` and `original_height` defined in the image so we could have a reference for how much we can crop, etc.

Given it's a generic element I think it can be a connection for the `Explore By` even for this static result

araujobarret (dismissed):

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
egdbear: My concern is more related to the endpoiont we are using, as we're currently requesting marketing data. The previous thought was focused on `marketing_categories` with potential API changes.

dblandin: This section instance ("Discover Something New") seems to be very MarketingCollection-specific, so it seems okay to use the marketingCollectionsLoader and fetch some of these values from the backend.

For categories, I agree, it's best to stick with static data as we don't yet have first class modeling or endpoints/loaders to use.

Does match your understanding as well? Or is there another reason to use the MarketingCollection loader here specifically?

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
dblandin: So far we've avoided using ` top-level `analytics` field in favor of contextModule and ownerType fields. These `entity` fields to me represent a way of exposing what type of entity/item/thing the cards are representing. That _could_ be used for internal navigation, as I think will be the case for the "Explore By" section instance where we aren't providing `href` but it could also be used for internal client implementation details (like analytics).

But I'd rather not enforce this at the schema level especially for a generic "card" section type like this.

dblandin: Otherwise, it seems tricky to remain generic _and_ give Eigen the context/data it needs in order to navigate to the right internal Marketing Collection **category** screen.

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
egdbear: That makes sense!

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
egdbear: Thanks for the explanation - entityID and entityType seemed a bit misleading to me at first, but this clarification definitely helps

src/schema/v2/homeView/sections/DiscoverSomethingNew.ts
anandaroop: Yeah I see that two of the display titles here differ from the collections’ current titles:

- **Figurative Art** instead of **Figuration**
- **Black-and-White Artworks** instead of **Black-and-White** 

We could re-title those (though there may be consistency considerations, with respect to how other similar collections are named — I think Jenna/Addy would be best to confirm that)

We could also add a `display_title` field to the `MarketingCollection` model if this seems like a common enough need. (`Gene` models have something similar)

anandaroop (approved):
Thanks for elaborating @dblandin — that all sounds good to me.

- **Re: images.** I agree that the true MP image type gives us the flexibility we need on the frontend for sizing the images per the design. 
  - One note, though: I had assumed that the `resized` and `cropped` fields would give us the final dimensions in the `width` and `height` subfields, but that only appears to be the case for `cropped`.

- **Re: connections:** Agree that connections are the most flexible and future-proof. We can still set an arbitrarily high `first` arg over in Eigen to ensure we get all the cards/chips we need in this first iteration.

anandaroop:
As discussed [here](https://artsy.slack.com/archives/C05EQL4R5N0/p1728316744508639), I'm going to push this PR forward. 

I've re-worked the PR to have more granular commits, updated tests, and a couple of touch-ups re: [images](https://github.com/artsy/metaphysics/pull/6087/commits/5ddf0429b98e0072ecb7447aa9dbb77b03e3ba2f) and [slugs](https://github.com/artsy/metaphysics/pull/6087/commits/9d443922197f3e35274751996d3cd90609953cf4).

I'm not removing any of the newly added sections/types because I don't want to break any Eigen WIP. But we should migrate soon to the new card/chip components in Eigen so that we can continue with the removals here.

PR #6092, opened 2024-10-03
fix: allow inquiry questions on inquireable artworks

starsirius:
https://artsyproduct.atlassian.net/browse/PBRW-72

There seems to be a regression after https://github.com/artsy/eigen/pull/10833 that causes the send inquiry button disabled on Eigen when there is no inquiry questions selected. The inquiry artworks associated with the Kamala sale ([example](https://www.artsy.net/artwork/hugh-hayden-potus)) do not return inquiry questions because of [existing condition check](https://github.com/artsy/metaphysics/blob/82f73dd0fcfdd033783cfeeebaadcbab27609c62/src/schema/v2/artwork/index.ts#L735-L736). I propose we decouple inquireability from auctions here, and rely on the main `artwork.inquireable` to determine the inquireability. I think after this change, app users should be able to see the questions and at least able to send an inquiry from the app.

We'll need to revisit the app issue separately.

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -728,12 +728,11 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         description:
           "Structured questions a collector can inquire on about this work",
         resolve: (
-          { sale_ids, id },
+          { id, inquireable },
           _params,
           { inquiryRequestQuestionsLoader }
         ) => {
-          // Sale artworks are not inquirable
-          if (!sale_ids.length) {
+          if (inquireable) {
             return inquiryRequestQuestionsLoader({
               inquireable_id: id,
               inquireable_type: "Artwork",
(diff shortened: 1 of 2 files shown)


damassi (approved):

src/schema/v2/artwork/index.ts
```diff
@@ -728,12 +728,11 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
           "Structured questions a collector can inquire on about this work",
         resolve: (
-          { sale_ids, id },
+          { id, inquireable },
           _params,
           { inquiryRequestQuestionsLoader }
         ) => {
-          // Sale artworks are not inquirable
-          if (!sale_ids.length) {
+          if (inquireable) {
```
starsirius, line 735: I think it makes sense to decouple inquireability from auctions.

starsirius: Yeah it seems correct to me, so will need more investigation.

PR #6102, opened 2024-10-08
refactor: re-organize section types

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-1349

This further re-organizes the home view directory in order to:
-  break up the `HomeViewSection.ts` file, which had become very unwieldy as more section types were added, into a new `sectionTypes/` directory

- further reinforce the first-class nature of thinking in terms _section instances and section types_:

- potentially provide a better file layout for adding tests as well
  - (right now everything is jammed into monolithic spec files inside `homeView/__tests__`, but it might make more sense to break those out into separate, more manageable spec files)

<table>
<tr >
 <th> section instances </th>
 <th> section types </th>
</tr>
<tr valign="top">
 <td> </td>
 <td> </td>
</tr>
</table>


src/schema/v2/homeView/sectionTypes/Activity.ts
```diff
@@ -0,0 +1,29 @@
...
+    notificationsConnection: {
+      type: NotificationsConnection.type,
+
+      args: pageable({}),
+      resolve: (parent, ...rest) => {
+        return parent.resolver ? parent.resolver(parent, ...rest) : []
+      },
+    },
+  },
+})
```
damassi, line 29: (Know this is still a draft so please ignore if this comment is premature)

It might be worthwhile to roll all of this up into a simple factory function. Aside from `fields` it seems as though each file below shares 90% of the same information. So something like 

```tsx
createHomeViewSection({
  name: ...
  description: ... 
  fields: {
    ...
  }
})
```
would be mostly sufficient, and we wouldn't need all of the duplicated file spread. 

And then a simple object could define things and roll it all up nicely, to be spread into the main type:

```tsx
const SECTIONS = [
  {
    name: ...
    description: ... 
    fields: ... 
  }
]

const SectionsTypes = SECTIONS.map(createHomeViewSection) 
``` 

This would probably be more maintainable later on through time, since the abundance of files creates the appearance of more underlying complexity than there really is.

anandaroop: Thanks for the 👀. For now, I think I'll keep it as is. One thing I've noticed as devs start to work in the schema is the [distinction between section _instances_ and section _types_](https://docs.google.com/presentation/d/1j9yFZmkdelJ6y3YQiuJ-KaUTv0Pc9-ZwyTdrR5HrU9Q/edit#slide=id.g3024881a6d2_1_126) needed to be clearer. 

That's _one_ of the motivations for the current refactor: being able to see in a clear way, reified in the filesystem, that those are the most important first-class concepts when working with the new schema. Hoping to help people fall into the pit of success, and I think that added explicitness outweighs the benefit of any extra DRY-ness that factory would provide.

But not opposed to this idea if our thinking changes 👍🏽

anandaroop: I think our disagreement falls along well-worn lines 😄  

On this question I'm [Team Sandi](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction) and not [the opposition](https://www.codewithjason.com/duplication-cheaper-wrong-abstraction/)

Also curious to hear what others think!

damassi: Yah, i'm not a fan of abstraction generally, but this is a _great_ candidate for it.

dblandin: I'm interested in how an approach to DRY this up would materialize but I'd rather not introduce that abstraction in this PR. This organizational change is a lot easier to grok, especially for new contributors, the current single file and I'd rather lean on the clear explicitness than potentially introduce a difficult abstraction to work with.

damassi: Aight, thanks all. Sounds good to me 🦅

dblandin (dismissed):
Nice! I think this will make it a lot easier for contributors to differentiate between shard section types and section instances.

Looking at our current tests, I feel like they might be a better fit within a `/homeView/sections/__tests__/` folder and specifically exercise section _instances_ rather than types.

anandaroop:
> Looking at our current tests, I feel like they might be a better fit within a /homeView/sections/__tests__/ folder and specifically exercise section instances rather than types.

Totally agree, I think that testing the types _through_ their instances makes sense 👍🏽

PR #6104, opened 2024-10-08
feat: add weaviate integration

egdbear:
This PR adds Weaviate integration to Metaphysics.  
It sets up unauthenticated loaders, currently accessible only through the staging VPN. I will follow up on resolver types once [this PR](https://github.com/artsy/quantum/pull/36/files) is merged.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -84,6 +84,7 @@ const {
   VORTEX_API_BASE,
   VORTEX_APP_ID,
   VORTEX_TOKEN,
+  WEAVIATE_API_BASE,
 } = process.env
 
 const mustHave = {
@@ -219,4 +220,5 @@ export default {
   VORTEX_API_BASE,
   VORTEX_APP_ID,
   VORTEX_TOKEN,
+  WEAVIATE_API_BASE,
 }
--- a/src/lib/loaders/loaders_without_authentication/weaviate.ts
+++ b/src/lib/loaders/loaders_without_authentication/weaviate.ts
@@ -1,0 +1,24 @@
+import factories from "../api"
+
+interface GraphQLArgs {
+  query: string
+  variables?: any
+}
+
+export default (opts) => {
+  const {
+    weaviateLoaderWithoutAuthenticationFactory: weaviateLoader,
+  } = factories(opts)
+
+  return {
+    weaviateGraphqlLoader: ({ query, variables }: GraphQLArgs) => {
+      return weaviateLoader(
+        "/graphql",
+        { query, variables: JSON.stringify(variables) },
+        {
+          method: "POST",
+        }
+      )
+    },
+  }
+}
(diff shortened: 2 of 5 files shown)


src/lib/loaders/loaders_without_authentication/weaviate.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export default (opts) => {
+  const {
+    weaviateLoaderWithoutAuthenticationFactory: weaviateLoader,
+  } = factories(opts)
+
+  return {
+    weaviateGraphqlLoader: ({ query, variables }: GraphQLArgs) => {
+      return weaviateLoader(
+        "/graphql",
```
mc-jones, line 16: How do we feel about using the [RESTful api](https://weaviate.io/developers/weaviate/api/rest)? I believe if we use their `graphql` api we would need to use stitching, which is in the process of being removed from MP.

egdbear: I don't think we need stitching since this is essentially a REST-like call but using a query format, right? We can simply use the query like this, for example:

```
const DiscoveryArtworks: GraphQLFieldConfig<void, ResolverContext> = {
  args: pageable({
    userId: { type: GraphQLString },
  }),
  resolve: async (_root, args, { weaviateGraphqlLoader, artworksLoader }) => {
    const { userId } = args

    const query = gql`
      query discoveryArtworksQuery {
        {
          Get {
            DiscoveryArtworks(where: { userId: "${userId}" }) {
              internalID
            }
          }
        }
      }
    `

    const body = await weaviateGraphqlLoader({
      query,
    })()

    if (!body.data.Get.DiscoveryArtworks) return connectionFromArray([], args)

    const artworkIds = body.data.Get.DiscoveryArtworks.map(
      (node) => node.internalID
    )

    const artworks =
      artworkIds?.length > 0 ? await artworksLoader({ ids: artworkIds }) : []

    return connectionFromArray(artworks, args)
  },
}
```

If this is not the preferred way we can def move to REST endpoints.

src/config.ts
```diff
@@ -219,4 +220,5 @@ export default {
   VORTEX_API_BASE,
   VORTEX_APP_ID,
   VORTEX_TOKEN,
+  WEAVIATE_API_BASE,
```
mc-jones, line 223: quick note on the URL. It may expect the internal address: `http://weaviate-internal.default.svc.cluster.local:8080`

src/lib/loaders/loaders_without_authentication/weaviate.ts
mc-jones: After chatting, we will stick with `graphql`. Turns out the RESTful api is primarily built for DB management, and [not querying](https://weaviate.io/developers/weaviate/concepts/interface#restful-api-and-graphql-api).

mc-jones (approved):

PR #6106, opened 2024-10-08
feat: add infinite discovery type and mutation

egdbear:
This PR introduces the `discoverArtworks` query type and the `createDiscoveryArtworkReference` mutation used for Infinite Discovery:

```graphql
discoverArtworks(
  after: String
  before: String
  first: Int
  last: Int
  userId: String
): ArtworkConnection
```

```graphql
createDiscoveryArtworkReference(
  input: CreateDiscoveryArtworkReferenceMutationInput!
): CreateDiscoveryArtworkReferenceMutationPayload
```

### DIFF
--- a/src/schema/v2/createDiscoveryArtworkReferenceMutation.ts
+++ b/src/schema/v2/createDiscoveryArtworkReferenceMutation.ts
@@ -4,6 +4,7 @@ import {
   GraphQLObjectType,
   GraphQLString,
   GraphQLUnionType,
+  GraphQLEnumType,
 } from "graphql"
 import { ResolverContext } from "types/graphql"
 import {
@@ -13,8 +14,20 @@ import {
 import { mutationWithClientMutationId } from "graphql-relay"
 import { generateUuid, generateBeacon } from "./discoverArtworks"
 
+export const ReferenceTypesEnum = new GraphQLEnumType({
+  name: "ReferenceTypes",
+  values: {
+    LIKED_ARTWORKS: {
+      value: "likedArtworks",
+    },
+    DISLIKED_ARTWORKS: {
+      value: "dislikedArtworks",
+    },
+  },
+})
+
 const SuccessType = new GraphQLObjectType<any, ResolverContext>({
-  name: "DiscoveryLikedArtworkMutationSuccess",
+  name: "DiscoveryArtworkReferenceMutationSuccess",
   isTypeOf: (data) => data.success,
   fields: () => ({
     success: {
@@ -25,7 +38,7 @@ const SuccessType = new GraphQLObjectType<any, ResolverContext>({
 })
 
 const FailureType = new GraphQLObjectType<any, ResolverContext>({
-  name: "DiscoveryLikedArtworkMutationFailure",
+  name: "DiscoveryArtworkReferenceMutationFailure",
   isTypeOf: (data) => {
     return data._type === "GravityMutationError"
   },
@@ -38,18 +51,17 @@ const FailureType = new GraphQLObjectType<any, ResolverContext>({
 })
 
 const ResponseOrErrorType = new GraphQLUnionType({
-  name: "CreateDiscoveryLikedArtworkResponseOrError",
+  name: "CreateDiscoveryArtworkReferenceResponseOrError",
   types: [SuccessType, FailureType],
 })
 
-export const createDiscoveryLikedArtworkMutation = mutationWithClientMutationId<
-  { userId: string; artworkId: string },
+export const CreateDiscoveryLikedArtworkMutation = mutationWithClientMutationId<
+  { userId: string; artworkId: string; reference: string },
   any,
   ResolverContext
 >({
-  name: "CreateDiscoveryLikedArtworkMutation",
-  description:
-    "Create a liked artwork cross reference for the user in weaviate",
+  name: "CreateDiscoveryArtworkReferenceMutation",
+  description: "Creates a cross reference for artwork and user in weaviate",
   inputFields: {
     userId: {
       description: "The user's ID",
@@ -59,16 +71,20 @@ export const createDiscoveryLikedArtworkMutation = mutationWithClientMutationId<
       description: "The artwork's ID",
       type: new GraphQLNonNull(GraphQLString),
     },
+    reference: {
+      description: "The reference type",
+      type: new GraphQLNonNull(ReferenceTypesEnum),
+    },
   },
   outputFields: {
-    createDiscoveryLikedArtworkResponseOrError: {
+    createDiscoveryArtworkReferenceResponseOrError: {
       type: ResponseOrErrorType,
       description: "On success: return boolean. On failure: MutationError.",
       resolve: (result) => result,
     },
   },
   mutateAndGetPayload: async (
-    { userId, artworkId },
+    { userId, artworkId, reference },
     { weaviateCreateCrossReferenceLoader }
   ) => {
     if (!weaviateCreateCrossReferenceLoader) {
@@ -82,9 +98,10 @@ export const createDiscoveryLikedArtworkMutation = mutationWithClientMutationId<
     )
 
     const userUUID = generateUuid(userId)
+    const referenceURL = `${userUUID}/references/${reference}`
 
     try {
-      await weaviateCreateCrossReferenceLoader(userUUID, {
+      await weaviateCreateCrossReferenceLoader(referenceURL, {
         beacon: artworkBeacon,
       })
       return { success: true }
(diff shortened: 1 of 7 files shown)


src/schema/v2/discoverArtworks.ts
```diff
@@ -0,0 +1,44 @@
...
+  args: pageable({
+    userId: { type: GraphQLString },
+  }),
+  resolve: async (_root, args, { weaviateGraphqlLoader, artworksLoader }) => {
+    if (!weaviateGraphqlLoader) return
+    const { userId } = args
+
+    const query = gql`
+      query discoveryArtworksQuery {
+        {
```
mc-jones, line 19: Think we need to remove one set of outer brackets. Also, just curious if there is a reason for using a named query in this instance?

src/schema/v2/discoverArtworks.ts
```diff
@@ -0,0 +1,44 @@
...
+          Get {
+            DiscoveryArtworks(where: { userId: "${userId}" }) {
```
mc-jones, line 21: Ok, so this is somewhat annoying, but the new class name is [`InfiniteDiscoveryArtworks`](https://github.com/artsy/quantum/blob/main/src/17-infinite-discovery/02-artworks.ts#L9). In our test weaviate instance, there is already a `DiscoveryArtworks`. 

Maybe this is fine, temporarily, but we'll need to make a decision around this. LMK what your thoughts are.

src/schema/v2/discoverArtworks.ts
mc-jones: This is also the query which will need to utilize the `nearObject` [search](https://weaviate.io/developers/weaviate/api/graphql/search-operators#nearobject), using a [`beacon`](https://weaviate.io/developers/weaviate/more-resources/glossary) as the identifying argument for the user (using just the id doesn't work across collections/classes in our experience). So the query will look something like this:

```
{
    Get {
         DiscoveryArtworks(
	    limit: 5
	    nearObject: {
	    beacon: "weaviate://weaviate.stg.artsy.systems/InfiniteDiscoveryUsers/${userID}"
	    }
	    ) {
	        title
	    }
	}
}
```

mc-jones:
So in working through these comments, I'm coming to realize that we may have to use both `graphql` and `REST` within our loaders OR commit to using their [typescript library](https://weaviate.io/developers/weaviate/client-libraries/typescript). It doesn't look like their `graphql` api supports mutations, which is where the `REST` would need to be used. Possible I'm missing something, but they even mention their client libraries utilize both `graphql` and `REST`. 

The client library seems simplest, but may be a new pattern in MP. Not sure if there we have an existing preference already? This seems to be their [recommendation](https://weaviate.io/developers/weaviate/concepts/interface#restful-api-and-graphql-api).

src/schema/v2/discoverArtworks.ts
egdbear: Yeah, that makes sense, do we need even the beacon?

egdbear:
> So in working through these comments, I'm coming to realize that we may have to use both `graphql` and `REST` within our loaders OR commit to using their [typescript library](https://weaviate.io/developers/weaviate/client-libraries/typescript). It doesn't look like their `graphql` api supports mutations, which is where the `REST` would need to be used. Possible I'm missing something, but they even mention their client libraries utilize both `graphql` and `REST`.
> 
> The client library seems simplest, but may be a new pattern in MP. Not sure if there we have an existing preference already? This seems to be their [recommendation](https://weaviate.io/developers/weaviate/concepts/interface#restful-api-and-graphql-api).

I think we can use REST for the mutation, as it will be similar to the API. I believe this is the same approach the [client library uses](https://github.com/weaviate/typescript-client/blob/e272b37eb515c2a7b29ab7046b6452761914dee5/src/collections/data/index.ts#L241).

src/schema/v2/discoverArtworks.ts
egdbear: No strong opinion on this, since it would involve offboarding the existing one and renaming the new one, right?

src/schema/v2/discoverArtworks.ts
egdbear: I think we can use just an UUID

image

mc-jones: When I last tested this, `UUID` only worked when a `nearObject` search used a search argument within the same class/collection as the class/collection being searched. So if I was using a `InfiniteDiscoveryArtwork` object as the search argument, this would work. But because we are using a object from a different class/collection (`InfiniteDiscoveryUsers`) as the search argument to search `InfiniteDiscoveryArtworks` a beacon is needed. I never found good documentation on this, but it was the only way I was previously able to get it to work.

src/schema/v2/discoverArtworks.ts
```diff
@@ -0,0 +1,59 @@
...
+  }),
+  resolve: async (_root, args, { weaviateGraphqlLoader, artworksLoader }) => {
+    if (!weaviateGraphqlLoader) return
+
+    const { userId, limit = 5 } = args
+
+    const userUUID = generateUuid(userId)
+
+    // TODO: Understand why localhost works here and weaviate://weaviate.stg.artsy.net doesn't
+    const beacon = `weaviate://localhost/InfiniteDiscoveryUsers/${userUUID}`
```
mc-jones, line 28: Need to look into this more...

src/schema/v2/createDiscoveryLikedArtworkMutation.ts
```diff
@@ -0,0 +1,102 @@
...
+  return uuid(userId, uuid.DNS).toString()
+}
+
+const SuccessType = new GraphQLObjectType<any, ResolverContext>({
+  name: "DiscoveryLikedArtworkMutationSuccess",
+  isTypeOf: (data) => data.success,
+  fields: () => ({
+    success: {
+      type: GraphQLBoolean,
+      resolve: (result) => result.success,
```
mc-jones, line 27: Keeping it super simple to start and just returning `true`.

mc-jones:
@egdbear - Working on the mutation, discovered duplicate cross references are accepted ([weaviate discussion](https://forum.weaviate.io/t/how-to-encode-frequencies-of-cross-references/1675/2) here that seems to confirm this), so we  would need to handle this.

Screenshot 2024-10-09 at 9 04 10 AM

egdbear:
@mc-jones maybe we can re-use one mutation with different reference as input type for liked and disliked artworks?

src/schema/v2/createDiscoveryArtworkReferenceMutation.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export const CreateDiscoveryLikedArtworkMutation = mutationWithClientMutationId<
+  { userId: string; artworkId: string; reference: string },
+  any,
+  ResolverContext
+>({
+  name: "CreateDiscoveryArtworkReferenceMutation",
+  description: "Creates a cross reference for artwork and user in weaviate",
+  inputFields: {
+    userId: {
```
araujobarret, line 66: don't think we need this here, it will increase some verbosity in the clients where they will have to get the userId before querying/mutating. We can just user the id coming from the header.

araujobarret (dismissed):
Looks good 💪 
my comment I don't consider it a blocker but will be better for the clients to not care about the userId params

src/schema/v2/createDiscoveryArtworkReferenceMutation.ts
egdbear: thanks for the early feedback and good point! We can use as non-required field while we are still in research phase.

mc-jones (approved):
Wooo. Think this is good starting point and enough to try and connect a test fronted (e.g. observable).

mc-jones:
> @mc-jones maybe we can re-use one mutation with different reference as input type for liked and disliked artworks?
> 
> Added [here](https://github.com/artsy/metaphysics/pull/6106/commits/1adac4d189f637468c6cb197d646272dc8642811)

Yup! Good call!

src/schema/v2/discoverArtworks.ts
joeyAghion: Curious about this. I've seen internal hostname references (like `weaviate-internal`) but not routing _by protocol_. How does this work?

egdbear: I'm not sure I have much more to add, aside from this resource: [Weaviate GraphQL Search Operators](https://weaviate.io/developers/weaviate/api/graphql/search-operators#variables-1).

image
From my understanding, if we have a multi-tenant strategy, it could be useful as an identifier. But I could be completely wrong about that.

PR #6107, opened 2024-10-08
feat(home view): add a showsConnection to HomeViewSectionShows

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-1340

`HomeViewSectionShows` is one of the three section types that wasn't yet using new-style data resolution. 

This PR addresses that.

First we modify the `HomeViewSectionShows` _section **type**_ by :

- giving it a `showsConnection`
- allowing that `showsConnection` to take a `near` argument

Then we modify the `ShowsForYou` _section **instance**_ by:

- wiring up the existing `meShowsLoader` (via `ShowsConnection.resolve`) which can make use of the `near` argument passed in from the client

### DIFF
--- a/src/schema/v2/homeView/sections/ShowsForYou.ts
+++ b/src/schema/v2/homeView/sections/ShowsForYou.ts
@@ -1,6 +1,13 @@
 import { ContextModule } from "@artsy/cohesion"
 import { HomeViewSection } from "."
 import { HomeViewSectionTypeNames } from "../sectionTypes/names"
+import { withHomeViewTimeout } from "../helpers/withHomeViewTimeout"
+import { ResolverContext } from "types/graphql"
+import { ShowsConnection } from "schema/v2/me/showsConnection"
+import { emptyConnection } from "schema/v2/fields/pagination"
+import { EventStatusEnums } from "schema/v2/input_fields/event_status"
+
+const SHOWS_TIMEOUT_MS = 6000
 
 export const ShowsForYou: HomeViewSection = {
   id: "home-view-section-shows-for-you",
@@ -10,4 +17,24 @@ export const ShowsForYou: HomeViewSection = {
     title: "Shows for You",
   },
   requiresAuthentication: true,
+
+  shouldBeDisplayed: (_context: ResolverContext) => true,
+
+  resolver: withHomeViewTimeout(async (parent, args, context, info) => {
+    if (!ShowsConnection.resolve) return emptyConnection
+
+    const finalArgs = {
+      ...args,
+      status: EventStatusEnums.getValue("RUNNING_AND_UPCOMING")?.value,
+    }
+
+    const result = await ShowsConnection.resolve(
+      parent,
+      finalArgs,
+      context,
+      info
+    )
+
+    return result
+  }, SHOWS_TIMEOUT_MS),
 }
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11414,6 +11414,12 @@ type HomeViewSectionShows implements HomeViewSectionGeneric & Node {
 
   # [Analytics] `owner type` analytics value for this scetion when displayed in a standalone UI, as defined in our schema (artsy/cohesion)
   ownerType: String
+  showsConnection(
+    after: String
+    before: String
+    first: Int
+    last: Int
+  ): ShowConnection
 }
 
 # A viewing rooms section in the home view
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11419,6 +11419,9 @@ type HomeViewSectionShows implements HomeViewSectionGeneric & Node {
     before: String
     first: Int
     last: Int
+
+    # Include shows within a radius of the provided location
+    near: Near
   ): ShowConnection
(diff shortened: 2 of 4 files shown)


src/schema/v2/homeView/sections/ShowsForYou.ts
```diff
@@ -1,6 +1,13 @@
 import { ContextModule } from "@artsy/cohesion"
 import { HomeViewSection } from "."
 import { HomeViewSectionTypeNames } from "../sectionTypes/names"
+import { withHomeViewTimeout } from "../helpers/withHomeViewTimeout"
+import { ResolverContext } from "types/graphql"
+import { ShowsConnection } from "schema/v2/me/showsConnection"
+import { emptyConnection } from "schema/v2/fields/pagination"
+import { EventStatusEnums } from "schema/v2/input_fields/event_status"
+
+const SHOWS_TIMEOUT_MS = 6000
```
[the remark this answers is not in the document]
olerichter00: Could we have different timeouts for production and staging? This way, we could optimize the timeouts for the production environment and, at the same time, avoid timeouts in staging that are coming from the environment being 🐌.

src/schema/v2/homeView/sections/ShowsForYou.ts
```diff
@@ -10,4 +17,24 @@ export const ShowsForYou: HomeViewSection = {
     title: "Shows for You",
   },
   requiresAuthentication: true,
+
+  shouldBeDisplayed: (_context: ResolverContext) => true,
+
+  resolver: withHomeViewTimeout(async (parent, args, context, info) => {
```
olerichter00, line 23: Is defining the resolver for a section's (main) connection in the parent (the section type) the preferred pattern?

olerichter00 (commented):
🌟 🌟 🌟

src/schema/v2/homeView/sections/ShowsForYou.ts
dblandin: The _section type_ (HomeViewSectionShows) defines the interface and the _section instance_ (ShowsForYou) defines the behavior/resolution.

This is the general pattern that allows different instances of the HomeViewSectionShows section type to specify different loading behavior but in the end still produce a ShowsConnection to clients.

src/schema/v2/homeView/sections/ShowsForYou.ts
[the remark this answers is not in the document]
dblandin: That's a good point. We have an env/config value that can be set differently between staging and production environments ([HOME_VIEW_RESOLVER_TIMEOUT_MS](https://github.com/artsy/metaphysics/blob/fc6a57aa2d25cb157b68e5620665714a80324b18/src/config.ts#L54)).

Looks like we could check the `PRODUCTION_ENV` var to provide different override values here.

```typescript
const SHOWS_TIMEOUT_MS = config.PRODUCTION_ENV ? 6000 : 10000

// To not specify a custom override in staging and defer to the default.
// I think this might need a slight change to `withHomeViewTimeout` though.
const SHOWS_TIMEOUT_MS = config.PRODUCTION_ENV ? 6000 : undefined
```

src/schema/v2/homeView/sections/ShowsForYou.ts
```diff
@@ -10,4 +17,24 @@ export const ShowsForYou: HomeViewSection = {
...
+      ...args,
+      status: EventStatusEnums.getValue("RUNNING_AND_UPCOMING")?.value,
+    }
+
+    const result = await ShowsConnection.resolve(
+      parent,
+      finalArgs,
+      context,
+      info
+    )
```
dblandin, line 36: **suggestion / non-blocking**: I know we use this approach elsewhere, mainly to avoid duplicating the resolver logic, but I wonder how much benefit we're really getting from reusing another field's resolver.

In this specific case the [me/showsConnection resolve function](https://github.com/artsy/metaphysics/blob/fc6a57aa2d25cb157b68e5620665714a80324b18/src/schema/v2/me/showsConnection.ts#L41-L94):
- Exits early if user is logged out
- Translates input args to meShowsLoader data loader args
- Injects an IP-backed `near` fallback
- Invokes the data loader
- Returns a connection

This seems pretty standard and reasonable for this section to handle itself and we wouldn't be coupling the implementation here with how shows are currently resolved under me/showsConnection.

For more complicated resolver logic, we've had success in the past extracting functions that can be used across different field resolvers.

dblandin (approved):
Looks good! 💯 

I left one non-blocking comment regarding shared resolve functions.

src/schema/v2/homeView/sections/ShowsForYou.ts
olerichter00: That makes sense! Thanks for the explanation 🙏

src/schema/v2/homeView/sections/ShowsForYou.ts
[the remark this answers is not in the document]
anandaroop: Addressed here: https://github.com/artsy/metaphysics/pull/6123

PR #6117, opened 2024-10-10
fix(home view): shape of empty connections

anandaroop:
As [pointed out a while back](https://artsy.slack.com/archives/C05EQL4R5N0/p1724099057516639?thread_ts=1724081856.316439&cid=C05EQL4R5N0) by @mzikherman, it's technically incorrect to return `[]` for an empty connection.

Instead we have an `emptyConnection` helper that returns the correct shape — 

https://github.com/artsy/metaphysics/blob/0f2e865fbe86b705cace2df46018daf1c7dedaec/src/schema/v2/fields/pagination.ts#L235-L238

We applied this a little bit after that comment, but not everywhere.

This PR fixes that. Confirmed that all sections still look good in simulator 👍🏽


mzikherman (approved):

dblandin (approved):
Awesome 👍🏼

PR #6126, opened 2024-10-11
feat: add Infinite Discovery object mutation

egdbear:
This PR adds the mutation for creating objects in Weaviate and organizes the infiniteDiscovery folder, structuring the mutations and types within it.


src/schema/v2/infiniteDiscovery/deleteDiscoveryArtworkReferencesMutation.ts
```diff
@@ -0,0 +1,164 @@
...
+          _type: "GravityMutationError",
+        }
+      } else {
+        throw new Error(error)
+      }
+    }
+  },
+})
+
+const getArtworkReferences = ({ data, reference, uuid }) =>
```
mc-jones, line 151: Not totally clear why we're doing this on the delete? Wouldn't we want to guard against double entries at the `create` stage of the cross reference?

Also, if we do need to do this here, I wonder if we can't simplify this method a bit? The nested iterators makes it a bit tough to read.

mc-jones (dismissed):

src/schema/v2/infiniteDiscovery/deleteDiscoveryArtworkReferencesMutation.ts
egdbear: > Not totally clear why we're doing this on the delete? Wouldn't we want to guard against double entries at the create stage of the cross-reference?

In this case, the goal during deletion is to ensure that all associated references (liked and disliked) are removed properly. The function:

1. Finds unique entries.
2. Prevents duplicate calls (in case duplicates were added via Weaviate GraphQL queries for example).

Added here to remind us that Weaviate isn't handling duplicate prevention automatically, so we need to manage it either here (I agree, it's not ideal) or at the create stage, or create a another way like a proxy API

PR #6130, opened 2024-10-11
chore: Add partner search criteria id to hits edge AMBER-1090

jonallured:
Building on top of https://github.com/artsy/gravity/pull/18243, this PR updates the edge of the hits connection to include this `PartnerSearchCriteria` id so that we can properly track from the "Matched Alerts" tab in CMS.

https://artsyproduct.atlassian.net/browse/AMBER-1090

/cc @artsy/amber-devs


jpotts244 (approved):
A+

PR #6078, opened 2024-10-01
feat(EMI-2088): dismiss me tasks

leamotta:
https://artsyproduct.atlassian.net/browse/EMI-2088

Opening https://github.com/artsy/gravity/pull/18194

Screenshot 2024-10-01 at 13 39 37

PR #6145, opened 2024-10-21
feat(EMI-2099): task acknowledgment

leamotta:
https://artsyproduct.atlassian.net/browse/EMI-2099

Opening https://github.com/artsy/gravity/pull/18248
Similar to #6078


src/schema/v2/schema.ts
```diff
@@ -238,6 +238,7 @@ import { DiscoverArtworks } from "./discoverArtworks"
 import { CreateDiscoveryLikedArtworkMutation } from "./createDiscoveryArtworkReferenceMutation"
 import { Invoice } from "./Invoice/invoice"
 import { createInvoicePaymentMutation } from "./Invoice/createInvoicePaymentMutation"
+import { ackTaskMutation } from "./me/ack_task_mutation"
```
MrSltun, line 241: Nitpick: can we change the name of the file to follow a Camel Case convention? just to be similar to the other files 👀? 
```suggestion
import { ackTaskMutation } from "./me/ackTaskMutation"
```

MrSltun (approved):
Looks good! I just left a small comment on the file name 👀

src/schema/v2/schema.ts
leamotta: there's plenty of others that use this convention, including the dismiss mutation. Is the recommended format CamelCase? Should I change the dismiss mutation as well?

MrSltun: Maybe that would be great to follow one convention 😅 but it's not part of this PR's scope, so feel free to ignore the other ones

PR #6149, opened 2024-10-22
feat: update query args for homeview > latest articles

MounirDhahri:
This PR fixes an issue with the new home view latest articles rail. We were missing the default sorting props and that resulted in returning different articles in the rail compared to the screen.

**Screenshot after changes**

Group 26


src/schema/v2/homeView/sections/LatestArticles.ts
```diff
@@ -19,5 +20,22 @@ export const LatestArticles: HomeViewSection = {
   },
   requiresAuthentication: false,
 
-  resolver: withHomeViewTimeout(ArticlesConnection.resolve!),
+  resolver: withHomeViewTimeout(async (parent, args, context, info) => {
```
MounirDhahri, line 23: Using logic similar to https://github.com/artsy/metaphysics/blob/35189608da893002ec7605c9edfea60e44988d6d/src/schema/v2/homeView/sections/News.ts#L26-L43

dblandin (approved):
Awesome. Thank you 🙏🏼

PR #6152, opened 2024-10-22
feat: support certainty config for discoverArtworks

mc-jones:
This PR does two things:

1. Adds the ability to pass in a `certainty` argument that gets passed along to `weaviate`. In effect, `certainty` is a[ similarity rating between 0 and 1](https://weaviate.io/developers/weaviate/api/graphql/search-operators#variables-1), where 1 is almost identical and 0 is an opposite. `certainty` and `distance` get at the same thing, so I am not including `distance`, for now, despite it being the newer approach recommended by `weaviate`.
2. Adds the ability to `sort` by `certainty` level. Defaults to returning more similar artworks first (i.e. artworks with higher `certainty` scores).

Together this gives us greater ability to configure the results coming back from `weaviate`. I'm particularly interested if we get better recommendation by moving slightly away from the most similar results. This could be part of a strategy to address the potential issue of `weaviate` recommending artworks that are too similar to each other.

### DIFF
--- a/src/schema/v2/infiniteDiscovery/discoverArtworks.ts
+++ b/src/schema/v2/infiniteDiscovery/discoverArtworks.ts
@@ -1,4 +1,9 @@
-import { GraphQLString, GraphQLFieldConfig, GraphQLInt } from "graphql"
+import {
+  GraphQLString,
+  GraphQLFieldConfig,
+  GraphQLInt,
+  GraphQLEnumType,
+} from "graphql"
 import { ResolverContext } from "types/graphql"
 import { artworkConnection } from "../artwork"
 import { connectionFromArray } from "graphql-relay"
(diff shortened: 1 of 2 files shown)


src/schema/v2/infiniteDiscovery/discoverArtworks.ts
```diff
@@ -21,19 +26,43 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
...
+          },
+        },
+      }),
+    },
   }),
   resolve: async (_root, args, { weaviateGraphqlLoader, artworksLoader }) => {
     if (!weaviateGraphqlLoader) return
 
-    const { userId, limit = 5 } = args
+    const { userId, limit = 5, certainty = 0.5, sort } = args
```
mc-jones, line 51: A somewhat arbitrary default but should be fine for testing. What this means is `weaviate` will only return `InfiniteDiscoveryArtworks` that have a `certainty` score greater than `0.5`.

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
```diff
@@ -54,6 +83,10 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
       (node) => node.internalID
     )
 
+    if (sort == "certainty") {
+      artworkIds.reverse()
+    }
+
```
egdbear, line 89: **question (non-blocking):** What is the reasoning behind reversing the artwork IDs? Shouldn't the vector certainty (distance) already resolve the ordering?

egdbear (approved):

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
mc-jones: Good question! Weaviate returns items in descending order based of `certainty` score (i.e. artworks that are the most similar) will be returned first. But I want to experiment with returning artworks that are a not the most similar. So, for example, artworks that have a certainty score of `0.8` instead of `.98`. With this, I can set a certainty score of `0.8` and sort in ascending order, which will give me the artworks first.

The theory here is this might help with the issue of weaviate not providing enough variation in its recommendations.

PR #6154, opened 2024-10-23
feat(ONYX-1294): add priceListed field to PartnerOfferToCollectorType

dariakoko:
Add priceListed field to PartnerOfferToCollectorType

According to [ONYX-1294](https://artsyproduct.atlassian.net/browse/ONYX-1294) we need to display priceListed on the partner offer sale details
Instead of getting this information from the artwork, let's expose another field in PartnerOfferToCollectorType in order to stay consistent


olerichter00 (commented):
I wonder why we don't use the list price from the artwork 🤔 But you're right, it makes sense to add it to `PartnerOfferToCollector` as well for consistency.

Could you add some tests for the new field [here](https://github.com/artsy/metaphysics/blob/49ef1b4b4ca32a3184feb1961d3622ed86c941bc/src/schema/v2/me/__tests__/index.test.ts#L708) please (this seems to be the only place that tests `PartnerOfferToCollector` offers?

olerichter00 (approved):

MrSltun (approved):

dariakoko:
@olerichter00 I had a closer look and I think using the list price from the artwork would actually be better - this ways we will have one source of truth for the artwork's original price

PR #6155, opened 2024-10-23
chore: remove action message

leamotta:
Follow up for https://github.com/artsy/gravity/pull/18261
We can remove instead of deprecating as no one is using it yet

### DIFF
--- a/src/schema/v2/me/task.ts
+++ b/src/schema/v2/me/task.ts
@@ -8,7 +8,6 @@ export interface Task {
   title: string
   created_at: string
   message: string
-  action_message: string
   action_link: string
   resolved_at: string
   dismissed_at: string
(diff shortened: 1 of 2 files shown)


src/schema/v2/me/task.ts
```diff
@@ -29,10 +28,6 @@ export const TaskType = new GraphQLObjectType<any, ResolverContext>({
     },
     title: { type: new GraphQLNonNull(GraphQLString) },
     message: { type: new GraphQLNonNull(GraphQLString) },
-    actionMessage: {
-      type: new GraphQLNonNull(GraphQLString),
-      resolve: ({ action_message }) => action_message,
-    },
```
MrSltun, line 35: I don't think we can remove a field once it has been released 😅 since it will break the schema, we can only mark it as deprecated

leamotta: We had a similar discussion here https://github.com/artsy/metaphysics/pull/5690
It is my understanding that if no clients are using it then we can remove, is that not right?

MrSltun: good point! and I think it's app only, so it makes sense to remove it since it hasn't been released yet 👍

MrSltun (approved):
🧹🧹

PR #6160, opened 2024-10-23
feat(art advisor): expose experiments in homeView response

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-1361

Co-authored-by: @dblandin 
Co-authored-by: @olerichter00 

Adds the ability to expose currently running experiments and variants in the homeView response.

### Request 

```graphql
{
  homeView {
    experiments {
      name
      variant
    }
    sectionsConnection(first: 3) {
      edges {
        node {
          internalID
        }
      }
    }
  }
}
```

### Response 

```json
{
  "data": {
    "homeView": {
      "experiments": [
        {
          "name": "onyx_experiment_home_view_test",
          "variant": "control"
        }
      ],
      "sectionsConnection": {
        "edges": [
          {
            "node": {
              "internalID": "home-view-section-tasks"
            }
          },
          {
            "node": {
              "internalID": "home-view-section-latest-activity"
            }
          },
          {
            "node": {
              "internalID": "home-view-section-new-works-for-you"
            }
          }
        ]
      }
    }
  }
}
```


src/schema/v2/homeView/experiments/HomeViewExperiments.ts
```diff
@@ -0,0 +1,80 @@
...
+import {
+  FeatureFlag,
+  getExperimentVariant,
+  getFeatureFlag,
+} from "lib/featureFlags"
+import { ResolverContext } from "types/graphql"
+import { CURRENTLY_RUNNING_EXPERIMENTS } from "./experiments"
+
+const HomeViewExperiment = new GraphQLObjectType<any, ResolverContext>({
+  name: "HomeViewExperiment",
```
mzikherman, line 18: This seems like a generic-y schema to get currently running experiments _in general_. Should this be moved out of `homeView` and made a more generic type? (and then can have a specific homeview field/resolver).

anandaroop: Yep, same thought here! 

I hadn't done it yet because I was staring side-eyed at this similar type… 

https://github.com/artsy/metaphysics/blob/c675d49b4c8ebafa2d2784b0db286cc32df4bbee/src/schema/v2/admin/featureFlags.ts#L17-L18

…which has a substantially different shape, I think because it's based on a different Unleash endpoint / has a different purpose. So it may make sense to keep the new type added in this PR, but move it to a common location, just as you are suggesting.

src/schema/v2/homeView/experiments/HomeViewExperiments.ts
```diff
@@ -0,0 +1,80 @@
...
+    const experiments = CURRENTLY_RUNNING_EXPERIMENTS.map(
+      (name: FeatureFlag) => {
+        const flag = getFeatureFlag(name)
+        const variant = getExperimentVariant(name, { userId: context.userID })
+
+        return {
+          ...flag,
+          service: "unleash",
+          variant: variant ? variant.name : null,
+        }
```
anandaroop, line 75: Wondering if we should maybe filter out the flags where `enabled: false` when assembling this list.

src/schema/v2/homeView/experiments/HomeViewExperiments.ts
```diff
@@ -0,0 +1,80 @@
...
+    description: {
+      type: GraphQLString,
+      description: "A description of the experiment",
+    },
+    variant: {
+      type: GraphQLString,
+      description: "The name of the experimental variant being shown currently",
+    },
+    variants: {
+      description: "The variants available within this experiment",
```
anandaroop, line 41: This type adds a few more fields (such as `variants` here) than we had [sketched out](https://www.notion.so/artsy/Ad-Hoc-Meeting-128cab0764a0801085c6f4cc4ae6fe1c). 

I found it helpful to have access to these extra fields as I was setting this up…

Screenshot 2024-10-23 at 11 30 15 PM


…but could be persuaded to slim this down. 

Then again, if we _do_ keep this as a more generic Unleash type as suggested above, it could be nice to retain these.

src/schema/v2/homeView/experiments/HomeViewExperiments.ts
olerichter00: Could we make this type independent from the home view and more generic? I guess, potentially it could be used in other places as well (`HomeViewExperiment` -> `Experiment`, or `userExperiment`, or something similar).

src/schema/v2/homeView/experiments/HomeViewExperiments.ts
dblandin: Yeah, I had a similar thought. We should filter to only enabled flags when assembling the _active_ list for the current user.

src/schema/v2/homeView/experiments/HomeViewExperiments.ts
dblandin: I could see a more general version of this client-facing `experiments` query field living outside of the homeView schema. Maybe this placement offers a pre-filtered selection of experiments to those only with `scope: homeview` or something like that.

That might also be useful eventually if we wanted to support dev menu awareness of all active experiments and offering override behavior, based on known variants.

dblandin: That said, maybe we could define a more generic client-facing "Experiment" type but keep the more focused placement within `homeView` to see how it feels in that context. We can always choose to expose a more generic placement later while keeping the types consistent.

The current experiment type / modeling I think is more geared towards dev/admin purposes and only used in Forque.

src/schema/v2/homeView/experiments/HomeViewExperiments.ts
anandaroop: I moved `HomeViewExperiment` to the more general `UnleashFeatureFlag` (named as such because it applies to `release` flags as well as `experiment` flags) in 3995f2df761c3e7d24d5d9b9b48d8fc6a56e28f0

src/schema/v2/homeView/experiments/HomeViewExperiments.ts
anandaroop: Added in 0f37be5454d3e6ba8209bebce6f75e51bb1fa908

src/schema/v2/homeView/experiments/HomeViewExperiments.ts
anandaroop: - Moved `HomeViewExperiment` to the more generic `UnleashFeatureFlag` type
- Kept the `HomeViewExperiments` _field_ which continues to serve up a list of those, but only if they are `enabled: true`.

I looked into using a `scope:home-view` to magically fetch only the relevant flags. This _ought_ to be supported by Unleash's [tags](https://docs.getunleash.io/reference/tags) mechanism, but in practice I found this did not work [as advertised](https://github.com/Unleash/unleash-client-node?tab=readme-ov-file#advanced-usage)  🤷🏽, so for now we'll maintain the explicit list of experiments in `homeView/experiments.ts`

src/schema/v2/unleashFeatureFlag.ts
```diff
@@ -0,0 +1,56 @@
...
+  GraphQLString,
+  GraphQLList,
+  GraphQLInt,
+} from "graphql"
+import { ResolverContext } from "types/graphql"
+
+/**
+ * A client-facing representation of an Unleash feature flag.
+ */
+export const UnleashFeatureFlag = new GraphQLObjectType<any, ResolverContext>({
```
damassi, line 13: Ah @anandaroop - we already have all of this stuff wired up for tools.artsy.net: 

https://github.com/artsy/metaphysics/blob/main/src/schema/v2/admin/featureFlags.ts

Can we update things there?

damassi: Also tho I don't see us abandoning feature flags just yet (though I have been exploring posthog) ideally we don't encode service names in variables and keep things generic.

damassi: Also, wanted to say, if `admin` isn't the right parent folder, feel free to move it out to just like `featureFlags`! But `admin` does feel kinda right for this kind of stuff (or something similar)

damassi: Also, wanted to say, if `admin` isn't the right parent folder, feel free to move it out to just like `featureFlags`! But `admin` does feel kinda right for this kind of stuff (or something similar)

anandaroop: Not actually the same type, see [above](https://github.com/artsy/metaphysics/pull/6160#discussion_r1814183706)!

- MP's Unleash-related admin fields use data loaders to hit the _admin_ REST API directly
  - e.g. `https://unleash.artsy.net/api/admin/projects/default/features` 
  - that^ returns data for administering flags

- MP's `lib/featureFlags` uses the Unleash API client to hit the _client_ REST API
  - e.g. `https://unleash.artsy.net/api/client/features/`
  - that^ returns data for clients which need only to evaluate flags

So: different endpoints with different [access](https://github.com/artsy/metaphysics/blob/4c8a2e1c556eeb251bbc575c65c55535c9392ecf/src/config.ts#L79) [credentials](https://github.com/artsy/metaphysics/blob/4c8a2e1c556eeb251bbc575c65c55535c9392ecf/src/config.ts#L82) and different response shapes, see e.g. [here](https://docs.getunleash.io/reference/api/unleash/get-features) vs [here](https://docs.getunleash.io/reference/api/unleash/get-all-client-features).

It _is_ confusing that these appear so similar, but they represent different views into Unleash.

damassi: So the question is: can this be consolidated? Whats the difference?

anandaroop: Here's what I'm going to do:

- consolidate the admin- and client-facing representations of the Unleash flags into a shared location:
  - `schema/v2/featureFlags/admin` 
  - `schema/v2/featureFlags/client`

- rename the new client facing type from `UnleashFeatureFlag` to `ClientFeatureFlag`

I didn't see specs for the admin stuff, but I manually verified via local Forque.

I still see these as separate types since they map to separate API resources, and I don't see much benefit of trying to merge them. (But I don't object if someone wants to try and merge these. That should be a little easier since everything is co-located now).

Gonna roll with this for now so we can move on to the Eigen side of this.

PR #6165, opened 2024-10-24
feat: updates the final list of the discover new collections

egdbear:
This PR updates the final list for the 'Discover something new' component.

Screenshot 2024-10-24 at 1 23 12 PM


araujobarret (approved):
💟

PR #6167, opened 2024-10-24
feat: add fallback for discovery users without liked artworks

mc-jones:
This PR adds a fallback for infinite discovery users who don't have any existing "likes", so that we can still produce some artworks for them to react to and start getting personalized recommendations. The fallback query, just gets the first `x` amount (set by `limit`) of artworks from `weaviate`. Additionally, I've added an `offset` argument so that we can get variation in our starting point.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -16852,6 +16852,7 @@ type Query {
     first: Int
     last: Int
     limit: Int
+    offset: Int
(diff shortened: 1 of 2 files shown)


mc-jones:
@egdbear - I'll likely self merge this one to get in staging and test more thoroughly before tomorrow.

PR #6168, opened 2024-10-24
ci(jest): Add CI parallelism

araujobarret:
Updates our jest CI config to run tests in parallel, and bumps Jest to 29.


package.json
```diff
@@ -141,7 +141,7 @@
     "eslint-plugin-promise": "4.0.1",
     "expect.js": "0.3.1",
     "husky": "3.1.0",
-    "jest": "24.9.0",
+    "jest": "^29.7.0",
```
anandaroop, line 144: **suggestion:** If we're bumping this, can we also bump `@types/jest` on L124? (Seems to have fallen quite behind 🥲)

src/schema/v2/__tests__/city.test.ts
```diff
@@ -481,12 +481,10 @@ describe("City", () => {
     it("includes shows and stub shows from a hard-coded list", async () => {
       const mockShows = [{ id: "sponsored-show" }]
 
-      const mockShowsLoader = jest.fn(() =>
-        Promise.resolve({
-          headers: { "x-total-count": "1" },
-          body: mockShows,
-        })
-      )
+      const mockShowsLoader = jest.fn().mockResolvedValue({
```
damassi, line 484: Jest 29 types have gotten way better, so these things started to fail

PR #6185, opened 2024-10-29
feat: use href for deep linking to explore by component

egdbear:
This PR adds `href` deep link to to the `ExploreByCategories` section of the home view. The goal is to enable curated sorting, which currently requires hardcoding the list here and resolving it in Eigen (TBD PR from @araujobarret)

Open to any suggestions! 😊

[Slack discussion](https://artsy.slack.com/archives/C05EEBNEF71/p1730199412491929)

PR #6186, opened 2024-10-29
feat: Support EXPLORE_BY sort for marketing collections

dblandin:
POC PR to demonstrate an alternate approach to https://github.com/artsy/metaphysics/pull/6185, to implement custom ordering of marketing collections filtered by a known category.

---

To support the new Explore By feature in the app, this allows the client to request an `EXPLORE_BY` sort on Marketing Collections, when also specifying a category, which switches the query to use a hard-coded and ordered list of Marketing Collection slugs.

cc/ @egdbear @araujobarret @anandaroop

PR #6187, opened 2024-10-29
feat: add custom sorting to marketing collections

egdbear:
This PR provides an alternative to [PR #6185](https://github.com/artsy/metaphysics/pull/6185). It introduces a static file, [marketingCollectionCategories](https://github.com/artsy/metaphysics/blob/9377d81a72851ac5cb086ee7a7e069486b39fc0c/src/lib/marketingCollectionCategories.ts), which serves as a single source of truth for the Explore by HomeView section and the marketing collection curated sorting.


_schemaV2.graphql
```diff
@@ -12667,6 +12667,7 @@ type MarketingCollectionQuery {
 enum MarketingCollectionsSorts {
   CREATED_AT_ASC
   CREATED_AT_DESC
+  CUSTOM
```
anandaroop, line 12670: **suggestion:** I think we'd want a more descriptive name for this. The other PR had `EXPLORE_BY`. I could also see other options (such as `CURATED`) which feel maybe less tied to the current UI but still express the fact that this is Artsy's manual ordering.

src/lib/marketing_collections_categories.ts
anandaroop, line 1: **suggestion:** We should make this filename camel case, that's a strong convention in MP.

src/schema/v2/__tests__/marketingCollections.test.ts
```diff
@@ -172,4 +172,48 @@ describe("MarketingCollections", () => {
...
+        }
+      }
+    `
+
+    const payload = [
+      {
+        slug: "percys-z-collection-1",
+      },
+      {
+        slug: "fiby-z-collection-2",
```
anandaroop, line 190: **question:** Who is Fiby??

src/schema/v2/__tests__/marketingCollections.test.ts
```diff
@@ -172,4 +172,48 @@ describe("MarketingCollections", () => {
...
+    const context = {
+      authenticatedLoaders: {},
+      marketingCollectionsLoader: () => Promise.resolve({ body: payload }),
+    } as any
+
+    const data = await runQuery(query, context)
+
+    expect(data).toEqual({
+      marketingCollections: payload,
+    })
```
anandaroop, line 203: **suggestion:** This test isn't really exercising anything other than that the mock data is being passed through. 

I think it would be better to test the actual side effects of passing the custom sort, e.g. assert that the final args to Gravity include a non-empty list of `slugs`, and do not include `category` or `sort`.

anandaroop (changes requested):
Wondering if there was a reason to depart from [the other branch](https://github.com/artsy/metaphysics/pull/6186)? Seems like we could have added more commits there.

Perhaps we didn't want to extract _all_ the marketing collection category metadata, but only the sorting info? I feel like it does make sense for all of that marketing collection category metadata to live together.

I leave it up to you, but in any case my feedback is below…

_schemaV2.graphql
egdbear: If the `CUSTOM` is not that descriptive let's go with CURATED then.

src/schema/v2/__tests__/marketingCollections.test.ts
egdbear: Sounds good!

egdbear:
> Wondering if there was a reason to depart from [the other branch](https://github.com/artsy/metaphysics/pull/6186)? Seems like we could have added more commits there.
> 
> Perhaps we didn't want to extract _all_ the marketing collection category metadata, but only the sorting info? I feel like it does make sense for all of that marketing collection category metadata to live together.
> 
> I leave it up to you, but in any case my feedback is below…

I wasn’t sure if the other branch (https://github.com/artsy/metaphysics/pull/6186) was ready or if it made sense to push the commits there yet. Since it was still in draft status, I decided to try out the code which was already on my local branch.

src/schema/v2/__tests__/marketingCollections.test.ts
egdbear: telegram-cloud-photo-size-4-5841689928617019050-y

src/lib/marketingCollectionsCategories.ts
```diff
@@ -0,0 +1,70 @@
+export const marketingCollectionsByCategory = {
```
dblandin, line 1: **suggestion**: This file will likely only export a single data structure. Thoughts on renamed this const to match the filename and exporting a default?

For the filename, why the double plural? "Collection categories" reads a bit better to me than "collections categories". Similar to "artwork medium" or "attribution classes".

```ts
const marketingCollectionCategories = {}

export default marketingCollectionCategories
```

dblandin: This diff introduces a second hardcoded data structures for the same concept. Can we migrate the previously defined structure in the [Explore By section instance definition](https://github.com/artsy/metaphysics/blob/3b7d326deb1b11130b5a849c5722dd3a6ab8dc9b/src/schema/v2/homeView/sections/ExploreByCategory.ts#L6-L40) and consolidate here?

src/lib/marketingCollectionsCategories.ts
```diff
@@ -0,0 +1,70 @@
...
+  Medium: {
+    slugs: [
```
dblandin, line 3: **suggestion**: It would be nice if this key name reflected that this is a selective, curated, and ordered list of marketing collections, identified by slug and not an exhaustive list of all marketing collections matching this category.

That underlying data and associations are owned by Gravity and could change. Maybe something like `curatedCollectionSlugs`? I think the key name should include `Collection` somewhere as these aren't slugs for the category, but for associated marketing collections.

src/schema/v2/__tests__/marketingCollections.test.ts
```diff
@@ -172,4 +172,105 @@ describe("MarketingCollections", () => {
...
+    `
+
+    const payload = [
+      {
+        slug: "percys-z-collection-1",
+      },
+      {
+        slug: "fiby-z-collection-2",
+      },
+    ]
```
dblandin, line 192: **suggestion**: These payloads don't need to be redefined. You can reference the `marketingCollectionsData` fixture defined above, which should be in scope here.

src/schema/v2/__tests__/marketingCollections.test.ts
```diff
@@ -172,4 +172,105 @@ describe("MarketingCollections", () => {
...
+
+    const context = {
+      authenticatedLoaders: {},
+    } as any
+
+    context.marketingCollectionsLoader = jest
+      .fn()
+      .mockReturnValueOnce(Promise.resolve({ body: payload }))
```
dblandin, line 200: **suggestion**: Instead of wrapping in a promise, you can use mockResolvedValue:

```ts
.mockResolvedValue({ body: marketingCollectionsData })
```

src/schema/v2/__tests__/marketingCollections.test.ts
```diff
@@ -172,4 +172,105 @@ describe("MarketingCollections", () => {
...
+    expect(context.marketingCollectionsLoader).toHaveBeenCalledWith({
+      size: 2,
+      slugs: [
+        "new-from-tastemaking-galleries",
+        "new-from-nonprofits-acaf27cc-2d39-4ed3-93dd-d7099e183691",
+        "new-from-small-galleries",
+        "new-from-leading-galleries",
+        "new-to-artsy",
+      ],
+    })
```
dblandin, line 213: **suggestion**: It reads like this test is focused on the switch from category+sort to slugs. I think that would be better accomplished with something like the following:

```ts
 expect(marketingCollectionsLoaderMock.mock.calls[0][0]).not.toContainKeys([
    "category",
    "sort",
  ])
  expect(marketingCollectionsLoaderMock).toHaveBeenCalledWith(
    expect.objectContaining({
      slugs: [
        "art-under-500-dollars",
        "art-under-1000-dollars",
        "art-under-2500-dollars",
        "art-under-5000-dollars",
        "art-under-10000-dollars",
        "art-under-25000-dollars",
        "art-under-50000-dollars",
      ],
    })
  )
})
```

src/schema/v2/__tests__/marketingCollections.test.ts
```diff
@@ -172,4 +172,105 @@ describe("MarketingCollections", () => {
...
+      category: "trending",
+      slugs: ["percys-z-collection-1", "fiby-z-collection-2"],
+    })
+
+    expect(data).toEqual({
+      marketingCollections: payload,
+    })
+  })
+
+  it("throws an error when used with a non-existent category", async () => {
```
dblandin, line 263: **suggestion**: Might also be worth adding a test exercising when `sort: CURATED` but category is not provided. I imagine that might exercise the same category missing handling, but we should confirm:

```
No curated sort available for category: undefined
```

src/schema/v2/marketingCollections.ts
```diff
@@ -329,10 +350,12 @@ export const MarketingCollections: GraphQLFieldConfig<void, ResolverContext> = {
...
     } = {
-      size,
-      artist_id: args.artistId,
-      is_featured_artist_content: args.isFeaturedArtistContent,
       ...args,
+      size,
+      ...(args.artistID && { artist_id: args.artistID }),
+      ...(args.isFeaturedArtistContent && {
+        is_featured_artist_content: args.is_featured_artist_content,
+      }),
```
dblandin, line 358: **question**: Was this additional null checking required for the feature or the get tests passing?

src/schema/v2/sorts/marketingCollectionsSort.ts
```diff
@@ -13,6 +13,9 @@ const MARKETING_COLLECTIONS_SORTS = {
   UPDATED_AT_DESC: {
     value: "-updated_at",
   },
+  CURATED: {
+    value: "curated",
```
dblandin, line 17: Nice 👍🏼

dblandin (changes requested):
Looking good!

_schemaV2.graphql
dblandin: Yeah, I agree that `CUSTOM` is a bit too vague but I like `CURATED` 👍🏼

src/lib/marketingCollectionsCategories.ts
dblandin: We have a single concept "Collection Categories" and we're creating a new file here to define those categories (matching those used in Gravity).

We already have some of their attributes defined in [ExploreByCategory.ts](https://github.com/artsy/metaphysics/blob/3b7d326deb1b11130b5a849c5722dd3a6ab8dc9b/src/schema/v2/homeView/sections/ExploreByCategory.ts#L6-L40), like id, title, imageUrl. Here's we're specifying another attribute: an ordered/curated set of collection slugs.

I'd recommend that we combine all of this metadata into this new file. If we need to still specify a separate ordering for the ExploreByCategory section instance, we can do that by referencing category definition in the file.

For example:

```ts
import marketingColletionCategories from "lib/marketingCollectionCategories"

const categoryKeys = [
  "Medium",
  "Movement",
  "Collect by Size",
  "Collect by Color",
  "Collect by Price",
  "Gallery"
]

export const ExploreByCategory: HomeViewSection = {
  id: "home-view-section-explore-by-category",
  featureFlag: "diamond_home-view-marketing-collection-categories",
  contextModule: ContextModule.exploreBy,
  type: HomeViewSectionTypeNames.HomeViewSectionCards,
  component: {
    title: "Explore by Category",
  },
  resolver: (_parent, args, _context, _info) => {
    const cards = categoryKeys.map((key) => {
      const category = marketingColletionCategories[key]

      return {
        entityID: category.id,
        entityType: OwnerType.collectionsCategory,
        image_url: category.imageUrl,
        title: category.title,
      }
    })

    return connectionFromArray(cards, args)
  },
}
```

src/lib/marketingCollectionsCategories.ts
egdbear: I'm a bit confused by this naming suggestion. The current variable name, `marketingCollectionsByCategory`, already reflects the idea of organized marketing collections by category. IMHO renaming this key might make it a bit more confusing and less clear about the what that key actual represents. Does it make sense?

src/schema/v2/__tests__/marketingCollections.test.ts
egdbear: They are within scope, but I find it helpful to ensure that the defined payload matches the expected response. This helps avoid any manipulation of the original `marketingCollectionsData` fixture. Does that make sense?

src/schema/v2/marketingCollections.ts
egdbear: It is to get test passing

src/lib/marketingCollectionsCategories.ts
dblandin: What is the goal of this file?

If it's only to represent the curated/sorted list of marketing collections by category, then we could change it's name and maybe it's structure to something that is focused on that purpose.

Like `orderedMarketingCollectionsByCategory.ts`:

```ts
export default {
  "Collect by Price": [
    "art-under-500-dollars",
    "art-under-1000-dollars",
    "art-under-2500-dollars",
  ]
}
```

If it's purpose is to more generally define Marketing Collection Categories, then we can keep a filename to reflect that and store associated metadata in that structure.

Like `marketingCollectionCategories.ts`:

```ts
export default {
  "Collect by Price": {
    id: "Collect by Price",
    title: "Price",
    orderedCollectionSlugs: [
      "art-under-500-dollars",
      "art-under-1000-dollars",
      "art-under-2500-dollars",
    ]
  }
}
```

src/schema/v2/__tests__/marketingCollections.test.ts
dblandin: I might agree if the fixture data were changing somehow but we're duplicating the same values and even the same count here. Without a need to specify different data, redefining it here just clutters the test and makes it harder to read.

I don't see any risk here in anything manipulating the original fixture. It's defined at it's current scope _specifically so_ it can be reused in nested spec definitions.

src/schema/v2/__tests__/marketingCollections.test.ts
egdbear: Okay, sounds good, adding the test case for arguments.

src/schema/v2/marketingCollections.ts
dblandin: I'd recommend making the test more selective [as suggested here](https://github.com/artsy/metaphysics/pull/6187#discussion_r1822195452) rather adding these additional checks and object spreads.

src/schema/v2/__tests__/marketingCollections.test.ts
```diff
@@ -172,4 +172,133 @@ describe("MarketingCollections", () => {
...
+
+    context.marketingCollectionsLoader = jest
+      .fn()
+      .mockResolvedValue({ body: marketingCollectionsData })
+
+    await runQuery(query, context)
+
+    expect(
+      context.marketingCollectionsLoader.mock.calls[0][0]
+    ).not.toContainKeys(["sort", "category"])
```
dblandin, line 302: **suggestion**: I would expect this response to fail with an error but the test dosstring suggests that it would pass.

Generally, I feel like there could be three tests introduced
  - it("returns an ordered list of collection by category when CURATED sort is requested")
    - happy path
  - it("returns an error when CURATED sort is requested and category is missing")
  - it("returns an error when CURATED sort is requested and category is unknown")

src/schema/v2/__tests__/marketingCollections.test.ts
```diff
@@ -172,4 +172,133 @@ describe("MarketingCollections", () => {
...
+      }
+    `
+
+    const context = {
+      authenticatedLoaders: {},
+    } as any
+
+    context.marketingCollectionsLoader = jest
+      .fn()
+      .mockResolvedValue({ body: marketingCollectionsData })
```
dblandin, line 296: You can easily collapse these section in a single const.

```suggestion
    const context = {
      marketingCollectionsLoader: jest.fn()
        .mockResolvedValue({ body: marketingCollectionsData }),
    } as any
```

dblandin: The only spec that really needs this mock setup and context defined is the happy path spec. The others introduced here should exercise earlier arg error handling and not rely on context at all.

src/lib/marketingCollectionsCategories.ts
egdbear: Thanks for clarifying that makes more sense!

src/schema/v2/__tests__/marketingCollections.test.ts
egdbear: Yeah, that was the idea, but the test case naming was a bit misleading.

src/schema/v2/__tests__/marketingCollections.test.ts
egdbear: Okay if it is the problem is around the cluttering of the test, I'll define that payload on the top and re-use it.

src/lib/marketingCollectionCategories.ts
```diff
@@ -0,0 +1,91 @@
...
+    ],
+  },
+  Gallery: {
+    id: "Gallery",
+    title: "Gallery",
+    imageUrl:
+      "https://files.artsy.net/images/collections-gallery-category.jpeg",
+    sortedCollectionSlugs: [
+      "new-from-tastemaking-galleries",
+      "new-from-nonprofits-acaf27cc-2d39-4ed3-93dd-d7099e183691",
```
dblandin, line 85: **suggestion / non-blocking**: These collection URLs are sharable from the app. We should probably update this collection to use a friendlier slug. I imagine it's conflicting with [this previously created collection](https://www.artsy.net/collection/new-from-nonprofits) with the same name.

dblandin (approved):
Nice — this looks good to me! 👍🏼

src/schema/v2/marketingCollections.ts
```diff
@@ -320,6 +324,24 @@ export const MarketingCollections: GraphQLFieldConfig<void, ResolverContext> = {
   }),
   resolve: async (_root, args, { marketingCollectionsLoader }) => {
     const { size } = convertConnectionArgsToGravityArgs(args)
+    // Enable curated sorting
+    if (args.sort === MARKETING_COLLECTIONS_SORTS.CURATED.value) {
```
dblandin, line 328: **suggestion / non-blocking**: Just realized that this would accomplish the same thing through the already exported enum type:

```ts
if (args.sort === MarketingCollectionsSorts.getValue('CURATED')?.value) {
```

PR #6193, opened 2024-10-31
chore: mark the genomic recs gravity loader as deprecated

cavvia:
Trying to mark the old genomic recs graphql loader/endpoint as deprecated, since it was decided in https://github.com/artsy/gravity/pull/18204 to not remove it entirely due to some old mobile clients.

### DIFF
--- a/src/schema/v2/me/index.ts
+++ b/src/schema/v2/me/index.ts
@@ -630,7 +630,8 @@ export const meType = new GraphQLObjectType<any, ResolverContext>({
       },
     },
     submissionsConnection: submissionsConnection,
-    // genomic recommendation
+    // DEPRECATED: This genomic recs loader is no longer in use.
+    // Use the artworkRecommendations connection to Vortex instead.
     recommendedArtworks: {
       type: artworkConnection.connectionType,
       args: pageable({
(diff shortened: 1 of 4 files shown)


src/schema/v2/me/index.ts
```diff
@@ -630,7 +630,8 @@ export const meType = new GraphQLObjectType<any, ResolverContext>({
       },
     },
     submissionsConnection: submissionsConnection,
-    // genomic recommendation
+    // DEPRECATED: This genomic recs loader is no longer in use.
+    // Use the artworkRecommendations connection to Vortex instead.
     recommendedArtworks: {
       type: artworkConnection.connectionType,
       args: pageable({
```
egdbear, line 637:
```suggestion
    recommendedArtworks: {
       deprecationReason: "This genomic recs loader is no longer in use. Use the artworkRecommendations instead.",
      type: artworkConnection.connectionType,
      args: pageable({
```

egdbear (commented):
Looks good! As a side note, I’d recommend using `deprecationReason` at the type level, so it's clear when we try to query it.

src/schema/v2/me/index.ts
cavvia: makes sense... thanks for the suggestion!

joeyAghion (dismissed):
👍 from me but needs a schema update according to CI:

> Schema is outdated. You might want to run `yarn dump:staging` and commit the changes.

egdbear (approved):
Seems legit, maybe a rebase will fix the CI

cavvia:
Schema has now been updated, thanks all.

PR #6204, opened 2024-11-05
feat: remove collections from homeView

egdbear:
This PR implements the recommendation from [this Figma](https://www.figma.com/board/kgm28u8rZLh3EfznOuOYSe/Homefeed-Versions?node-id=0-1&node-type=canvas&t=t6JR0d1jtHJzEXmH-0) by removing collections from the HomeView.


dblandin (changes requested):
I don't think we should remove the entire MarketingCollections section type. We may have a need for it later if we want to ship another rail of marketing collections in the future.

Based on the Slack conversation, it sounds like the desired change is here is to remove the "Collections" section instance which is the specific section/rail that we're sending to clients.

dblandin (approved):
Perfect. Thank you!! 💜

PR #6211, opened 2024-11-07
refactor: weaviate discovery and cross reference reset

mc-jones:
This PR refactors two things: 

1. It changes the `discoveryArtwork` implementation to create a user if one is not found in `weaviate` on initial lookup. This is primarily a convenience for us during development, as it prevents us from having to have the user manually create their own `weaviate` user. 
2. It attempts to simplify how we reset a user back to a "cold start" state; i.e. they have no liked or seen artworks.


src/schema/v2/infiniteDiscovery/resetDiscoveryArtworkReferencesMutation.ts
```diff
@@ -0,0 +1,101 @@
...
+            Accept: "application/json",
+          },
+          body: JSON.stringify({
+            properties: {
+              likedArtworks: null,
+              seenArtworks: null,
+            },
+          }),
+        }
+      )
```
mc-jones, line 87: Obvious anti-pattern here of using fetch directly instead of a defined loader. Our canonical pattern would require us making changes to `metaphysics` `fetch` library, which seemed overkill and potentially risky. Specifically, it would require us to add support for the `PATCH` method (not a big deal) and we would have to change how [we parse query params](https://github.com/artsy/metaphysics/blob/main/src/lib/apis/fetch.ts#L22) to [use `strictNullHandling: true`](https://github.com/ljharb/qs/blob/d185cee45f17f1a69705d034eb370dbbbc9fba2d/README.md?plain=1#L598), otherwise, our `null` values here are converted to empty strings by `qs`, which breaks `weaviate`. Same issue if we try to pass an empty array. 

If we decide we are going to add production support for weaviate, we will need fix this.

egdbear (dismissed):
🚀

src/schema/v2/infiniteDiscovery/resetDiscoveryArtworkReferencesMutation.ts
mc-jones: 👍  done in [a6b6bae](https://github.com/artsy/metaphysics/pull/6211/commits/a6b6baef8083a02601bc45be7b201123fda31756)

src/schema/v2/infiniteDiscovery/resetDiscoveryArtworkReferencesMutation.ts
```diff
@@ -0,0 +1,101 @@
...
+        return result
+      },
+    },
+  },
+  mutateAndGetPayload: async ({ userId }, {}) => {
+    const weaviteUserId = generateUuid(userId)
+
+    try {
+      await fetch(
+        `https://weaviate.stg.artsy.systems/v1/objects/InfiniteDiscoveryUsers/${weaviteUserId}`,
```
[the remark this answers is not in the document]
mc-jones: I hear you, but until we make a positive decision to continue with weaviate I'm hesitant to add anything to global config, as it just add a bit of cleanup and I think any final approach will deviate quite a bit from this.

PR #6217, opened 2024-11-08
feat: add likedDiscoveryArtworks

mc-jones:
This PR adds a `likedDiscoveryArtworks` so that we can easily get the artworks a user has liked from weaviate (not just the recommended artworks). Primary motivation for this is to be able to display a list of artworks they have "liked" in the testing prototype.

### DIFF
--- a/src/schema/v2/infiniteDiscovery/likedDiscoveryArtworks.ts
+++ b/src/schema/v2/infiniteDiscovery/likedDiscoveryArtworks.ts
@@ -1,0 +1,60 @@
+import {
+  GraphQLString,
+  GraphQLFieldConfig,
+  GraphQLNonNull,
+  GraphQLInt,
+} from "graphql"
+import { ResolverContext } from "types/graphql"
+import { artworkConnection } from "../artwork"
+import { connectionFromArray } from "graphql-relay"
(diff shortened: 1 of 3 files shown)


src/schema/v2/infiniteDiscovery/likedDiscoveryArtworks.ts
```diff
@@ -0,0 +1,60 @@
...
+  type: artworkConnection.connectionType,
+  args: pageable({
+    userId: { type: GraphQLNonNull(GraphQLString) },
+    limit: { type: GraphQLInt },
+  }),
+  resolve: async (_root, args, { weaviateGraphqlLoader, artworksLoader }) => {
+    if (!weaviateGraphqlLoader || !artworksLoader) {
+      new Error("A loader is not available")
+    }
+    const { userId, limit } = args
```
egdbear, line 26: **nitpick**: set default limit here?

egdbear (approved):

src/schema/v2/infiniteDiscovery/likedDiscoveryArtworks.ts
mc-jones: oh, good catch.

PR #6220, opened 2024-11-11
feat: remove partner offer from primary label

starsirius:
~DON'T MERGE YET. Will need to coordinate deploy with https://github.com/artsy/force/pull/14897, and we might want to include tracking changes.~ As this work becomes too complex to manage now it requires analytics changes. I decided to guard it behind a [feature flag](https://unleash.artsy.net/projects/default/features/emerald_clientside-collector-signals) so we can continuously and incrementally deliver (vs. releasing all at once).

To [load and render the partner offer signals on the client-side](https://github.com/artsy/force/pull/14897) and still render other primary labels (e.g. "increased interest" and "curators' pick") server-side, we'd want to remove partner offer from the `primaryLabel` field.

The separation of authenticated and unauthenticated data into different fields is also [more aligned with schema design](https://github.com/artsy/force/pull/14807#discussion_r1838837387) (although I agree the name now can be a bit confusing).

Note that I didn't [remove it from the enum](https://github.com/artsy/metaphysics/blob/374b26aef4ad052e011aa7b1f09de24cecd11cfe/src/schema/v2/artwork/collectorSignals.ts#L95) as I don't want to introduce breaking schema changes. I can follow up separately once the main work is in.

### DIFF
--- a/src/lib/featureFlags.ts
+++ b/src/lib/featureFlags.ts
@@ -13,6 +13,7 @@ const FEATURE_FLAGS_LIST = [
   "onyx_enable-home-view-section-featured-fairs",
   "diamond_home-view-marketing-collection-categories",
   "onyx_experiment_home_view_test",
+  "emerald_clientside-collector-signals",
 ] as const
 
 export type FeatureFlag = typeof FEATURE_FLAGS_LIST[number]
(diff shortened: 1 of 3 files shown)


starsirius:
OK I moved this change behind a feature flag, so we can release things incrementally. This should be ready for review and hopefully merge! 🙏

damassi (commented):
🙏

PR #6226, opened 2024-11-13
fix: update the artwork params when querying for infinite discovery

egdbear:
This PR updates the artwork parameters used for querying the artworks from Elastic Search.

Referring to [this comment](https://github.com/artsy/metaphysics/pull/6219/files#r1837943508), it would make sense during the research phase to include something closer to what we’d ultimately deploy in production


cavvia (approved):

cavvia (commented):
Will we be rotating the artworks in the query between batches? Assuming no new saves have come in from the previous batch of artworks, it would require some explicit rotation.

egdbear:
> Will we be rotating the artworks in the query between batches? Assuming no new saves have come in from the previous batch of artworks, it would require some explicit rotation.

Good point, I will note that, and add in the upcoming PR!

PR #6229, opened 2024-11-13
feat: infinite disco return curated artworks

mc-jones:
1. Return curated works when user has no liked artworks.
2. Inject 2 random "challenge" artworks, also coming from the curated collection.

### DIFF
--- a/src/schema/v2/infiniteDiscovery/discoverArtworks.ts
+++ b/src/schema/v2/infiniteDiscovery/discoverArtworks.ts
@@ -237,8 +237,7 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
       ).forEach((challengeArtwork) => {
         const length = response.length
         const randomIndex = random(length)
-        // Splice mutates the array
-        response.splice(randomIndex, 0, challengeArtwork)
+        response = response.toSpliced(randomIndex, 0, challengeArtwork)
       })
     } else {
       response = await weaviateGraphqlLoader({
--- a/src/schema/v2/infiniteDiscovery/discoverArtworks.ts
+++ b/src/schema/v2/infiniteDiscovery/discoverArtworks.ts
@@ -13,7 +13,7 @@ import { connectionFromArray } from "graphql-relay"
 import { pageable } from "relay-cursor-paging"
 import gql from "lib/gql"
 import uuid from "uuid/v5"
-import { sampleSize, shuffle } from "lodash"
+import { random, sampleSize, shuffle } from "lodash"
 
 export const generateUuid = (userId: string) => {
   if (!userId) return ""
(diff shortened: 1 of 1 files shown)


src/schema/v2/infiniteDiscovery/discoverArtworks.ts
```diff
@@ -179,55 +179,96 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
...
-    )
+      // Insert the challenge artwork at a random index
+      sampleSize(
+        challengeResponse.data?.Get?.InfiniteDiscoveryArtworks,
+        2
+      ).forEach((challengeArtwork) => {
+        const length = response.length
+        const randomIndex = random(length)
+        // Splice mutates the array
+        response.splice(randomIndex, 0, challengeArtwork)
```
egdbear, line 241: **suggestion**: we shouldn't mutate the response to avoid any potential side effects (due to many factors like for example schema change for example or similar)

```suggestion
      // consider renaming this to curated
      const challengeArtworks = sampleSize(
        challengeResponse.data?.Get?.InfiniteDiscoveryArtworks,
        2
      );
      // Combine response and challenge artworks, then shuffle
      const shuffledArtworks = shuffle(response.concat(challengeArtworks));
```

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
```diff
@@ -179,55 +179,96 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
...
+                  beacon: "${beacon}",
+                  certainty: ${certainty}
+                },
+                limit: ${limit},
+                offset: ${offset},
+            ) {
+              internalID
+              _additional {
+                id
+              }
```
egdbear, line 198: **suggestion**: `searchQuery` and `challengeQuery` could work without declaring them in the resolver? I think it would make sense to move them out to make this simpler?

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
```diff
@@ -179,55 +179,96 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
       user.seenArtworks?.map((node) => node.internalID) || []
     const artworkIdsToFilter = [...likedArtworkIds, ...seenArtworkIds]
 
-    let searchQuery
-
+    let response
     if (likedArtworkIds.length > 0) {
```
egdbear, line 183: **suggestion:** Rather than declaring response here, and later mutating the response we could instead, get all data (in both cases we'd need curated artworks), and then generate the shuffled response based on this case. Similar to [this](https://github.com/artsy/metaphysics/blob/be271119240cf778fb1cdd1836789bbac73dbf41/src/schema/v2/infiniteDiscovery/discoverArtworks.ts#L113)

egdbear (changes requested):
This looks 🚀 Let's give a small polish before merging!

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
mc-jones: >we shouldn't mutate the response to avoid any potential side effects 

I don't follow. In this instance we're just mutating an array of "internalIDs", which shouldn't be affected by schema changes? But will update to copy the array instead of mutating directly. 

>const shuffledArtworks = shuffle(response.concat(challengeArtworks));

This doesn't quite work since order matters in this instance. The `nearObject` search is returning works in order of most recommended to less, so we need to preserve that order.

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
```diff
@@ -179,55 +179,96 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
...
-          InfiniteDiscoveryArtworks(
-              limit: ${limit},
-              offset: ${offset},
-          ) {
-            internalID
-            _additional {
-              id
+      `
+
+      const challengeQuery = gql`
```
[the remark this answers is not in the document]
mc-jones: resolved in [b0f3355](https://github.com/artsy/metaphysics/pull/6229/commits/b0f3355870ac3d217158a8a48c6f5360c2e23363)

PR #6241, opened 2024-11-19
refactor: infinite discovery

mc-jones:
This PR cleans and simplifies the existing infinite discovery code, focusing on the `weaviate` centric code. 

Additionally, it swaps out `DeleteDiscoveryUserReferencesMutation` for `DeleteDiscoveryUser`. This follows #6231, and uses more appropriate naming. 

**NOTE:** I still need to double check I didn't introduce any regressions, but I wanted to get your thoughts on the approach.


mc-jones:
Taking this out of Draft mode. Let me know if any thoughts @egdbear !

src/lib/infiniteDiscovery/weaviate.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  const curatedSample = sampleSize(curatedWorks, sampleAmount)
+
+  curatedSample.forEach((challengeArtwork) => {
+    const length = target.length
+    const randomIndex = random(length)
+    mixedArtworks.splice(randomIndex, 0, challengeArtwork)
+  })
+
+  return mixedArtworks
```
egdbear, line 90: **suggestion:** This could be simplified like: 
```suggestion
  const curatedSample = sampleSize(curatedWorks, sampleAmount)
  const mixedArtworks = shuffle([...curatedWorks, ...target])

  return mixedArtworks
```

egdbear (commented):
Thanks for clean-up! It looks a better than it was! 

I'm curious, how are we going to handle duplicates? As a consumer of the weaviate API I'm not forbidden to create duplicates for any reference type, for example: 

```
mutation {
	createDiscoveryArtworkReference(input: {
		userId: "any",
		reference: LIKED_ARTWORKS
		artworkId: "66e05b9579a525000e32a9ef"
	}) {
		clientMutationId
	}
}
```

src/lib/infiniteDiscovery/weaviate.ts
mc-jones: This has come up a few times so let me know if I'm missing something. But the current goal is to preserve the order of the artworks coming back from weaviate which have been ordered by similarity. The suggestion would not preserve the order of the weaviate recommended artworks, right?

egdbear: Oh, my bad, again, I haven't been following the recent comms, and I haven't found that in the ACs in [the notion](https://www.notion.so/artsy/Infinite-Discovery-Qualitative-QA-12fcab0764a0805c921bfe0146881f9c). 
In that case you can dismiss this comment!

src/lib/infiniteDiscovery/weaviate.ts
```diff
@@ -0,0 +1,251 @@
...
+  return responseArray.InfiniteDiscoveryUsers[0]
+}
+
+/**
+ * Extracts the artwork internalIDs from the Weaviate response
+ *
+ * @param response from the Weaviate API containing InfiniteDiscoveryArtworks
+ * @returns an array of the artwork internalIDs
+ */
+export function GetArtworkIds(response: ArtworkResponse): string[] {
```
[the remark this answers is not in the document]
mc-jones: good catch. Fixed in [dbe7883](https://github.com/artsy/metaphysics/pull/6241/commits/dbe78831de9529a7b93fe8872d3d3da275be8e03)

mc-jones:
Yeah, right now for testing this is just being handled client side. But we can add a check here, too. Will follow up with this in a different PR.

PR #6251, opened 2024-11-20
refactor(onyx-1406): add partner artwork to viewing room type

nickskalkin:
Thanks again @mzikherman for suggesting to try `require` thingy, couldn't figure out where the error was coming from!

Follow-up: https://github.com/artsy/metaphysics/pull/6065#discussion_r1846229641

This PR adds partner artworks connection to viewing rooms:

```graphql
{
  viewingRoom(id: "leslie-feely-alice-baber") {
    partnerArtworksConnection(first: 1) {
      totalCount
      edges {
        node {
          title
        }
      }
    }
  }
}
```

### DIFF
--- a/src/schema/v2/viewingRoom.ts
+++ b/src/schema/v2/viewingRoom.ts
@@ -12,7 +12,7 @@ import {
   convertConnectionArgsToGravityArgs,
   defineCustomLocale,
 } from "lib/helpers"
-import { createPageCursors } from "./fields/pagination"
+import { createPageCursors, emptyConnection } from "./fields/pagination"
 import { connectionFromArray } from "graphql-relay"
 import { artworkConnection } from "./artwork"
 import { pageable } from "relay-cursor-paging"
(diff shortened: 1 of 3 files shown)


src/schema/v2/viewingRoom.ts
```diff
@@ -66,6 +65,8 @@ defineCustomLocale(LocaleEnViewingRoomRelativeLong, {
 export const ViewingRoomType = new GraphQLObjectType<any, ResolverContext>({
   name: "ViewingRoom",
   fields: () => {
+    const { PartnerArtworks } = require("schema/v2/partner/partnerArtworks")
+
```
anandaroop, line 69: **question (non-blocking):** Trying to get my head around the `require` (vs `import`) — this is because it's being referenced inside of the thunk declared on L67?

anandaroop (dismissed):
LGTM 👍🏽  just looking for one clarification below…

PR #6249, opened 2024-11-20
refactor(onyx-1404): add subsections to viewing room type

nickskalkin:
Continuation of https://github.com/artsy/metaphysics/pull/6065 (viewing rooms unstitching)

This PR adds `viewingRoom.subsections` field:

```graphql
{
  viewingRoom(id: "leslie-feely-alice-baber") {
    subsections {
      __typename
      body
      caption
      image {
        __typename
        internalID
        height
        width
        imageURLs {
          __typename
          normalized
        }
      }
      internalID
      title
    }
  }
}
```

Response:

```json
{
  "data": {
    "viewingRoom": {
      "subsections": [
        {
          "__typename": "ViewingRoomSubsection",
          "body": "",
          "caption": "Alice Baber, 1972, photo by Richard Galef, Alice Baber Papers, AAA",
          "image": {
            "__typename": "ARImage",
            "internalID": "e2185542-6cad-4916-b39c-4e5458f6066b",
            "height": 1007,
            "width": 1024,
            "imageURLs": {
              "__typename": "ImageURLs",
              "normalized": "https://d32dm0rphc51dk.cloudfront.net/w6qTHWN4EWvAZh7Dxnto5A/normalized.jpg"
            }
          },
          "internalID": "fbfadaed-caba-4006-a19f-f131811b2f07",
          "title": "Baber, in Her Studio"
        }
      ]
    }
  }
}
```

PR #6254, opened 2024-11-21
refactor: don't repeat InternalIDField definition

nickskalkin:
Follow-up of this code review comment: https://github.com/artsy/metaphysics/pull/6249#discussion_r1850676046

Before I was repeating definition of `InternalID` field in multiple places, now I simply spread the definition from a single place.

### DIFF
--- a/src/schema/v2/object_identification.ts
+++ b/src/schema/v2/object_identification.ts
@@ -245,6 +245,14 @@ export const InternalIDFields: GraphQLFieldConfigMap<any, ResolverContext> = {
   },
 }
 
+export const InternalIDField: GraphQLFieldConfigMap<any, ResolverContext> = {
+  internalID: {
+    description: "A type-specific ID likely used as a database ID.",
+    type: new GraphQLNonNull(GraphQLID),
+    resolve: ({ id }) => id,
+  },
+}
+
 export default {
   GlobalIDField,
   GravityIDFields,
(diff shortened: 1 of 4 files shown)


src/schema/v2/object_identification.ts
```diff
@@ -245,6 +245,14 @@ export const InternalIDFields: GraphQLFieldConfigMap<any, ResolverContext> = {
   },
 }
 
+export const InternalIDField: GraphQLFieldConfigMap<any, ResolverContext> = {
```
dblandin, line 248: **suggestion / non-blocking**: Can we also reuse this definition within the `GraphQLFieldConfigMap`s defined above?

For example:

```ts
export const InternalIDFields: GraphQLFieldConfigMap<any, ResolverContext> = {
  id: GlobalIDField,
  internalID: InternalIDField
}
```

src/schema/v2/viewingRoom.ts
```diff
@@ -68,11 +69,7 @@ export const ViewingRoomType = new GraphQLObjectType<any, ResolverContext>({
     const { PartnerArtworks } = require("schema/v2/partner/partnerArtworks")
 
     return {
-      internalID: {
-        description: "A type-specific ID likely used as a database ID.",
-        type: new GraphQLNonNull(GraphQLID),
-        resolve: ({ id }) => id,
-      },
+      ...InternalIDField,
```
dblandin, line 72: **suggestion / non-blocking**: Since this is just a single field and we're not spreading multiple fields, pulling in the specific field definition might be fine too:

```ts
internalID: InternalIDField
```

dblandin (approved):
Nice 👍🏼

dblandin:
@nickskalkin Do you want to keep this open or should we close for now?

PR #6282, opened 2024-12-03
feat(DIA-1014): infinite discovery home view section

iskounen:
This PR adds a new home view section for [Infinite Discovery](https://www.figma.com/design/BzXubI34mJlLU9CmmYip0p/Infinite-Discovery?node-id=917-27991&t=BJwsDAs7piAAOw6G-4) so that Diamond can start building it.

card

### DIFF
--- a/src/schema/v2/homeView/sections/InfiniteDiscovery.ts
+++ b/src/schema/v2/homeView/sections/InfiniteDiscovery.ts
@@ -1,9 +1,27 @@
+import { ContextModule, OwnerType } from "@artsy/cohesion"
 import { HomeViewSection } from "."
+import { withHomeViewTimeout } from "../helpers/withHomeViewTimeout"
+import { HomeViewCard } from "../sectionTypes/Card"
 import { HomeViewSectionTypeNames } from "../sectionTypes/names"
 
 export const InfiniteDiscovery: HomeViewSection = {
-  id: "home-view-section-infinite-discovery",
+  contextModule: ContextModule.infiniteDiscoveryBanner,
   featureFlag: "diamond_home-view-infinite-discovery",
-  type: HomeViewSectionTypeNames.HomeViewSectionCard,
+  id: "home-view-section-infinite-discovery",
   requiresAuthentication: true,
+  ownerType: OwnerType.infiniteDiscovery,
+  type: HomeViewSectionTypeNames.HomeViewSectionCard,
+
+  resolver: withHomeViewTimeout(async (parent, _args, _context, _info) => {
+    const card: HomeViewCard = {
+      title: "Discover art, tailored to you",
+      subtitle:
+        "Our new feature learns your unique tastes as you explore, delivering personalized recommendations effortlessly.",
+      buttonText: "Explore Art",
+      image_url: "https://files.artsy.net/images/infinite_discovery.png",
+      entityType: "Page",
+      entityID: parent.ownerType,
+    }
+    return card
+  }),
 }
--- a/src/schema/v2/homeView/sections/InfiniteDiscovery.ts
+++ b/src/schema/v2/homeView/sections/InfiniteDiscovery.ts
@@ -1,0 +1,9 @@
+import { HomeViewSection } from "."
+import { HomeViewSectionTypeNames } from "../sectionTypes/names"
+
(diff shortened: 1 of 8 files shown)


src/schema/v2/homeView/sections/InfiniteDiscovery.ts
```diff
@@ -0,0 +1,9 @@
+import { HomeViewSection } from "."
+import { HomeViewSectionTypeNames } from "../sectionTypes/names"
+
+export const InfiniteDiscovery: HomeViewSection = {
+  id: "home-view-section-infinite-discovery",
+  featureFlag: "diamond_home-view-infinite-discovery",
+  type: HomeViewSectionTypeNames.HomeViewSectionCard,
+  requiresAuthentication: true,
```
iskounen, line 8: This section will act as a link to open Infinite Discovery so it is very simple. I foresee adding analytics properties as the project progresses. Please tell me if there are other properties that I should consider.

src/schema/v2/homeView/sections/InfiniteDiscovery.ts
```diff
@@ -0,0 +1,9 @@
+import { HomeViewSection } from "."
+import { HomeViewSectionTypeNames } from "../sectionTypes/names"
+
+export const InfiniteDiscovery: HomeViewSection = {
+  id: "home-view-section-infinite-discovery",
+  featureFlag: "diamond_home-view-infinite-discovery",
+  type: HomeViewSectionTypeNames.HomeViewSectionCard,
```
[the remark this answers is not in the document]
iskounen: > not sure if we're gonna use the element HomeViewSectionCard to display it

I agree that we can iterate on this section's details, but I'd like to have some confidence in my choice of section type. Who can confirm that this (or an alternative) is the right choice?

olerichter00 (approved):
Looks great! 🌟 

ℹ️ A thought on the front-end implementation of the app: The new banner appears to have a fixed height. We should make sure it looks good with iOS font scaling (see this PR as an example of adding font scaling support to artwork rails: https://github.com/artsy/eigen/pull/11214).

src/schema/v2/homeView/sections/InfiniteDiscovery.ts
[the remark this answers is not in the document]
anandaroop: Yeah that's right. It comes down to a judgement call between:

- use an **existing section type**, as-is and let it render in the default way
- use an **existing section type**, but prescribe a **custom client-side component** 
  - (when the section type's data is adequate but we desire a different presentation)
- devise a **new section type** altogether 
  - (when no existing section type meets our data requirements)

_Hopefully_, the first option meets many if not most of our needs, but we have the other two options as well. 

In particular we hope that `Card` can meet many of the one-off banner-type presentation needs, which is why it was recommended for this use case.

src/schema/v2/homeView/sections/InfiniteDiscovery.ts
anandaroop: Yeah I think you will likely want to add a resolver that returns the card's data before merging this.

Right now the card section is wired up correctly — but it doesn't return any of the card-y data. This means that over in Eigen this section won't even render because of [this early return](https://github.com/artsy/eigen/blob/89c48d20c04126bcc43f97b242b84f3878b9125a/src/app/Scenes/HomeView/Sections/HomeViewSectionCard.tsx#L39-L41).

So a minimal PR here would likely also include a resolver that looks something [like this](https://github.com/artsy/metaphysics/blob/15bd7d2a4fa31c10d0b78ee918ed65e491076d63/src/schema/v2/homeView/sections/GalleriesNearYou.ts#L19-L29).

Let me know if you want to pair, happy to do so.

anandaroop (commented):
Looking good so far 👍🏽 but you'd probably want to add a resolver, see comments below…

src/schema/v2/homeView/sections/InfiniteDiscovery.ts
iskounen: Thanks for the advice. I didn't understand how Eigen consumes sections when I drafted this PR, and assumed that Eigen would be in charge of defining its contents. I've added the resolver now.

PR #6283, opened 2024-12-03
refactor: release un-stitched viewing rooms

nickskalkin:
About to finally release it


nickskalkin:
peril complains about breaking schema changes, but I did `yarn relay` in Force / Eigen / Volt and it didn't complain - should be alright, right?

_schemaV2.graphql
```diff
@@ -22198,15 +22153,21 @@ type Viewer {
...
   viewingRoomsConnection(
     after: String
+    before: String
+    featured: Boolean
     first: Int
+    ids: [ID!]
+    last: Int
     partnerID: ID
-    statuses: [ViewingRoomStatusEnum!]
+    statuses: [ViewingRoomStatusEnum!] = [live]
```
olerichter00, line 22167: Is `[live]` the new default?

olerichter00 (approved):
🌟

_schemaV2.graphql
nickskalkin: I was actually always a default: https://github.com/artsy/gravity/blob/main/app/graphql/types/query_type.rb#L97

but stitching code was not specifying it: https://github.com/artsy/metaphysics/blob/main/src/lib/stitching/gravity/v2/stitching.ts#L129

I understand it as this schema change shouldn't change anything, right? I will double-check

nickskalkin: yeah, I made such request:

```graphql
{
  viewer {
    viewingRoomsConnection(first: 1) {
      totalCount
    }
  }
}
```

And I see this is introspection:

```
{
            "query": "query ($_v0_first: Int, $_v0_statuses: [ViewingRoomStatusEnum!]) {\n  viewingRoomsConnection(first: $_v0_first, statuses: $_v0_statuses) {\n    totalCount\n  }\n}\n",
            "vars": {
              "_v0_first": 1,
              "_v0_statuses": [
                "**live**"
              ]
            }
          }
```

olerichter00: Thanks for testing 🙏  Good the have the default in the schema.

PR #6288, opened 2024-12-05
chore: add is editable field to ArtworkConsignmentSubmission

dariakoko:
Ass discussed, add isEditable field that indicates if an artwork can be edited to MP: https://github.com/artsy/eigen/pull/11249#pullrequestreview-2480985833

### DIFF
--- a/src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
+++ b/src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
@@ -39,6 +39,8 @@ const ArtworkConsignmentSubmissionType = new GraphQLObjectType<
       },
       isEditable: {
         type: GraphQLBoolean,
+        description:
+          "Whether the user is allowed to edit the associated My Collection artwork.",
         resolve: ({ state }) => {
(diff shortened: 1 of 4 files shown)


src/schema/v2/artwork/artworkConsignmentSubmissionType.ts
```diff
@@ -37,6 +37,12 @@ const ArtworkConsignmentSubmissionType = new GraphQLObjectType<
         type: GraphQLString,
         resolve: ({ externalId }) => externalId,
       },
+      isEditable: {
```
olerichter00, line 40:
```suggestion
      isEditable: {
        description: "Whether the user is allowed to edit the associated My Collection artwork.",
```

olerichter00 (commented):
Good idea to add the field to the `ArtworkConsignmentSubmission` instead of `Artwork` 👍 

I've added a comment with a suggestion to include a description for the new field and I think some (simple) tests would be great as well.

MounirDhahri (dismissed):
Lovely! thanks for the addition this is sooo useful

PR #6297, opened 2024-12-10
feat(home view): allow section declaration to specify a minimum required Eigen version

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-1444

This originated in a [discussion over this Eigen PR](https://github.com/artsy/eigen/pull/11248#issuecomment-2529641707) about the question of how to handle this tricky scenario:

- 🟢  A section type such as `HomeViewSectionCard` has been available for a long time
- 🟢  So old and new clients will know how to render that card correctly
- 🟡  **BUT** that card navigates the user to a feature which is WIP and won't be ready for release till later

  tfw

In this situation we _could_ add a second layer of Echo feature flagging over in Eigen to conditionally suppress the card within the home view. But this adds a lot of paperwork, and is sure to be repeated over and over as we add entry points in the home view for various WIP features.

Instead we propose to control this with a simple declarative syntax in the home view definition here in Metaphysics.

```diff
export const CardThatTakesMeToAFancyFeature: HomeViewSection = {
  id: "home-view-section-fancy-entry-point",
  type: HomeViewSectionTypeNames.HomeViewSectionCard,
+  minimumEigenVersion: { major: 42, minor: 4, patch: 2 },

  resolver: withHomeViewTimeout(async (parent, _args, _context, _info) => {
    // ...
  }),
}
```

This allows us to peg the visibility of a section to a specific minimum Eigen version with a one-line PR to Metaphysics.


src/lib/semantic-versioning.ts
```diff
@@ -0,0 +1,39 @@
+import { isArray } from "lodash"
+
+export type SemanticVersionNumber = {
+  major: number
+  minor: number
+  patch: number
+}
```
anandaroop, line 7: There was some talk of specifying the version string more simply e.g. `8.59.0` as we do elsewhere.

But since I want to work with the semver components I want this to be more strongly typed.

I also considered using a Typescript [template literal type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) — but I got spooked by its apparent lack of accurate enforcement.

For example, it does not complain about this malformed semver string,  which could result in a consequential parsing error:

TS why?

src/lib/semantic-versioning.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export function getEigenVersionNumber(
+  userAgent?: string | string[]
+): SemanticVersionNumber | null {
+  if (!userAgent) return null
+  if (!userAgent.includes("Artsy-Mobile")) return null
+
+  const agent = isArray(userAgent) ? userAgent[0] : userAgent
+  const parts = agent.split("/")
+  const version = parts.at(-1)
```
anandaroop, line 17: **L15** is interesting — [MP seems convinced](https://github.com/artsy/metaphysics/blob/70fd9642465eaaaceff2babb493d0f60d7a7fd84/src/types/graphql.d.ts#L23-L24) that a user agent could be an _array_ of strings as well as just a single string — but I'm not sure why. 

So I just consider the first string in such a case — safe?

**L16-17** is based on my reading of this code in Eigen, and seems to work out fine IRL:

https://github.com/artsy/eigen/blob/396ab8c1cf4a16c8095a3ce240e59a112b496ce0/src/app/store/GlobalStore.tsx#L111-L113

src/schema/v2/homeView/helpers/__tests__/isSectionDisplayable.test.ts
```diff
@@ -121,4 +121,81 @@ describe("isSectionDisplayable", () => {
...
+
+      expect(
+        isSectionDisplayable(
+          section as HomeViewSection,
+          context as ResolverContext
+        )
+      ).toBe(true)
+    })
+
+    it("returns true if an Eigen version is not recognized", () => {
```
anandaroop, line 183: There seemed to be some uncertainty in KS when we mentioned this case. 

My thinking here is:

- **If** we know a request is coming from Eigen then we gate the section content as needed
- **Else** we allow the section content through since we don't want to suppress it from non-Eigen clients

src/schema/v2/homeView/sections/InfiniteDiscovery.ts
```diff
@@ -9,6 +9,8 @@ export const InfiniteDiscovery: HomeViewSection = {
   featureFlag: "diamond_home-view-infinite-discovery",
   id: "home-view-section-infinite-discovery",
   requiresAuthentication: true,
+  // TODO: update this to match the first release that can support Infinite Discovery
+  minimumEigenVersion: { major: 8, minor: 59, patch: 0 },
```
anandaroop, line 13: Here I actually wire this up for infinite disco, but can certainly remove this if that's preferable (it might be!).

Thinking of how this will work in the future — it is a simple thing to add this constraint, but it must be done at the right time, in order to not interfere with development, and to not serve up to users prematurely.

src/lib/fetchPersistedQuery.ts
```diff
@@ -12,7 +12,8 @@ export const fetchPersistedQuery: RequestHandler = (req, res, next) => {
     } else {
       const message = `Unable to serve persisted query with ID ${documentID}`
       error(message)
-      return res.status(404).send(message).end()
+      res.status(404).send(message).end()
+      return
```
anandaroop, line 16: One necessary but unrelated change, see https://github.com/artsy/metaphysics/pull/6297/commits/6c800ab80e9335789a3a7ada5dba13addef303a3

anandaroop:
Looking over this, wondering if it might be wise to add some Android focused coverage as well…

src/lib/semantic-versioning.ts
mzikherman: I kinda like the more structured approach as you've done in the PR. While it might be a tad more verbose, it's clearly structured and typing the right realistic abstraction - the Eigen version.

src/lib/semantic-versioning.ts
mzikherman: I suppose the extra snippet that checks `isArray` is safe to include, but I think that typing is possibly wrong. There's a decent amount of uses of `userAgent` around the codebase (not typed with `ResolverContext`) that treat it as a string and I don't ever recall an issue. It's just a value from a header which should be a string - https://github.com/artsy/metaphysics/blob/f05d2dcf28b915156ae900b185a29830f5b5e378/src/index.ts#L152

mzikherman (approved):
I dig it!

src/lib/semantic-versioning.ts
```diff
@@ -0,0 +1,39 @@
+import { isArray } from "lodash"
```
damassi, line 1: #superminor - most files in the `lib` directory follow conventional `camelCase` file naming

src/schema/v2/homeView/helpers/__tests__/isSectionDisplayable.test.ts
MounirDhahri: good thought!

MounirDhahri (approved):
LGTM! thanks!

joeyAghion:
> that card navigates the user to a feature which is WIP and won't be ready for release till later

Will that feature end up relying on other Echo or feature toggles? If so, I sort of wish the corresponding home view could rely on that as well, rather than introducing yet another variable to the already complicated context.

> we could add a second layer of Echo feature flagging over in Eigen to conditionally suppress the card within the home view...

This was my first thought too. I guess it would look like a `{"name":"AREnableFoo","minVersion":"..."}`?

damassi (dismissed):

src/lib/semantic-versioning.ts
anandaroop: Cool, casting that to a singular string in 772de17

src/lib/semantic-versioning.ts
anandaroop: Fixed in 439e01d

anandaroop (commented):
> I sort of wish the corresponding home view could rely on [existing toggles] as well

I do like the idea of using the same Unleash flag that gates the feature to also gate the home view section — but my understanding (as well as our [current documentation](https://github.com/artsy/eigen/blob/4e7c837d2457fa80c6f8e16e53e1aaeaf6140154/docs/a_b_testing_best_practices.md?plain=1#L11)) suggests we don't use Unleash flags as feature toggles in Eigen; we only use them to manage experiments.

So I'll plan on rolling with this approach but am open to revisiting if this^ changes.

PR #6306, opened 2024-12-12
feat(dia-1033): infiniteDiscovery queries and temporal direct integration with OpenSearch

nickskalkin:
[Jira](https://artsyproduct.atlassian.net/browse/DIA-1033)

The schema (and underlying code) is not final yet; it is meant to enable us to test OpenSearch-based artwork similarity on staging. You will notice that Metaphysics talks to OpenSearch directly. In the future, we aim to move such communication to Gravity. 

# discoverArtworksConnection

Returns a list of "similar" artworks. Options are optional, if `likedArtworkIds` and `tasteProfileVector` both are empty - this query will return an initial batch of artworks (which are curator's picks currently, but will be changed in future) for user to start his discovery journey. 

```graphql
{
  discoverArtworksConnection(
    first: 10
    excludeArtworkIds: ["artwork-1"]
    likedArtworkIds: ["artwork-2", "artwork-3"]
  ) {
    edges {
      node {
        internalID
        title
        imageUrl
      }
    }
  }
}
```

# Pre-deploy

- [ ] Add env variables to staging


package.json
```diff
@@ -86,6 +86,7 @@
     "lodash": "4.17.21",
     "longjohn": "0.2.12",
     "marked": "2.0.1",
+    "mathjs": "^14.0.1",
```
nickskalkin, line 89: we need this for the `mean` function to calculate mean of multiple vectors. It brings with it some dependancies - I think we can implement our own `mean`. But at the same time - this code is expected to be short-lived - such computations will live in Gravity

joeyAghion (commented):
Some minor questions, which you can feel free to consider in future PRs:
* The `excludeArtworkIds` list sounds like it will only work within a single client instance or session. Is the idea that the mobile app would retain that list for a long time?
* I'd like to avoid project labels like "infinite discovery" in the schema, since the actual feature tends to evolve significantly beyond those. "Infinite" also has a specific meaning which doesn't exactly fit here, so maybe [eventually] this can be organized under "discover" or "discovery?"

egdbear (commented):
- We planned to use a single type regardless of the number of APIs. This approach simplifies switching between them, even from the client side.
- Naming: I don't have a strong opinion about discoverArtworks. It seems fine, but if it's not ideal, we should consider finding a better name and updating the types accordingly. @joeyAghion
- Curious if we can calculate the mean using any existing libraries. I haven't checked yet, but it’s worth exploring this before shipping. 
- If you have a strong opinion about exposing the mean calculation, I'd love to hear more about that! 

Overall, great progress! 🚀

_schemaV2.graphql
```diff
@@ -11842,6 +11842,25 @@ type ImageURLs {
   normalized: String
 }
 
+# Schema for everything infinite discovery-related
+type InfiniteDiscovery {
```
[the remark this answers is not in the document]
nickskalkin: again, not strongly opinionated, but I intentionally introduced a new field, because:

- it returns a connection and I think it should be explicit in the name
- `discoverArtworks` seemed too generic to me, but @joeyAghion suggests to not use a name which is too project-specific.

src/schema/v2/infiniteDiscovery/tasteProfileVector.ts
```diff
@@ -0,0 +1,23 @@
+import {
```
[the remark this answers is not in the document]
nickskalkin: We don't need it. I initially added it to help me test things and when It came down to deleting it I've decided to keep it a reason which doesn't makes sense to me now. I will delete it

src/lib/apis/opensearch.ts
```diff
@@ -0,0 +1,26 @@
...
+    Accept: "application/json",
+    "Content-Type": "application/json",
+  }
+
+  const response = await (
+    await fetch(
+      urljoin(OPENSEARCH_API_BASE, path),
+      assign({}, fetchOptions, { headers })
+    )
+  ).json()
```
damassi, line 23: This feels like a nice spike-y approach; get it working here, eventually move it to where it needs to be (when those addition ecosystem (grav) features are required)

nickskalkin:
@joeyAghion @egdbear got rid of `infinite` part, but kept `Connection`. 

@egdbear we can remove weaviate-backed code completely in follow-up PR, I think there is no point in storing it anymore.

src/schema/v2/schema.ts
```diff
@@ -354,6 +355,7 @@ const rootFields = {
   homeView: HomeView,
   identityVerification: IdentityVerification,
   identityVerificationsConnection,
+  infiniteDiscovery: InfiniteDiscovery,
```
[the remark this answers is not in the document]
nickskalkin: At this point it it easier to delete this type since there is not much common between them. I think we should do it in a separate PR

egdbear:
> @joeyAghion @egdbear got rid of `infinite` part, but kept `Connection`.
> 
> @egdbear we can remove weaviate-backed code completely in follow-up PR, I think there is no point in storing it anymore.

I think we should keep the `discoverArtworks` type since it’s essentially the same as `discoverArtworksConnection` in the sense that both encapsulate the `artworkConnection`. Also, I feel it’s a good idea to keep the Weaviate code as it is and replace the Elasticsearch-backed approach with the OpenSearch approach. This way we can maintain a single type/resolver for both APIs while keeping the Weaviate implementation running, which could be useful for any comparisons down the line.

nickskalkin:
@egdbear

- **[this is important to agree upon now]** how likely is it that we are going compare weaviate / opensearch implementations? I thought it was decided that we stick to opensearch and weaviate did it part.
- **[this is important to agree upon now]** I am ok to move OS code into the old type, but old type has arguments that in my mind do not have sense now, is there a reason to keep them (userId, sort, useRelatedArtworks for example)? The interface between the old type and the new is completely different. If we don't need old arguments + we won't use (very likely) weaviate in future - why not simply kill the old type?
- **[not important, but still want to discuss]** on naming, I still feel that any `Connection` has to have `Connection` in it's name for transparency, but I would like to hear other opinions (@mzikherman)

egdbear:
> [this is important to agree upon now] how likely is it that we are going compare weaviate / opensearch implementations? I thought it was decided that we stick to opensearch and weaviate did it part.

Weaviate notebook is still using this, and it might be useful to keep it for reference. I _would_ keep it until we start implementing the backend for OpenSearch. Why should we keep it? I think it could be helpful for folks to check out or compare the searches as we transition to the OpenSearch approach. I could def be wrong about this, but @joeyAghion might also chime in on this.


> [this is important to agree upon now] I am ok to move OS code into the old type, but old type has arguments that in my mind do not have sense now, is there a reason to keep them (userId, sort, useRelatedArtworks for example)? The interface between the old type and the new is completely different. If we don't need old arguments + we won't use (very likely) weaviate in future - why not simply kill the old type? 

The part about removing it has already been addressed in the previous comment. As for the API and arguments, I think we can add the existing arguments used in `discoverArtworksConnection` and replace everything after [this line](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/infiniteDiscovery/discoverArtworks.ts#L82)


> [not important, but still want to discuss] on naming, I still feel that any Connection has to have Connection in it's name for transparency, but I would like to hear other opinions (@mzikherman)

Yeah that's a good callout, when we were building the API for the Weaviate I was referring to the [artworksRecommendations](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/me/artworkRecommendations.ts) which is also artworkConnection.

nickskalkin:
Not the request is:

```graphql
{
  discoverArtworks(
    limit: 10
    useOpenSearch: true
    excludeArtworkIds: ["artwork-1"]
    likedArtworkIds: ["artwork-2", "artwork-3"]
  ) {
    edges {
      node {
        internalID
        title
        imageUrl
      }
    }
  }
}

```

src/lib/infiniteDiscovery/getInitialArtworksSample.ts
```diff
@@ -0,0 +1,17 @@
+import { opensearch } from "lib/apis/opensearch"
+
+export const getInitialArtworksSample = async (limit, artworksLoader) => {
```
egdbear, line 3: **suggestion:** We should probably add a constant for the limit here (`size=135`), as I believe the default size is 10. This way, we can ensure we get a random selection of 10 artworks from the entire collection.

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
```diff
@@ -48,19 +51,22 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
         },
       }),
     },
-    useRelatedArtworks: { type: GraphQLBoolean, defaultValue: false },
+    useOpenSearch: { type: GraphQLBoolean, defaultValue: false },
+    excludeArtworkIds: {
+      type: new GraphQLList(GraphQLString),
+      description:
+        "(Only for when useOpenSearch is true) Exclude these artworks from the response, currently works only in combination with tasteProfileVector",
```
egdbear, line 58: **question:** Is this comment still relevant, now that we don't have the tasteProfileVector?

egdbear (dismissed):

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
nickskalkin: thanks, not relevant anymore

src/lib/infiniteDiscovery/getInitialArtworksSample.ts
nickskalkin: I don't think that passing size will randomise the results (but you probably mean that we can randomise them via javascript and pick N we need). I've implemented randomization using elasticsearch's function score

PR #6315, opened 2024-12-16
fix: add the constant values for args

egdbear:
This PR adds constant values for args to avoid issues when there are no expected args.

Screenshot 2024-12-16 at 9 49 29 AM

✅ Added the `OPENSEARCH_API_BASE` and `OPENSEARCH_ARTWORKS_INFINITE_DISCOVERY_INDEX` env variablwes to staging.


nickskalkin (approved):
thanks!

PR #6319, opened 2024-12-16
feat: add backfill for search results

egdbear:
This PR adds backfill functionality for cases when there are fewer than 8 artworks. The requirement is always to have: - 8 artworks, plus 2 random picks (currently limited to curated collection).

### DIFF
--- a/src/schema/v2/infiniteDiscovery/discoverArtworks.ts
+++ b/src/schema/v2/infiniteDiscovery/discoverArtworks.ts
@@ -88,7 +88,7 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
     if (useOpenSearch) {
       const { excludeArtworkIds = [], likedArtworkIds = [] } = args
 
-      let result = []
+      let result: any = []
 
       if (likedArtworkIds.length < 3) {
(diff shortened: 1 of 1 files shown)


src/schema/v2/infiniteDiscovery/discoverArtworks.ts
```diff
@@ -105,10 +105,18 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
...
         result = await findSimilarArtworks(
           tasteProfileVector,
-          limit,
+          8,
+          excludeArtworkIds,
+          artworksLoader
+        )
+
+        // backfill with random curated picks if we don't have enough similar artworks
+        const randomArtworks = await getInitialArtworksSample(
```
nickskalkin, line 114: Maybe we can use `relatedArtworks` here? @egdbear

cavvia: Right, we're already using MLT with categorical similarity on OpenSearch (OS)... so I don't see why we would hit ES. The idea is to use a hybrid of visual and categorical similarity on the OS backend.

PR #6324, opened 2024-12-18
feat: add improvements for discoverArtworks query

egdbear:
This PR introduces two new parameters for discoverArtworks:

- **`useMltQuery`**  
  - When set to `true`, combines k-NN search (visual similarity) with the "More Like This" (MLT) content similarity search.  

- **`mltFields`**  
  - An array of fields that the MLT query will use to calculate similarity.  

**Visuals:**  

Example:

image

**Before** - knn only search 

Screenshot 2024-12-18 at 10 43 35 AM

**After** - mlt + knn search

Screenshot 2024-12-18 at 10 36 19 AM


egdbear:
I'm merging this and I'll follow-up if there are any major issues!

PR #6328, opened 2024-12-19
feat(dia-1038): infinite discovery - make initial arworks batch index configurable

nickskalkin:
@egdbear let me know what you think. I suppose we may want to experiment with different initial artwork datasets in future, and I think it will be easy if `discoverArtworks` query field has an option to specify which artworks collection to use. For now, I've indexed Most Loved, Street Art, and New this Week marketing collections to the `infinite_discovery_initial_artworks` index (zenith PR to come).

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -16812,6 +16812,9 @@ type Query {
     # (Only for when useOpenSearch is true) Exclude these artworks from the response
     excludeArtworkIds: [String]
     first: Int
+
+    # (Only for when useOpenSearch is true) Which index to use to display initial batch of artworks
+    initialArtworksIndexName: String = "infinite_discovery_initial_artworks"
     last: Int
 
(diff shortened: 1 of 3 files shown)


egdbear (approved):
This seems ✅ to me. n addition to a daily job for indexing only new artworks (I believe we can reuse the existing Zenith `infinite_discovery` job), we will also need a weekly job to update the initial collection for cold-start users. I think we can parameterize the existing job and have both jobs scheduled to run. /cc @cavvia

cavvia:
@nickskalkin Any reason why we need a separate index for these? Managing two indexes seems like more maintenance long-term and more complexity in our indexing code. These could just be collection id fields on the main artworks index, for example.

nickskalkin:
Yes, @cavvia , I agree with you. Initially, I wanted to have them indexed separately to allow switching between them easily for testing purposes, but now I doubt that we will need it. I’ll switch the implementation to use only one index. In any case, I’m porting this code to Gravity now.”

PR #6334, opened 2025-01-07
feat: update open search artworks index

egdbear:
This PR updates the latest index built using [the OpenSearch pipeline](https://github.com/artsy/zenith/tree/main/projects/infinite_discovery/src).


damassi:
Interesting re opensearch!

PR #6343, opened 2025-01-09
chore(EMI-2226): Add homeViewTasksSection to success response when clearing a task.

erikdstock:
Part of [EMI-2226].

This PR adds the `homeViewTasksSection` field to our success response type so that we can refresh the home view with an updated task list when dismissing or acknowledging a task. Eigen PR to consume this new field and optimistically update forthcoming.

Somewhat confusingly I realized that I added this field to the `taskOrError` field. It is adjacent to the task on the non-error type but the field name implies that it is one thing - a task. Not sure if there is a better way here.

cc @artsy/emerald-devs
paired with @MrSltun on some of it.

[EMI-2026]: https://artsyproduct.atlassian.net/browse/EMI-2226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EMI-2226]: https://artsyproduct.atlassian.net/browse/EMI-2226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/me/dismiss_task_mutation.ts
+++ b/src/schema/v2/me/dismiss_task_mutation.ts
@@ -28,10 +28,6 @@ const SuccessType = new GraphQLObjectType<any, ResolverContext>({
         return response
       },
     },
-    homeViewTasksSection: {
-      type: HomeViewTasksSectionType,
-      resolve: () => Tasks,
-    },
   }),
 })
 
@@ -70,6 +66,10 @@ export const dismissTaskMutation = mutationWithClientMutationId<
       type: new GraphQLNonNull(ResponseOrErrorType),
       description: "On success: the new state of the Task",
       resolve: (result) => result,
+    },
+    homeViewTasksSection: {
+      type: HomeViewTasksSectionType,
+      resolve: () => Tasks,
     },
   },
   mutateAndGetPayload: async ({ id }, { meDismissTaskLoader }) => {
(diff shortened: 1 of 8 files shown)


olerichter00 (commented):
Good idea 🌟 

An alternative could be to just return [homeView](https://github.com/artsy/metaphysics/blob/4f5e74ca13cd7972ac2094b49fe40dee2230d35b/src/schema/v2/schema.ts#L357) instead and then get the section from there in the client. Similar to how a lot of mutations return `me` ([example](https://github.com/artsy/metaphysics/blob/4f5e74ca13cd7972ac2094b49fe40dee2230d35b/src/schema/v2/me/updateMyPasswordMutation.ts#L38)). Maybe this simplifies the return type 🤔

src/schema/v2/me/dismiss_task_mutation.ts
```diff
@@ -74,7 +80,7 @@ export const dismissTaskMutation = mutationWithClientMutationId<
     try {
       const task: Task = await meDismissTaskLoader?.(id)
 
-      return { ...task, __typename: "SuccessType" }
+      return { ...task, _type: "SuccessType" }
```
MrSltun, line 83: **Question:** was this intentional 👀

MrSltun (approved):
Looks good! 🌟  I just left one question :)

src/schema/v2/me/dismiss_task_mutation.ts
erikdstock: Yes, if you look through how this file was written this is actually a throwaway value that can be anything but a magic 'gravity error' string

PR #6354, opened 2025-01-15
feat(ONYX-1479): add quick links section

dariakoko:
Addresses ONYX-1479

Adds QuickLinks section to the list of HomeView sections
image


src/schema/v2/homeView/sectionTypes/QuickLinks.ts
```diff
@@ -0,0 +1,50 @@
...
+    },
+  }),
+})
+
+/*
+ * A section type in the home view is specified declaratively
+ * as a GraphQL object that implements the HomeViewGenericSectionInterface
+ *
+ * Below we will configure its various fields.
+ */
```
olerichter00, line 34: 🧹 

```suggestion
```

src/schema/v2/homeView/sections/QuickLinks.ts
```diff
@@ -0,0 +1,32 @@
+// import { ContextModule, OwnerType } from "@artsy/cohesion"
+import { HomeViewSection } from "."
+import { HomeViewSectionTypeNames } from "../sectionTypes/names"
+
+export const QuickLinks: HomeViewSection = {
+  id: "home-view-section-quick-links",
+  // TODO: add ContextModule quickLinks
```
olerichter00, line 7: 👍

olerichter00 (dismissed):
Looks great 🌟

src/schema/v2/homeView/sectionTypes/QuickLinks.ts
```diff
@@ -0,0 +1,51 @@
...
+      type: new GraphQLNonNull(GraphQLString),
+      description: "Quick link URL",
+    },
+    ownerType: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: "The context module for analytics",
+    },
+    contextScreenOwnerId: {
+      type: GraphQLString,
+      description: "The owner ID for the context module",
```
MounirDhahri, line 31: @anandaroop @dblandin what do you think about this?

src/schema/v2/homeView/sectionTypes/QuickLinks.ts
```diff
@@ -0,0 +1,51 @@
...
+>({
+  name: HomeViewSectionTypeNames.HomeViewSectionQuickLinks,
+  description: "A selection of quick links in the home view",
+  interfaces: [HomeViewGenericSectionInterface, NodeInterface],
+  fields: {
+    ...standardSectionFields,
+    quickLinks: {
+      type: new GraphQLList(QuickLinkType),
+      resolve: (parent, ...rest) =>
+        parent.resolver ? parent.resolver(parent, ...rest) : "",
```
MounirDhahri, line 48: We forgot to change this back to 
```suggestion
        parent.resolver ? parent.resolver(parent, ...rest) : [],
```

src/schema/v2/homeView/sectionTypes/QuickLinks.ts
```diff
@@ -0,0 +1,51 @@
...
+  any,
+  ResolverContext
+>({
+  name: HomeViewSectionTypeNames.HomeViewSectionQuickLinks,
+  description: "A selection of quick links in the home view",
+  interfaces: [HomeViewGenericSectionInterface, NodeInterface],
+  fields: {
+    ...standardSectionFields,
+    quickLinks: {
+      type: new GraphQLList(QuickLinkType),
```
MounirDhahri, line 46: we can even make this safer
```suggestion
      type: new GraphQLNonNull(new GraphQLList(QuickLinkType)),
```

MounirDhahri (dismissed):

src/schema/v2/homeView/sectionTypes/QuickLinks.ts
dariakoko: yeah right

src/schema/v2/homeView/sectionTypes/QuickLinks.ts
dblandin: What is this used for? The only specific value I see currently is "/collect?price_range=*-1000" which doesn't read as an ID to me.

Is this needed value need for the analytics support?

dblandin (commented):
Looking good!

Some high-level feedback:

I like "QuickLinks" as the name of the specific _section instance_ but I wonder if we can abstract the _section type_ a bit more. Maybe "NavigationPills"?

That way, we can potentially have more than one instance of this section type in the schema. For example, maybe later on we have a need for two different section instances (e.g. "ForYouQuickLinks" and "ExploreQuickLinks") and each can be backed by the same "NavigationPills" section type.

What do you think?

src/schema/v2/homeView/sectionTypes/QuickLinks.ts
MounirDhahri: yes, that's what we have in mind. We can also use the href as a payload for the tracking event. it's not clear so far.

src/schema/v2/homeView/sections/NavigationPills.ts
```diff
@@ -2,25 +2,25 @@ import { ContextModule, OwnerType } from "@artsy/cohesion"
 import { HomeViewSection } from "."
 import { HomeViewSectionTypeNames } from "../sectionTypes/names"
 
-export const QuickLinks: HomeViewSection = {
+export const NavigationPills: HomeViewSection = {
```
MounirDhahri, line 5: I believe that we don't need to rename this. Renaming the section type should be enough

src/schema/v2/homeView/sectionTypes/QuickLinks.ts
dblandin: To capture the `href` value, you could use `destination_path` which we've included in similar events.

https://github.com/artsy/cohesion/pull/525

src/schema/v2/homeView/sections/NavigationPills.ts
dblandin: Cross-linking this comment: https://github.com/artsy/cohesion/pull/548#issuecomment-2595851381

src/schema/v2/homeView/sectionTypes/NavigationPills.ts
```diff
@@ -0,0 +1,51 @@
...
+} from "graphql"
+import { ResolverContext } from "types/graphql"
+import { NodeInterface } from "../../object_identification"
+import { HomeViewGenericSectionInterface } from "./GenericSectionInterface"
+import { HomeViewSectionTypeNames } from "./names"
+import { standardSectionFields } from "./GenericSectionInterface"
+import { NavigationPill } from "../sections/NavigationPills"
+
+const QuickLinkType = new GraphQLObjectType<NavigationPill, ResolverContext>({
+  name: "NavigationPill",
```
anandaroop, line 15: **question:** What do you think about bringing this naming into sync? Can we rename this `QuickLinkType` → `NavigationPillType`?

(This also relates to the comments elsewhere about distinguishing the naming of the section type and the section instance.)

src/schema/v2/homeView/sections/NavigationPills.ts
anandaroop: Yeah, I think `NavigationPills` is a good name for the section _type_ — but for this section _instance_ I like the idea of calling it something else like `QuickLinks`. That way we decouple the section type from the specific, concrete section that uses it, as we have for the previous sections mentioned in Devon's linked comment.

This is of greater value for section types that can be re-used and I'm not sure if this one _will_ be, but I think this is a good default for us to follow.

src/schema/v2/homeView/sections/NavigationPills.ts
```diff
@@ -0,0 +1,39 @@
+import { ContextModule, OwnerType } from "@artsy/cohesion"
+import { HomeViewSection } from "."
+import { HomeViewSectionTypeNames } from "../sectionTypes/names"
+
+export const NavigationPills: HomeViewSection = {
+  id: "home-view-section-navigation-pills",
+  contextModule: ContextModule.quickLinks,
+  ownerType: OwnerType.quickLinks,
```
anandaroop, line 8: **question:** This `ownerType` optional attribute is [intended for a section's view-all screen](https://github.com/artsy/metaphysics/blob/b8a4d4c2e438395de8e33eff3a0c65c7a7080840/src/schema/v2/homeView/sections/_ExampleSection.ts#L68-L87) — I think that's not actually relevant or required here, right?

src/schema/v2/homeView/sections/NavigationPills.ts
```diff
@@ -0,0 +1,39 @@
...
+  type: HomeViewSectionTypeNames.HomeViewSectionNavigationPills,
+  requiresAuthentication: true,
+  resolver: () => {
+    return QUICK_LINKS
+  },
+}
+
+export interface NavigationPill {
```
anandaroop, line 16: **question:** I wonder if it makes more sense for this interface to live over in the `NavigationPill` section type and be imported here? This interface is very coupled to that section type's response shape.

anandaroop (commented):
Looking good! 

I've just added some thoughts about naming and organization, similar to others' comments.

I'd be in favor of de-coupling the section type (`NavigationPills`) from the specific section instance (`QuickLinks`?) even if there isn't an immediately obvious opportunity for re-use of the section type.

The type:

```ts
// sectionTypes/NavigationPills.ts

export interface NavigationPill {
  …
}

const NavigationPillType = new GraphQLObjectType<NavigationPill, …>({
  name: "NavigationPill",
  ...
})

export const HomeViewNavigationPillsSectionType = new GraphQLObjectType({
  name: HomeViewSectionTypeNames.HomeViewSectionNavigationPills,
  description: "A selection of navigational pills in the home view",
  ...
})
```

The instance:

```ts
// sections/QuickLinks.ts

export const QuickLinks: HomeViewSection = {
  id: "home-view-section-quick-links",
  …
}


```

src/schema/v2/homeView/sectionTypes/NavigationPills.ts
```diff
@@ -0,0 +1,47 @@
...
+export const HomeViewNavigationPillsSectionType = new GraphQLObjectType<
+  any,
+  ResolverContext
+>({
+  name: HomeViewSectionTypeNames.HomeViewSectionNavigationPills,
+  description: "A selection of quick links in the home view",
+  interfaces: [HomeViewGenericSectionInterface, NodeInterface],
+  fields: {
+    ...standardSectionFields,
+    quickLinks: {
```
anandaroop, line 41: **suggestion:** I think it would still be best to have consistent naming here, i.e. `HomeViewSectionNavigationPills` would expose an attribute called `navigationPills` or `pills` (but not ~quickLinks~ since that's the naming associated with a specific _instance_ of this `HomeViewSectionNavigationPills` type)

anandaroop (commented):
Whoops I didn't get a chance to comment again before it was merged, but I do have one suggestion below. 

It should still be early enough for this not to a breaking change in Eigen, correct?

src/schema/v2/homeView/sectionTypes/NavigationPills.ts
anandaroop: I did a [quick PR here](https://github.com/artsy/metaphysics/pull/6405) to show what I mean.

PR #6355, opened 2025-01-15
fix: infinite discovery - overfetch artworks to avoid blank results, extract initial artworks index name to an env var

nickskalkin:
- over-fetch artworks to avoid < 5 artworks displayed (happens when an artwork gets unpublished, but still in index)
- extract initial artworks index name to an ENV variable (because I want to point to staging index from my local)


src/lib/infiniteDiscovery/findSimilarArtworks.ts
```diff
@@ -10,6 +10,9 @@ interface FindSimilarArtworksOptions {
   weights?: number[]
 }
 
+const DEFAULT_FETCH_SIZE = 5
+const OVERFETCH = 10
```
egdbear, line 14: **suggestion:** This is not an issue, but I’m curious if the naming can be improved:
- `DEFAULT_FETCH_SIZE`: Perhaps `DEFAULT_BATCH_SIZE` might add more clarity, as "batch" better explains the context here.
- `OVERFETCH`: - This name doesn’t clearly emphasize its purpose. Maybe consider a name that reflects its intent, like `EXTRA_FETCH_SIZE` or `FETCH_BUFFER`.

egdbear (approved):
I have some naming concerns but overall 🚀 🚀 🚀

nickskalkin:
thanks @egdbear addressed your feedback, gonna merge

PR #6359, opened 2025-01-16
chore: limit the amount of recommendations

egdbear:
This PR adds hardcodes the limit for infinite discovery recommendations


iskounen (approved):
🙌

PR #6367, opened 2025-01-22
chore: consistent artwork money fields resolver

erikdstock:
Downstream from some hackathon work... This PR updates our artwork and editionSet fields to use a consistent resolver for the money type, avoiding missing display values.

In order to preserve some behavior I clobbered that display value on one field where we are using the gravity `price_display` json value instead. There might be other cases where we want to verify that nothing is changing drastically - for example, are we querying any `display` fields where the format is hardcoded.

cc @artsy/emerald-devs

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -568,7 +568,7 @@ describe("Artwork type", () => {
         expect(data).toEqual({
           artwork: {
             pricePaid: {
-              display: "$210",
+              display: "US$210",
               currencyCode: "USD",
             },
           },
@@ -654,7 +654,7 @@ describe("Artwork type", () => {
               maxPrice: {
                 minor: 42000,
                 major: 420,
-                display: null,
+                display: "US$420",
                 currencyCode: "USD",
               },
             },
@@ -676,7 +676,7 @@ describe("Artwork type", () => {
               minPrice: {
                 minor: 42000,
                 major: 420,
-                display: null,
+                display: "US$420",
                 currencyCode: "USD",
               },
               maxPrice: null,
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -568,7 +568,7 @@ describe("Artwork type", () => {
         expect(data).toEqual({
           artwork: {
             pricePaid: {
-              display: "US$210",
+              display: "$210",
               currencyCode: "USD",
             },
           },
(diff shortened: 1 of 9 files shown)


src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -676,7 +676,7 @@ describe("Artwork type", () => {
               minPrice: {
                 minor: 42000,
                 major: 420,
-                display: null,
+                display: "US$420",
```
oxaudo, line 679: ❤️

src/schema/v2/artwork/index.ts
```diff
@@ -1542,18 +1571,27 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         type: Money,
         description:
           "The price paid for the artwork in a user's 'my collection'",
-        resolve: (artwork) => {
-          const { price_paid_cents } = artwork
+        resolve: async (artwork, args, ctx, info) => {
+          const { price_paid_cents, price_paid_currency = "USD" } = artwork
```
oxaudo, line 1575: Slightly surprised that we ask collectors to enter price in cents, but I guess it was here already.

src/schema/v2/artwork/index.ts
```diff
@@ -1542,18 +1571,27 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+            (await resolvePriceAndCurrencyFieldsToMoney(
+              { minor: price_paid_cents, currencyCode: price_paid_currency },
+              args,
+              ctx,
+              info
+            )) || {}
+          return (
+            moneyFields && {
+              ...moneyFields,
+              // TODO: Display field legacy implementation maintained until we verify
```
oxaudo, line 1587: Should we create a separate card for this?

oxaudo (approved):
This looks great to me. Thanks @erikdstock !

erikdstock:
Now wondering if I should use the `amount()` [helper](https://github.com/artsy/metaphysics/blob/d4007f9fe93c02dad4e6a5eb56701aa2723086f9/src/schema/v2/fields/money.ts#L31-L85) to let money display values be formatted better...

oxaudo:
> Now wondering if I should use the `amount()` [helper](https://github.com/artsy/metaphysics/blob/d4007f9fe93c02dad4e6a5eb56701aa2723086f9/src/schema/v2/fields/money.ts#L31-L85) to let money display values be formatted better...

Good point @erikdstock. Think we should generate the same string as get's added to SippingInfo below in the same file: https://github.com/artsy/metaphysics/pull/6367/files#diff-54b84e0c164089fc202e3439d8ded89ab4925442fb613f4e36a9c22acdb66cb8R1455

src/schema/v2/artwork/index.ts
erikdstock: It is probably calculated on the artwork from some input using the money helper gem

erikdstock:
I'm closing this and making a note to return because changes in #6714 will make this work a little easier in the future.

PR #6381, opened 2025-01-23
fix: throw exception for case with no curated artworks

egdbear:
This PR introduces an error handler for cases where no curated artworks are available.  In the future, we'll refine the error message, and it will comply with `ErrorType` once it becomes available from Gravity.


iskounen (approved):
Thank you 🎉

egdbear:
du bist willkomen @iskounen

PR #6383, opened 2025-01-23
feat: update shippingInfo logic

oxaudo:
This will update the actual shipping display to what is covered by this diagram:

Screenshot 2025-01-23 at 5 26 13 PM

I did update the specs to follow this diagram. So fairly confident in the logic here. It can be refactored I think but will do for now.

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -1397,7 +1397,7 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           }
 
           let domesticLine = ""
-          let internationalLine = "Unavailable"
+          let internationalLine = "Contact gallery"
 
           const fullCountryName = artwork.shipping_origin
             ? COUNTRIES[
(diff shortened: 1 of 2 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -1389,76 +1389,58 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
...
+                international_shipping_fee_cents || null
+            ).resolve(artwork, {
+              precision: 0,
+              symbol: symbolFromCurrencyCode(artwork.price_currency),
+            })
+            internationalLine = `${formattedPrice}`
+          }
 
-          return `Shipping: ${domesticShipping} ${shippingRegion}, ${internationalShipping} rest of world`
+          return `Domestic: ${domesticLine} \nInternational: ${internationalLine}`
```
egdbear, line 1443: **suggestion:** I'm not sure if this will format the way we want on the client. Maybe we could try using the [markdown helper](https://github.com/artsy/metaphysics/blob/33786b0710971be0ae73479ec5b7e651277d9122/src/schema/v2/fields/markdown.ts#L38) to make sure it formats as expected? 😊

oxaudo: I was checking and it seems to work. But def need to verify once this is merged. But yea - could also look at adding this with `markdown`.

erikdstock (approved):

PR #6389, opened 2025-01-27
feat(dia-1069): use gravity instead of opensearch for infinite discovery

nickskalkin:
[Jira](https://artsyproduct.atlassian.net/browse/DIA-1069)

> [!IMPORTANT]
> Do not merge until https://github.com/artsy/gravity/pull/18497 is deployed on staging

Previously, we were querying OpenSearch directly—we did it only to save some time to build a POC. There is an open [PR](https://github.com/artsy/gravity/pull/18497) in Gravity that moves all logic from Metaphysics.


egdbear (dismissed):

nickskalkin:
@egdbear should be reviewable now

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
```diff
@@ -9,15 +9,11 @@ import { ResolverContext } from "types/graphql"
...
-import { getInitialArtworksSample } from "lib/infiniteDiscovery/getInitialArtworksSample"
-import { calculateMeanArtworksVector } from "lib/infiniteDiscovery/calculateMeanArtworksVector"
-import { findSimilarArtworks } from "lib/infiniteDiscovery/findSimilarArtworks"
-import config from "config"
 
 export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
   type: artworkConnection.connectionType,
   args: pageable({
-    limit: { type: GraphQLInt },
+    limit: { type: GraphQLInt, defaultValue: 5 },
```
egdbear, line 16: 👍

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
```diff
@@ -39,71 +35,24 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
...
-        initialArtworksIndexName
-      )
-      result.push(...randomArtworks)
-    }
+    const gravityResponse = await artworksDiscoveryLoader(gravityArgs)
 
-    return connectionFromArray(result, args)
+    return connectionFromArray(gravityResponse, {
+      first: args.limit,
+    })
```
egdbear, line 56: **issue (non-blocking):** We should handle and return an appropriate error to clients when no curated artworks are available due to a user having seen all curated artworks any saves. This can happen, and we should address it gracefully. Possibly in Gravity [here](https://github.com/artsy/gravity/blob/31fc48f446b1d1e7b64e276264a558a24e694e6f/app/services/artworks_discovery/curated_artworks.rb#L2)

egdbear (approved):
🚀

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
nickskalkin: I would put such error message on client side, but let's discuss

PR #6391, opened 2025-01-28
fix(markdown): ensures internal hardcoded http links are https for seo purposes

dzucconi:
While we currently redirect HTTP to HTTPS in production, our SEO reports flagged hard-coded HTTP links as an issue. This change automatically updates HTTP links found in markdown inputs to HTTPS, addressing all likely instances where these might appear.

### DIFF
--- a/src/schema/v2/fields/__tests__/markdown.test.ts
+++ b/src/schema/v2/fields/__tests__/markdown.test.ts
@@ -1,4 +1,8 @@
-import { formatMarkdownValue, markdown } from "../markdown"
+import {
+  enforceInternalHTTPS,
+  formatMarkdownValue,
+  markdown,
+} from "../markdown"
 
 describe("formatMarkdownValue", () => {
   it("formats markdown as html", () => {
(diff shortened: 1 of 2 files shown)


joeyAghion (commented):
This is OK-ish but ideally any automatic transformation would happen at the back-end, where it could apply to other surfaces like app and maybe only have to run once instead of on each read.

dzucconi:
@joeyAghion I mean, this is the backend? I do think this is the most appropriate place to catch it because it's just anything that touches Markdown. 

However, I think we might be able to get away with a one-time task that just searches for these URLs throughout the database and updates them with the assumption that we're pretty much never going to see that happen again unless somebody manually types "http". The tricky part there is just the legwork to find all of the Markdown fields that it might be in and then managing to efficiently query them. Though we can do most of this just manually, and it certainly doesn't need to be fast.

joeyAghion:
This is... mid-end (?). Although you're right that this shared location can be leveraged by the mobile app as well.

PR #6405, opened 2025-02-03
refactor(homeView): naming within new quick link navigation pills

anandaroop:
To elaborate on my comments in https://github.com/artsy/metaphysics/pull/6354 I'm showing what I mean by separating the concepts of section instance and section type.

Just as…
- the `Artists` section type exposes a list of `artists`
-  the `Cards` section type exposes a list of `cards`

I propose that…
- `NavigationPills` section type exposes a list of `navigationPills`

This decouples the `NavigationPills` naming from the specific instance that makes use of this section type (`QuickLinks`). 

Just like the `Cards` naming is decoupled from the specific instances of `DiscoverSomethingNew` and `ExploreByCategory`.

(This may end up being a lot of fuss for nothing if `NavigationPills` never enjoys any re-use, but I still think this is the right default mindset for naming in the home view)

### DIFF
--- a/src/schema/v2/homeView/sections/__tests__/QuickLinks.test.ts
+++ b/src/schema/v2/homeView/sections/__tests__/QuickLinks.test.ts
@@ -12,7 +12,7 @@ describe("QuickLinks", () => {
             contextModule
             ownerType
             ... on HomeViewSectionNavigationPills {
-              quickLinks {
+              navigationPills {
                 title
                 href
                 ownerType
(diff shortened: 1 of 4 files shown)


src/schema/v2/homeView/sections/__tests__/QuickLinks.test.ts
```diff
@@ -12,7 +12,7 @@ describe("QuickLinks", () => {
             contextModule
             ownerType
             ... on HomeViewSectionNavigationPills {
-              quickLinks {
+              navigationPills {
```
MounirDhahri, line 15: totally makes sense!

MounirDhahri (approved):

PR #6412, opened 2025-02-05
chore: add artwork framed dimensions and weight.

erikdstock:
The type of this pr is **chore**
Part of [EMI-2278]

This PR adds the following properties to our artwork type for use in our client-side shipping estimates widget:
`depthCm, diameterCm, shippingWeight, shippingWeightMetric, framedDiameter`

It depends on artsy/gravity#18560 for some of the fields.

Some related adjacent fields (like `framedHeight` and `framedWidth` are typed as GraphQL strings instead of floats. I'd argue we should convert these, which would be a breaking schema change, or re-implement these fields to follow the `_Cm` naming convention of our non-framed dimensions fields.

[EMI-2278]: https://artsyproduct.atlassian.net/browse/EMI-2278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2439,7 +2439,9 @@ type Artwork implements Node & Searchable & Sellable {
 
   # The depth as expressed by the original input metric
   depth: String
+  depthCm: Float
   description(format: Format): String
+  diameterCm: Float
   dimensions: dimensions
   displayArtistBio: Boolean
   displayLabel: String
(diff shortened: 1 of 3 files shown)


_schemaV2.graphql
```diff
@@ -2474,7 +2476,10 @@ type Artwork implements Node & Searchable & Sellable {
   framed: ArtworkInfoRow
     @deprecated(reason: "Consider using isFramed field (boolean) instead")
   framedDepth: String
+  framedDiameter: Float
```
oxaudo, line 2479: I'm confused by why this `framedDiameter` is Float and the above `framedDepth` is String? They both defined as Floats in gravity. We should probably keep it consistent one way or the other?

oxaudo (approved):
This looks good to me.

I don't really understand the type difference on some fields. Also wonder why we are doing `_cm` for the regular w/h/d but not doing this for the framed stuff. Maybe instead of passing Metrics around we should convert to `_cm` for framed as well?

PR #6413, opened 2025-02-06
chore: remove active bids home view section

MounirDhahri:
Jira ticket: https://artsyproduct.atlassian.net/browse/ONYX-1515

This PR removes active bids from the home view


olerichter00 (approved):
🧹 🧹 🧹

PR #6418, opened 2025-02-07
feat: do not allow submissions from any users

MounirDhahri:
**Description:**

This PR is a temporary solution to disable submissions on both app and web. It will allow us to not receive any additional submissions

**Note:**
- This PR should be reverted once we remove the SWA experience from both App and Web.

**App**

https://github.com/user-attachments/assets/becfc40f-897e-4aa5-abb1-2aac5a7c464e

**Web**

https://github.com/user-attachments/assets/6abf0cf6-f3a0-40af-94eb-d2e27f60d84d

### DIFF
--- a/src/schema/v2/artist/targetSupply/__tests__/index.test.ts
+++ b/src/schema/v2/artist/targetSupply/__tests__/index.test.ts
@@ -16,12 +16,12 @@ describe("ArtistTargetSupply", () => {
         artistLoader: () => {
           return Promise.resolve({
             id: "andy-warhol",
-            target_supply: true,
+            target_supply: false,
           })
         },
       }
       const response = await runQuery(query, context)
-      expect(response.artist.targetSupply.isTargetSupply).toEqual(true)
+      expect(response.artist.targetSupply.isTargetSupply).toEqual(false)
     })
   })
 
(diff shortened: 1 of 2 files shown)


anandaroop:
Because the "target supply" concept has a meaning beyond secondary market sales, this will have a [wider effect](https://github.com/search?q=org%3Aartsy%20isTargetSupply&type=code) than just SWA. 

If it's _very_ temporary until a more targeted solution can be delivered, I think it's ok. But I do think it would be better to come up with a more targeted fix.

MounirDhahri:
That's a good call. @anandaroop those side effects are intentional in this case because it will mean that we won't be showing "submit this artwork" or "interested in selling" banners.

MounirDhahri:
@joeyAghion that's a safer route that I can also take. I think anyway the users will be frustrated, so I will update to use your suggested approach

anandaroop:
Yep, I assumed that was the case.

I'm thinking more of side effects like [this](https://github.com/artsy/forque/blob/8db7e389b38d6992c94df87205d6f6588723c80a/src/pages/artists/dedupe/components/manager/SelectBadRecords.tsx#L40) — where deleting an artist record is, by design, prevented when that record represents a target supply artist.

Maybe a quick chat is on order to discuss options?

MounirDhahri:
We decided over zoom that we would make the change from Convection and close this one.

PR #6433, opened 2025-02-13
chore: Allow artworks to be updated in bulk (Batch Edits)

lidimayra:
[AMBER-1272]

Pairing with @jpotts244

Kick-off the MP mutation for triggering Bulk Update operations (Batch Edits)

Example:
```graphql
mutation bulkUpdateArtworksMetadata(
  $id: String!
  $metadata: BulkUpdateArtworksMetadataInput!
) {
  bulkUpdateArtworksMetadata(
    input: {
      id: $id
      metadata: $metadata
    }
  ) {
    __typename
    bulkUpdateArtworksMetadataOrError {
    __typename
    ... on BulkUpdateArtworksMetadataMutationSuccess {
	updatedPartnerArtworks {
          ids
       }
     }
      ... on BulkUpdateArtworksMetadataMutationFailure {
        mutationError {
          message
        }
      }
    }
  }
}
```

Variables used on local tests:
```json
{
  "id": "almine-rech",
  "metadata": {
    "location_id": "67acd02ac4a53e0007b51cc0",
    "category": "foobar"
  }
}
```

_Intentionally using a random value for category given we have no validations for this field_

[AMBER-1272]: https://artsyproduct.atlassian.net/browse/AMBER-1272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/schema.ts
+++ b/src/schema/v2/schema.ts
@@ -144,6 +144,7 @@ import { updateArtworkMutation } from "./artwork/updateArtworkMutation"
 import { artworksForUser } from "./artworksForUser"
 import { authenticationStatus } from "./authenticationStatus"
 import { BankAccount } from "./bank_account"
+import { bulkUpdateArtworksMetadataMutation } from "./partner/BulkOperation/bulkUpdateArtworksMetadataMutation"
 import { bulkUpdatePartnerArtworksMutation } from "./bulkUpdatePartnerArtworksMutation"
 import { CollectorProfileForUser } from "./CollectorProfile/collectorProfile"
 import { CollectorProfilesConnection } from "./CollectorProfile/collectorProfiles"
@@ -445,6 +446,7 @@ export default new GraphQLSchema({
       adminUpdateFeatureFlag: updateFeatureFlagMutation,
       artworksCollectionsBatchUpdate: artworksCollectionsBatchUpdateMutation,
       assignArtworkImportArtist: AssignArtworkImportArtistMutation,
+      bulkUpdateArtworksMetadata: bulkUpdateArtworksMetadataMutation,
       bulkUpdatePartnerArtworks: bulkUpdatePartnerArtworksMutation,
       commerceOptIn: commerceOptInMutation,
       commerceOptInReport: commerceOptInReportMutation,
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -4149,6 +4149,11 @@ type BidderPositionSuggestedNextBid {
 # exceed the size of a 32-bit integer, it's encoded as a string.
 scalar BigInt
 
+input BulkUpdatePartnerArtworksMetadataInput {
+  # The partner location ID to assign
+  locationId: String
+}
+
 type BulkUpdatePartnerArtworksMutationFailure {
   mutationError: GravityMutationError
 }
@@ -4164,8 +4169,8 @@ input BulkUpdatePartnerArtworksMutationInput {
   # ID of the partner
   id: String!
 
-  # The partner location ID to assign
-  location: String
+  # Metadata to be updated
+  metadata: BulkUpdatePartnerArtworksMetadataInput
 }
 
 type BulkUpdatePartnerArtworksMutationPayload {
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -4149,9 +4149,42 @@ type BidderPositionSuggestedNextBid {
 # exceed the size of a 32-bit integer, it's encoded as a string.
 scalar BigInt
 
-input BulkUpdatePartnerArtworksMetadataInput {
+input BulkUpdateArtworksMetadataInput {
   # The partner location ID to assign
   location_id: String
+}
+
+type BulkUpdateArtworksMetadataMutationFailure {
+  mutationError: GravityMutationError
+}
+
+input BulkUpdateArtworksMetadataMutationInput {
+  clientMutationId: String
+
+  # ID of the partner
+  id: String!
+
+  # Metadata to be updated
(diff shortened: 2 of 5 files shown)


src/schema/v2/schema.ts
```diff
@@ -443,6 +444,7 @@ export default new GraphQLSchema({
       adminUpdateFeatureFlag: updateFeatureFlagMutation,
       artworksCollectionsBatchUpdate: artworksCollectionsBatchUpdateMutation,
       assignArtworkImportArtist: AssignArtworkImportArtistMutation,
+      bulkUpdateArtworksMetadata: bulkUpdateArtworksMetadataMutation,
```
damassi, line 447: Wondering thoughts on the term `metadata` here (and throughout) -- why not just `bulkUpdateArtworks` and keep it simple?

And then in the mutation, we can keep it pretty standard api-wise with a traditional `import` variable:

```tsx
bulkUpdateArtworks(input: $input) {
  ... 
}
```

(When i think 'metadata' i think of some subset of data around a specific artwork, vs artwork fields themselves)

thoughts?

lidimayra: Yes, that's where we've been struggling a bit.

The thing is: we already know this endpoint/mutation won't be a fit for 100% of the scenarios.
We want to use it for the majority of the cases, mostly when it comes to metadata that's considered low complexity on the fields correlation list. Some examples: location, medium type, listed price, classification, etc.

However, by the time we move to more complex fields (listing options, shipping options, etc) we'll need additional endpoints/mutations more specific for those operations (those will require way more handling in Gravity and it will accept a very specific subset of acceptable params).

The `metadata` suffix here would be an attempt to indicate this mutation will be used for "generic" metadata. In parallel, if one wants to update a list of artworks to BNMO, we'd probably need a different one.

damassi: Ah that makes sense 👍 its a 'slot' for those things we know we can update now, but which are different from other key artwork fields. 

One thought (though not necessarily recommending it) would be to have conditions inside to keep things transparent. Eg:

```tsx
const { foo, bar, baz, ...metadata } = args 

if (foo) {
  ...
}
if (bar) {
  ...
}
if (baz) {
  ...
}
if (metadata) {
  await updateSomeOtherField(...)
}
```
or even lift this up to a service endpoint that distinguishes between each at the Gravity API level, and then we can call 

```
api/v1/artworks/bulk_update
```

and delegate from there. 

That way, in the future, there's a one-stop-shop for managing this stuff, without a lot of need from the dev to understand internals -- those are all managed internal to the endpoint itself. For complex workflows like this this approach, I think, will be less cumbersome to manage.

A similar example is the filterArtworks ES endpoint, via MP. The internals there are pretty complex, but from a consumer perspective (devs) all we need to know is that that one field exists, and everything we could possibly need can be gathered from there, without needing to orchestrate a number of different behaviors. 

(Def non-blocking! just food for thought)

damassi: The other thought is around having a lot of mutations that are _specific_ to the bulk artworks update flow. Q: would this mutation ever be used outside of this context? If yes, one can imagine it being left around; if no, then this mutation will then need to be coordinated _on on the client, along with the other mutations_ as mutations are typically client/consuming app centric. 

That then means its less portable should the app change in the future, or we need to invoke said mutation from, say, folio; requiring a copy of client-side logic.

lidimayra: Those are very good points for us to discuss. Especially interested in what would be the best approach when it comes to the mutation usage on the clients.

Example:
If an artwork is submitted for a regular bulk update, it could have any values for fields like availability, price display, currency and so on. But if an artwork is submitted for an update to turn it into BN, it is much more restricted.

The initial thought was that by providing specific mutations for the special scenarios, the expectation is that clients wouldn't need to know which params can be passed, the schema should provide that right away. On the other hand, I felt like a unified approach could end up requiring more work on the client-side when calling the mutation. Also thought that we'd prefer to keep our mutations 1:1 with the loaders/endpoints for the sake of clarity. But I'm not that sure now if that's really the best call.

damassi: Totally. In general, good to keep in mind that GraphQL is an API gateway -- its meant to coordinate multiple api responses. Sometimes it makes sense to keep things 1:1 with the underlying api, as in an example like:

```
batchUpdateArtworksMutation -> api/v1/artworks/batch_update
```

In many other cases (most perhaps) its a coordinating layer, so that one can design **ideal** APIs for client consumers. When things are complex like this this is where GraphQL shines. 
 
But I'll leave this up to y'all 👍

lidimayra: I created https://artsyproduct.atlassian.net/browse/AMBER-1300 so we can keep track of this discussion on a higher level

jpotts244: thanks for ticketing Lidi! and for comments Chris!


Yeah would love to get thoughts on schema design, I have a WIP cleanup pass coming in hot which might expose some of these discussions (we have the artsy shipping bulk, commerce opt in bulk, and metadata stuff bulk how do we design MP layer etc etc). I'm looking to consolidate some of that in the next PR!

lidimayra: @jpotts244 Love this! 😻

damassi: Down with that 👍

damassi (approved):
👍 and noted about iteration! If things might switch around a bit we may want to hold off on tests a little bit until we're sure things are stable; at that point we'll want to add (lots of copy / paste examples around for this mutation style)

PR #6443, opened 2025-02-18
feat: add user seen artworks mutation

egdbear:
This PR introduces a new mutation, createUserSeenArtwork, which will be used to track seen artworks for Infinite Discovery. [Gravity PR](https://github.com/artsy/gravity/pull/18598) that introduced the endpoint for tracking seen artworks.

Ticket [DIA-1155](https://artsyproduct.atlassian.net/browse/DIA-1155)

Screenshot 2025-02-19 at 2 17 08 PM

[DIA-1155]: https://artsyproduct.atlassian.net/browse/DIA-1155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/infiniteDiscovery/createUserSeenArtworkMutation.ts
+++ b/src/schema/v2/infiniteDiscovery/createUserSeenArtworkMutation.ts
@@ -57,7 +57,8 @@ export const createUserSeenArtworkMutation = mutationWithClientMutationId<
   ResolverContext
 >({
   name: "CreateUserSeenArtwork",
-  description: "Creates User Seen Artwork.",
+  description:
+    "Marks an artwork as seen when a user swipes through Infinite Discovery.",
   inputFields,
   outputFields: {
     userSeenArtworkOrError: {
--- a/src/schema/v2/infiniteDiscovery/createUserSeenArtworkMutation.ts
+++ b/src/schema/v2/infiniteDiscovery/createUserSeenArtworkMutation.ts
@@ -1,0 +1,97 @@
+import { mutationWithClientMutationId } from "graphql-relay"
+import {
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLString,
+  GraphQLUnionType,
+} from "graphql"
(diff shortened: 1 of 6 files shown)


src/schema/v2/infiniteDiscovery/createUserSeenArtworkMutation.ts
```diff
@@ -0,0 +1,97 @@
...
+  types: [SuccessType, FailureType],
+})
+
+export const createUserSeenArtworkMutation = mutationWithClientMutationId<
+  Input,
+  any,
+  ResolverContext
+>({
+  name: "CreateUserSeenArtwork",
+  description: "Creates User Seen Artwork.",
```
nickskalkin, line 60: **nitpick**: I would add more human description, like "Marks an artwork as seen when a user swipes through Infinite Discovery"

nickskalkin (approved):
🚀

src/schema/v2/infiniteDiscovery/discoverArtworks.ts
```diff
@@ -18,6 +19,11 @@ export const DiscoverArtworks: GraphQLFieldConfig<void, ResolverContext> = {
       type: new GraphQLList(GraphQLString),
       description: "Exclude these artworks from the response",
     },
+    useInternalTracking: {
```
nickskalkin, line 22: @egdbear sudden idea — let's instead of this param use an ENV var in line number 61? This way backend will be able to control behavior without frontend changes

egdbear: Do you think that is necessary if we add this change to Eigen?

joeyAghion: (Way-late comment:) I agree, it would be best to avoid leaking this decision or details about redis to the client. The back-end should decide and I think we can even deprecate the `excludeArtworkIds` arg (shortly).

PR #6454, opened 2025-02-20
feat: env variable for infinite discovery use internal tracking of seen artworks

nickskalkin:
Allows dynamically control behaviour of `discoverArtworks` endpoint—this should give us more flexibility when switching from storage-based tracking of seen artworks to backend-based tracking.


egdbear (approved):

egdbear:
Do we need to introduce this interim environment variable for now, or can we just connect with the new seen artworks API?

nickskalkin:
@egdbear I think it's worth introducing it (it costs nothing, but I will give us some control to switch between not-tested backend approach and the tested one). Once we have some confidence in the backend approach—we can get rid of it, as well as from `use_internal_tracking` param on Gravity side

PR #6465, opened 2025-02-24
feat(homeView): allow QuickLinks to enforce minimum Eigen version requirements

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-1558

The approach here is quite similar to https://github.com/artsy/metaphysics/pull/6297 and makes use of the versioning helpers already introduced in that PR.

QuickLinks now can individually set a minimum required Eigen version:

```diff
{
  title: "Featured Fairs",
  href: "/featured-fairs",
  ownerType: OwnerType.fairs,
  icon: "FairIcon",
+ minimumEigenVersion: { major: 42, minor: 0, patch: 0 },
},
```

That one-line change should be sufficient to **release** a quick link to supported versions of Eigen.

Note that this does not solve the question of how to return a WIP quick link for **dev or testing** purposes. For that we might have to rely on other techniques like feature flags; or limiting to staging by checking `config.SYSTEM_ENVIRONMENT`. And once we are satisfied that a quick link is releasable, only then would we set a `minimumEigenVersion` to release it. 

That two-step sequence doesn't seem _so_ bad to me, but I'm open to ideas for streamlining.

### DIFF
--- a/src/schema/v2/homeView/sectionTypes/NavigationPills.ts
+++ b/src/schema/v2/homeView/sectionTypes/NavigationPills.ts
@@ -10,12 +10,14 @@ import { HomeViewGenericSectionInterface } from "./GenericSectionInterface"
 import { HomeViewSectionTypeNames } from "./names"
 import { standardSectionFields } from "./GenericSectionInterface"
 import { OwnerType } from "@artsy/cohesion"
+import { SemanticVersionNumber } from "lib/semanticVersioning"
 
 export interface NavigationPill {
   title: string
   href: string
   ownerType: OwnerType
   icon?: string
+  minimumEigenVersion?: SemanticVersionNumber
 }
 
 const NavigationPillType = new GraphQLObjectType<
--- a/src/schema/v2/homeView/sectionTypes/NavigationPills.ts
+++ b/src/schema/v2/homeView/sectionTypes/NavigationPills.ts
@@ -17,6 +17,11 @@ export interface NavigationPill {
   href: string
   ownerType: OwnerType
   icon?: string
+
+  /**
+   * The implementing section must enforce this check.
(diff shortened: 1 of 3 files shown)


src/schema/v2/homeView/sections/QuickLinks.ts
```diff
@@ -2,19 +2,37 @@ import { ContextModule, OwnerType } from "@artsy/cohesion"
...
+      context.userAgent as string
+    )
+    if (actualEigenVersion && quickLink.minimumEigenVersion) {
+      isDisplayable = isAtLeastVersion(
+        actualEigenVersion,
+        quickLink.minimumEigenVersion
+      )
+    }
+    return isDisplayable
+  })
```
anandaroop, line 32: This is the gist of it.

Similar to the versioning we introduced at the home view section level, this logic considers all quick links displayable for _non_-Eigen clients.

src/schema/v2/homeView/sectionTypes/NavigationPills.ts
```diff
@@ -10,12 +10,19 @@ import { HomeViewGenericSectionInterface } from "./GenericSectionInterface"
...
   title: string
   href: string
   ownerType: OwnerType
   icon?: string
+
+  /**
+   * The implementing section must enforce this check.
+   * See QuickLinks for an example
+   */
+  minimumEigenVersion?: SemanticVersionNumber
```
anandaroop, line 25: Added this comment to make this very explicit. 

Unlike with the versioning at the root home view level (see #6297 ), you don't get Eigen version enforcement "for free" within a section of type `NavigationPills` — that is left to the implementing section.

dariakoko (approved):
Looks great, thanks Roop!

PR #6477, opened 2025-02-28
feat: add ownerType to the FeaturedFairs home view section

dariakoko:
FollowUp on https://github.com/artsy/eigen/pull/11615: to make sure we track the FeaturedFairs section title tap correctly we need to add ownerType


dblandin (approved):
👍🏼

PR #6481, opened 2025-02-28
feat: add fairs and meduim quick links

dariakoko:
Add two last quick links to MP: Featured fairs and Medium

https://github.com/user-attachments/assets/5636ea5f-5b5c-4a39-b49b-6e9593c0eadb

### DIFF
--- a/src/schema/v2/homeView/sections/QuickLinks.ts
+++ b/src/schema/v2/homeView/sections/QuickLinks.ts
@@ -63,26 +63,23 @@ export const QUICK_LINKS: Array<NavigationPill> = [
     ownerType: OwnerType.collection,
     icon: undefined,
   },
-
-  // TODO: enable once navigation is sorted out, via minimumEigenVersion
-  // {
-  //   title: "Medium",
-  //   href: "/collections-by-category/Medium",
-  //   ownerType: OwnerType.collectionsCategory,
-  //   icon: "ArtworkIcon",
-  // },
+  {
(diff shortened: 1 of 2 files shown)


src/schema/v2/homeView/sections/QuickLinks.ts
```diff
@@ -63,26 +63,23 @@ export const QUICK_LINKS: Array<NavigationPill> = [
...
-  // {
-  //   title: "Medium",
-  //   href: "/collections-by-category/Medium",
-  //   ownerType: OwnerType.collectionsCategory,
-  //   icon: "ArtworkIcon",
-  // },
+  {
+    title: "Medium",
+    href:
+      "/collections-by-category/Medium?homeViewSectionId=home-view-section-explore-by-category&entityID=Medium",
```
dariakoko, line 69: Looks very disturbing 👀  I don't think this is the best approach, but it's working. Ideally we would have `/collections-by-category/Medium`, which we cannot use right now due to [the current setup on Eigen](https://github.com/artsy/eigen/blob/main/src/app/Scenes/CollectionsByCategory/Footer.tsx) where  `homeViewSectionId` and `entityID` are required to fetch the category data

dblandin: Yep, I agree. This coupling is on our radar to address/refactor but the href shown here should be a fairly low risk way to support the quick link in the meantime.

PR #6500, opened 2025-03-12
feat(DIA-1194): move infinite discovery above discover something new

iskounen:
This PR updates the order of home view elements so that the "Infinite Discovery" entry point appears above the "Discover Something New" section.


MounirDhahri (approved):
LGTM!

PR #6503, opened 2025-03-13
chore: artwork import errors are now an enum, as is state

Mitch-Henson:
Adding some MP niceties that we can take advantage of in Volt. This avoids the direct lookup needed to Gravity of what is possible and helps with development.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2970,13 +2970,13 @@ type ArtworkImport implements Node {
     before: String
     blockersOnly: Boolean
     createdOnly: Boolean
-    errorTypes: [String!]
-    excludeErrorTypes: [String!]
+    errorTypes: [ArtworkImportError]
+    excludeErrorTypes: [ArtworkImportError]
     first: Int
     hasErrors: Boolean
     last: Int
   ): ArtworkImportRowConnection
-  state: String!
+  state: ArtworkImportState
   unmatchedArtistNames: [String!]!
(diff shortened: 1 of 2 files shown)


Mitch-Henson:
@damassi as Force doesn't use this part of MP, it's fine to ignore the breaking changes, yeah?

lidimayra:
It's not super clear to me if we reached consensus on where we stand regarding graphql enums, but I guess the recent discussions were guiding us to stay away from it?

Related thread [here](https://artsy.slack.com/archives/C05F8TNKGAV/p1739560749874999)

Mitch-Henson:
Thanks for bringing up that context, I must have missed it. Funnily enough it seemed to be in consensus when it was raised [here](https://artsy.slack.com/archives/C07UYRFAVPU/p1741808073392229) (the reason that I'm actually tackling this work)

damassi:
For the record (and sorry if there was some confusion!) We should be using enums whenever we see concrete values. Its fine to avoid them for prototyping and just spinning stuff up quick (especially if there's dev going on over on the gravity side) but enums just make everything better and more safe, so we should lean in 👍

damassi (approved):

damassi:
And yep, totally fine to avoid breaking changes. A lot of it is just "lets be safe" stuff -- which mostly just applies to eigen (mobile versions can't retroactively change their schema) vs web, because we can always update web.

PR #6513, opened 2025-03-18
fix(artist): closes artist title tag a/b test

dzucconi:
The control branch won so this reverts the title tag back to what it originally was.


src/schema/v2/artist/meta.ts
```diff
@@ -93,505 +89,3 @@ const Meta: GraphQLFieldConfig<void, ResolverContext> = {
 }
 
 export default Meta
-
-// https://artsyproduct.atlassian.net/browse/DIA-1131
-const EXPERIMENT_GROUP = [
```
JanaeHijaz, line 98: oh interesting. I wonder how we determine which artists go into this control group?

PR #6537, opened 2025-03-21
feat: bump dd-trace

egdbear:
[future-friday] I noticed that our dd-trace version `0.36.2` is EOL for a while.

image

### DIFF
--- a/src/lib/tracer.ts
+++ b/src/lib/tracer.ts
@@ -3,7 +3,7 @@ import { verbose as debug, error } from "./loggers"
 import tracer from "dd-trace"
 import { Tags } from "opentracing"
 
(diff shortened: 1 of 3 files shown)


src/lib/tracer.ts
```diff
@@ -20,7 +20,6 @@ export function init() {
         // Noop
       },
     },
-    debug: !PRODUCTION_ENV,
```
egdbear, line 23: `PRODUCTION_ENV` is `true` for both staging and production, so this was likely never enabled.  
In this version, debug mode is disabled by default.  

Reference: https://github.com/DataDog/dd-trace-js/blob/v5.x/docs/API.md#custom-logging  

To enable debug mode, set `DD_TRACE_DEBUG=true`.

mc-jones:
🙏

PR #6543, opened 2025-03-24
chore: artwork import price listed currency

Mitch-Henson:
This will enable us to be able to highlight the `currencyPrefix` when showing price without having to do additional work on the front end. An example us of this can be found in [this figma file](https://www.figma.com/design/y31Q2l6ggPJEtUfWWtvBuE/Listing-experience?node-id=2654-12379&t=FxVBrxKUGtoid3sy-4).

An example response with these changes is as follows

```
"priceListed": {
  "major": 5000,
  "minor": 500000,
  "currencyCode": "USD",
  "display": "US$5,000",
  "currencyPrefix": "US$",
  "amount": "5,000"
}
```

### DIFF
--- a/src/schema/v2/fields/money.ts
+++ b/src/schema/v2/fields/money.ts
@@ -14,7 +14,7 @@ import { ResolverContext } from "types/graphql"
 // Taken from https://github.com/RubyMoney/money/blob/master/config/currency_iso.json
 import currencyCodes from "lib/currency_codes.json"
 import { GraphQLLong } from "lib/customTypes/GraphQLLong"
-import { priceDisplayText } from "lib/moneyHelpers"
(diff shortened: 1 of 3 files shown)


src/schema/v2/fields/money.ts
```diff
@@ -227,6 +232,8 @@ export const resolveMinorAndCurrencyFieldsToMoney = async (
       cents: minor,
       currency: currencyCode,
       display: priceDisplayText(minor, currencyCode, ""),
+      amount: priceAmount(minor, currencyCode, ""),
+      currencyPrefix: currencyPrefix(currencyCode),
```
damassi, line 236: Oh nice, so this is already a helper? Nicee

damassi (approved):
Looks good to me, but wondering if @mzikherman might have any comments?

PR #6544, opened 2025-03-24
feat: add isFixedShippingFeeOnly property to artwork

oxaudo:
In preparation to check in Force (EMI-2357) for fixed shipping fee for the order adding this property here. 

Was debating if just request a raw data and do this check in Force but figured it might be helpful in other contexts and also did not want a lot of logic on the client for this one.

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -3213,6 +3213,237 @@ describe("Artwork type", () => {
     })
   })
 
+  describe("#isFixedShippingFeeOnly", () => {
+    const query = `
+      {
+        artwork(id: "richard-prince-untitled-portrait") {
+          isFixedShippingFeeOnly
+        }
+      }
+    `
+    describe("when artsy domestic shipping enabled", () => {
+      beforeEach(() => {
+        artwork.process_with_artsy_shipping_domestic = true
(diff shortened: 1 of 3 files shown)


src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -3213,6 +3213,237 @@ describe("Artwork type", () => {
...
+            })
+          })
+        })
+      })
+    })
+
+    describe("when artsy domestic shipping disabled", () => {
+      beforeEach(() => {
+        artwork.process_with_artsy_shipping_domestic = false
+        artwork.artsy_shipping_international = false
```
erikdstock, line 3319: Are we covering all cases of overlap in the presence of `process_with_artsy_shipping_domestic` and `domestic_shipping_fee_cents`?

erikdstock (approved):
LGTM. As discussed in slack we can move this to the gravity artwork if we need the property in other areas (like exchange)

src/schema/v2/artwork/__tests__/artwork.test.js
oxaudo: I tried to think of all cases. But not sure what is the best way to organize the spec. Followed the pattern from the `onlyShipsDomestically`. Happy to add if I'm missing some.

PR #6551, opened 2025-03-25
feat(feature): exposes meta title

dzucconi:
Exposes the `meta_title` field / supports editing.


src/types/runtime/gravity/Feature.ts
```diff
@@ -29,6 +29,7 @@ export const Feature = Record({
     source: String.Or(Undefined),
   }).Or(Null),
   created_at: String,
+  meta_title: String.Or(Null).Or(Undefined),
```
dzucconi, line 32: Funny but this is the only Gravity response we have that's properly typed and I must say: it's nice!

egdbear (approved):

PR #6565, opened 2025-03-27
feat: submit an ApplePay order

starsirius:
https://artsyproduct.atlassian.net/browse/EMI-2304

This is the MP side of the [Exchange PR](https://github.com/artsy/exchange/pull/2470) to support submitting an Apple Pay order.

### DIFF
--- a/src/schema/v2/order/submitOrderMutation.ts
+++ b/src/schema/v2/order/submitOrderMutation.ts
@@ -1,0 +1,49 @@
+import { GraphQLString, GraphQLNonNull, GraphQLID } from "graphql"
+import {
+  ORDER_MUTATION_FLAGS,
+  OrderMutationResponseType,
+} from "./sharedOrderTypes"
+import { mutationWithClientMutationId } from "graphql-relay"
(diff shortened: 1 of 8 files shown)


src/schema/v2/order/submitOrderMutation.ts
```diff
@@ -0,0 +1,49 @@
...
+      description: "Stripe confirmation token.",
+    },
+  },
+  outputFields: {
+    orderOrError: {
+      type: OrderMutationResponseType,
+      resolve: (response) => response,
+    },
+  },
+  mutateAndGetPayload: async (input, { meOrderSubmitLoader }) => {
```
damassi, line 34: We should be using our canonical `ThingOrError` patterns for mutations (see here: https://github.com/artsy/metaphysics/pull/6564)

starsirius: I think it is using the `orderOrError` pattern (see tests below), but happy to update if I missed something!

starsirius:
Going to merge this in preparation of the QA! Feedback still welcome!!

PR #6566, opened 2025-03-27
fix: correctly handle known errors from exchange

erikdstock:
type: **fix**
resolves [EMI-2358]

This PR adds a lightweight error handling pattern so we can forward known error types from exchange. Currently it is only concerned with validation errors: these have a `message` and a `code` which are by default the same (unless the error gets a [details](https://github.com/artsy/exchange/blob/main/spec/api/me/orders_endpoint_spec.rb#L246-L249) as well).

We could get more clever about this:
- we could also send the error `type`, ie `validation`
- we could make the various error types a union type
- we could make the various error codes an enum type

Curious on others thoughts but this works for today.

cc @artsy/emerald-devs 

https://github.com/artsy/exchange/blob/main/app/api/concerns/grape_error_handlers.rb#L50-L52

[EMI-2358]: https://artsyproduct.atlassian.net/browse/EMI-2358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/order/exchangeErrorHandling.ts
+++ b/src/schema/v2/order/exchangeErrorHandling.ts
@@ -1,0 +1,28 @@
+import { ORDER_MUTATION_FLAGS } from "./sharedOrderTypes"
+
+type ExchangeError = Error & {
+  statusCode: number
+  body?: {
+    message: string
+    code: string
+  }
+}
+export const handleExchangeError = (error: ExchangeError) => {
+  let errorProperties: { message: string; code: string }
+
+  if (error.statusCode === 422 && typeof error.body === "object") {
+    errorProperties = {
+      message: error.body.message,
+      code: error.body.code,
+    }
+  } else {
+    errorProperties = {
+      message: "An error occurred",
+      code: "internal_error",
+    }
+  }
+  return {
+    ...errorProperties,
+    _type: ORDER_MUTATION_FLAGS.ERROR,
+  }
+}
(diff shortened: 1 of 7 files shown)


starsirius (approved):
Let's do it and keep iterating!

src/schema/v2/order/exchangeErrorHandling.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  if (error.statusCode === 422 && typeof error.body === "object") {
+    errorProperties = {
+      message: error.body.message,
+      code: error.body.code,
+    }
+  } else {
+    errorProperties = {
+      message: "An error occurred",
+      code: "internal_error",
```
starsirius, line 21: I think we'll keep iterating on the types of error down the road!

src/schema/v2/order/exchangeErrorHandling.ts
```diff
@@ -0,0 +1,28 @@
+import { ORDER_MUTATION_FLAGS } from "./sharedOrderTypes"
+
+type ExchangeError = Error & {
+  statusCode: number
+  body?: {
+    message: string
+    code: string
```
starsirius, line 7: I like how Stripe aims for the [message](https://docs.stripe.com/api/errors#errors-message) to be human readable and for end users.

> message nullable string
A human-readable message providing more details about the error. For card errors, these messages can be shown to your users.

PR #6581, opened 2025-04-01
feat: add seller field to order

starsirius:
https://artsyproduct.atlassian.net/browse/EMI-2362

Following https://github.com/artsy/exchange/pull/2486, this adds a `seller` field on the order type in which we'll be able fetch the stripe account id. It's basically to replace the current [stitched version](https://github.com/artsy/metaphysics/blob/c83d5b17aa3cb4cceb9b18978a243b6d36f6e87b/src/lib/stitching/exchange/v2/stitching.ts#L607).

cc @artsy/emerald-devs

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -15896,6 +15896,9 @@ type Order {
   # The selected fulfillment option for the order
   selectedFulfillmentOption: FulfillmentOption
 
+  # The seller of the order
+  seller: SellerType
+
   # The total amount for shipping
   shippingTotal: Money
 
@@ -20102,6 +20105,8 @@ interface Sellable {
   isSold: Boolean
   saleMessage: String
 }
+
+union SellerType = Partner
 
 type SendConfirmationEmailMutationFailure {
   mutationError: GravityMutationError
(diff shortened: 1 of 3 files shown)


_schemaV2.graphql
```diff
@@ -20091,6 +20094,8 @@ interface Sellable {
   saleMessage: String
 }
 
+union SellerType = Partner
```
starsirius, line 20109: Making the seller type a union, aligned with Exchange assumption, to (hopefully) avoid breaking changes when seller can be other than partners.

leamotta (approved):

src/schema/v2/order/sharedOrderTypes.ts
```diff
@@ -415,6 +428,15 @@ export const OrderType = new GraphQLObjectType<OrderJSON, ResolverContext>({
...
       },
     },
+    seller: {
+      type: SellerType,
+      description: "The seller of the order",
+      resolve: ({ seller_id }, _args, { partnerLoader }) => {
+        if (!seller_id) return null
+
+        return partnerLoader(seller_id).catch(() => null)
+      },
```
erikdstock, line 438: non-blocking: For completeness, should we verify seller type and log/return null if it is not a seller?

erikdstock (commented):
L G T M

🥇

src/schema/v2/order/sharedOrderTypes.ts
starsirius: I'd hope the validation can be provided upstream, but yeah I assume eventually we'll have some logic in place for different seller types.

PR #6582, opened 2025-04-01
fix: remove leading and trailing whitespace characters from the search query

nickskalkin:
When I type "Claudia" and "Claudia " (with a trailing space) into the search input, I get completely different results:

| "Claudia" | "Claudia " |
|---|---|
|  |  |

We believe this is worth fixing.

## Tech notes

The issue occurs because we apply a strong boost to the `name_exact` field, which is a `keyword` field. Keyword fields in Elasticsearch don’t apply any normalization — for example, they don’t trim surrounding whitespace. You can see the difference by comparing the analyzer results between the `name_exact` (keyword) and `name` (text) fields:

### name_exact

```
POST /artists_production/_analyze
{
  "field": "name_exact",
  "text": "   Claudia   "
}
```

Returns:

```json
{
  "tokens": [
    {
      "token": "   Claudia   "
     }
  ]
}
```

### name

```
POST /artists_production/_analyze
{
  "field": "name",
  "text": "   Claudia   "
}
```

Returns:

```json
{
  "tokens": [
    {
      "token": "claudia"
     }
  ]
}
```

One potential solution would be to apply normalization logic (like trimming) at index or query time in Elasticsearch. However, I think the simplest fix would be to trim leading/trailing spaces in Metaphysics before sending the query.

### DIFF
--- a/src/schema/v2/__tests__/search.test.ts
+++ b/src/schema/v2/__tests__/search.test.ts
@@ -356,4 +356,24 @@ describe("Search", () => {
       expect(geneNode.name).toBe("Minimalism")
     })
   })
+
+  it("Removes leading and trailing whitespace characters from the query", async () => {
+    const query = `
+      {
+        searchConnection(query: "  David Bowie  ", first: 10) {
+          edges {
+            node {
+              __typename
+            }
+          }
+        }
+      }
+    `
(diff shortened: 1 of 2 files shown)


anandaroop (approved):
Nice!

PR #6583, opened 2025-04-01
feat(settingsV2): Mutation to update and delete Partner Contact records

jpotts244:
Mutation to update and delete Partner Contact records

## Partner contact update structure
```graphql

mutation {
  updatePartnerContact(input: { 
  partnerId: "5f80bfefe8d808000ea212c1"
  contactId: "b9fa1baa-f8f2-4beb-90f6-80361103462c"
  email: "fun2@test.com"
  }) {
    partnerContactOrError {
      ... on UpdatePartnerContactSuccess {
        partnerContact {
          name
          internalID
          phone
          email
          location { 
            display
          }
        }
      }
      ... on UpdatePartnerContactFailure {
        mutationError {
          message
        }
      }
    }
  }
}

```
Screenshot 2025-04-01 at 12 04 20 PM

---

## Partner contact delete structure
```graphql

mutation {
  deletePartnerContact(input: { 
  partnerId: "5f80bfefe8d808000ea212c1"
	contactId: "b9fa1baa-f8f2-4beb-90f6-80361103462c"
  }) {
    partnerContactOrError {
      ... on DeletePartnerContactSuccess {
        partnerContact {
          name
          internalID
          phone
          email
          location { 
            display
          }
        }
      }
      ... on DeletePartnerContactFailure {
        mutationError {
          message
        }
      }
    }
  }
}
```

Screenshot 2025-04-01 at 12 40 41 PM


src/schema/v2/partner/Settings/updatePartnerContactMutation.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export const UpdatePartnerContactMutation = mutationWithClientMutationId<
+  UpdatePartnerContactInputProps,
+  any,
+  ResolverContext
+>({
+  name: "UpdatePartnerContact",
+  description: "Updates an existing contact for a partner",
+  inputFields: {
+    contactId: {
```
jpotts244, line 62: What are your thoughts on input id naming convention?

`contactId` vs `contactID`? (I'll make this consistent with my previous PRs in a clean up pass)

mzikherman: We're all over the place in MP - I wouldn't worry too much about it. Because of the types/Relay support - you'll get the appropriate error/auto-complete if you try to use one when a mutation expects the other.

We should be consistent within a single mutation (just for a bit of sanity - why not), but in general in MP we have lots of examples of both formats.

src/schema/v2/partner/Settings/updatePartnerContactMutation.ts
```diff
@@ -0,0 +1,142 @@
...
+    try {
+      const response = await updatePartnerContactLoader(
+        { partnerId, contactId },
+        {
+          name,
+          position,
+          can_contact: canContact,
+          email,
+          phone,
+          partner_location_id: locationID,
```
mzikherman, line 128: Something that might be worth double-checking, just for completeness - does this let you _unset_ a location (assuming that's even possible). I'd imagine `locationID: null` being sent to the mutation would unset a location, whereas not including `locationID` in the mutation input at all wouldn't affect the location of the existing contact.

mzikherman: (that's assuming the endpoint even allows unsetting a location - which it might not! Just one of those quirks I've noticed quite a lot at Artsy - we typically build something that lets you 'set', but don't consider an 'unset' use case). This is so minor that it could easily be followed up/don't want to distract you - just lobbing it up for 🤯

jpotts244: Great idea to test, ok, going for it! (if its a big issue I'll follow up as a fast follow)

jpotts244: Great success 🧵 

## Unset a location
Sending in `locationId: null` looks to unset the current location
Screenshot 2025-04-01 at 1 00 31 PM


## Edit a location
Sending in a different locationId updates the association correctly ✅ 

## Keeping the location param out
All good, only updates input fields ✅

src/schema/v2/partner/Settings/updatePartnerContactMutation.ts
```diff
@@ -0,0 +1,142 @@
...
+  position?: string
+  canContact?: boolean
+  email?: string
+  phone?: string
+  locationId?: string
+}
+
+const SuccessType = new GraphQLObjectType<any, ResolverContext>({
+  name: "UpdatePartnerContactSuccess",
+  isTypeOf: (data) => !!data.id,
```
jpotts244, line 29: One thing I ran into here ^^

The PartnerContact model was just recently migrated from `Mongo` to `PQSL`
Records coming from psql dont seem to have a `_id`, will adjust the others regarding partner contact to `id`

damassi: @jonallured - This is probably out of my pay grade, and i'm certainly misunderstanding something, but i'm curious why we don't keep these things backwards compatible (so these things issues don't come up?)

damassi (dismissed):

src/schema/v2/partner/Settings/updatePartnerContactMutation.ts
jpotts244: Hm, for some reason I was getting an error when trying to access `._id` in MP land (let me check for user error)

Jon did make it backwards compatible: https://github.com/artsy/gravity/commit/51adfb1050943b847787969c4ecaef2ba31836b1

jpotts244: Ah ok, so I'm using `..IDFields` in my Partner contact type

Which is converting `id` (a slug) into `internalID` [here](https://github.com/artsy/metaphysics/blob/f7286b1eade1fbbf3840d2c80ec2972ea6c7e593/src/schema/v2/object_identification.ts#L200-L217)

damassi: Ok cool! I was suprised and didn't think Jon would make a rookie mistake like that 😆 

There's a few different ID field options, which vary depending on the use-case: https://github.com/artsy/metaphysics/blob/f7286b1eade1fbbf3840d2c80ec2972ea6c7e593/src/schema/v2/object_identification.ts#L176-L216

@mzikherman could explain it better than I

mzikherman: Yup! Anytime the JSON payload does the `id` / `_id` dance (that was just the pattern Gravity started using long ago) - most of the time you prob want `GravityIDFields`, or maybe even `SlugAndInternalIDFields` as that supports `_id` (database id -> `internalID) and `id` (slug -> `slug`).

jpotts244: Good tips!! Cool I'll adjust to a more fruitful `..Fields` option, thanks!

jonallured: yes please note that anytime you see a regression like a missing _id then that's a mistake not intentional 😅

PR #6590, opened 2025-04-03
feat(edition-sets): Share fields with sellable

damassi:
Ran into an issue where EditionSets are _not_ Artworks, but `Artwork` implements the `Sellable` interface. So expand the `Sellable` interface with some additional `Artwork` types, and implement them via a shared object in `EditionSets` 😵‍💫 

Something feels sloppy here, and i'm not quite sure about naming, but it does smooth things out nicely on the GraphQL consumer side and makes things feel more "artwork like". 

Thoughts?

cc @artsy/amber-devs

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10212,6 +10212,8 @@ input EditableLocation {
 type EditionSet implements Sellable {
   availability: String
   dimensions: dimensions
+
+  # The edition set parent-artwork display label (title)
   displayLabel: String
   displayPriceRange: Boolean
   editionOf: String
@@ -10230,7 +10232,10 @@ type EditionSet implements Sellable {
   isAcquireable: Boolean
   isForSale: Boolean
 
-  # Do we want to encourage inquiries on this work?
+  # Is the edition set parent-artwork part of an auction?
+  isInAuction: Boolean
+
+  # Is the edition set parent-artwork inquireable?
   isInquireable: Boolean
   isOfferable: Boolean
   isOfferableFromInquiry: Boolean
@@ -10241,7 +10246,7 @@ type EditionSet implements Sellable {
   priceDisplay: String
   priceListed: Money
 
-  # Whether this artwork is published or not
+  # Is the edition set parent-artwork published?
   published: Boolean
   saleMessage: String
 
@@ -20149,19 +20154,33 @@ union SecondFactorOrErrorsUnion = AppSecondFactor | Errors | SmsSecondFactor
 
 # A piece that can be sold
 interface Sellable {
+  availability: String
   dimensions: dimensions
+  displayLabel: String
+  displayPriceRange: Boolean
   editionOf: String
+
+  # A globally unique ID.
+  id: ID!
+  internalDisplayPrice: String
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
 
   # Whether a piece can be purchased through e-commerce
   isAcquireable: Boolean
   isForSale: Boolean
+  isInAuction: Boolean
+  isInquireable: Boolean
 
   # Whether a user can make an offer on the work
   isOfferable: Boolean
 
   # Whether a user can make an offer on the work through inquiry
   isOfferableFromInquiry: Boolean
+  isPriceHidden: Boolean
   isSold: Boolean
+  published: Boolean
   saleMessage: String
 }
 
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10212,6 +10212,7 @@ input EditableLocation {
 type EditionSet implements Sellable {
   availability: String
   dimensions: dimensions
+  displayLabel: String
   displayPriceRange: Boolean
   editionOf: String
 
@@ -10228,13 +10229,23 @@ type EditionSet implements Sellable {
   internalID: ID!
   isAcquireable: Boolean
   isForSale: Boolean
+
+  # Is this artwork part of an auction?
+  isInAuction: Boolean
+
+  # Do we want to encourage inquiries on this work?
+  isInquireable: Boolean
   isOfferable: Boolean
   isOfferableFromInquiry: Boolean
+  isPriceHidden: Boolean
   isSold: Boolean
   listPrice: ListPrice
   price: String
   priceDisplay: String
   priceListed: Money
+
+  # Whether this artwork is published or not
+  published: Boolean
   saleMessage: String
 
   # size bucket assigned to an artwork based on its dimensions
@@ -20141,19 +20152,41 @@ union SecondFactorOrErrorsUnion = AppSecondFactor | Errors | SmsSecondFactor
 
 # A piece that can be sold
 interface Sellable {
+  availability: String
   dimensions: dimensions
+  displayLabel: String
+  displayPriceRange: Boolean
   editionOf: String
+
+  # A globally unique ID.
+  id: ID!
+
+  # Price for internal partner display, requires partner access
+  internalDisplayPrice: String
+
+  # A type-specific ID likely used as a database ID.
+  internalID: ID!
 
   # Whether a piece can be purchased through e-commerce
   isAcquireable: Boolean
   isForSale: Boolean
+
+  # Is this artwork part of an auction?
+  isInAuction: Boolean
+
+  # Do we want to encourage inquiries on this work?
+  isInquireable: Boolean
 
   # Whether a user can make an offer on the work
   isOfferable: Boolean
 
   # Whether a user can make an offer on the work through inquiry
   isOfferableFromInquiry: Boolean
+  isPriceHidden: Boolean
   isSold: Boolean
+
+  # Whether this artwork is published or not
+  published: Boolean
   saleMessage: String
 }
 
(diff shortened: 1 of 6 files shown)


src/schema/v2/sellable.ts
```diff
@@ -39,5 +41,54 @@ export const Sellable = new GraphQLInterfaceType({
       type: GraphQLString,
       resolve: ({ sale_message }) => sale_message,
     },
-  },
+  }),
 })
+
+export const sharedSellableFields = {
+  ...InternalIDFields,
```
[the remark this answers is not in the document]
damassi: Thanks 👍 This ID field is mostly for having an identifier to update in the mutation, but i'll update the code with this comment so that this is totally clear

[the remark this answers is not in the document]
jpotts244: Excellent!

_schemaV2.graphql
```diff
@@ -10225,13 +10226,23 @@ type EditionSet implements Sellable {
   internalID: ID!
   isAcquireable: Boolean
   isForSale: Boolean
+
+  # Is this artwork part of an auction?
```
oxaudo, line 10230: Super #nitpick but might be worth not to use artwork here as we are looking at edition set and the info is read from artwork but does not exist on EditionSet itself. So maybe 'is this edition set belongs to the artwork in an auction' is more accurate description? Or just `Is this object part of an auction?`

_schemaV2.graphql
```diff
@@ -10225,13 +10226,23 @@ type EditionSet implements Sellable {
...
+  isInAuction: Boolean
+
+  # Do we want to encourage inquiries on this work?
```
oxaudo, line 10233: #suggestion - `Can collectors submit inquiries on this work/object' ?

_schemaV2.graphql
```diff
@@ -20133,19 +20144,41 @@ union SecondFactorOrErrorsUnion = AppSecondFactor | Errors | SmsSecondFactor
...
+  displayPriceRange: Boolean
   editionOf: String
 
+  # A globally unique ID.
+  id: ID!
+
+  # Price for internal partner display, requires partner access
+  internalDisplayPrice: String
+
+  # A type-specific ID likely used as a database ID.
```
oxaudo, line 20159: This one is interesting - it's an editionSet id? Or the artworkId?

src/schema/v2/sellable.ts
```diff
@@ -39,5 +41,57 @@ export const Sellable = new GraphQLInterfaceType({
...
+  },
+  internalDisplayPrice: {
+    type: GraphQLString,
+    resolve: ({ internal_display_price }) => internal_display_price,
+    description: "Price for internal partner display, requires partner access",
+  },
+  isInAuction: {
+    type: GraphQLBoolean,
+    description: "Is this artwork part of an auction?",
+    resolve: async ({ sale_ids }, _options, { salesLoader }) => {
```
oxaudo, line 71: How would this work for EditionSets? Don't thing `sale_ids` are accessible from looking at EditionSet in isolation?

oxaudo (commented):
Not sure my comments are useful here but those are immediate reactions from looking at this stuff.

_schemaV2.graphql
damassi: Ah @oxaudo - i just copied and pasted this stuff right out of the artwork page. Didn't write any of those comments and are historical, so didn't want to change anything.

_schemaV2.graphql
damassi: Its a "sellable" ID (since it implements the interface) -- which just plucks out the internalID, however it may be

src/schema/v2/sellable.ts
mzikherman: yea - we should make sure all of these fields _can_ be correctly resolved for both an artwork and an edition set - otherwise just asking for bugs/confusion later.

it might be premature to even consolidate all of these fields (and resolvers) in `sharedSellableFields`. Maybe instead, artwork and edition set should _each_ define all of these fields within their types/files, along with the appropriate resolvers?

oxaudo: Since this an MP type I *suspect* it's ok to make it easier to get the artwork properties from each edition set and define this as 'sellable' interface which can be either artwork or edition set. In fact we already  use it as such in some cases: https://github.com/artsy/gravity/blob/8aab8f403c395c03630d543dd564c407ca450552/app/models/domain/invoice_line_item.rb#L74 However for resolving I believe we need to be able to read the property from edition set object, no?  That is why we've added bunch of those `delegate` references: https://github.com/artsy/gravity/blob/8aab8f403c395c03630d543dd564c407ca450552/app/models/domain/edition_set.rb#L35

`sale_ids` are not accessible from EditionSet object directly as far as I can tell...

mzikherman: Oh yes - that's what I'm getting at! Sharing the _resolvers_ here likely won't work b/c artwork + edition set, while they have a lot of overlap in JSON fields, aren't identical.

So we may still want to leave the field declarations here for documentation, but likely a bunch/all of these fields should exist in _both_ `Artwork` and `EditionSet` MP types.

And furthermore, b/c some of this needs to be resolved using artwork-level fields that aren't available on the edition set JSON (like `sale_ids`) - in addition to creating the correct resolvers on `Artwork` and `EditionSet`, we _also_ want to inject the `artwork` JSON into the edition set JSON for `EditionSet` resolvers to have access to.

src/schema/v2/artwork/updateArtworkMutation.ts
```diff
@@ -43,25 +45,40 @@ const ResponseOrErrorType = new GraphQLUnionType({
...
   offer?: boolean
+  priceCurrency?: string
   priceHidden?: boolean
+  priceIncludesTax?: boolean
   priceListed?: string
-  editionSets?: Omit<UpdateArtworkMutationInputProps, "editionSets">[]
+  priceMax?: number
+  priceMin?: number
+  shippingWeight?: number
+  shippingWeightMetric?: string
```
damassi, line 70: Filling out the artwork mutation options

src/schema/v2/artwork/index.ts
```diff
@@ -1860,7 +1858,7 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         resolve: ({ framed_width }) => framed_width,
       },
       framedDiameter: {
-        type: GraphQLFloat,
+        type: GraphQLString,
```
mzikherman, line 1861: Backwards-safe?

src/schema/v2/edition_set.ts
```diff
@@ -43,10 +47,31 @@ export const EditionSetType = new GraphQLObjectType<any, ResolverContext>({
...
       resolve: ({ acquireable }) => acquireable,
     },
+    isForSale: {
+      type: GraphQLBoolean,
+      resolve: ({ forsale }) => forsale,
+    },
+    isInquireable: {
+      type: GraphQLBoolean,
+      description: "Do we want to encourage inquiries on this work?",
+      resolve: ({ inquireable }) => inquireable,
```
mzikherman, line 73: I don't see this in the edition set JSON - https://github.com/artsy/gravity/blob/b7ea76c854c1d79175f89daa3e4e810a2b722d8c/app/models/domain/edition_set.rb#L82

src/schema/v2/edition_set.ts
```diff
@@ -86,6 +105,10 @@ export const EditionSetType = new GraphQLObjectType<any, ResolverContext>({
         return { cents, currency }
       },
     },
+    published: {
```
mzikherman, line 108: Same - this isn't in the edition set JSON.

mzikherman (dismissed):
Nice and minimal. I think a couple of those of fields may still not be quite right - may need the artwork field injected in, I would recommend injecting the entire `artwork` into each edition set to make those (and any future ones) easier.

src/schema/v2/artwork/index.ts
damassi: Nope, but been chatting with @erikdstock over here [about this earlier](https://artsy.slack.com/archives/C02JHHHKP5K/p1743705306244979)

src/schema/v2/sellable.ts
damassi: Resolved these, and others, to the artwork

src/schema/v2/edition_set.ts
damassi: Resolved to the artwork item and noted in description

src/schema/v2/edition_set.ts
damassi: ditto

src/schema/v2/artwork/index.ts
```diff
@@ -611,10 +611,18 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
       editionSets: {
         type: new GraphQLList(EditionSet.type),
         args: { sort: EditionSetSorts },
-        resolve: ({ edition_sets }, { sort }) => {
+        resolve: ({ edition_sets, ...artwork }, { sort }) => {
+          // Inject a copy of the artwork into each edition set
```
mzikherman, line 615: Nice

mzikherman (approved):
Cool - this looks exactly like what i was thinking!

src/schema/v2/artwork/index.ts
damassi: Arta types accept a string or a number, so things should just work in force 👍

damassi:
Aight, merging this on 🍏 Thanks everyone 👍

PR #6595, opened 2025-04-03
fix(featured fairs): add missing criteria

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-1658

A colleague [reported](https://artsy.slack.com/archives/C03N12SR0RK/p1743691566357749) that the app was displaying **Featured Fairs** that were not visible on web.

This brings the `featuredFairs` field's filtering criteria closer inline with the [existing query](https://github.com/artsy/force/blob/f651e59f4c09ca1e5909802dd745d4a2b93fe168/src/Apps/Fairs/Routes/FairsIndex.tsx#L335) that Force uses to generate the fairs listing on the /fairs landing page.


olerichter00 (commented):
Thanks! 🌟

PR #6602, opened 2025-04-07
feat(edition-sets): Add more fields

damassi:
Adds a few more fields to the `EditionSets` type and related network. 

cc @artsy/amber-devs

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2648,6 +2648,9 @@ type Artwork implements Node & Searchable & Sellable {
   isComparableWithAuctionResults: Boolean
   isDisliked: Boolean!
   isDownloadable: Boolean
+
+  # Whether this artwork is eligible for an ecommerce transaction
+  isEcommerce: Boolean
   isEdition: Boolean
 
   # Artwork is eligible for the Artsy Guarantee
@@ -10230,6 +10233,7 @@ type EditionSet implements Sellable {
   # A type-specific ID likely used as a database ID.
   internalID: ID!
   isAcquireable: Boolean
+  isEcommerce: Boolean
   isForSale: Boolean
 
   # Is the edition set parent-artwork part of an auction?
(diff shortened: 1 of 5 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -913,6 +913,12 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
           return comparables_count > 0 && category !== "Architecture"
         },
       },
+      isEcommerce: {
```
damassi, line 916: Its a bit odd to me that this has never been added to our schema before, but maybe its mostly just useful in CMS.

src/schema/v2/fields/money.ts
```diff
@@ -190,6 +190,7 @@ export const Money = new GraphQLObjectType<any, ResolverContext>({
     currencyPrefix: {
       type: GraphQLString,
       description: "The symbol used for the currency",
+      resolve: ({ currency }) => currencyPrefix(currency),
```
damassi, line 193: @Mitch-Henson, note I had to resolve one more spot to make this work universally

mzikherman (approved):

_schemaV2.graphql
```diff
@@ -2648,6 +2648,9 @@ type Artwork implements Node & Searchable & Sellable {
   isComparableWithAuctionResults: Boolean
   isDisliked: Boolean!
   isDownloadable: Boolean
+
+  # Whether this artwork is eligible for an ecommerce transaction
+  isEcommerce: Boolean
```
oxaudo, line 2653: I think this is a raw field. More like 'galley signals intention' if it wants the work to be `buy now`. The `isAcquireable` is the one that clients use to actually see if `buy now` transaction is possible.

src/schema/v2/artwork/index.ts
oxaudo: I think this is a raw field. More like 'galley signals intention' if it wants the work to be `buy now`. The `isAcquireable` is the one that clients use to actually see if `buy now` transaction is possible.

PR #6603, opened 2025-04-07
chore(dia-1224): removes blurhash feature flag and related field

dzucconi:
In the effort to reduce the number of future flags that 💎  has.

This field is being removed entirely as it's not used by either Eigen or Force and because it appears to be at an attack vector. Usage of it in Force with larger arguments appears to have caused [inc-219](https://artsy.app.opsgenie.com/reports/post-mortem/154274cc-497e-4b30-92a6-6dcc85f45bf1/detail)

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -12785,7 +12785,6 @@ type Image {
 
   # Blurhash code for the image
   blurhash: String
-  blurhashDataURL(width: Int = 64): String
   caption: String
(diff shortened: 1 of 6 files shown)


dzucconi:
Also to be clear about why this caused an issue: the size of the data returned will grow roughly with the square of the width. And we can accept arbitrary widths.

damassi (approved):

PR #6631, opened 2025-04-14
feat(sellable): Add listingOptions CMS pass-through

damassi:
With the `isEcommerce` option being better understood, this should more accurately be reflected via the `listingOptions` props on the artwork model, which directly point to:

- ecommerce -> isBuyNow
- offer -> isMakeOffer

So this updates things and passes it through:

```
query Artwork {
  artwork(id: "hi") {
    listingOptions {
      isBuyNow
      isMakeOffer
    }
  }
}
```
cc @artsy/amber-devs


lidimayra:
Thank you!! 🙏

PR #6632, opened 2025-04-14
chore: Add mutation to update partner flags

jonallured:
This PR takes the work done over on this one: https://github.com/artsy/gravity/pull/18820 and uses it to create a general purpose mutation that can update multiple partner flags. I'll use this mutation in the work I'm doing to migrate the CMS preferences page.

I worked with Claude to do this so I'm not 100% sure if it's following patterns/best practices so def open to feedback!!

/cc @artsy/amber-devs

### DIFF
--- a/src/schema/v2/partner/updatePartnerFlagsMutation.ts
+++ b/src/schema/v2/partner/updatePartnerFlagsMutation.ts
@@ -3,6 +3,7 @@ import {
   GraphQLNonNull,
   GraphQLObjectType,
   GraphQLUnionType,
+  GraphQLBoolean,
 } from "graphql"
 import { mutationWithClientMutationId } from "graphql-relay"
 import {
@@ -11,11 +12,13 @@ import {
 } from "lib/gravityErrorHandler"
 import Partner from "./partner"
 import { ResolverContext } from "types/graphql"
-import GraphQLJSON from "graphql-type-json"
 
 interface UpdatePartnerFlagsMutationInputProps {
   id: string
-  flags: Record<string, string>
+  inquireAvailabilityPriceDisplayEnabledByPartner?: boolean | null
+  artworksDefaultMetric?: string | null
+  artworksDefaultCurrency?: string | null
+  artworksDefaultPartnerLocationId?: string | null
 }
 
 const SuccessType = new GraphQLObjectType<any, ResolverContext>({
@@ -57,9 +60,21 @@ export const updatePartnerFlagsMutation = mutationWithClientMutationId<
       type: new GraphQLNonNull(GraphQLString),
       description: "The id of the partner to update.",
     },
-    flags: {
-      type: new GraphQLNonNull(GraphQLJSON),
-      description: "An object containing flag keys and values to update. If a value is empty, the flag will be unset.",
+    inquireAvailabilityPriceDisplayEnabledByPartner: {
+      type: GraphQLBoolean,
+      description: "Controls whether the partner has enabled the inquire availability price display. If null, the flag will be unset.",
+    },
+    artworksDefaultMetric: {
+      type: GraphQLString,
+      description: "The default metric system to use for artworks. If null, the flag will be unset.",
+    },
+    artworksDefaultCurrency: {
+      type: GraphQLString,
+      description: "The default currency to use for artworks. If null, the flag will be unset.",
+    },
+    artworksDefaultPartnerLocationId: {
+      type: GraphQLString,
+      description: "The default partner location ID to use for artworks. If null, the flag will be unset.",
     },
   },
   outputFields: {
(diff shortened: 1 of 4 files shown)


src/schema/v2/partner/updatePartnerFlagsMutation.ts
```diff
@@ -0,0 +1,90 @@
...
+>({
+  name: "UpdatePartnerFlagsMutation",
+  description: "Updates multiple flags on a partner simultaneously.",
+  inputFields: {
+    id: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: "The id of the partner to update.",
+    },
+    flags: {
+      type: new GraphQLNonNull(GraphQLJSON),
```
jonallured, line 61: Is this `GraphQLJSON` the right thing to use? I wasn't sure if Claude was off track here... 🤔

damassi: The non-null is ok, because if we're calling the mutation we want to be sure that the dev is passing in the correct flags

But! We should be explicit about the fields that we can update, vs using `GraphQLJSON` - we should always avoid that as that a) kills type-safety/intellisense at the DX level; and b) forces devs to go to gravity to see just _what_ they can update.

So yah, ideally, it looks something like this:

https://github.com/artsy/metaphysics/blob/main/src/schema/v2/startIdentityVerificationMutation.ts#L15-L23

and each of those subfields don't necessarily need to be NonNull -- because we're probably not updating them all at once, every single one. flexiblity is key. So basically just the ID needs to be nonnull

damassi: Well the validation happens in MP, its our "ideal user interface" for data, so we can do whatever. 

If there are three settings on the preference page, and those are clear, we can encode them here. Under no circumstances could an MP consumer send things we don't necessarily want, and when we're looking for what we can change, its there and self-documented GraphiQL docs. Feel free to follow the UI. As we add new UI we can update this concrete type.

This was something that came up recently with the Batch Import app -- JSON types in the schema are great for moving fast on a POC, but as soon as anything real starts to emerge they're unwound in favor of type-safety and explicitness, regardless of whatever gravity is doing.

src/schema/v2/partner/updatePartnerFlagsMutation.ts
```diff
@@ -70,12 +85,38 @@ export const updatePartnerFlagsMutation = mutationWithClientMutationId<
...
+        flags["artworks_default_metric"] = artworksDefaultMetric
+      }
+      
+      if (artworksDefaultCurrency !== undefined) {
+        flags["artworks_default_currency"] = artworksDefaultCurrency
+      }
+      
+      if (artworksDefaultPartnerLocationId !== undefined) {
+        flags["artworks_default_partner_location_id"] = artworksDefaultPartnerLocationId
+      }
```
damassi, line 118: #non-blocking: This can probably be made a bit more MP conventional tho, something like:

```tsx
const gravityArgs = {
  flags: {
    inquire_availability_price_display_enabled_by_partner: inquireAvailabilityPriceDisplayEnabledByPartner,
    artworks_default_partner_location_id: artworksDefaultPartnerLocationId
    ...
  }
}
```

and just fallback on js treating non-values as undefined by default. No need to guard everything.

jonallured: Hrm but here's the logic:

```
value => update key to that value
null => unset key
undefined => skip it
```

src/schema/v2/partner/updatePartnerFlagsMutation.ts
jonallured: More chatter here: https://artsy.slack.com/archives/C05F8TNKGAV/p1744658179386129

But yeah I updated the PR to be less general and instead hard-coded the partner flags that will be used in the CMS page that I'm migrating.

src/schema/v2/partner/updatePartnerFlagsMutation.ts
damassi: 👍

damassi (approved):

src/schema/v2/partner/updatePartnerFlagsMutation.ts
jpotts244: Hmm is it possible to send `null` over from the UI side? I think they are handled as dropdowns?

damassi: Sorry, hard to know whats a claudism (!== undefined is unusual and never seen in javascript code) vs what's idiomatic in the codebase. 

Slightly burned by some of the stuff that's happening via claude over in /artist2 so trying to at least call out abnormalities, even if its functionally correct.

PR #6635, opened 2025-04-15
feat(onyx-1663): purchasesConnection

nickskalkin:
As part of our effort to finalize the Torque deprecation, we are migrating the Purchases section to Forque. To display the table, we need a new field called `purchasesConnection`:

```graphql
{
  purchasesConnection(first: 10) {
    edges {
      node {
        internalID
        user {
          internalID 
        }
        artwork {
          internalID
        }
        saleDate
        salePrice
      }
    }
  }
}
```

### DIFF
--- a/src/schema/v2/purchases.ts
+++ b/src/schema/v2/purchases.ts
@@ -1,0 +1,127 @@
+import type { ResolverContext } from "types/graphql"
+import { GraphQLString, GraphQLFloat, GraphQLObjectType } from "graphql"
+import type { GraphQLFieldConfig } from "graphql"
+import { IDFields, NodeInterface } from "./object_identification"
+import { ArtworkType } from "./artwork"
+import {
+  connectionWithCursorInfo,
+  createPageCursors,
+} from "schema/v2/fields/pagination"
+import { date } from "./fields/date"
+import { convertConnectionArgsToGravityArgs } from "lib/helpers"
+import { connectionFromArraySlice } from "graphql-relay"
+import { pageable } from "relay-cursor-paging"
+import { UserType } from "./user"
+import { FairType } from "./fair"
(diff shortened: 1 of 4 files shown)


src/schema/v2/purchases.ts
```diff
@@ -0,0 +1,127 @@
...
+      type: UserType,
+    },
+  },
+})
+
+export const PurchasesConnection: GraphQLFieldConfig<void, ResolverContext> = {
+  type: connectionWithCursorInfo({
+    name: "Purchases",
+    nodeType: PurchaseType,
+  }).connectionType,
```
anandaroop, line 72: This connection definition mostly looks right to me, but it diverges a bit from our pagination recommendation in [pagination.md](https://github.com/artsy/metaphysics/blob/main/docs/pagination.md) — is that intentional?

src/schema/v2/purchases.ts
```diff
@@ -0,0 +1,127 @@
...
+      resolve: ({ artsy_commission }) => artsy_commission,
+    },
+    artwork: {
+      type: ArtworkType,
+    },
+    createdAt: date(({ created_at }) => created_at),
+    discoverAdmin: {
+      type: UserType,
+      resolve: ({ discover_admin }) => discover_admin,
+    },
```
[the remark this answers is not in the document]
nickskalkin: Thanks! Copied descriptions from Gravity

src/schema/v2/purchases.ts
nickskalkin: Good catch! At first I couldn't notice the difference to be honest 😅  I've added page and size options. I've also added paginationResolver to the end, or do I miss something else?

anandaroop (approved):
Looks good now 👍🏽

PR #6636, opened 2025-04-15
chore: Enable filters for bulk update artworks metadata

lidimayra:
These filters (artwork ids, partner artist and location) are already available on the Gravity endpoint. Changes here allow them on the mutation as well.

cc @artsy/amber-devs

### DIFF
--- a/src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
+++ b/src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
@@ -22,6 +22,11 @@ interface Input {
     category: string | null
     price_listed: number | null
   } | null
+  filters: {
+    artwork_ids: string[] | null
+    location_id: string | null
+    partner_artist_id: string | null
+  } | null
 }
 
 const BulkUpdateArtworksMetadataInput = new GraphQLInputObjectType({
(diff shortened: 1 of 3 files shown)


damassi (dismissed):
One comment about following MP idioms, but I'd certainly not block it

src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
```diff
@@ -42,6 +47,24 @@ const BulkUpdateArtworksMetadataInput = new GraphQLInputObjectType({
   },
 })
 
+const BulkUpdateArtworksFilterInput = new GraphQLInputObjectType({
+  name: "BulkUpdateArtworksFilterInput",
+  fields: {
+    artwork_ids: {
```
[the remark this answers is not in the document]
lidimayra: Oops, forgot about this! Fixed it in [2891901](https://github.com/artsy/metaphysics/pull/6636/commits/28919018f3cd171fb772825149d3062945c9533f), thanks for bringing this up! 🙏

damassi (approved):
🙏

src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
```diff
@@ -164,9 +164,22 @@ export const bulkUpdateArtworksMetadataMutation = mutationWithClientMutationId<
     { id, metadata, filters },
     { updatePartnerArtworksMetadataLoader }
   ) => {
-    const gravityOptions = {
-      metadata: metadata,
-      filters: filters,
+    const gravityOptions: any = {}
```
lidimayra, line 167: Is this `any` here alright? Would it be better to come up with `GravityOptions` interface for defining this? It feels like a special case (in most queries/mutations, we do not have these wrapped objects at this point, wonder what would the ideal here?

damassi: IMO not necessary -- we're defining the contract via input args via our mutation, which makes things safe on the consumer side (for the most part). To double type it in this way feels kinda like overkill, and we don't typically do it. 

Here's an example that types the inputs for a function that formats the gravity args -- this is mostly for DX: https://github.com/artsy/metaphysics/blob/main/src/schema/v2/artwork/updateArtworkMutation.ts#L187

So you can see how sometimes it makes sense -- especially for inputs that are quite long, and so on. But for a few args here and there, its a use yr best judgement kinda thing 👍

PR #6638, opened 2025-04-15
chore: Add mutation to update partners

jonallured:
This PR adds a new mutation for updating a partner record in the API. I exported the documentation for that REST endpoint and used that as a starting point for this mutation. I asked Claude to follow our typical mutation patterns and I think what emerged looks pretty good but lmk if there's anything that seems off!

/cc @artsy/amber-devs


damassi (approved):
Looks great 👍

PR #6653, opened 2025-04-18
feat(EMI-2208): Rich phone number field for `Order.fulfillmentDetails` type

erikdstock:
type: **Feat**
relates to [EMI-2208] and artsy/force#15460

This PR adds the phone number type to an order's saved `fulfillmentDetails` object. I made minimal modifications to the existing type but it might deserve a larger conversation to make sure this wouldn't break something.

This assumes we would store country codes as the alpha2 code (technically case doesn't matter to the phone validation layer), not the `+{number}` code. This matches the behavior of the settings/shipping address modal which uses downcased alpha2.

This also includes a breaking schema change to the existing field it is replacing but I don't think we use it yet.

[EMI-2208]: https://artsyproduct.atlassian.net/browse/EMI-2208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11975,7 +11975,9 @@ type FulfillmentDetails {
 
   # Country code of the buyer's phone number
   phoneNumberCountryCode: String
-    @deprecated(reason: "Use `order.fulfillmentDetails.phoneNumber.regionCode`")
+    @deprecated(
+      reason: "Use `phoneNumber.regionCode` for the alpha-2 country code or phoneNumber.countryCode for the numeric country code"
+    )
 
   # Shipping address postal code
   postalCode: String
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -11971,10 +11971,13 @@ type FulfillmentDetails {
   name: String
 
   # Phone number of the buyer
-  phoneNumber: String
+  phoneNumber: PhoneNumberType
 
   # Country code of the buyer's phone number
   phoneNumberCountryCode: String
+    @deprecated(
+      reason: "Use `order.fulfillmentDetails.phoneNumber.countryCode`"
+    )
 
   # Shipping address postal code
   postalCode: String
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -17951,6 +17951,7 @@ enum PhoneNumberFormats {
 }
 
 type PhoneNumberType {
+  # Numeric phone number country code
   countryCode: String
 
   # A formatted phone number if the number could be parsed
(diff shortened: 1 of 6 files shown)


src/schema/v2/order/__tests__/MeOrder.test.ts
```diff
@@ -348,5 +352,207 @@ describe("Me", () => {
...
+                  }
+                }
+              }
+            }
+          }
+        `
+        const result = await runAuthenticatedQuery(query, context)
+        expect(result.me.order.fulfillmentDetails.phoneNumber).toEqual({
+          countryCode: "1",
+          regionCode: null,
```
starsirius, line 440: Is it confusing that the persisted "us" is not returned in `regionCode` but "1" is returned in the `countryCode`?

src/schema/v2/phoneNumber.ts
```diff
@@ -108,12 +108,36 @@ const PhoneNumberType: GraphQLObjectType<
       args: {
         format: PhoneNumberFormats,
       },
+      description: "A formatted phone number. Null if isValid is false",
       resolve: ({ parsedPhone, phoneUtil }, { format }) =>
-        parsedPhone && phoneUtil.format(parsedPhone, format),
+        parsedPhone && phoneUtil.isValidNumber(parsedPhone)
+          ? phoneUtil.format(parsedPhone, format)
+          : null,
```
starsirius, line 115: Is this a behavioral change?

_schemaV2.graphql
```diff
@@ -11925,10 +11925,13 @@ type FulfillmentDetails {
   name: String
 
   # Phone number of the buyer
-  phoneNumber: String
+  phoneNumber: PhoneNumberType
```
starsirius, line 11928: 🎉

_schemaV2.graphql
```diff
@@ -11925,10 +11925,13 @@ type FulfillmentDetails {
...
 
   # Country code of the buyer's phone number
   phoneNumberCountryCode: String
+    @deprecated(
+      reason: "Use `order.fulfillmentDetails.phoneNumber.countryCode`"
```
starsirius, line 11933: I think it's actually `regionCode`.

src/schema/v2/phoneNumber.ts
erikdstock: youre right. we don't need this change. an invalid number can still be parsed and show while being invalid:
image
image

I'll update this.

src/schema/v2/order/__tests__/MeOrder.test.ts
erikdstock: yes... yes it is.

erikdstock: yes. I think this was actually the library not working as expected. it seems like if the library 

Good number, the backing code works:
image

Bad number it doesn't - unless we use the country code (+1) to get the region code ("us")
image
image

PR #6659, opened 2025-04-21
feat(settingsv2): Exposes mutation to update partner profile type images

jpotts244:
Exposes mutation to update partner profile type images
 
 Wires in Gravity's new endpoint ~[here](https://github.com/artsy/gravity/pull/18850/files)~ [here](https://github.com/artsy/gravity/pull/18852)
 
 Now using `partnerId` on the input type and returning the full partner object (`owner`) from the resolver
 
 Ex:
 
Screenshot 2025-04-22 at 9 17 24 AM

### DIFF
--- a/src/schema/v2/partner/Settings/__tests__/updatePartnerProfileImageMutation.test.js
+++ b/src/schema/v2/partner/Settings/__tests__/updatePartnerProfileImageMutation.test.js
@@ -6,16 +6,16 @@ describe("UpdatePartnerProfileImageMutation", () => {
     mutation {
       updatePartnerProfileImage(
         input: {
-          profileId: "foo"
+          partnerId: "foo"
           type: "icon"
           remoteImageS3Key: "s3-key-here"
           remoteImageS3Bucket: "s3-bucket-here"
         }
       ) {
-        imageOrError {
+        partnerOrError {
           __typename
           ... on UpdatePartnerProfileImageSuccess {
-            image {
+            partner {
               internalID
             }
           }
@@ -33,7 +33,10 @@ describe("UpdatePartnerProfileImageMutation", () => {
     const context = {
       updatePartnerProfileImageLoader: () =>
         Promise.resolve({
-          id: "newly-created-image-id",
+          id: "profile-id",
+          owner: {
+            _id: "foo",
+          },
         }),
     }
 
@@ -41,10 +44,10 @@ describe("UpdatePartnerProfileImageMutation", () => {
 
     expect(updatedProfile).toEqual({
       updatePartnerProfileImage: {
-        imageOrError: {
+        partnerOrError: {
           __typename: "UpdatePartnerProfileImageSuccess",
-          image: {
-            internalID: "newly-created-image-id",
+          partner: {
+            internalID: "foo",
           },
         },
       },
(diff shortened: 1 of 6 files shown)


damassi (dismissed):

jpotts244:
Just tested this and it works!!

Screenshot 2025-04-21 at 3 59 19 PM


I'm going to go in and change the return type to `image` instead of profile for simplicity

src/schema/v2/partner/Settings/updatePartnerProfileImageMutation.ts
```diff
@@ -0,0 +1,105 @@
...
+  type: string
+  remoteImageS3Key: string
+  remoteImageS3Bucket: string
+}
+
+const SuccessType = new GraphQLObjectType<any, ResolverContext>({
+  name: "UpdatePartnerProfileImageSuccess",
+  isTypeOf: (data) => !!data.id,
+  fields: () => ({
+    image: {
```
damassi, line 26: I _think_ you'd want `partner` here just to refresh the UI after the image is done uploaded. Totally fine!

damassi (approved):
One comment, but if that case is handled in yr thinking then all good 👍

src/schema/v2/partner/Settings/updatePartnerProfileImageMutation.ts
damassi: (Also kinda nice / easy conceptually -- just being like: i've updated some attribute of a thing; now return the thing -- because i might want to refresh something _else_ related to the thing after the attribute of the thing was updated, which is common. Things in things)

damassi: Cool 👍 trying to think through how it would be possible to refresh the page without that ID, but I don't think its possible

mzikherman: Re: this discussion, could even accept a `partnerID` as the input argument for the mutation instead of profile id. That might be more convenient (one less layer of abstraction from partner to profile that the client needs to know).

And/or, the gravity API could return the `profile/owner` back instead of or along with the image if returning just the 'image' from the gravity API isn't helpful.

jpotts244: Ah `partnerID`, good idea, that sounds clearer that this, I'll adjust the api

jpotts244: WDYT: https://github.com/artsy/gravity/pull/18852

src/schema/v2/partner/Settings/__tests__/updatePartnerProfileImageMutation.test.js
```diff
@@ -0,0 +1,82 @@
...
+          partnerId: "foo"
+          type: "icon"
+          remoteImageS3Key: "s3-key-here"
+          remoteImageS3Bucket: "s3-bucket-here"
+        }
+      ) {
+        partnerOrError {
+          __typename
+          ... on UpdatePartnerProfileImageSuccess {
+            partner {
```
damassi, line 18: 👍

damassi (commented):
🚢

damassi (approved):

PR #6662, opened 2025-04-22
chore: Add mutation to update profiles

jonallured:
This PR adds a basic mutation to update a `Profile` record over in Gravity. I used Claude to come up with a plan to add this mutation as it did over on https://github.com/artsy/metaphysics/pull/6632 and https://github.com/artsy/metaphysics/pull/6638 but something seems to have gone a little sideways. Will fix what's wrong and get this back on track!

/cc @artsy/amber-devs

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,24 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Commands
+- Build: `yarn build`
+- Dev server: `yarn dev` or `yarn verbose-dev` for more logs
+- Lint: `yarn lint` (fix with `yarn lint:fix`)
+- Type check: `yarn type-check`
+- Test all: `yarn test`
+- Test single file: `yarn jest path/to/file.test.ts`
+- Watch tests: `yarn jest --watch`
+- Format code: `yarn prettier-project`
+
+## Code Style
+- TypeScript with strict typing
+- Prettier formatting: no semicolons, double quotes, trailing commas
+- Use camelCase for variables, PascalCase for types/interfaces
+- Avoid any when possible, but allowed when necessary
+- Unused variables should be prefixed with `_`
+- Jest for testing with descriptive test names
+- GraphQL queries use the `gql` tag from lib/gql
+- Organize imports logically (internal vs external)
+- Follow React best practices for component structure
--- a/src/schema/v2/profile/updateProfileMutation.ts
+++ b/src/schema/v2/profile/updateProfileMutation.ts
@@ -1,0 +1,135 @@
+import {
+  GraphQLString,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLUnionType,
+  GraphQLBoolean,
+} from "graphql"
+import { mutationWithClientMutationId } from "graphql-relay"
+import {
+  formatGravityError,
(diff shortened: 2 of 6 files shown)


src/schema/v2/profile/updateProfileMutation.ts
```diff
@@ -0,0 +1,142 @@
...
+    },
+    location: {
+      type: GraphQLString,
+      description: "Location.",
+    },
+    isPrivate: {
+      type: GraphQLBoolean,
+      description: "Private profiles hide certain features for non admins.",
+    },
+    menuColorClass: {
```
jpotts244, line 90: ooo who is this `menuColorClass`? I dont see any reference in the UI?

src/schema/v2/profile/index.ts
```diff
@@ -0,0 +1,3 @@
+import { updateProfileMutation as _updateProfileMutation } from "./updateProfileMutation"
```
jpotts244, line 1: I dont think this index file is needed?

We should just be able to use the [exported Mutation](https://github.com/artsy/metaphysics/pull/6662/files#diff-75ada1fd5dfdc7e06e9f55082a7b050e62ca10b719608401e9ce68a5d06740feR54) in the schema as youve dont below without this!

src/schema/v2/profile/updateProfileMutation.ts
jpotts244: I dont see anyone setting this value in my very swift and biased [search](https://github.com/search?q=org%3Aartsy+menu_color_class&type=code) maybe we can ✂️

src/schema/v2/profile/index.ts
damassi: Claude is drunk 😆

damassi: But yep 👍 As an aside, we've been moving away from index files for a few years now, whatever the context (and not sure about this weird _aliasing that's going on)

src/schema/v2/profile/updateProfileMutation.ts
jonallured: Removed!

src/schema/v2/profile/index.ts
jonallured: Removed!

jonallured:
Ok so I _think_ part of the problem may have been that Claude had not been inited on this project so I did that and asked it to address the PR feedback as well. This is back to green now. But just wanted to call your attention to that Claude MD file @damassi in case you have any comments on what it generated.

jpotts244 (dismissed):

CLAUDE.md
```diff
@@ -0,0 +1,25 @@
...
+- Test all: `yarn test`
+- Test single file: `yarn jest path/to/file.test.ts`
+- Watch tests: `yarn jest --watch`
+- Format code: `yarn prettier-project`
+
+## Code Style
+- TypeScript with strict typing
+- Prettier formatting: no semicolons, double quotes, trailing commas
+- Use camelCase for variables, PascalCase for types/interfaces
+- Prefix interfaces with `I` when appropriate
```
damassi, line 19: Probably don't need this one ^

damassi (dismissed):
Looks good! 👍

CLAUDE.md
jonallured: Removed!

PR #6668, opened 2025-04-23
feat(onyx-1669): add purchase root field

nickskalkin:
> [!NOTE]  
> Temporarily depends on https://github.com/artsy/metaphysics/pull/6635

As part of our effort to finalize the Torque deprecation, we are migrating the Purchases section to Forque. This PR adds `purchase(id: ...)` root field so that we could render /purchase/:id page there

```graphql
{
  purchase(id: "purchase-id") {
    internalID
}
```

### DIFF
--- a/src/schema/v2/__tests__/purchase.test.ts
+++ b/src/schema/v2/__tests__/purchase.test.ts
@@ -1,0 +1,121 @@
+import { runQuery } from "schema/v2/test/utils"
+import gql from "lib/gql"
+import sinon from "sinon"
+
+describe("purchase", () => {
+  const purchaseMockData = {
+    id: "purchase-id",
+    artsy_commission: 25.5,
+    artwork: { _id: "artwork-id" },
+    created_at: "2025-04-14T00:00:00Z",
+    discover_admin: { id: "discover-user-id" },
+    email: "email@example.com",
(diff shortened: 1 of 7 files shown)


src/schema/v2/__tests__/purchase.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  it("resolves a purchase", async () => {
+    const context = {
+      purchaseLoader: sinon
+        .stub()
+        .withArgs({
+          id: "purchase-id",
+        })
+        .returns(Promise.resolve(purchaseMockData)),
+    }
```
anandaroop, line 32: We don't really use Sinon these days  — so I'm wondering if we grabbed this from an older field/spec? 

**suggestion:** Let's use Jest mocks instead, e.g.:

```ts
const context = {
  purchaseLoader: jest.fn().mockResolvedValueOnce(purchaseMockData),
};
```

Or, if you want to check the args before returning:

```ts
const context = {
  purchaseLoader: jest.fn().mockImplementation(args => {…}),
};
```

Something like ^that…

anandaroop (dismissed):
Looks good, but one suggestion re: test style

nickskalkin:
Going to close this one and open another PR with fixes @anandaroop

PR #6672, opened 2025-04-24
feat(onyx-1669): add purchase root field

nickskalkin:
As part of our effort to finalize the Torque deprecation, we are migrating the Purchases section to Forque. This PR adds `purchase(id: ...)` root field so that we could render /purchase/:id page there

```graphql
{
  purchase(id: "purchase-id") {
    internalID
}
```


nickskalkin:
It was approved here https://github.com/artsy/metaphysics/pull/6668, going to merge without adding reviewers

PR #6676, opened 2025-04-24
chore: update home view tests after removing discover daily feature flag

iskounen:
These tests started failing when we deprecated the Discover Daily feature flag because that section was hidden by default. This PR updates these tests to look at the `card` title instead of the `component` title when a section is a HomeViewSectionCard type. (This also impacted the assertion for "Galleries Near You".)


anandaroop (approved):
👍🏽

PR #6683, opened 2025-04-28
feat: implement QuickLinks v2

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-1689

- Implements the new selection of Quick Links names and destinations for the v2 iteration.
- Places this behind the [onyx_enable-quick-links-v2](https://unleash.artsy.net/projects/default/features/onyx_enable-quick-links-v2) Unleash flag (enabled in staging)

When Eigen receives the updated response (and [supports the newest icons](https://github.com/artsy/eigen/pull/12003) ) it will look like this:

pills

### DIFF
--- a/src/schema/v2/homeView/sections/QuickLinks.ts
+++ b/src/schema/v2/homeView/sections/QuickLinks.ts
@@ -4,6 +4,7 @@ import { HomeViewSectionTypeNames } from "../sectionTypes/names"
 import type { NavigationPill } from "../sectionTypes/NavigationPills"
 import { ResolverContext } from "types/graphql"
 import { getEigenVersionNumber, isAtLeastVersion } from "lib/semanticVersioning"
+import { isFeatureFlagEnabled } from "lib/featureFlags"
 
 export const QuickLinks: HomeViewSection = {
   id: "home-view-section-quick-links",
@@ -17,7 +18,11 @@ export const QuickLinks: HomeViewSection = {
 }
 
 function getDisplayableQuickLinks(context: ResolverContext) {
-  return QUICK_LINKS.filter((quickLink) => {
+  const quickLinks = isFeatureFlagEnabled("onyx_enable-quick-links-v2")
+    ? QUICK_LINKS_V2
+    : QUICK_LINKS
+
+  return quickLinks.filter((quickLink) => {
     let isDisplayable = true
     const actualEigenVersion = getEigenVersionNumber(
       context.userAgent as string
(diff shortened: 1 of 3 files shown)


src/schema/v2/homeView/sections/QuickLinks.ts
```diff
@@ -32,6 +37,64 @@ function getDisplayableQuickLinks(context: ResolverContext) {
...
 }
 
+export const QUICK_LINKS_V2: Array<NavigationPill> = [
+  {
+    title: "Discover Daily",
+    href: "/infinite-discovery",
+    ownerType: OwnerType.infiniteDiscovery,
+    icon: "ImageSetIcon",
+    minimumEigenVersion: { major: 8, minor: 67, patch: 0 }, // same constraint as InfiniteDiscovery section
+  },
```
anandaroop, line 47: @iskounen Since we are now going to link to Discover Daily from the quick links, I want to double-check this version constraint.

src/schema/v2/homeView/sections/QuickLinks.ts
```diff
@@ -32,6 +37,64 @@ function getDisplayableQuickLinks(context: ResolverContext) {
...
+    href: "/collection/paintings",
+    ownerType: OwnerType.collection,
+    icon: "ArtworkIcon",
+  },
+  {
+    title: "Galleries for You",
+    href: "galleries-for-you",
+    ownerType: OwnerType.galleriesForYou,
+    icon: "InstitutionIcon",
+  },
```
anandaroop, line 83: Note that this is one of two icons — `InstitutionIcon` and `ImageSetIcon` — that will need to be supported in the corresponding Eigen component. 

That will be in a separate PR (and in the meantime it's safe to release this, because the pills will simply render without those two icons)

src/schema/v2/homeView/sections/QuickLinks.ts
```diff
@@ -32,6 +37,64 @@ function getDisplayableQuickLinks(context: ResolverContext) {
...
+  {
+    title: "Shows for You",
+    href: "/shows-for-you",
+    ownerType: OwnerType.shows,
+    icon: undefined,
```
olerichter00, line 88: 👀

olerichter00 (approved):
🌟

src/schema/v2/homeView/sections/QuickLinks.ts
MounirDhahri: this is a nice addition but because we were careful in Eigen as well, we should be safe also without this check

MounirDhahri (approved):
Nice work! also good thought to be careful here and hide this behind a ff

nickskalkin (approved):

src/schema/v2/homeView/sections/QuickLinks.ts
iskounen: If you're looking to remove that minimum version, I think it's OK because enough users are on a newer version than that.

anandaroop: Sounds like it's fine either way, so I'm just gonna merge as-is. Thanks!

PR #6694, opened 2025-05-01
feat: add Your Bids quick link for users with active bids

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-1691

For Quick Links v2, this Implements the conditional **Your Bids** link that is shown only to users with active bids:

```js
{
  title: "Your Bids",
  href: "/inbox",
  ownerType: OwnerType.inbox,
  icon: undefined,
}
```

Tapping on the pill takes you to `/inbox`, which in turn defaults to the **Bids** view if there are active bids.

| Pill | Destination |
|--------|--------|
|  |  |

### DIFF
--- a/src/schema/v2/homeView/sections/QuickLinks.ts
+++ b/src/schema/v2/homeView/sections/QuickLinks.ts
@@ -12,8 +12,10 @@ export const QuickLinks: HomeViewSection = {
   ownerType: OwnerType.quickLinks,
   type: HomeViewSectionTypeNames.HomeViewSectionNavigationPills,
   requiresAuthentication: true,
-  resolver: (_parent, _args, context, _info) => {
-    return getDisplayableQuickLinks(context)
+  resolver: async (_parent, _args, context, _info) => {
+    let links = getDisplayableQuickLinks(context)
+    links = await maybeInsertYourBidsLink(links, context)
+    return links
   },
 }
 
@@ -51,6 +53,9 @@ export const QUICK_LINKS_V2: Array<NavigationPill> = [
     ownerType: OwnerType.auctions,
     icon: "GavelIcon",
   },
+  /*
+   * see below for optional Your Bids link inserted here
+   */
   {
     title: "New This Week",
     href: "/collection/new-this-week",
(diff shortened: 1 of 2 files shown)


src/schema/v2/homeView/sections/QuickLinks.ts
```diff
@@ -146,3 +151,26 @@ export const QUICK_LINKS: Array<NavigationPill> = [
...
+
+async function maybeInsertYourBidsLink(
+  links: NavigationPill[],
+  context: ResolverContext
+) {
+  if (isFeatureFlagEnabled("onyx_enable-quick-links-v2")) {
+    const { MyBids } = require("schema/v2/me/myBids")
+    const bids = await MyBids.resolve?.({}, {}, context, {} as any)
+    const hasBids = bids?.active?.length > 0
+    const auctionsIndex = links.findIndex((link) => link.href === "/auctions")
```
anandaroop, line 163: Noting that once this is enabled, this `MyBids.resolve(…)` check will be performed on every home view request. We should keep an eye on performance.

src/schema/v2/homeView/sections/QuickLinks.ts
```diff
@@ -146,3 +151,26 @@ export const QUICK_LINKS: Array<NavigationPill> = [
     icon: "FairIcon",
   },
 ]
+
+async function maybeInsertYourBidsLink(
```
dariakoko, line 155: **minor:** maybe `shouldInsertYourBidsLink` could be a better name for this function?

dariakoko (dismissed):

src/schema/v2/homeView/sections/QuickLinks.ts
```diff
@@ -146,3 +151,26 @@ export const QUICK_LINKS: Array<NavigationPill> = [
...
+
+    if (hasBids && auctionsIndex >= 0) {
+      const yourBids: NavigationPill = {
+        title: "Your Bids",
+        href: "/inbox",
+        ownerType: OwnerType.inbox,
+        icon: undefined,
+      }
+      links.splice(auctionsIndex + 1, 0, yourBids)
+    }
```
olerichter00, line 173: A comment would be helpful to explain where and under what conditions the navigation pill is inserted.

src/schema/v2/homeView/sections/QuickLinks.ts
anandaroop: If this were a predicate function I would agree, but since this actually mutates the list, this naming convention is pretty standard (we [use it a lot](https://github.com/search?q=org%3Aartsy+%2Fmaybe%5Cw%7B5%2C%7D%2F&type=code), in fact)

PR #6702, opened 2025-05-02
feat: add displayTexts and resolver for titleText to me.order

oxaudo:
Most of the changes here is just alphabetizing properties as I had hard time tracking them otherwise:)

The real change is adding DisplayTexts type that only include titleText for now but will add strings needed for the message.

Debating if I should try to cover each state in specs but feel like those would be pretty performative now. Maybe just add cases if we encounter bugs? I'll sleep on it:)

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10547,6 +10547,12 @@ type DismissTaskSuccess {
   task: Task!
 }
 
+# Display texts for the order based on its state
+type DisplayTexts {
+  # Title text to display for the order
+  titleText: String!
+}
+
 type DoNotUseThisPartner {
   displayName: String
   display_name: String @deprecated(reason: "Use displayName")
@@ -16388,6 +16394,9 @@ type Order {
 
   # Order code
   code: String!
+
+  # Display texts for the order based on its buyer_state
+  displayTexts: DisplayTexts!
 
   # Buyer fulfillment details for order
   fulfillmentDetails: FulfillmentDetails
--- a/src/schema/v2/order/sharedOrderTypes.ts
+++ b/src/schema/v2/order/sharedOrderTypes.ts
@@ -43,6 +43,8 @@ interface OrderJSON {
   shipping_region?: string
   shipping_country?: string
   tax_total_cents?: number
+  buyer_state?: string
+  fulfillment_type?: string
   fulfillment_options: Array<{
     type: string
     amount_minor: number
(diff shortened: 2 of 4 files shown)


_schemaV2.graphql
```diff
@@ -10547,6 +10547,12 @@ type DismissTaskSuccess {
   task: Task!
 }
 
+# Display texts for the order based on its state
+type DisplayTexts {
+  # Title text to display for the order
+  titleText: String!
+}
+
```
erikdstock, line 10555: If we have a solid buyer state type then I think we can keep a completely visual concern like a text string on the client. For example we might show something different in a text view based on the viewport size.

src/schema/v2/order/sharedOrderTypes.ts
```diff
@@ -269,25 +271,63 @@ const SellerType = new GraphQLUnionType({
...
+    },
+  },
+})
+
+const resolveDisplayTexts = (order: OrderJSON) => {
+  switch (order.buyer_state) {
+    case "submitted":
+      return {
+        titleText: "Great choice!",
+      }
```
erikdstock, line 290: I really think `buyerState` should be the main field that comes from the order's `buyer_state`. It could be a graphql string enum (eg `SUBMITTED | PAYMENT_FAILED | SHIPPED`) or a richer graphql union ` {__typename: 'Shipped', trackingNumber: "asdf1234", carrier: UPS }`. This would let us give clients everything they need to build the rest up.

_schemaV2.graphql
oxaudo: We already have at least 2 places Force and Eigen where we need to show exact same strings. And the message part will have a bit more logic in integrating other order properties (shipping/dates) into the display. The idea here is that we don't have to repeat this code on the client and will just keep adding whatever needed here in this DisplayTexts.

src/schema/v2/order/sharedOrderTypes.ts
erikdstock: I don't think I understand the problem or how putting copy in metaphysics solves the problem.

My thinking is that a buyer state should be a clear situation. Just from the name of that state as a buyer I should know if I need to wait
- 'the order is submitted [and waiting for seller response]'
- 'the order is approved [and waiting for shipment]'
- 'the order is approved [and payment is processing]'

or if I need to do something
- 'the payment failed [and i need to update it]'
- 'i have a counteroffer [and i need to respond]'
- 'the order is approved [and i need to submit a wire transfer]'

All of these things can be communicated with a single value and can imply the existence of any related data like payment type. In any case, sending strings from metaphysics presents challenges for clients that have to blindly render some string. Let's talk in person when we all have a chance.

oxaudo: That is exactly what we are going for here with this states. And this part is the logic to generate the string in one spot so that can be shared between Eigen and Force.

PR #6721, opened 2025-05-07
feat(WIP): add more properties to order.DisplayTexts

oxaudo:
I was trying to think how to add this additional texts that will be displayed in message for each state and make it shareable between Force and Eigen. Basically we need to cover all states in this table: https://www.figma.com/board/3y8s4tNuSvq69cs4DCb0Mm/Status-Page-Copy-Audit?node-id=64-158&t=MNHfmSmdNhHOcDSZ-4. (there might be some more updates there based on our discussions about arta state). 

Here is my initial proposal.
I tried to go through all the cases and assign them to the appropriate `buyer_state` that we have created previously. And seems like there is some additional sharing between state if we break the whole message into paragraphs. Added those strings into this table: https://docs.google.com/spreadsheets/d/1GOd7UHdyeLOSJTJRppR30ZW2vXq4JwGq_wKE5FFAzAY/edit?gid=346921686#gid=346921686 (G,H,I column).

In this pr I'm adding simple implementation for `messageText1`. The value can be displayed as is I believe with one caveat that the links should be properly implemented. My intention here is that clients can look up links by appropriate `data-link` id and do the transform on the client. Not sure it's the best approach - happy to figure out alternatives. 

For the messageText2 (2nd paragraph) we'll need to get into more details about specific shipping. This will need to wait till https://github.com/artsy/exchange/pull/2558 is resolved. I'll also add a separate issue to define all the rules for available Arta shipping options.

This is def WIP - will need to cover more specs. But wonder if the approach seems reasonable to all: @MrSltun @erikdstock @starsirius @fladson @leamotta ?

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -16421,7 +16421,7 @@ type Order {
   # Order code
   code: String!
 
-  # Display texts for the order based on its buyer_state
+  # Display texts for the order based on its buyer_state and order shipping/payment states
   displayTexts: DisplayTexts!
 
   # Buyer fulfillment details for order
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10558,6 +10558,15 @@ type DismissTaskSuccess {
 
 # Display texts for the order based on its state
 type DisplayTexts {
+  # First paragraph of the message text for the order
+  messageText1(format: Format = MARKDOWN): String
+
+  # Second paragraph of the message text for the order
+  messageText2(format: Format = MARKDOWN): String
+
+  # Third paragraph of the message text for the order
+  messageText3(format: Format = MARKDOWN): String
+
   # Title text to display for the order
   titleText: String!
 }
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10559,13 +10559,7 @@ type DismissTaskSuccess {
 # Display texts for the order based on its state
 type DisplayTexts {
   # First paragraph of the message text for the order
-  messageText1(format: Format = MARKDOWN): String
-
-  # Second paragraph of the message text for the order
-  messageText2(format: Format = MARKDOWN): String
-
-  # Third paragraph of the message text for the order
-  messageText3(format: Format = MARKDOWN): String
+  messageText(format: Format = MARKDOWN): String
 
   # Title text to display for the order
   titleText: String!
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10556,7 +10556,7 @@ type DismissTaskSuccess {
   task: Task!
 }
 
-# Display texts for the order based on its state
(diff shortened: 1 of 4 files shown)


src/schema/v2/order/sharedOrderTypes.ts
```diff
@@ -288,47 +290,82 @@ const DisplayTextsType = new GraphQLObjectType<any, ResolverContext>({
...
-    default:
-      return {
-        titleText: "Your order",
-      }
+      titleText = "Your order was canceled"
+  }
+
+  return {
+    titleText: titleText,
+    messageText: messageParts.join("<br/><br/>"),
```
damassi, line 368: This is slightly faffy, but switches with returns are a bit nicer and more safe, since it precludes the ability to make mistakes around `break` statements (while also being a bit easier to read). 

Totally cool to just call `.join` when/where you need it (on messageParts), or define a helper

oxaudo:
Ok. Believe most is addressed here. Merging.

_schemaV2.graphql
```diff
@@ -10556,8 +10556,11 @@ type DismissTaskSuccess {
   task: Task!
 }
 
-# Display texts for the order based on its state
+# Display texts for the order based on its state and order shipping/payment states
 type DisplayTexts {
+  # First paragraph of the message text for the order
+  messageText(format: Format = MARKDOWN): String
```
starsirius, line 10562: This might be late but I feel `message` and `title` are pretty self explanatory (and scoped under `DisplayTexts` already) and more concise, without the `Text` suffix.

src/schema/v2/order/types/DisplayTexts.ts
```diff
@@ -0,0 +1,112 @@
...
+})
+
+export const DisplayTexts: GraphQLFieldConfig<OrderJSON, ResolverContext> = {
+  description:
+    "Display texts for the order based on its buyer_state and order shipping/payment states",
+  type: new GraphQLNonNull(DisplayTextsType),
+  resolve: (order) => resolveDisplayTexts(order),
+}
+
+const formatMessage = (parts: string[]) => parts.join("<br/><br/>")
```
starsirius, line 45: Looks like this helper is not necessary?

oxaudo: @starsirius  - It is used as my idea is to make message as an array or parts and those parts are joined by beak line. For now it's only one line there so it's a bit redundant - but I'll add to the message in a follow up pr. That is where we need to use shipping details so was waiting for the other exchange PR.

_schemaV2.graphql
oxaudo: Totally open to rename it. Not late as it's barely used for now and good to do it before Eigen actually start using it.

oxaudo (commented):
Will address the comments in a follow up Prs.

PR #6737, opened 2025-05-13
feat: update fields on displayTexts for the order

oxaudo:
This will be used for display for https://artsyproduct.atlassian.net/browse/EMI-2436

Currently only has one happy pass test here. We need to figure out more comprehensive strategy for testing here. Created placeholder story for testing part: https://artsyproduct.atlassian.net/jira/software/c/projects/EMI/boards/114

@starsirius - also changed the field names to drop the `test` part per your suggestion. It's a breaking change for Force page but as it's behind feature flag - don't think is a big issue. Will update force as soon as this is merged.

### DIFF
--- a/src/schema/v2/order/types/DisplayTexts.ts
+++ b/src/schema/v2/order/types/DisplayTexts.ts
@@ -8,17 +8,18 @@ import type { ResolverContext } from "types/graphql"
 import { OrderJSON } from "./exchangeJson"
 import { formatMarkdownValue } from "../../fields/markdown"
 import { FormatEnums } from "../../input_fields/format"
+import moment from "moment-timezone"
 
 const DisplayTextsType = new GraphQLObjectType<any, ResolverContext>({
   name: "DisplayTexts",
   description:
     "Display texts for the order based on its state and order shipping/payment states",
   fields: {
-    titleText: {
+    title: {
       type: new GraphQLNonNull(GraphQLString),
       description: "Title text to display for the order",
     },
-    messageText: {
+    message: {
       type: GraphQLString,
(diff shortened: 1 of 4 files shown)


src/schema/v2/order/types/DisplayTexts.ts
```diff
@@ -42,71 +43,123 @@ export const DisplayTexts: GraphQLFieldConfig<OrderJSON, ResolverContext> = {
...
+const formatMessage = (parts: string[], joinWith = "<br/><br/>") =>
+  parts.join(joinWith)
+
 const resolveDisplayTexts = (order: OrderJSON) => {
   const isPickup = order.fulfillment_type == "pickup"
+  const stateExpireTime =
+    order.buyer_state_expires_at && moment.tz(order.buyer_state_expires_at)
+  const formattedStateExpireTime =
+    stateExpireTime &&
+    `${stateExpireTime.format("MMM D")}, ${stateExpireTime.format("h:mma z")}`
```
starsirius, line 55: Do want to display in local time zone or the fixed America/New_York time zone? There is this [helper](https://github.com/artsy/metaphysics/blob/07be03ae492c0a87ddf97289474acfe295425bf2/src/lib/date.ts#L55) that can be reused, with slightly different format.

oxaudo: I looked at that one and was not sure if we want to reuse ir. We can follow up with exact formatting adjustments. If you are ok with general direction here - I'm inclined to merge it for now and have this going all the way to UI. So we can start sharing this with Eigen.

The good thing - adjustment will be just here if all works as intended :)

PR #6738, opened 2025-05-14
feat(DIA-1303): add hasFollows to profile

araujobarret:
Resolves [DIA-1303]

Shifts this logic that is being used in Eigen and Force to MP. Will create the PRs for the respective repos as soon as this one is approved.

[DIA-1303]: https://artsyproduct.atlassian.net/browse/DIA-1303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/schema/v2/profile.ts
```diff
@@ -30,6 +30,11 @@ export const ProfileType = new GraphQLObjectType<any, ResolverContext>({
         name: "ProfileCounts",
         fields: {
           follows: numeral(({ follows_count }) => follows_count),
+          hasFollows: {
+            type: GraphQLNonNull(GraphQLBoolean),
+            description: "If the profile has more than 500 follows",
+            resolve: ({ follows_count }) => follows_count > 500,
+          },
```
egdbear, line 37: **question:** could this naming be a bit misleading, since it only applies to partner follows? Maybe it could be moved or clarified elsewhere?

egdbear: Since we're already exposing the profile from the Partner type, and this logic only applies for partners. I’m wondering if that’s the right place for this kind of data?

egdbear: That works! Options like `hasSignificantFollows` could open up another can of worms, but this seems to be the all right 🙏

egdbear (approved):

src/schema/v2/profile.ts
joeyAghion: I'm also confused by this name, though I understand why we'd want to centralize its logic in Metaphysics. Maybe `visibleFollowsCount` (or `counts.visibleFollows` if within an existing counts object) would be more clear and useful?

PR #6740, opened 2025-05-14
chore: add back fields for automatic schema update

oxaudo:
This is just to get auto deploy possible. Will follow with actual removal after Force update.

### DIFF
--- a/src/schema/v2/order/types/DisplayTexts.ts
+++ b/src/schema/v2/order/types/DisplayTexts.ts
@@ -35,12 +35,18 @@ const DisplayTextsType = new GraphQLObjectType<any, ResolverContext>({
     },
     // TODO: deprecated - remove after deploy
     titleText: {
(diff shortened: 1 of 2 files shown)


src/schema/v2/order/types/DisplayTexts.ts
```diff
@@ -33,6 +33,21 @@ const DisplayTextsType = new GraphQLObjectType<any, ResolverContext>({
         return formatMarkdownValue(message, format)
       },
     },
+    // TODO: deprecated - remove after deploy
```
starsirius, line 36: I think there is a GraphQL way to [deprecate](https://github.com/artsy/metaphysics/blob/254b7dae27a1845391bab2fab04e38f84c62cb67/src/schema/v2/conversation/index.ts#L414) fields. Hopefully quick change and regenerate the schema.

PR #6743, opened 2025-05-15
feat: rename hasFollows to hasVisibleFollowsCount

araujobarret:
related to #6738 

Renames `Partner#hasFollows` to `hasVisibleFollowsCount`


joeyAghion:
This is better than `hasFollows`. Personally I'd still prefer combining the boolean and count into a single `visibleFollowsCount` or `counts.visibleFollows` for simplicity, but if this matches other similar fields in Metaphysics I'm OK with it.

PR #6746, opened 2025-05-15
chore: delete unused field as force schema was updated

oxaudo:
Quick clean up before we try to add some more experimental things to this type


MrSltun (approved):
🧹

PR #6748, opened 2025-05-15
feat: add messageType as an alternative to displayMessage

oxaudo:
Discussed with @MrSltun that it might be really hard (next to impossible) to share html between Force and Eigen. So we might need to repeat texts for the message part as html elements integrated there. 

However we still want to keep the logic of what message is applicable based on all the 'mode', 'payment', 'shipment' etc combinations and create a very specific type just for this messageDisplay.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10589,11 +10589,33 @@ type DismissTaskSuccess {
 
 # Display texts for the order based on its state and order shipping/payment states
 type DisplayTexts {
-  # First paragraph of the message text for the order
+  # Fomatted full message text for the order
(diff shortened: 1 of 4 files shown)


MrSltun (approved):
Looks great! 🌟 

But after reviewing [react-native-html-renderer](https://meliorence.github.io/react-native-render-html/) (the package we use to render html in Eigen), it seems that it could be possible to write custom renderers html content in Eigen, I'm fine with either approaches now :)

oxaudo:
I'll merge it to see how it'll look with the new way in force. 

Maybe let's bring this up for the group to see what wider group thinks?

PR #6751, opened 2025-05-16
feat: adds "Art Under $X" quick pill for users with price preferences

nickskalkin:
This implements the conditional "Art Under $X" link that is shown only to users with [computed price preferences](https://github.com/artsy/vortex/pull/709). For this feature I've added a new field under `me`, which is called `pricePreference`.

Price buckets are the same as on the website under the "What's new" dropdown:

Request:

```graphql
{
  me {
    pricePreference
  }
  
  homeView {
    sectionsConnection(first: 1) {
      edges {
        node {
          ... on HomeViewSectionNavigationPills {
            navigationPills {
              title
              href
            }
          }
        }
      }
    }
  }
}
```

Response:

```json
{
  "data": {
    "me": {
      "email": "nikita.skalkin@artsymail.com",
      "pricePreference": 250
    },
    "homeView": {
      "sectionsConnection": {
        "edges": [
          {
            "node": {
              "navigationPills": [
                {
                  "title": "Discover Daily",
                  "href": "/infinite-discovery"
                },
                {
                  "title": "Auctions",
                  "href": "/auctions"
                },
                {
                  "title": "New This Week",
                  "href": "/collection/new-this-week"
                },
                {
                  "title": "Art Under $500",
                  "href": "/collect?price_range=0-500"
                },
                ...
              ]
            }
          }
        ]
      }
    }
  }
}
```

### DIFF
--- a/src/schema/v2/homeView/helpers/priceBucketBasedOnPricePreference.ts
+++ b/src/schema/v2/homeView/helpers/priceBucketBasedOnPricePreference.ts
@@ -2,33 +2,34 @@ export const priceBucketBasedOnPricePreference = (pricePreference: number) => {
   // Price buckets are aligned with the dropdown options on the website's top menu's dropdown
   const priceBuckets = [
     { priceRange: "0-500", min: 0, max: 500, text: "Art Under $500" },
-    { priceRange: "501-1000", min: 501, max: 1000, text: "Art Under $1000" },
-    { priceRange: "1001-2500", min: 1001, max: 2500, text: "Art Under $2500" },
-    { priceRange: "2501-5000", min: 2501, max: 5000, text: "Art Under $5000" },
+    { priceRange: "500-1000", min: 500, max: 1000, text: "Art Under $1000" },
+    { priceRange: "1000-2500", min: 1000, max: 2500, text: "Art Under $2500" },
+    { priceRange: "2500-5000", min: 2500, max: 5000, text: "Art Under $5000" },
     {
-      priceRange: "5001-10000",
-      min: 5001,
+      priceRange: "5000-10000",
+      min: 5000,
       max: 10000,
       text: "Art Under $10000",
     },
     {
-      priceRange: "10001-25000",
-      min: 10001,
+      priceRange: "10000-25000",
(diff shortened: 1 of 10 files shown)


MounirDhahri (approved):
LGTM, we could even consider showing this based on app version if we don't want users to see the collect route on a webview

src/schema/v2/homeView/helpers/priceBucketBasedOnPricePreference.ts
```diff
@@ -0,0 +1,35 @@
...
+      max: Number.POSITIVE_INFINITY,
+      text: "Art Above $25000",
+    },
+  ]
+
+  if (pricePreference > 25000) return priceBuckets[priceBuckets.length - 1]
+
+  const priceBucket = priceBuckets.find(({ min, max }) => {
+    return pricePreference >= min && pricePreference <= max
+  })
```
anandaroop, line 32: **question:** Is `pricePreference` guaranteed to be an integer?

I think the logic here would fail, for example, to bucket a value of `500.5` or `1000.5` correctly.

If we are _absolutely_ certain that this value will always be an integer, then this might be safe.

But it might be better to make this a little more resilient by using [half-open intervals](https://chatgpt.com/share/682c9c9b-5a24-8004-b4f0-5a205a65f6b7) for buckets which fully span the range of numbers, i.e…

`0-500`
`500–1000`
etc

combined with logic something like… 

```js
pricePreference > min && pricePreference <= max  // note open interval on one side
```

anandaroop (commented):
Looks great 🎉  

Just one question about the bucketing logic…

anandaroop:
> we could even consider showing this based on app version if we don't want users to see the collect route on a webview 

Nice idea. If so, we have [precedent](https://github.com/artsy/metaphysics/blob/d19173a96e352b343350ed66109efddc5c6b6e24/src/schema/v2/homeView/sections/QuickLinks.ts#L48) for that.

nickskalkin:
- added min eigen version
- fixed price buckets logic to make them work with floating point numbers

PR #6753, opened 2025-05-19
feat: Allow flat fee update and published filter in bulk

lidimayra:
Relies on the changes added during:
https://github.com/artsy/gravity/pull/18926

Allow bulk operations to filter artworks by published status and allow domestic shipping fee to be updated in batch.

cc @artsy/amber-devs


olerichter00 (approved):
🌟

PR #6755, opened 2025-05-19
feat(DIA-1303): enable unauth access to the profile loader

araujobarret:
Remove auth necessity from `hasVisibleFollowsCount`.


egdbear:
Could you clarify why we're removing it? profileLoader is an authorized loader. Another option might be to use               `.catch(() => null)`

egdbear:
> Because this is the way it works, we expose the follow counts for unauthenticated sessions, we want to keep the same behaviour.

I don't see that we expose follow counts for unauthenticated sessions.. What I'm missing? 😅 
image

PR #6758, opened 2025-05-20
chore: Apply camelCase to flat fee argument

lidimayra:
Just noticed while working in Volt that I added the flat fee argument as snake_case. Replace it with camelCase instead.

cc @artsy/amber-devs


olerichter00 (approved):
🐫

Mitch-Henson (approved):
Happens to all of us rubyists 😅

PR #6759, opened 2025-05-20
chore: remove message from displayTexts as we are using  instead

oxaudo:
Unfortunately looks like sharing html straight out of MP for both Force and Eigen proved difficult. Would still require significant client code to format and interpret it properly and might make tracking and proper formatting more complicated.

So ended up creating the very 'custom' `messageType` here that will be mapped 1x1 to the view variations on client. And refactored it for Force.

Nothing else uses this `message` field so cleaning it up here to keep things simpler.

### DIFF
--- a/src/schema/v2/order/types/DisplayTexts.ts
+++ b/src/schema/v2/order/types/DisplayTexts.ts
@@ -7,9 +7,6 @@ import {
 } from "graphql"
 import type { ResolverContext } from "types/graphql"
 import { OrderJSON } from "./exchangeJson"
-import { formatMarkdownValue } from "../../fields/markdown"
-import { FormatEnums } from "../../input_fields/format"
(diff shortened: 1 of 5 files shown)


src/schema/v2/order/types/DisplayTexts.ts
```diff
@@ -228,22 +147,16 @@ const resolveDisplayTexts = (order: OrderJSON) => {
...
       return {
         title: "Your order was canceled",
-        message:
-          "You can contact <a href='#' data-link='contact-orders'>orders@artsy.net</a> with any questions.",
         messageType: "CANCELLED_ORDER",
       }
     default:
       return {
-        titlet: "Your order",
+        title: "Your order",
```
starsirius, line 159: 🙈

starsirius (approved):

PR #6769, opened 2025-05-22
refactor: always use Vortex's app token

mc-jones:
This PR follows https://github.com/artsy/force/pull/15609 and ensures that we always use the `VORTEX_TOKEN`.


mzikherman (approved):

egdbear:
If we disable the appToken and rely solely on the user token or vortex, will Braze queries under the ‘me’ endpoint still function correctly?

PR #6773, opened 2025-05-22
refactor: fallback to vortex token based on role

mc-jones:
This PR patches #6772 . 

Looks like we inspected an incorrect token testing and `force`'s  does include a subject application so we need to fallback on another criteria. In this case, we will use the `email_provider` role.


ovasdi (approved):

mzikherman (approved):
Good catch.

PR #6780, opened 2025-05-23
feat: improve rate limiting logic

MounirDhahri:
### POC

**Description:**
After the incident today, we noticed that we have a rate limiting middleware that is not being used in MP. We enabled it shortly in staging, but it wasn't working fine although the limit was 60 requests per minute (we were doing a few hundreds of requests per minute).
The rate limiting kicks in by default based on the user ip. This PR adjusts it to be a combination of IP + queried interfaces. So the identifier will be something like
```typescript
123.23.413.4:query:artist
```
```typescript
123.23.413.4:query:artwork
```
```typescript
123.23.413.4:query:artist,query:artwork
```
This could theoretically help us protect further our interfaces. 

What comes next?
We need to try out what is a good rate limit here, we can try it out in staging first. We can set it as below:
```
hokusai staging env set RATE_LIMIT_MAX=60
hokusai staging env set RATE_LIMIT_WINDOW_MS=60000
hokusai staging refresh
```

Example of this in practice and how the identifier will look like

https://github.com/user-attachments/assets/46617fdf-b00a-46f6-a561-aeeff5844789


MounirDhahri:
We tried this out on local Force and it doesn't play well with the prefetching since we trigger a bunch of queries to the same endpoint. The same thing applies also to Eigen where we also do prefetching (both can be disabled without much effort).

Maybe the limit that I set was too little, maybe the id should be also part of the unique identifier 🤷

PR #6789, opened 2025-05-29
feat(EMI-2486): Order shipment properties with generated tracking URLs

erikdstock:
Type: FEAT

Related to [EMI-2486]
Requires https://github.com/artsy/exchange/pull/2594 to work good

This PR adds a shipment type that includes tracking numbers and the url to visit that number for the most common carriers. It relies on the [ts-tracking-number](https://github.com/rjbrooksjr/ts-tracking-number) package, which is recommended by the tracking number reference package linked... from that repo. Overall pretty straightforward.

I tried to be conservative in how much i re-interpret what comes from exchange. Artsy shipping orders should often already have tracking numbers so I leave those alone, and I only match against carriers that match some text (like DHL or DHL Global).

[EMI-2486]: https://artsyproduct.atlassian.net/browse/EMI-2486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -8616,21 +8616,6 @@ enum ConversationsInputMode {
   USER
 }
 
-# Enum for different courier codes
-enum CourierCode {
-  # DHL Express
-  DHL
-
-  # Federal Express
-  FEDEX
-
-  # United Parcel Service
-  UPS
-
-  # United States Postal Service
-  USPS
-}
-
 type CreateAccountRequestMutationFailure {
   mutationError: GravityMutationError
 }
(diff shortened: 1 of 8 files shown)


_schemaV2.graphql
```diff
@@ -8613,6 +8613,21 @@ enum ConversationsInputMode {
...
+  DHL
+
+  # Federal Express
+  FEDEX
+
+  # United Parcel Service
+  UPS
+
+  # United States Postal Service
+  USPS
```
erikdstock: I think i realized a bug: Because of [this line](https://github.com/artsy/metaphysics/pull/6789/files#diff-e079f6119f7138d5807eb6abbd27286d3b7679d62db93e5d5f14544c357f7dc7R126) we probably can't guarantee the courier code will only be one of these. Maybe we could just add an OTHER.

erikdstock:
closing because we have decided this might be better off in exchange.

PR #6795, opened 2025-06-03
chore: add deliveryInfo to order json

oxaudo:
Just cherryPicking parts of what you added in https://github.com/artsy/metaphysics/pull/6789 minus this fancy shipperCode/url generation.

Debated if we want to consolidate `estimatedDelivery` and `estimatedDeliveryWindow'` into something like `estimatedDeliveryDisplay` but for now decided against it as clients might want to only display one. Also one is more of a string and the other one is a data format but not necessarily with timezone so might need to tweak it on the client.

Thanks @erikdstock for pushing the Exchange addition. And we can follow up with fancier url generation

### DIFF
--- a/src/schema/v2/order/types/DeliveryInfo.ts
+++ b/src/schema/v2/order/types/DeliveryInfo.ts
@@ -22,12 +22,13 @@ const DeliveryInfoType = new GraphQLObjectType({
 
     estimatedDelivery: {
       type: GraphQLString,
-      description: "Eesimated delivery date as saved on the order",
+      description: "Estimated delivery date saved on the order as date stringr",
     },
 
     estimatedDeliveryWindow: {
       type: GraphQLString,
-      description: "Eesimated delivery window as saved on the order",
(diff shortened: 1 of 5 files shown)


src/schema/v2/order/types/DeliveryInfo.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    estimatedDelivery: {
+      type: GraphQLString,
+      description: "Eesimated delivery date as saved on the order",
+    },
+
+    estimatedDeliveryWindow: {
+      type: GraphQLString,
+      description: "Eesimated delivery window as saved on the order",
+    },
```
erikdstock: Estimated*

If one of these is an iso 8601 or similar date string it would be good to put in the descriptions field.

erikdstock (commented):
Minor comments, looks good otherwise

oxaudo:
Updated description and merging here.

PR #6800, opened 2025-06-05
refactor(homeView): limit auction engagement rule to AuctionLotsForYou only

dblandin:
Previously the AuctionEngagementRule repositioned three auction-related sections (Auctions, AuctionLotsForYou, and LatestAuctionResults) after the NewWorksForYou section for auction-adjacent and auction-engaged users. This change limits the rule to only reposition the AuctionLotsForYou section, allowing the other auction sections to remain in their default positions.

Quote from our A/B test data analysis (via Xander):

> - **We should only prioritize the Auction Lots For You rail, and move down the Auction and Auction Results rail**
    - 💡 Auction and Auction Results do not update as frequently. Also, because of Quick Links, users can now access relevant auctions without scrolling, fulfilling the IA need for auctions.

🤖 Generated with [Claude Code](https://claude.ai/code)


anandaroop (approved):
Awesome 💯

PR #6806, opened 2025-06-10
chore: use node 22

mc-jones:
This PR bumps `node` to version `22`. `node` version `18` is [EOL](https://nodejs.org/en/about/previous-releases).

[PHIRE-1856]

[PHIRE-1856]: https://artsyproduct.atlassian.net/browse/PHIRE-1856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -45,7 +45,6 @@ commands:
           command: |
             sudo apt update
             sudo apt install --assume-yes python3-pip
-            sudo ln -s /usr/bin/python3 /usr/bin/python
```
mc-jones, line 48: Added in #3338 due to change in the node executor. But looks like this is no longer needed with the `22.5` version.

src/lib/__tests__/graphqlErrorHandler.test.ts
```diff
@@ -86,7 +86,7 @@ describe("graphqlErrorHandler", () => {
       it("does include a HTTP status code when a response is present", () => {
         const originalError: ServerError = {
           message: "underlying",
-          response: { status: 404 },
+          response: { status: 404 } as Response,
```
mc-jones, line 89: Use type assertion here to avoid having to mock out the entire response object. Should be sufficient to just check for status for the purposes of testing.

src/lib/__tests__/apis/fetch.test.ts
```diff
@@ -40,7 +40,7 @@ it("tries to parse the response when there is a String and resolves with it (2)"
     callback(null, reqResponse)
   })
 
-  return expectPromiseRejectionToMatch(fetch("foo/bar"), /Unexpected token/)
+  return expectPromiseRejectionToMatch(fetch("foo/bar"), /(Expected property)/)
```
mc-jones, line 43: This appears to be a change in `node`'s json parser error message, but trying to get more confirmation.

ovasdi (dismissed):

src/lib/__tests__/apis/fetch.test.ts
narikazu: Interesting. What the new error message says is completely the opposite of the previous one.

Dockerfile
```diff
@@ -1,7 +1,7 @@
 # ---------------------------------------------------------
 # Base build dependencies
 # ---------------------------------------------------------
-FROM node:18.15-alpine as builder-base
+FROM node:22-alpine as builder-base
```
narikazu: question(non-blocking): I'm wondering if we should specify the minor version of node here? If I check the version of node on `node:22-alpine`, it's `22.16.0`, which is different than the one we specified in other places. 

```bash
$ docker inspect node:22-alpine | grep VERSION
                "NODE_VERSION=22.16.0",
```

narikazu (dismissed):
Looks good to me.

Dockerfile
mc-jones: Yeah, good call. We should be explicit in our versioning. At the very least consistent. Will push up a commit with these changes.

src/lib/__tests__/apis/fetch.test.ts
mc-jones: Sorry, should have been more specific here. I am struggling to find the docs that say it explicitly, but my understanding is that at some point the `v8` engine that `node` is built on, and what implements `JSON.parse`, updated their error implementation to have more descriptive error messages. So with the node bump, we also received a bump in the `v8` engine. 

So the full error would be something like `"Expected property name or '}' in JSON at position 2 (line 1 column 3)"`.

PR #6820, opened 2025-06-13
fix: require node 22.5.1

nickskalkin:
Seems like there is a bug in 22.5 node version, which sometimes doesn't let to install packages. I've described the issue here: https://artsy.slack.com/archives/CP9P4KR35/p1749743052843689


mc-jones (approved):
🙏 Thank you!

PR #6830, opened 2025-06-18
feat: add excludeArtistFromDiscoveryLoader

egdbear:
This PR adds a new GraphQL mutation that allows users to exclude artists when dismissing them in Discover Daily.

image

Changes:
- Created excludeArtistFromDiscovery mutation
- Added excludeArtistFromDiscoveryLoader gravity loader with endpoint artworks_discovery/artist/exclude

This mutation follows the same pattern as createUserSeenArtwork but is used when users want to exclude/dismiss artists from Discover Daily.

- [x] - Merge [Gravity PR](https://github.com/artsy/gravity/pull/19025) first

/cc @artsy/diamond-devs

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10989,6 +10989,30 @@ type ExchangeError {
   message: String!
 }
 
+type ExcludeArtistFromDiscoveryFailure {
+  mutationError: GravityMutationError
+}
+
+input ExcludeArtistFromDiscoveryInput {
+  artistId: String!
+  clientMutationId: String
+}
+
+type ExcludeArtistFromDiscoveryPayload {
+  clientMutationId: String
+
+  # On success: the excluded artist information.
+  excludeArtistFromDiscoveryOrError: ExcludeArtistFromDiscoveryResponseOrError
+}
+
+union ExcludeArtistFromDiscoveryResponseOrError =
+    ExcludeArtistFromDiscoveryFailure
+  | ExcludeArtistFromDiscoverySuccess
+
+type ExcludeArtistFromDiscoverySuccess {
+  artistId: String
+}
+
 enum ExhibitionPeriodFormat {
   # Long formatted period e.g. February 25 – May 24, 2015
   LONG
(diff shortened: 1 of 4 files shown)


src/schema/v2/infiniteDiscovery/createUserExcludeArtworkMutation.ts
```diff
@@ -0,0 +1,98 @@
+import { mutationWithClientMutationId } from "graphql-relay"
```
iskounen: **issue:** This file looks like a duplicate of creatUserExcludeArtworkMutation.ts. Is it needed?

src/schema/v2/infiniteDiscovery/createUserExcludeArtistMutation.ts
```diff
@@ -0,0 +1,98 @@
...
+        "You need to pass a X-Access-Token header to perform this action"
+      )
+    }
+
+    const createUserExcludeArtistLoaderPayload = Object.keys(args)
+      .filter((key) => key !== "id")
+      .reduce(
+        (acc, key) => ({ ...acc, [snakeCase(key)]: args[key] }),
+        {} as GravityInput
+      )
```
iskounen: **question:** I assume that this code is taking `artistId` from the args and converting it to `artist_id`. If that assumption is correct, does it need to be this complex? Can you just do:

```ts
const createUserExcludeArtistLoaderPayload = {
  artist_id: args.artistId
}
```

iskounen (commented):
I'd like to suggest a clearer name for this mutation. I see that you're matching the naming convention of `CreateUserSeenArtworkMutation`, but neither of those is clear to users of the Metaphysics schema unless you've worked on our team.

Some ideas:

- `ExcludeArtistFromDiscovery`
- `CreateUserPreferenceArtistExclusion`

egdbear:
> * ExcludeArtistFromDiscovery



> I'd like to suggest a clearer name for this mutation. I see that you're matching the naming convention of `CreateUserSeenArtworkMutation`, but neither of those is clear to users of the Metaphysics schema unless you've worked on our team.
> 
> Some ideas:
> 
> * `ExcludeArtistFromDiscovery`
> * `CreateUserPreferenceArtistExclusion`

Sounds good! I like the idea to rename this type to `ExcludeArtistFromDiscovery`

src/schema/v2/infiniteDiscovery/createUserExcludeArtworkMutation.ts
egdbear: 🗑️

iskounen (dismissed):

PR #6832, opened 2025-06-18
feat(EMI-2578): Update subtotal display text

MrSltun:
This PR resolves [EMI-2578]

>[!NOTE]
> This PR requires artsy/exchange#2614 to be merged

### Description
This PR updates the Subtotal line to match the design, where it displays
- `Price` for buy now orders
- `Gallery offer` for partner offers
- `Your offer` on all make offer orders
- `Seller's offer` on a counteroffer from the partner

### DIFF
--- a/src/schema/v2/order/__tests__/MeOrder.test.ts
+++ b/src/schema/v2/order/__tests__/MeOrder.test.ts
@@ -672,7 +672,7 @@ describe("Me", () => {
         expect(result.me.order.pricingBreakdownLines).toEqual([
           {
             __typename: "SubtotalLine",
-            displayName: "Subtotal",
+            displayName: "Price",
             amount: {
               display: "US$5,000",
               currencySymbol: "$",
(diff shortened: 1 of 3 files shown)


oxaudo:
Like I wonder here is those texts make sense for offers that are already approved and shipped? Did not think about that when looking at designs but seeing it in real life and testing with all the different states makes me realize that we now have this strange switch when for example the flow:
1. You made an offer - price says `Your offer`
2. Partner counteroffered - price says `Seller's offer`
3. You accepted seller's offer. The order is approved. What do we expect to see on details page? Looked at it with @cami-w and the answer is:

Screenshot 2025-06-24 at 5 25 40 PM

Think this back and forth with names based on mode/source/awaiting reply from is even more complicated with MOOEA orders and offers based on Limited Partner Offers. We probably need to go through all the states here with those artworks and see if this pricing makes sense. Believe we have a follow up story for that.

PR #6833, opened 2025-06-18
feat(DIA-1363): add maximum eigen version to home view sections

iskounen:
This PR adds a `maximumEigenVersion` to the existing logic that checks whether a Home View section is displayable or not.

This is needed because we will be moving Discover Something New and Explore by Category from the Home tab to the Search tab in an upcoming release, but we still need to guarantee that old versions of the app continue showing them on the Home tab.

For context, we added `minimumEigenVersion` to prevent the Discover Daily entry point from displaying on older versions of the app (that did not know about Discover Daily).

cc: @artsy/diamond-devs

### DIFF
--- a/src/schema/v2/homeView/sections/index.ts
+++ b/src/schema/v2/homeView/sections/index.ts
@@ -48,6 +48,7 @@ export type HomeViewSection = {
   ownerType?: OwnerType
   requiresAuthentication: boolean
   minimumEigenVersion?: SemanticVersionNumber
+  maximumEigenVersion?: SemanticVersionNumber
   shouldBeDisplayed?: (context: ResolverContext) => boolean
   resolver?: GraphQLFieldResolver<any, ResolverContext>
   type: keyof typeof HomeViewSectionTypeNames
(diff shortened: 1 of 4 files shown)


src/schema/v2/homeView/sections/index.ts
```diff
@@ -48,6 +48,7 @@ export type HomeViewSection = {
   ownerType?: OwnerType
   requiresAuthentication: boolean
   minimumEigenVersion?: SemanticVersionNumber
+  maximumEigenVersion?: SemanticVersionNumber
```
MounirDhahri, line 51: **nitpick:** maybe we can consolidate maximum and minimum into one property here. Something like
```
eigenVersion: { maximum, minimum}
```

MounirDhahri (approved):
LGTM, just a minor nitpick

src/schema/v2/homeView/sections/index.ts
iskounen: If we were to continue adding fooEigenVersion, barEigenVersion, ... then I think your suggest makes sense. However, I think maximumEigenVersion will be the last one that we add, so I think it's OK the way it is. (I also don't want to introduce more nesting or break backwards compatibility.)

MounirDhahri: sounds reasonable to me

PR #6837, opened 2025-06-24
feat(DIA-1363): add discoveryMarketingCollections query

iskounen:
This PR adds a `discoveryMarketingCollections` query to the Metaphysics schema. This will allow Eigen to query a subset of our marketing collections for the _Discover Something New_ widget, while allowing us to update its contents without an app deployment.

```graphql
{
  discoveryMarketingCollections {
    slug
    title
    category
    thumbnail
  }
}
```

cc: @artsy/diamond-devs


MounirDhahri:
Thanks Adam, this makes sense

PR #6839, opened 2025-06-24
feat(EMI-2591): expose Order#buyer_state field

fladson:
We need this data available in Force for the redirect logic.

https://artsyproduct.atlassian.net/browse/EMI-2591

@artsy/emerald-devs

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -16632,6 +16632,9 @@ type Order {
   buyerPhoneNumber: String
     @deprecated(reason: "Use `order.fulfillmentDetails.phoneNumber`")
 
+  # Current state of the order
+  buyerState: OrderBuyerStateEnum
+
   # Expiration for the current state of the order
   buyerStateExpiresAt: String
 
@@ -16697,6 +16700,22 @@ type Order {
 
   # The total list price of items (accounting for limited partner offer if applicable)
   totalListPrice: Money
+}
+
+enum OrderBuyerStateEnum {
+  APPROVED
+  CANCELLED
+  COMPLETED
+  DECLINED_BY_BUYER
+  DECLINED_BY_SELLER
+  INCOMPLETE
+  PAYMENT_FAILED
+  PROCESSING_OFFLINE_PAYMENT
+  PROCESSING_PAYMENT
+  REFUNDED
+  SHIPPED
+  SUBMITTED
+  UNKNOWN
 }
 
 enum OrderCreditCardWalletTypeEnum {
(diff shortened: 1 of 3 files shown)


erikdstock (commented):
2 general, minor comments, otherwise seems good to merge when passing.
- if possible to annotate the meaning of each state enum value in the gql schema that would be helpful
- If we add a state that can’t be resolved I wonder if that is an error on the order or a null- and if it causes an issue for the order can we have a fallback to something like UNKNOWN? Relay typescript does this automatically by adding something like `% future added value` to its generated types. Basically I don’t want to ruin the whole order query. 


Tagged @oxaudo because she did a lot of work here

oxaudo (approved):
👍

_schemaV2.graphql
```diff
@@ -16632,6 +16632,9 @@ type Order {
   buyerPhoneNumber: String
     @deprecated(reason: "Use `order.fulfillmentDetails.phoneNumber`")
 
+  # Current state of the order
```
oxaudo: #super minor - but maybe something like 'calculated state of the order that defines buyer facing state/actions'. Or something along those lines to covey that it's order state that ment to encapsulate buyer facing state of the order?

fladson:
> If we add a state that can’t be resolved I wonder if that is an error on the order or a null- and if it causes an issue for the order can we have a fallback to something like UNKNOWN? Relay typescript does this automatically by adding something like % future added value to its generated types. Basically I don’t want to ruin the whole order query.

We already are [defaulting](https://github.com/artsy/exchange/blob/main/app/models/order.rb#L538) to `UNKNOWN` in Exchange, so it should not raise anything.

erikdstock:
> > If we add a state that can’t be resolved I wonder if that is an error on the order or a null- and if it causes an issue for the order can we have a fallback to something like UNKNOWN? Relay typescript does this automatically by adding something like % future added value to its generated types. Basically I don’t want to ruin the whole order query.
> 
> We already are [defaulting](https://github.com/artsy/exchange/blob/main/app/models/order.rb#L538) to `UNKNOWN` in Exchange, so it should not raise anything.

What i mean (and again, non blocking) is: say we add a new buyer state in exchange called `'artwork_lost'`. That won't be able to resolve to a valid value, which will create some kind of graphql resolution error - in fact probably the easiest way to verify this would be writing a simple test.

fladson:
> What i mean (and again, non blocking) is: say we add a new buyer state in exchange called 'artwork_lost'. That won't be able to resolve to a valid value, which will create some kind of graphql resolution error - in fact probably the easiest way to verify this would be writing a simple test.

I see now, yeah, that's a good point. I pushed a resolver to cover this, let me know what you think.

erikdstock (approved):

PR #6842, opened 2025-06-24
fix(AMBER-1714): Create new field to hit /match/fairs

ansor4:
attempts to fix https://artsyproduct.atlassian.net/browse/AMBER-1714

🐛 Partners are no longer able to search for fairs that are not live yet

Partners need to do this in order to create their fair booths _prior_ to the fairs release to the public internet so we've added an authenticated loader for partners to hit to search for fairs at `/match/fairs`

⚠️ Heads up, I did try to get the `matchConnection` to work for an authenticated user but that got too futsy, thinking this scoped to partner might be a nice separation of concerns

Query structure will look something like this
Screenshot 2025-06-24 at 11 56 11 AM

### DIFF
--- a/src/schema/v2/partner/partnerFairConnection.ts
+++ b/src/schema/v2/partner/partnerFairConnection.ts
@@ -36,20 +36,25 @@ export const PartnerFairConnection: GraphQLFieldConfig<
       description: "Exclude these MongoDB ids from results",
     },
   }),
-  resolve: async (_root, { excludeIDs, ...args }, { matchFairsLoader }) => {
+  resolve: async (
+    _root,
+    { excludeIDs, term, ...args },
+    { matchFairsLoader }
+  ) => {
     if (!matchFairsLoader) {
       return null
     }
+    console.log(excludeIDs, term, args)
 
     const { page, size, offset } = convertConnectionArgsToGravityArgs(args)
-    const gravityOptions = {
-      size,
-      offset,
-      total_count: true,
-      term: args.term,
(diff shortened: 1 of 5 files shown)


src/schema/v2/partner/partnerFairConnection.ts
```diff
@@ -0,0 +1,55 @@
...
+import { FairType } from "schema/v2/fair"
+
+export const PartnerFairConnection: GraphQLFieldConfig<
+  { id: string },
+  ResolverContext
+> = {
+  type: connectionWithCursorInfo({
+    name: "PartnerFair",
+    nodeType: FairType,
+  }).connectionType,
```
mzikherman, line 19: This should be reusing the `FairConnection` type instead of defining an entirely new type here, which is essentially a dupe of `FairConnection`, except being called `PartnerFairConnection`.

Import https://github.com/artsy/metaphysics/blob/60f8a06aff87e77da0bf352d89da9389c457fddb/src/schema/v2/fair.ts#L575 and use that type here.

jpotts244: coming in hot over here: https://github.com/artsy/metaphysics/compare/jpotts244/allow-partners-to-query-non-published-fails?expand=1

jpotts244: also figured this doesnt need to be nested under partner! I'll explain more in the other pr

jpotts244:
closing in favor of: https://github.com/artsy/metaphysics/pull/6844

PR #6844, opened 2025-06-24
fix: Allow authenticated users to query for `visibile_to_public:false` fairs

jpotts244:
Stemming off this PR: https://github.com/artsy/metaphysics/pull/6842

Diagnosing from @lidimayra:

> It looks like the older version relied on the specific match endpoint for fairs ([ref](https://github.com/artsy/volt/blob/36dac6f61cb29b38458922257ddb122f1c25ca08/app/controllers/partner_shows_controller.rb#L33)).
> The new one relies on the generic match ([ref](https://github.com/artsy/volt/blob/36dac6f61cb29b38458922257ddb122f1c25ca08/app/javascript/components/AutocompleteInputs/FairAutocomplete.tsx#L250)).
> When switching from one to the other, we ended up losing the automatic access we had to non-public data if the request came from a user with special permissions ([ref](https://github.com/artsy/gravity/blob/a2f8ddd7960d2da26710f10a67c3da16989f8d32/app/api/v1/match_endpoint.rb#L79-L80)).

We need to support this endpoint, and allow authenticated partners to be able to search for `published: false` fairs
See details [here](https://artsy.slack.com/archives/C07PRTJSD6G/p1750698231171439)

I've hooked into the `fairsConnection` and allow it to accept a `term` for authenticated users only to [hit this endpoint](https://github.com/artsy/gravity/blob/main/app/api/v1/match_endpoint.rb#L72-L84)

Query structure will look like:
```graphql
{
  fairsConnection(term: "ART ASIA DELHI 2025", first: 15) {
    edges {
      node {
        isPublished
        name
        internalID
      }
    }
  }
}
```

Screenshot 2025-06-24 at 3 01 55 PM

## Next steps
- Adjust volt to use the above instead of [matchConnection](https://github.com/artsy/volt/blob/main/app/javascript/components/AutocompleteInputs/FairAutocomplete.tsx#L247-L253)

cc @artsy/amber-devs

### DIFF
--- a/src/lib/loaders/loaders_with_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_with_authentication/gravity.ts
@@ -756,6 +756,7 @@ export default (accessToken, userID, opts) => {
       {},
       { headers: true }
     ),
+    matchFairsLoader: gravityLoader("match/fairs", {}, { headers: true }),
     matchUsersLoader: gravityLoader("match/users", {}, { headers: true }),
     mergeArtistLoader: gravityLoader("artists/merge", {}, { method: "POST" }),
     meAlertLoader: gravityLoader((id) => `me/alert/${id}`),
--- a/src/schema/v2/fairs.ts
+++ b/src/schema/v2/fairs.ts
@@ -100,6 +100,7 @@ export const fairsConnection: GraphQLFieldConfig<
     near?: NearType
     sort?: FairSortsType
     status?: EventStatusType
+    term?: string
   } & CursorPageable
 > = {
   type: fairConnection.connectionType,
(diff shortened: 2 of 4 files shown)


src/schema/v2/fairs.ts
```diff
@@ -116,11 +117,53 @@ export const fairsConnection: GraphQLFieldConfig<
...
+      const { body, headers } = await matchFairsLoader({
+        term: args.term,
+        page,
+        size,
+        total_count: true,
+      })
+
+      const totalCount = parseInt(headers["x-total-count"] || "0", 10)
+
+      return {
```
mzikherman: While you're here...this should use the `paginationResolver` helper - that's preferred and basically just wraps all this up for you.

jpotts244: good call 🧹

src/lib/loaders/loaders_with_authentication/gravity.ts
```diff
@@ -756,6 +756,7 @@ export default (accessToken, userID, opts) => {
       {},
       { headers: true }
     ),
+    matchFairsLoader: gravityLoader("match/fairs", {}, { headers: true }),
```
jpotts244, line 759: Hits [this Gravity endpoint](https://github.com/artsy/gravity/blob/main/app/api/v1/match_endpoint.rb#L72-L84)

ansor4 (approved):
thanks!

mzikherman (approved):

PR #6846, opened 2025-06-25
feat(EMI-2577): add artworkOrEditionSet for orders

leamotta:
https://artsyproduct.atlassian.net/browse/EMI-2577

We'd like to display the correct artwork/edition set price or dimensions in the order details page. This allows access to edition set info through an order's line item.

Screenshot 2025-06-25 at 10 32 41
Screenshot 2025-06-25 at 10 32 50


fladson (approved):

oxaudo:
Thanks for adding this!

PR #6853, opened 2025-06-26
chore(AMBER-1724): Allow price adjustment for Batch Edits

lidimayra:
[AMBER-1724]

This PR relies on https://github.com/artsy/gravity/pull/19071

It allows the `priceAdjustment` arg to be passed to the bulk metadata update mutation. When passed, it applies the adjustment to the artwork / edition sets prices, min prices and max prices

cc @artsy/amber-devs


olerichter00 (approved):
💯%

Mitch-Henson (approved):

PR #6857, opened 2025-06-26
refactor(DIA-1308): deprecate partnerBio arg to biographyBlurb

iskounen:
This PR deprecates the `partnerBio` argument that could be passed to an artist's `biographyBlurb` field. The intention of this was that it would prefer partner-supplied bios over Artsy-supplied ones, but in practice it was never used.

A side-effect of this change is that (featured) partner-supplied bios will now be returned for artists that have no Artsy-supplied one. Before this change, those were only returned when `partnerBio` was true.

cc: @artsy/diamond-devs

### DIFF
--- a/src/schema/v2/artist/index.ts
+++ b/src/schema/v2/artist/index.ts
@@ -444,6 +444,8 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
             type: GraphQLBoolean,
             description: "If true, will return featured bio over Artsy one.",
             defaultValue: true,
+            deprecationReason:
+              "Artsy bios are always returned over featured bios.",
           },
           ...markdown().args,
         },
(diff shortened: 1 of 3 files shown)


src/schema/v2/artist/index.ts
```diff
@@ -495,16 +493,19 @@ export const ArtistType = new GraphQLObjectType<any, ResolverContext>({
...
+                credit: `Submitted by ${partner.name}`,
+                partner_id: partner.id,
+                partner: partner,
+              }
             }
           }
 
           // Fall back to the default bio
-          return { text: formatMarkdownValue(blurb, format) }
+          return null
```
anandaroop, line 507: **suggestion (non-blocking):** The comment on L496 makes sense, but I feel like the ones here on L507 and L486 have gotten out of sync with the code. Might be nice to update those.

anandaroop (dismissed):
Looks good 👍🏽 just some minor feedback below about the inline code comments.

General question: nothing here accounts for the rules we employ to prefer one partner bio over another, right? That is assumed to have already happened elsewhere and be reflected by the `featured: true` flag?

src/schema/v2/artist/index.ts
iskounen: I cleaned the comments up. Thanks for the advice.

iskounen:
> General question: nothing here accounts for the rules we employ to prefer one partner bio over another, right? That is assumed to have already happened elsewhere and be reflected by the featured: true flag?

@anandaroop That's correct - this code takes whatever featured partner is returned by Gravity at face-value. 

I still have to add the featured: true flag for 50K existing artists using those rules, and I also need to update CMS to observe those rules when partners add new artists. However, this change alone is expected to make about 1,000 partner bios visible.

PR #6864, opened 2025-06-30
chore: AMBER-1735 - artwork import preferences

Mitch-Henson:
Partially solves [AMBER-1735]

Ensure that we are displaying the dimension metric as this will be used on the front end as each Artwork Import has their own denormalised currency and measurement.

cc: @artsy/amber-devs 

[AMBER-1735]: https://artsyproduct.atlassian.net/browse/AMBER-1735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -3179,7 +3179,8 @@ type ArtworkImport implements Node {
     timezone: String
   ): String
   createdBy: ArtworkImportCreatedBy
-  currency: String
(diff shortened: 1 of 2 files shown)


olerichter00 (approved):
👍

PR #6872, opened 2025-07-02
feat: add trackItemImpressions to HomeViewSectionArtworks

MounirDhahri:
Solves [ONYX-1780]

This PR adds `enableItemsImpressionTracking` to artwork sections
| section with enableItemsImpressionTracking disabled | section with enableItemsImpressionTracking enabled |
|--------|--------|
| Screenshot 2025-07-02 at 10 45 51 | Screenshot 2025-07-02 at 10 46 05 | 

[ONYX-1780]: https://artsyproduct.atlassian.net/browse/ONYX-1780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/homeView/sectionTypes/Artworks.ts
+++ b/src/schema/v2/homeView/sectionTypes/Artworks.ts
@@ -1,4 +1,4 @@
-import { GraphQLObjectType } from "graphql"
+import { GraphQLBoolean, GraphQLNonNull, GraphQLObjectType } from "graphql"
 import { pageable } from "relay-cursor-paging"
 import { ResolverContext } from "types/graphql"
 import { artworkConnection } from "../../artwork"
@@ -7,6 +7,11 @@ import { NodeInterface } from "../../object_identification"
 import { HomeViewGenericSectionInterface } from "./GenericSectionInterface"
 import { HomeViewSectionTypeNames } from "./names"
 import { standardSectionFields } from "./GenericSectionInterface"
+import { HomeViewSection } from "../sections"
+
+export interface HomeViewArtworksSection extends HomeViewSection {
+  enableItemsImpressionTracking?: boolean
+}
 
 export const HomeViewArtworksSectionType = new GraphQLObjectType<
   any,
@@ -18,6 +23,10 @@ export const HomeViewArtworksSectionType = new GraphQLObjectType<
   fields: {
     ...standardSectionFields,
 
+    enableItemsImpressionTracking: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      resolve: (parent) => !!parent.enableItemsImpressionTracking,
+    },
     artworksConnection: {
       type: artworkConnection.connectionType,
       args: pageable({}),
--- a/src/schema/v2/homeView/sectionTypes/Artworks.ts
+++ b/src/schema/v2/homeView/sectionTypes/Artworks.ts
@@ -10,11 +10,11 @@ import { standardSectionFields } from "./GenericSectionInterface"
 import { HomeViewSection } from "../sections"
 
 export interface HomeViewArtworksSection extends HomeViewSection {
-  enableItemsImpressionTracking?: boolean
+  trackItemImpressions?: boolean
 }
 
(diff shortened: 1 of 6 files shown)


src/schema/v2/homeView/sectionTypes/Artworks.ts
```diff
@@ -7,6 +7,11 @@ import { NodeInterface } from "../../object_identification"
 import { HomeViewGenericSectionInterface } from "./GenericSectionInterface"
 import { HomeViewSectionTypeNames } from "./names"
 import { standardSectionFields } from "./GenericSectionInterface"
+import { HomeViewSection } from "../sections"
+
+export interface HomeViewArtworksSection extends HomeViewSection {
```
MounirDhahri, line 12: I am interested in hearing your thoughts about this type - we didn't need to export this so far because we always needed only one additional field that is hanled by a resolver. In this case that wasn't enough because of the new hard typed `enabledImpressionsTracking`. Please let me know if there is a better way to achieve this

src/schema/v2/homeView/sectionTypes/Artworks.ts
```diff
@@ -18,6 +23,10 @@ export const HomeViewArtworksSectionType = new GraphQLObjectType<
   fields: {
     ...standardSectionFields,
 
+    enableItemsImpressionTracking: {
```
dblandin: Another option might be `trackItemImpressions: true/false`

src/schema/v2/homeView/sectionTypes/Artworks.ts
dblandin: I think this makes sense to me! We're now diverging a bit from the parent interface in terms of configuration options so makes sense that we need to extend it.

dblandin (approved):
Nice! One suggested alternative on the field name, but this is looking good to me overall 👍🏼

src/schema/v2/homeView/sections/NewWorksForYou.ts
```diff
@@ -19,7 +19,7 @@ export const NewWorksForYou: HomeViewSection = {
   },
   ownerType: OwnerType.newWorksForYou,
   requiresAuthentication: true,
-
+  trackItemImpressions: true,
```
anandaroop, line 22: **suggestion:** Would you mind adding this to the `_ExampleSection.ts` template? 

Following the configuration for this section you could add it right after [the existing](https://github.com/artsy/metaphysics/blob/a58f91ba112132e5105cbe5536bccede48b2e8cb/src/schema/v2/homeView/sections/_ExampleSection.ts#L236-L240) `requiresAuthentication` field. 

(I would also make clear that this new option is specific to artwork section types that are typed as `HomeViewArtworksSection`.)

anandaroop (dismissed):
Looks good to me! Just one little request for the benefit of future-us...

src/schema/v2/homeView/sections/NewWorksForYou.ts
MounirDhahri: good idea!

MounirDhahri: Okey I might have gone too far, I also updated `_ExampleSectionType` to include the new property and exported the interface so it can raises errors when something is wrong. I am happy to change if you think that's too much ts (if there is such a thing 😄 )

Screenshot 2025-07-03 at 10 28 32

src/schema/v2/homeView/sections/_ExampleSection.ts
```diff
@@ -239,6 +239,12 @@ export const ExampleSection: HomeViewSection = {
    */
   requiresAuthentication: false,
 
+  /*
+   * Whether the section should track item impressions.
+   *
+   */
+  trackItemImpressions: true,
```
anandaroop, line 246: Since we recommend that devs actually copy this file to kick off a new section I wonder if we should actually default this to `false`? We discussed in Onyx office hrs that this results in very chatty analytics events, so we should deliberately opt in to this.

PR #6874, opened 2025-07-02
feat: add artwork caption

egdbear:
This PR introduces a new caption field to the Artwork type, intended for use for alt tags only. 

image

This is a follow-up for https://github.com/artsy/vortex/pull/713

Next steps: 
- [ ] Create a feature flag on Force

/cc @artsy/diamond-devs

### DIFF
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -409,6 +409,22 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         description: "The location of the artwork in My Collection",
       },
       availability: { type: GraphQLString },
+      caption: {
+        // This field is intended for use exclusively on the individual Artwork page for alt tags
+        // Do not use this field in any other context, like artwork rails
+        type: GraphQLString,
+        description:
+          "This field is intended for use exclusively on the individual Artwork page for alt tags",
+        resolve: ({ _id }, _options, { artworkCaptionsLoader }) => {
+          if (!artworkCaptionsLoader) return null
+
+          return artworkCaptionsLoader({
+            artwork_id: _id,
+          }).then(({ data }) => {
+            return data?.caption
+          })
+        },
+      },
       category: {
         type: GraphQLString,
         deprecationReason: "Prefer to use `mediumType`.",
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2623,6 +2623,9 @@ type Artwork implements Node & Searchable & Sellable {
   # Can a user request a lot conditions report for this artwork?
   canRequestLotConditionsReport: Boolean
 
+  # This field is intended for use exclusively on the individual Artwork page for alt tags
+  caption: String
+
   # Represents the "**medium type**", such as _Painting_. (This field is also commonly referred to as just "medium", but should not be confused with the artwork attribute called `medium`.)
   category: String @deprecated(reason: "Prefer to use `mediumType`.")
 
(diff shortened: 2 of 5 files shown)


src/schema/v2/artwork/index.ts
```diff
@@ -409,6 +409,22 @@ export const ArtworkType = new GraphQLObjectType<any, ResolverContext>({
         description: "The location of the artwork in My Collection",
       },
       availability: { type: GraphQLString },
+      caption: {
+        // This field is intended for use exclusively on the Artworks page
+        // Do not use this field in any other context, like artwork rails
+        type: GraphQLString,
+        description:
+          "This field is intended for use exclusively on the Artworks page",
+        resolve: ({ internalID }, _options, { artworkCaptionsLoader }) => {
```
anandaroop: **suggestion:** "Artworks page" → "individual Artwork page" to clarify that this is meant for use in an individual artwork context.

anandaroop (commented):
Re: whether this belongs on the image or the artwork —

That's a very interesting point, because partners _can and do_ change the default image associated with an artwork, which means a caption associated with an artwork could get out of sync with the actual image that we display.

I was curious how often this actually happens:

default image changes

The daily average number of such changes is much higher than I thought it would be: >500 changes per weekday.

Some of those may be simple re-uploads that don't require a caption change, but surely not all of them. We'd need to dig further if we wanted a proper answer to that question.

In such cases, do we have any expectation of re-generating the caption so that it is still accurate at the artwork level?

egdbear:
> Re: whether this belongs on the image or the artwork —
> 
> That's a very interesting point, because partners _can and do_ change the default image associated with an artwork, which means a caption associated with an artwork could get out of sync with the actual image that we display.
> 
> I was curious how often this actually happens:
> 
> default image changes
> The daily average number of such changes is much higher than I thought it would be: >500 changes per weekday.
> 
> Some of those may be simple re-uploads that don't require a caption change, but surely not all of them. We'd need to dig further if we wanted a proper answer to that question.
> 
> In such cases, do we have any expectation of re-generating the caption so that it is still accurate at the artwork level?

Can we confirm that these updates are actually different images for the artworks?

I agree that we'll be slightly out of sync, since our logic runs [daily](https://github.com/artsy/zenith/blob/main/projects/artwork_captions/src/sql/artworks.sql#L18). We always update artworks that were modified the previous day, along with query to handle backfills.

egdbear:
> Might it make more sense to expose this `caption` on an Image type?
> 
> I understand the source data is imported and indexed by artwork id today, but that seems mostly efficiency-driven. This data is actually a property of the image.

`Image.caption` appears to exist already and serves different purpose - For describing individual images (like installation shots in partner shows). More on Force [here](https://github.com/artsy/force/blob/dfe815448751dac6e979b6bc8c6c731d5ba56c2e/src/Apps/Show/Components/ShowInstallShots.tsx#L89) and Eigen [here](https://github.com/artsy/force/blob/dfe815448751dac6e979b6bc8c6c731d5ba56c2e/src/Apps/Show/Components/ShowInstallShots.tsx#L89). 
I guess we can follow-up on that and try to consolidate artwork caption and partner submitted captions/descriptions if necessary.

joeyAghion:
That `Image.caption` might exist in the schema, but I don't think it has a source when it comes to artworks. Maybe these new captions should be made available via that existing property.

anandaroop:
> I agree that we'll be slightly out of sync, since our logic runs [daily](https://github.com/artsy/zenith/blob/main/projects/artwork_captions/src/sql/artworks.sql#L18). We always update artworks that were modified the previous day.

Oh that's good to know, and basically addresses my staleness concern.

> Can we confirm that these updates are actually different images for the artworks?

I compiled the last 1000 changes [here](https://docs.google.com/spreadsheets/d/1G3TWikVkAyI1XJRvdYXhJKyWAE7d-unrhEMqHYBcTu4/edit?usp=sharing) — in some cases there are missing images, as well as some repetition in the change log data. Overall the scope of the problem seems smaller than my the initial query above suggested. And the daily update window sounds like it will take care of the remainder 👍🏽

PR #6875, opened 2025-07-02
feat: migrate ingress to new controllers (staging)

artsyjian:
The existing ingress controllers are not scalable. We have [created a new set of Ingress Controllers](https://github.com/artsy/substance/pull/420) that are scalable.

We want to migrate the ingress to the new controllers without disruption to traffic.

This PR adds a parallel ingress that is tied to the new controllers.

After Staging deploy:

- [ ] Point the app's DNS to the new controllers. This should be done via Infrastructure repo.
- [ ] Verify that the app still works.
- [ ] Delete the old ingress via kubectl. (`kubectl --context staging delete ingress <old-ingress-name>`)

### DIFF
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -157,11 +157,11 @@ spec:
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-  name: metaphysics
+  name: metaphysics-2025
   annotations:
     nginx.ingress.kubernetes.io/whitelist-source-range: {{ externalIngressAllowSourceIP|join(',') }}
 spec:
-  ingressClassName: nginx
+  ingressClassName: external-nginx
   rules:
   - host: metaphysics-staging.artsy.net
     http:


narikazu (approved):

hokusai/staging.yml
```diff
@@ -157,11 +157,11 @@ spec:
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-  name: metaphysics
+  name: metaphysics-2025
```
narikazu, line 160: note to self: I can confirm that migrating ingress to new contollers doesn't affect metaphysics works because [it's pointing to a URL](https://dash.cloudflare.com/0373426be7be649ff052277fb5377c4f/workers/services/view/metaphysics-cdn-staging/production/settings).

artsyjian: ooo, i didn't know that existed. thanks for pointing out.

PR #6879, opened 2025-07-02
fix: Remove default values for page, size arguments for partnerArtworksMat…

jpotts244:
…chConnection

Exposed as part of this [PR](https://github.com/artsy/volt/pull/9406)

Removing the defaults to `page` and `size` and only applying them if no cursor arguments are present
Having a default of `page=1` at the arg level was overriding my cursor arguments when trying to paginate 
and `offset` was always being set to `0` when trying to paginate on the volt side

Ex:
```
/match/partner/gagosian/artworks?offset=0&size=10&term=cecily&total_count=true
```

So I was only able to see the first page of data always!

This new fork will:
- Backwards compatibility for old clients (Folio): If neither `first` nor `after` are present, it applies the old defaults
- For relay backed clients: If `first` or `after` are present, it lets `convertConnectionArgsToGravityArgs()` handle the Relay cursor logic
 

cc @artsy/amber-devs

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -17145,9 +17145,9 @@ type Partner implements Node {
     before: String
     first: Int
     last: Int
-    page: Int = 1
+    page: Int
     query: String!
-    size: Int = 10
+    size: Int
   ): ArtworkConnection
   cached: Int
   categories: [PartnerCategory]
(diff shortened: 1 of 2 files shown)


mzikherman (approved):
Nice!

PR #6880, opened 2025-07-03
fix: make alert fetching consistent across the graph

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-1777

So this Eigen alert-saving bug was a bit of a saga, and required looking into how MP fetches alert data in different contexts.
 
We observed that modifying and then saving an alert in Eigen could fail or succeed depending on how you end up there:

- ❌ Going to your your **activity feed** (`🔔`  → `Alerts` → `any activity item`) and editing the criteria there _fails_
- ✅ Going to your **list of alerts** (`💙 Favorites` → `🔔 Alerts`) and editing the criteria there _succeeds_

The relevant difference in Eigen is that depending on the screen, the alert data is fetched from different parts of MP's graph, as annotated in https://github.com/artsy/eigen/commit/8e44db4c731ae7ecd488189831f7245d9163b933:

annotated

Turns out these resolvers, even though they return the same `AlertType`, resolve their data quite differently:

### ❌ The one that fails

In the failing version, the alert data is nested at `me > notification > item > alert` and the corresponding resolver looks like this:

https://github.com/artsy/metaphysics/blob/a58f91ba112132e5105cbe5536bccede48b2e8cb/src/schema/v2/notifications/Item/AlertNotificationItem.ts#L25-L27

### ✅ The one that succeeds

In the success version, the alert data is nested at `me > alertsConnection > alert` and the corresponding resolver looks like this:

https://github.com/artsy/metaphysics/blob/a58f91ba112132e5105cbe5536bccede48b2e8cb/src/schema/v2/me/index.ts#L163-L165

---

The failing one is the odd one out, the only time we return an `AlertType` that doesn't use the newfangled "fetch from the `meAlertLoader` and then transform the response with `resolveAlertFromJSON`" approach. We see that pattern in 4 of the 5 instances where we return an `AlertType` node. This is the only exception

There surely are reasons this code evolved the way it did (un-stitching etc), but long story short: I think it _should_ be safe to resolve the failing one in the same way as the successful one. They both declare themselves to be the same `AlertType` after all.

Testing this locally seems to work - both mutations succeed in Eigen. This probably warrants a deploy block and some more thorough testing in app + web.


src/schema/v2/notifications/Item/__tests__/index.test.ts
```diff
@@ -176,7 +182,7 @@ describe("NotificationItem", () => {
                     "open edition",
                     "unique",
                   ],
-                  "internalID": "search-criteria-id",
+                  "internalID": "alert-id",
```
anandaroop, line 185: This change in the response is what will ultimately fix the currently broken mutation in Eigen.

nickskalkin (approved):
nice catch!

mzikherman (approved):
Nice investigation! Was most likely an oversight.

anandaroop:
No need for deploy block, I just checked this locally against Force and Eigen for uses of the `NotificationItem.item.alert` field which was modified in this PR.

- **Force**
  - one usage at `src/Components/Notifications/AlertNotification.tsx`, and it still works
- **Eigen**
  - just the one broken usage at `src/app/Scenes/Activity/components/AlertNotification.tsx`, and it now works 🎉 

Merging!

dariakoko:
Thanks Roop!

PR #6882, opened 2025-07-03
fix: remove deprecationReason from partnerBio arg

iskounen:
This PR removes the `@deprecationReason` annotation from the `partnerBio` arg on the `biographyBlurb` of an `artist` because it was blocking a Force deployment https://github.com/artsy/force/pull/15781.

```
Directive "deprecated" may not be used on ARGUMENT_DEFINITION.
```

I moved the deprecation reason to the `@description` annotation until Force can be updated to support that annotation.


leamotta:
@iskounen thank you!

PR #6887, opened 2025-07-03
docs: minor updates to home view examples

anandaroop:
In the interest of keeping the home view schema and docs as tightly scoped to real-world usage as possible I've made a couple of minor tweaks here.

- Defaulting `trackItemImpressions` to `false`, since we only want to opt into the extra analytics traffic deliberately

- Removing the `trackItemImpressions` implementation from the example section type, since as of now it only pertains to _artwork_ section types, which have already been implemented nicely  in https://github.com/artsy/metaphysics/pull/6872. My feeling is that we can expand this capability to other section types as needed, rather than preemptively.


MounirDhahri (approved):
Thank you 🙏

PR #6894, opened 2025-07-07
chore: remove emerald_clientside-collector-signals feature flag

oxaudo:
Removing emerald_clientside-collector-signals flag as it's 100% released to all.

My understanding is that we would not return 'Partner Offer' now for collector signals here any more but that is how it's working now.


erikdstock (approved):
Thank you.

PR #6900, opened 2025-07-09
fix: gracefully handle missing caption errors

egdbear:
This PR introduces a catch to gracefully handle missing caption errors

/cc @artsy/diamond-devs

### DIFF
--- a/src/schema/v2/artwork/__tests__/artwork.test.js
+++ b/src/schema/v2/artwork/__tests__/artwork.test.js
@@ -5738,47 +5738,29 @@ describe("Artwork caption field", () => {
         },
       })
     })
-  })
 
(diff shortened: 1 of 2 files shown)


src/schema/v2/artwork/__tests__/artwork.test.js
```diff
@@ -5738,47 +5738,29 @@ describe("Artwork caption field", () => {
         },
       })
     })
-  })
 
-  describe("with authenticated context", () => {
-    it("returns caption when artworkCaptionsLoader succeeds", async () => {
+    it("returns null when artworkCaptionsLoader throws an error", async () => {
```
MounirDhahri, line 5742: thanks for adding this test

PR #6905, opened 2025-07-09
feat: Adding mutation level support for bulk add to show

jpotts244:
Adding mutation level support for bulk add to show

This will power the experience in our bulk edits flow to allow users to batch add artworks to a show

Ex of the spiked behavior end to end can be seen in #bulk_publishing channel

cc @artsy/amber-devs

### DIFF
--- a/src/schema/v2/partner/BulkOperation/shared.ts
+++ b/src/schema/v2/partner/BulkOperation/shared.ts
@@ -1,0 +1,37 @@
+import {
+  GraphQLBoolean,
+  GraphQLInputObjectType,
+  GraphQLList,
+  GraphQLString,
+} from "graphql"
+import { Availability } from "schema/v2/types/availability"
+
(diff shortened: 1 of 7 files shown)


src/schema/v2/partner/BulkOperation/shared.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  GraphQLBoolean,
+  GraphQLInputObjectType,
+  GraphQLList,
+  GraphQLString,
+} from "graphql"
+import { Availability } from "schema/v2/types/availability"
+
+export const BulkArtworkFilterInput = new GraphQLInputObjectType({
```
olerichter00, line 9: 🌟

olerichter00 (approved):
Looks great! 🌟

lidimayra (approved):
Nice!!!! 🌟

PR #6910, opened 2025-07-15
feat(onyx-1832): new home view section 'Based on your saves'

nickskalkin:
👋  In this PR, I’m introducing a new rail called "Based on Your Recent Saves."

The resolver fetches the user’s last 3 saves and sends them to Gravity’s `related/artworks` endpoint, which returns similar artworks based on genes, tags, and other signals.

Some TODOs are still pending (rail position, analytics events, tests), as we’re still aligning with product on a few open questions that may influence the final implementation.

I’d suggest merging this PR (it’s behind a feature flag), so we can enable it on staging. This will make it easier for the designer and other team members to explore the rail in the app and help us make more informed decisions moving forward.

## Request:

```graphql
{
  homeView {
    section(id: "home-view-section-based-on-your-recent-saves") {
      component {
        title
      }
      
      ... on HomeViewSectionArtworks {
          edges {
            node {
              title
            }
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/src/schema/v2/homeView/sections/__tests__/BasedOnYourRecentSaves.test.ts
+++ b/src/schema/v2/homeView/sections/__tests__/BasedOnYourRecentSaves.test.ts
@@ -98,23 +98,18 @@ describe("BasedOnYourRecentSaves", () => {
       ],
     }
 
-    const similarArtworksResponse = {
-      body: [
-        {
-          _id: "similar-artwork-1",
-          id: "similar-artwork-1",
-          slug: "similar-artwork-1-slug",
-        },
-        {
-          _id: "similar-artwork-2",
-          id: "similar-artwork-2",
-          slug: "similar-artwork-2-slug",
-        },
-      ],
-      headers: {
-        "x-total-count": "10",
+    const similarArtworksResponse = [
+      {
+        _id: "similar-artwork-1",
+        id: "similar-artwork-1",
+        slug: "similar-artwork-1-slug",
       },
-    }
+      {
+        _id: "similar-artwork-2",
+        id: "similar-artwork-2",
+        slug: "similar-artwork-2-slug",
+      },
+    ]
 
     const savedArtworksLoader = jest.fn(async () => savedArtworksResponse)
     const similarArtworksLoader = jest.fn(async () => similarArtworksResponse)


src/schema/v2/homeView/sections/BasedOnYourRecentSaves.ts
```diff
@@ -0,0 +1,53 @@
...
+        artwork_id: map(works, "_id"),
+        for_sale: true,
+        size: SIMILAR_ARTWORKS_SIZE,
+      })
+
+      if (similarArtworks.length === 0) return null
+
+      return {
+        totalCount: similarArtworks.length,
+        ...connectionFromArray(similarArtworks, args),
```
olerichter00: I'm not 100% sure about this, but we might need `pageCursors` as well ([example](https://github.com/artsy/metaphysics/blob/000934644eeab1b397961a415a30648282d36ab7/src/schema/v2/viewingRoom.ts#L104)) 🤔
```suggestion
        pageCursors: createPageCursors({ page, size }, totalCount),
        ...connectionFromArray(similarArtworks, args),
```

src/schema/v2/homeView/sectionTypes/names.ts
```diff
@@ -8,6 +8,8 @@ export const HomeViewSectionTypeNames = {
   HomeViewSectionArtists: "HomeViewSectionArtists",
   HomeViewSectionArtworks: "HomeViewSectionArtworks",
   HomeViewSectionAuctionResults: "HomeViewSectionAuctionResults",
+  HomeViewSectionBasedOnYourRecentSaves:
+    "HomeViewSectionBasedOnYourRecentSaves",
```
anandaroop: **question:** I see that the new section is defined as having a section type of `HomeViewSectionArtworks` which is what I would expect 👍🏽 — but then why do we define a whole new section _type_ here? (Doesn't look like it is being used either)

src/schema/v2/homeView/sections/BasedOnYourRecentSaves.ts
anandaroop: See also: https://github.com/artsy/metaphysics/blob/main/docs/pagination.md — off the top of my head I'm not sure why the cited example differs from this, but I think we regard this doc as best-practice.

src/schema/v2/homeView/sectionTypes/names.ts
nickskalkin: thanks, this change doesn't makes sense. Will revert

src/schema/v2/homeView/sections/BasedOnYourRecentSaves.ts
nickskalkin: thanks, fixed!

olerichter00 (dismissed):

PR #6916, opened 2025-07-16
feat: Allow source to be sent in bulk update mutations

jpotts244:
Allow clients to pass in a `source` arg into our batch operation mutations

This allows us to diagnose and record where these batches are getting triggered from

Options include:
- admin
- artworks_list
- partner_artist_artworks_list
- show_artworks_list

See gravity API here
- https://github.com/artsy/gravity/pull/19124
- https://github.com/artsy/gravity/pull/19099

cc @artsy/amber-devs

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -2562,6 +2562,9 @@ input ArtsyShippingOptInMutationInput {
 
   # ID of the partner
   id: String!
+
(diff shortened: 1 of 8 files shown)


lidimayra (approved):
🌟

olerichter00 (approved):
Very good idea 💡

PR #6927, opened 2025-07-18
feat(NWFY): serve version C with artist diversity re-reranking to all users

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-1844

- The [Artist Diversity re-ranking experiment](https://www.notion.so/Artist-Diversity-Re-Ranking-in-NWFY-225cab0764a08056b47ed08d283cf4ab?pvs=21) that was added in https://github.com/artsy/metaphysics/pull/6892 has passed 🎉
- So this removes the experiment flag and serves Version C ([now updated](https://github.com/artsy/zenith/pull/535) to incorporate the winning variant's logic)

TODO

- [x] Merge only after Version C has been rebuilt with updated logic from [zenith/535](https://github.com/artsy/zenith/pull/535)
- [ ] Remove the Unleash flag ( ⚠️ only after this PR is merged)

### DIFF
--- a/src/schema/v2/homeView/sections/NewWorksForYou.ts
+++ b/src/schema/v2/homeView/sections/NewWorksForYou.ts
@@ -1,5 +1,4 @@
 import { ContextModule, OwnerType } from "@artsy/cohesion"
-import { getExperimentVariant } from "lib/featureFlags"
 import { artworksForUser } from "schema/v2/artworksForUser/artworksForUser"
 import { withHomeViewTimeout } from "../helpers/withHomeViewTimeout"
 import { HomeViewArtworksSection } from "../sectionTypes/Artworks"
@@ -21,18 +20,7 @@ export const NewWorksForYou: HomeViewArtworksSection = {
   requiresAuthentication: true,
   trackItemImpressions: true,
   resolver: withHomeViewTimeout(async (parent, args, context, info) => {
-    const variant = getExperimentVariant(
-      "onyx_nwfy-artist-diversity-experiment",
-      {
-        userId: context.userID,
-      }
-    )
-
-    let recommendationsVersion = "C"
-
-    if (variant && variant.enabled && variant.name === "variant-a") {
-      recommendationsVersion = "A"
-    }
+    const recommendationsVersion = "C"
 
     const finalArgs = {
       // formerly specified client-side
--- a/src/schema/v2/homeView/experiments/experiments.ts
+++ b/src/schema/v2/homeView/experiments/experiments.ts
@@ -6,5 +6,4 @@ import { FeatureFlag } from "lib/featureFlags"
  */
 export const CURRENTLY_RUNNING_EXPERIMENTS: FeatureFlag[] = [
   "onyx_experiment_home_view_test",
-  "onyx_nwfy-artist-diversity-experiment",
 ]
--- a/src/schema/v2/homeView/sections/__tests__/NewWorksForYou.test.ts
+++ b/src/schema/v2/homeView/sections/__tests__/NewWorksForYou.test.ts
@@ -1,13 +1,10 @@
 import gql from "lib/gql"
 import { runQuery } from "schema/v2/test/utils"
-import { getExperimentVariant } from "lib/featureFlags"
 import "schema/v2/homeView/experiments/experiments"
 
 jest.mock("lib/featureFlags", () => ({
   getExperimentVariant: jest.fn(),
 }))
-
-const mockGetExperimentVariant = getExperimentVariant as jest.Mock
 
 describe("NewWorksForYou", () => {
   it("returns the section's metadata", async () => {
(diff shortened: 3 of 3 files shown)


src/schema/v2/homeView/sections/NewWorksForYou.ts
```diff
@@ -21,18 +20,7 @@ export const NewWorksForYou: HomeViewArtworksSection = {
...
-        userId: context.userID,
-      }
-    )
-
-    let recommendationsVersion = "C"
-
-    if (variant && variant.enabled && variant.name === "variant-a") {
-      recommendationsVersion = "A"
-    }
+    const recommendationsVersion = "A" // the winning variant but maybe we wait until Version C has been updated with the Version A logic
```
anandaroop: Keeping this in `draft` because maybe we prefer instead to wait until Zenith is updated to copy the **Version A** logic _into_ **Version C**, which then once again becomes the version we serve to everyone.

If so, we can hold and then update this line appropriately before merging.

anandaroop: cc @cavvia — I can ticket or take this^ on if you agree.

cavvia: I think it's more intuitive to convert version A into the production version and we can start using C as our experiment version in future.

cavvia:
I'm assuming this would switch Web to Version A as well? We currently have different versions rolled out on Web/App.

MounirDhahri (approved):
LGTM 👍 

@cavvia this won't affect web afaik. We would need to also update Force to use the same variant, or maybe even better remove the hard coded version from the arguments and move that over to Metaphysics
https://github.com/artsy/force/blob/dc5fb4f3edaef3c682a68831bf95a1fdaf433f7b/src/Apps/Home/Components/HomeNewWorksForYouRail.tsx#L127

cavvia:
@MounirDhahri I see... I'm happy with either approach, I'm not sure where we land on client-side vs graphql server responsibilities on these kind of parameters. As long as we update web to make it consistent with this.

src/schema/v2/homeView/sections/NewWorksForYou.ts
anandaroop: That's fine with me but I believe that involves propagating this change elsewhere such as…

- [eigen/src/app/Scenes/NewWorksForYou/NewWorksForYou.tsx](https://github.com/artsy/eigen/blob/4f06b58a648d59b36fff516ef44c33c590551c5a/src/app/Scenes/NewWorksForYou/NewWorksForYou.tsx#L17)

- [force/src/Apps/NewForYou/newForYouRoutes.tsx](https://github.com/artsy/force/blob/c9a3af690469db3e4d5b4e52137a76894555c263/src/Apps/NewForYou/newForYouRoutes.tsx#L12) 

- [force/src/Apps/Home/Components/HomeNewWorksForYouRail.tsx](https://github.com/artsy/force/blob/c9a3af690469db3e4d5b4e52137a76894555c263/src/Apps/Home/Components/HomeNewWorksForYouRail.tsx#L127)

It's because of code like that^ that I figured Version C might be our "forever control group." If we instead promote other versions round-robin style as they get promoted from experiments, we'll have to be diligent about propagating these changes to code paths like the above.

dblandin: I wonder why we're pinning to certain versions here in the client code. I assume that these clients would just want the currently defined default version of recs, unless signaled otherwise (like via a query param or experiment enrollment).

Is it possible to define C as the current default server-side and simply remove these explicit version args in the client code, rather than need to propagate changes now and in the future?

anandaroop: Looks like…

-  the `version` argument to MP's `artworksForUser` field [is optional](https://github.com/artsy/metaphysics/blob/e52cc7ad0c488b76aff975615e8d738bb171fa22/_schemaV2.graphql#L19198)
- and if not supplied, recs would would [default to to Version B](https://github.com/artsy/vortex/blob/8ba40161dfb2af90b88ea8dda8ea1a2a93b26a44/app/models/new_for_you_recommendation.rb#L2), as defined in Vortex

We _could_ change that Vortex default on an ongoing basis as we promote different winning variants, and that might allow us to remove the client-side defaults in Force and Eigen. We'd need to be sure no client is relying on that current Vortex default of Version B.

nickskalkin: > We'd need to be sure no client is relying on that current Vortex default of Version B.

Are there other clients apart from Force/Eigen that query NWFY? Only Braze comes to mind

cavvia: I'm not suggesting we promote versions round robin... I'm suggesting we stick to Version A in future for web/app and reserve C for experiments. It's confusing to have C as production.

cavvia: It seems like this requires an Eigen update and an app release though, in which case probably not worth making the change, as it slows us down.

cavvia: Promoting Version A to production version here: https://github.com/artsy/zenith/pull/535

anandaroop: Ah ok, so…

- A one-time sweep of those clients to specify Version A as a default (or no default, if we rely on Vortex's default). 
- And then we'd stick to the pattern of keeping Version A as production/control from now on.

Sounds good to me, and with your Zenith change, it should be safe to start that sweep now.

I'll start with making Version A the client-side default (and we can optionally continue with setting and respecting the Vortex-side default.)

> Are there other clients apart from Force/Eigen that query NWFY? Only Braze comes to mind

That's the only one I can think of, and I believe Braze explicitly requests Version B already.

anandaroop: Ok, actually reading through https://github.com/artsy/zenith/pull/535 I see that I misunderstood. 

I guess we need not bother with any of ^that sweep for now. In which case I will change L23 here  to Version C and get this merged.

anandaroop:
After some [clarification](https://artsy.slack.com/archives/C05EQL4R5N0/p1753197429303349?thread_ts=1753123561.966869&cid=C05EQL4R5N0) we will return to using Version C as the production / control version for now. This PR only to be merged tomorrow after Version C has been rebuilt.

PR #6931, opened 2025-07-21
feat: Amber 1813 support batch updating signature field

rabidpaul:
This change resolves [AMBER-1813](https://artsyproduct.atlassian.net/browse/AMBER-1813)

[AMBER-1813]: https://artsyproduct.atlassian.net/browse/AMBER-1813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
+++ b/src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
@@ -34,6 +34,7 @@ interface Input {
     priceListed?: number
     provenance?: string
     published?: boolean
+    signature?: string
   }
   filters?: {
     artistId?: string
@@ -93,6 +94,10 @@ const BulkUpdateArtworksMetadataInput = new GraphQLInputObjectType({
     published: {
       type: GraphQLBoolean,
       description: "Publish or unpublish artworks",
+    },
+    signatureDetails: {
+      type: GraphQLString,
+      description: "Details about the signature",
     },
   },
 })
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -64,7 +64,7 @@ With your dependencies set up, you can run Metaphysics by running:
 yarn start
 ```
 
-Which will start the server on http://localhost:3000
```
jpotts244, line 67: nice ✅

damassi:
@rabidpaul - One request is to make sure that the PR description reflects the work. Its unlikely that we'd ever migrate away from JIRA, but the main thing is we don't want to leave PR pull request context in a 3rd party source, where GH only contains a link. Then GH becomes less crawlable, and search doesn't work as it should.

src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
```diff
@@ -94,6 +95,10 @@ const BulkUpdateArtworksMetadataInput = new GraphQLInputObjectType({
       type: GraphQLBoolean,
       description: "Publish or unpublish artworks",
     },
+    signatureDetails: {
```
jpotts244, line 98: CMS batch edit has a nice `dev` mode interface

It also shows us how the client is sending over this field to MP, looks a like its just `signature`!

Screenshot 2025-07-21 at 2 39 08 PM


Shall we fast follow to rename?

jpotts244: let me know if you want to test together!

PR #6935, opened 2025-07-22
feat: adjust `listPrice` handling on artwork and edition set for more consistent currencyPrefix+price formatting

jpotts244:
Stemming off [this bug report from folio](https://artsy.slack.com/archives/C07PRTJSD6G/p1752155644122979), we've had feedback that it would be helpful to see an artworks price formatted with currency code on folio

While digging in the space, ⚠️ we noticed we dont surface currency prefix for any edition set pricing across artsy.net ⚠️ 

This can be confusing to users (as reported by the partner above too) for currencies that use the same symbol (USD and AUD's use of `$` for example)

More discussion [here](https://artsy.slack.com/archives/C05F8TNKGAV/p1753202844188899)

On Artsy.net, For unique works we format the price as:
```
US$123
```
and for edition works we format the price:
```
$123
```

Some examples below
- [Unique artwork](https://staging.artsy.net/artwork/testy-mctest-tital-6) (see the US$<price> )
- [Editioned artwork](https://staging.artsy.net/artwork/testy-mctest-stuff) (see the $<price>)
- [Another non price range editioned set](https://staging.artsy.net/artwork/testy-mctest-jackies-es-test-with-price-listed)

Screenshot 2025-07-22 at 12 01 24 PM

Screenshot 2025-07-22 at 12 43 59 PM

## Fix

In this PR, we are adjusting the `listedPrice` interface to return currency prefix as part of the `display` field, this will adjust both artworks and edition sets

```graphql
{
  artwork(id: "testy-mctest-tital-6") {
    saleMessage
    priceListedDisplay
    editionSets {
      saleMessage
      priceDisplay
      price
      listPrice {
         ... on PriceRange {
          display
        } 
        ... on Money {
          display
        }
      }
    }
  }
}


```

### Further examples

Exact price edition set artwork:
Screenshot 2025-07-22 at 3 31 21 PM

Price Range edition set artwork:
Screenshot 2025-07-22 at 3 49 59 PM

cc @artsy/amber-devs and adding team Emerald for feedback too

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -10920,7 +10920,6 @@ type EditionSet implements Sellable {
   price: String
   priceDisplay: String
   priceListed: Money
-  priceListedDisplay: String
   prototypes: String
 
   # Is the edition set parent-artwork published?
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -3269,6 +3269,7 @@ type ArtworkImportRow {
   currency: String!
   dimensionMetric: String!
   errors: [ArtworkImportRowError!]!
+  excludedFromImport: Boolean!
 
   # A globally unique ID.
   id: ID!
@@ -4319,6 +4320,7 @@ type AuthenticationType {
 }
 
 type Author {
+  articles: [Article!]!
   bio(format: Format): String
 
   # A globally unique ID.
@@ -4328,7 +4330,7 @@ type Author {
 
   # A type-specific ID.
   internalID: ID!
-  name: String
+  name: String!
   twitterHandle: String
 }
 
@@ -4613,6 +4615,9 @@ input BulkArtworkFilterInput {
 }
 
 input BulkUpdateArtworksMetadataInput {
+  # The artist IDs to be assigned
+  artistIds: [String]
+
   # The availaiblity to be assigned
   availability: Availability
 
@@ -16099,6 +16104,9 @@ type Mutation {
 
   # Submit an order
   submitOrder(input: submitOrderInput!): submitOrderPayload
+  toggleArtworkImportRowExclusion(
+    input: ToggleArtworkImportRowExclusionInput!
+  ): ToggleArtworkImportRowExclusionPayload
   transferMyCollection(
     # Parameters for TransferMyCollection
     input: TransferMyCollectionInput!
@@ -16757,6 +16765,9 @@ union OpeningHoursUnion = OpeningHoursArray | OpeningHoursText
 
 # Buyer's representation of an order
 type Order {
+  # List of available payment methods for the order
+  availablePaymentMethods: [OrderPaymentMethodEnum!]!
+
   # List of alpha-2 country codes to which the order can be shipped
   availableShippingCountries: [String!]!
 
@@ -18992,6 +19003,7 @@ type Query {
   # A list of Articles
   articles(
     auctionID: String
+    authorID: String
     channelID: String
     featured: Boolean
 
@@ -19243,6 +19255,12 @@ type Query {
 
   # If user is logged out; status is `LOGGED_OUT`. If user is logged in; status is `LOGGED_IN`. If user is logged in with invalid authentication (401); 'Promise' resolves to 'Status.Invalid'. All other status codes will resolve to `LOGGED_IN` because we don't know whether or not the authentication is valid (error could be something else).
   authenticationStatus: AuthenticationStatus!
+
+  # An Editorial author
+  author(
+    # The ID of the author
+    id: String!
+  ): Author
 
   # A user's bank account
   bankAccount(
@@ -20711,6 +20729,14 @@ type ResizedImageUrl {
   srcSet: String!
   url: String!
   width: Int
+}
+
+input S3LocationInput {
+  # The S3 bucket name where the image is stored.
+  bucket: String!
+
+  # The S3 key (object path) for the image.
+  key: String!
 }
 
 # The conditions for uploading assets to media.artsy.net
@@ -22564,6 +22590,30 @@ type TaxMoreInfo {
   url: String!
 }
 
+type ToggleArtworkImportRowExclusionFailure {
+  mutationError: GravityMutationError
+}
+
+input ToggleArtworkImportRowExclusionInput {
+  artworkImportID: String!
+  artworkImportRowID: String!
+  clientMutationId: String
+  excludedFromImport: Boolean!
+}
+
+type ToggleArtworkImportRowExclusionPayload {
+  clientMutationId: String
+  toggleArtworkImportRowExclusionOrError: ToggleArtworkImportRowExclusionResponseOrError
+}
+
+union ToggleArtworkImportRowExclusionResponseOrError =
+    ToggleArtworkImportRowExclusionFailure
+  | ToggleArtworkImportRowExclusionSuccess
+
+type ToggleArtworkImportRowExclusionSuccess {
+  artworkImport: ArtworkImport
+}
+
 # Total line
 type TotalLine {
   # The monetary amount for the line
@@ -22824,7 +22874,7 @@ input UpdateArtworkImportRowInput {
   artworkImportRowID: String!
   clientMutationId: String
   fieldName: String!
-  fieldValue: String!
+  fieldValue: String
 }
 
 type UpdateArtworkImportRowPayload {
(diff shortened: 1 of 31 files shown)


_schemaV2.graphql
```diff
@@ -10920,6 +10920,7 @@ type EditionSet implements Sellable {
   price: String
   priceDisplay: String
   priceListed: Money
+  priceListedDisplay: String
```
jpotts244: This seems to work for edition sets, now we have a diff in how the `saleMessage` is calculated

Will adjust, looks like artwork handles [this in MP directly ](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/artwork/index.ts#L1686-L1710)

Screenshot 2025-07-22 at 1 13 26 PM

src/schema/v2/edition_set.ts
```diff
@@ -189,13 +190,35 @@ export const EditionSetType = new GraphQLObjectType<any, ResolverContext>({
...
+    //   },
+    // },
     priceListed: {
       type: Money,
-      resolve: ({ price_listed: price_listed, price_currency: currency }) => {
-        const factor =
-          currencyCodes[currency?.toLowerCase()]?.subunit_to_unit ?? 100
-        const cents = price_listed * factor
-        return { cents, currency }
+      resolve: ({ price_cents, price_currency }, args, context, info) => {
```
jpotts244: Allows for us to query `display` on `money` for edition sets

```
{
  artwork(id: "testy-mctest-jackies-es-test-with-price-listed") {
    saleMessage
    priceListedDisplay
    editionSets {
      saleMessage
      priceDisplay
      priceListedDisplay
      price
      priceListed {
        ... on Money {
          amount
          major
          currencyPrefix
          currencySymbol
          display
        }
      }
    }
  }
}

```
Screenshot 2025-07-22 at 2 23 55 PM



⚠️ edition set `price_cents` come back from the API as an array ⚠️ Need a better way to resolve this

⚠️ Also note the diff in `saleMessage` for edition set vs for artwork, need to play with that too ⚠️

src/schema/v2/fields/listPrice.ts
```diff
@@ -50,7 +51,7 @@ export const listPrice: GraphQLFieldConfig<any, any> = {
       ? {
           __typename: Money.name,
           cents: price_cents[0],
-          display: price,
+          display: priceDisplayText(price_cents[0], price_currency, ""),
```
jpotts244, line 54: now were def not backwards compatible, so still need to think a bit here

Clients asking for 
```graphql
listPrice {
  ... on PriceRange {
    display
  } 
  ... on Money {
     display
  }
}
```

will get
```
US$123
```
instead of 
```
$123
```

mzikherman: It might be ok! `US$` vs `$` feels nicely disambiguated. I think there are likely only a few currencies where we'd only display the symbol instead of the more fully-qualified form (EUR, GBP) - and it's _probably_ fine to make that update to disambiguate those (even if they don't _need_ it, technically). For USD, there are other currencies that are 'dollars' (like AUD), so `US$` feels more correct than _just_ `$`. And for EUR and GBP...I guess technically those don't need disambiguation, but i don't see a compelling reason to handle those differently at this point.

jpotts244: That is true!! Ill also add in emerald just for awareness



I do have one expectation that is failing for a different reason than expecting:

Screenshot 2025-07-22 at 4 43 59 PM

will dig a bit there^

src/lib/moneyHelpers.ts
```diff
@@ -84,7 +84,7 @@ export const priceRangeDisplayText = (
 
   if (higherPriceCents == null) {
     const amount = priceAmount(lowerPriceCents as number, currency, format)
-    return `${fullPrefix}${amount} and up`
+    return `Starting at ${fullPrefix}${amount}`
   }
```
jpotts244, line 88: Adjusted this helper to match the expectation defined [here](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/artwork/__tests__/artwork.test.js#L658-L680)

src/lib/__tests__/moneyHelpers.test.ts
```diff
@@ -57,7 +57,11 @@ describe("priceRangeDisplayText", () => {
   })
 
   it("appends text 'and up' when high price is not given", () => {
```
jpotts244, line 59: Yell if objections to this change, trying to consolidate some of the various string handling to simplify this logic and keep the treatment the same for artwork/edition set etc

mzikherman (dismissed):

damassi (approved):
Amazed that this has been just now caught. Partners create a lot of edition sets!

leamotta (dismissed):

Mitch-Henson (approved):
Love it!

jpotts244:
~One thing I'm noticing prior to merging this~

UPDATE: this new code respects this conditionality here!


Looks like for certain currencies we dont surface the currency prefix (GBP for example) on any artwork:
https://staging.artsy.net/artwork/david-shrigley-i-cannot-live-without-you-83

Screenshot 2025-07-23 at 10 36 41 AM

~I think this change will adjust these to also be changed, I wonder if that'll cause more harm than good~

jpotts244:
> Amazed that this has been just now caught. Partners create a lot of edition sets!


Yes same, seemed sus, and dug around a bit this morning, I think this is actually only having an issue if there are _multiple_  edition sets for works with a single edition set it seems to work!

- single edition set: https://staging.artsy.net/artwork/robert-cottingham-hawk-eye-1
- multiple edition sets: https://staging.artsy.net/artwork/jon-allured-editioned-bn-mo-usd-us-4

Tested both cases with the new logic and all seems to be respected 🫡

leamotta:
> One thing I'm noticing prior to merging this
> 
> Looks like for certain currencies we dont surface the currency prefix (GBP for example) on any artwork: https://staging.artsy.net/artwork/david-shrigley-i-cannot-live-without-you-83
> 
> Screenshot 2025-07-23 at 10 36 41 AM
> I think this change will adjust these to also be changed, I wonder if that'll cause more harm than good

Would be good to confirm! For Emerald's concerns, I think we'd like to avoid showing GBP₤ or EUR€ in the artwork page. We've done work in the past to remove that prefix from emails and there is logic in Gravity to reflect this https://github.com/artsy/gravity/blob/main/app/models/concerns/money_extensions.rb#L50

jpotts244:
@leamotta just updated the thread above, looks like this change respects that correctly (thanks to this logic that gets run in the money helper:
https://github.com/artsy/metaphysics/blob/main/src/lib/moneyHelpers.ts#L4)

Here's the query Im running
Screenshot 2025-07-23 at 10 55 43 AM

leamotta:
@jpotts244 Thanks for verifying! 🏅

src/schema/v2/fields/listPrice.ts
```diff
@@ -40,8 +41,8 @@ export const listPrice: GraphQLFieldConfig<any, any> = {
     name: "ListPrice",
     types: [PriceRange, Money],
   }),
-  resolve: ({ price_cents, price, price_currency }) => {
-    if (!price_cents || price_cents.length === 0) {
+  resolve: ({ price_cents, price_currency }) => {
+    if (!price_cents || price_cents.length === 0 || !price_currency) {
```
jpotts244, line 45: apparently on the partner side (folio) price currency can be `null`!! adding one extra null check

mzikherman:
@leamotta curious why that's the case? While I agree there's a handful of currencies that don't _need_ disambiguation since they use unique symbols, it feels like a bit of an optimization that makes things different enough (requires special handling) - and I'm not quite clear what the _benefit_ is? (besides _technically_ you don't need to disambiguate).

leamotta:
@mzikherman mainly for readability - if a currency symbol is unique, skipping disambiguation keeps things cleaner. The special handling impact is minor; it just avoids unnecessary extra info. Plus some currencies like CHF have the same symbol as the disambiguation, we wouldn't want to show `CHF CHF3000`

jpotts244:
Ok thanks all for the discussion, I've been waiting until I can QA this before merging. Finally just did a run through side by side with local and staging
- browsing
- searching
- orders
all look OK

damassi:
Shipp itt 👍

PR #6952, opened 2025-07-25
chore: add mutation to deselect an artwork import row from import

mzikherman:
Mutation to use https://github.com/artsy/gravity/pull/19171 to deselect an artwork import row from import.

I'll also need to raise another small Gravity PR to make sure those deselected rows are not returned to `rowsConnection`.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -3269,6 +3269,7 @@ type ArtworkImportRow {
   currency: String!
   dimensionMetric: String!
   errors: [ArtworkImportRowError!]!
+  excludedFromImport: Boolean!
 
   # A globally unique ID.
   id: ID!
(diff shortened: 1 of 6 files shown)


_schemaV2.graphql
```diff
@@ -15929,6 +15953,9 @@ type Mutation {
   deliverSecondFactor(
     input: DeliverSecondFactorInput!
   ): DeliverSecondFactorPayload
+  deselectArtworkImportRow(
```
damassi: Should this be `toggleArtworkImportRow`? Because at some point we might want to enable / disable it, vs just disable it. Thoughts?

damassi (approved):
One q about the naming to make things a bit more future-proof (and aligned with gravity API naming) but otherwise LGTM

_schemaV2.graphql
mzikherman: Eh, its so trivial to update/expand if needed - you can add the `toggleArtworkExclude` argument. I actually initially had that all as toggle and with the naming, but then assumed you'd request a _simpler_ version where you don't even pass in a true/false 😅

mzikherman: But I can rename if you think it's clearer.

mzikherman:
Pushed the rename.

PR #6968, opened 2025-07-31
feat: add isLiveOpenHappened attribute to sale to know that live sale is past it's live bidding open time

oxaudo:
Step for https://artsyproduct.atlassian.net/browse/EMI-2582

The current available `isLiveOpen` on sale is only true during the lime live bidding is open. But we want to hide bidding info for collectors in history and other pages after life open happened as we can not guarantee that our 'bid status' is correct as the auction enters Live bidding part.

Instead of manually testing for live_start_at being after now in multiple clients adding the attribute here to be reused in Force and Eigen.


erikdstock (approved):
Seems reasonable!

PR #6973, opened 2025-08-01
feat(ONYX-1860): expose priceRangeUpdatedAt

dariakoko:
Add `priceRangeUpdatedAt` introduced in following Gravity PR: https://github.com/artsy/gravity/pull/19198

### DIFF
--- a/src/schema/v2/me/index.ts
+++ b/src/schema/v2/me/index.ts
@@ -394,10 +394,15 @@ export const meType = new GraphQLObjectType<any, ResolverContext>({
         "Indicates whether the Artwork Budget was updated within the past three months",
       type: new GraphQLNonNull(GraphQLBoolean),
       resolve: ({ price_range_updated_at }) => {
-        if (!price_range_updated_at) return false
-        const threeMonthTime = 90 * 24 * 60 * 60 * 1000
-        new Date().getTime() - new Date(price_range_updated_at).getTime() <
-          threeMonthTime
+        // if the timestamp is unset, also consider the budget "stale" and needing an update
+        if (!price_range_updated_at) return true
+
+        const staleCutOff = moment().subtract(3, "months")
+        const updatedAt = moment(price_range_updated_at)
+
+        const isStale = updatedAt < staleCutOff
+
+        return isStale
       },
     },
     hasQualifiedCreditCards: {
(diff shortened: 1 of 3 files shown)


dblandin (commented):
Thoughts on abstracting this a bit? I think we could avoid passing the raw timestamp to clients, keep some control of [the logic here](https://www.figma.com/design/xqc2S6v0iOtz8iayTlhacp/Price-Preference?node-id=306-11040&m=dev) in Metaphysics, and avoid doubling down on the "price range" language now that the UI has shifted towards "Artwork Budget" and a single max value.

Something like:

```grapqhl
{
  me {
    hasStaleArtworkBudget # true / false
  }
}
```

An alternative could be `shouldUpdateArtworkBudget`... but generally I think exposing a boolean value here would satisfy the client needs while not leaking too much of the internal implementation details.

dariakoko:
Good point @dblandin 
Addressed
`new Date(price_range_updated_at).getTime()` bothers me a bit. Is it okay to use it like this or is it better to adjust the date format on gravity side?

src/schema/v2/me/index.ts
```diff
@@ -389,6 +389,17 @@ export const meType = new GraphQLObjectType<any, ResolverContext>({
...
     },
+    hasStaleArtworkBudget: {
+      description:
+        "Indicates whether the Artwork Budget was updated within the past three months",
+      type: new GraphQLNonNull(GraphQLBoolean),
+      resolve: ({ price_range_updated_at }) => {
+        if (!price_range_updated_at) return false
+        const threeMonthTime = 90 * 24 * 60 * 60 * 1000
+        new Date().getTime() - new Date(price_range_updated_at).getTime() <
+          threeMonthTime
```
dblandin: We can express this more succinctly using some momentJS magic:

```typescript
// if the timestamp is unset, also consider the budget "stale" and needing an update
if (!price_range_updated_at) return true

// define the cutoff
const staleCutOff = moment().subtract(3, "months")
const updatedAt = moment(price_range_updated_at)

// if last recorded update was before this cutoff, consider it stale
const isStale = updatedAt < staleCutOff

return isStale
```

In this example, I'm assuming the following logic:
- If the timestamp is unset => stale
- If the timestamp is set more 3 months old ago => stale
- If the timestamp is set less than 3 months ago => not stale

This would put every current user into the "stale" group, which might be fine as even if they have a non-default price range set, as we don't know how recently this was provided.

We might consider a small backfill of this timestamp within Gravity so that any user account created within the past three months with a non-default price range will have this timestamp set and we'll avoid prompting them again until another three months have past.

dariakoko:
I'm gonna go ahead and merge it, thanks Devon!

dblandin:
@dariakoko When you have a moment, it would be great to follow-up and add some specific test cases for this logic, now that it'll live here in Metaphysics. Should be pretty straightforward to mock a few timestamp examples coming from Gravity and assert that we get the expected boolean result.

PR #6982, opened 2025-08-04
feat(mcp): [POC] Add Apollo MCP Server example

damassi:
Follow-up experiment to https://github.com/artsy/mcp-analytics. 

This implements a bare bones config for https://github.com/apollographql/apollo-mcp-server. 

### Setup:
- Followed the [installation methods](https://www.apollographql.com/docs/apollo-mcp-server/command-reference#installation-methods) to install a local version. 
```bash
cd metaphysics
curl -sSL https://mcp.apollo.dev/download/nix/latest | sh
```

Can also use the docker container. 

- Add this `mcp-config.yml` config to the root of the project:
```yml
endpoint: http://localhost:5001/v2
transport:
  type: streamable_http
operations:
  source: local
  paths:
    - ./mcp/operations/analytics
schema:
  source: local
  path: ./_schemaV2.graphql
overrides:
  mutation_mode: "all" # enable mutations! 
introspection:
  execute:
    enabled: true
  introspect:
    enabled: true
  search:
    enabled: true
headers:
  "x-access-token": "fill in"
  "x-user-id": "fill in"
```
- Update `mcp-config.yml` with access tokens and user ID 
- Start the MCP server in one terminal pane: `./apollo-mcp-server mcp-config.yml`
- Open another terminal pane and boot local metaphysics: `yarn start` 
- Add this config to Claude Desktop or Claude code (via `~/.claude.json`):
```json
"artsy-mcp": {
  "command": "npx",
  "args": [
    "mcp-remote",
    "http://127.0.0.1:5000/mcp"
  ]
}
```
- Boot Claude Desktop or start `claude-code`

More information in this [slack thread](https://artsy.slack.com/archives/CA8SANW3W/p1754335676656499)

PR #6989, opened 2025-08-06
feat(mcp update artist): add update artist mcp mutation

egdbear:
This PR adds the update artist mutation as a addition to original PR https://github.com/artsy/metaphysics/pull/6982

### DIFF
--- a/mcp-config.yml
+++ b/mcp-config.yml
@@ -5,9 +5,12 @@ operations:
   source: local
   paths:
     - ./mcp/operations/analytics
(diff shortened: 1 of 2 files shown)


damassi (approved):
Nice -- and thanks for using an operation 👍 

Worth calling out that we can get real specific here since the names are fed to the LLM. If artsy support frequently need target supply updated, we could name this `updateArtistTargetSupply`, etc. 

But for now, generic is good. The server will see 'updateArtist` and instantly have enough info to take action.

PR #7006, opened 2025-08-12
feat(batch-edit): Support Certificate of Authenticity related boolean fields

jpotts244:
feat: Support Certificate of Authenticity related boolean fields in Batch edits

Exposes the backend logic [here](https://github.com/artsy/gravity/blob/main/app/services/partner_artworks/batch_edit_builder_service.rb#L10-L30)

Allows client to send over values for
- certificate of authenticity
- coa by gallery
- coa by 3rd party

AKA: These boolean fields on the artwork form:
Screenshot 2025-08-12 at 2 34 00 PM

cc @artsy/amber-devs

### DIFF
--- a/src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
+++ b/src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
@@ -26,6 +26,9 @@ interface Input {
     artistIds?: string[]
     availability?: string
     category?: string
+    certificateOfAuthenticity?: boolean
+    coaByGallery?: boolean
+    coaByAuthenticatingBody?: boolean
     conditionDescription?: string
     domesticShippingFeeCents?: number
     ecommerce: boolean
@@ -65,6 +68,19 @@ const BulkUpdateArtworksMetadataInput = new GraphQLInputObjectType({
     category: {
       type: GraphQLString,
       description: "The category (medium type) to be assigned",
+    },
+    certificateOfAuthenticity: {
+      type: GraphQLBoolean,
+      description:
+        "Whether a certificate of authenticity is provided for these artworks.",
+    },
+    coaByGallery: {
+      type: GraphQLBoolean,
+      description: "If COA is provided by the gallery.",
+    },
+    coaByAuthenticatingBody: {
+      type: GraphQLBoolean,
+      description: "If COA is provided by a third-party authenticating body.",
     },
     conditionDescription: {
       type: GraphQLString,
@@ -244,6 +260,9 @@ export const bulkUpdateArtworksMetadataMutation = mutationWithClientMutationId<
         artist_ids: metadata.artistIds,
         availability: metadata.availability,
         category: metadata.category,
+        certificate_of_authenticity: metadata.certificateOfAuthenticity,
+        coa_by_gallery: metadata.coaByGallery,
+        coa_by_authenticating_body: metadata.coaByAuthenticatingBody,
         condition_description: metadata.conditionDescription,
         domestic_shipping_fee_cents: metadata.domesticShippingFeeCents,
         ecommerce: metadata.ecommerce,
--- a/src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
+++ b/src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
@@ -26,7 +26,7 @@ interface Input {
     artistIds?: string[]
     availability?: string
     category?: string
-    certificateOfAuthenticity?: boolean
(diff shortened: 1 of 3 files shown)


jpotts244:
Assigning you @olerichter00 as you are working on the client side 🫡 !

We can change this layer to infer a value for `certificate_of_authenticity` if any `coa_by_*` fields are present, etc. I believe `certificate_of_authenticity` is required on the backend 

Looking at your designs here:
Screenshot 2025-08-12 at 2 39 11 PM

src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
```diff
@@ -26,6 +26,9 @@ interface Input {
     artistIds?: string[]
     availability?: string
     category?: string
+    certificateOfAuthenticity?: boolean
```
olerichter00:
```suggestion
    hasCertificateOfAuthenticity?: boolean
```

**Minor:** We could change it to `hasCertificateOfAuthenticity` to be consistent with the [artwork field](https://github.com/artsy/metaphysics/blob/0ce2f793466f93964c524bed136d138dcea3d510/src/schema/v2/artwork/index.ts#L1991) and the [My Collection artwork mutations](https://github.com/artsy/metaphysics/pull/5770/files#diff-3a304b8e0fa043b030a2e5b3040547c6e4008ebebed2a9207ba67d27dc18192dR13848).

olerichter00 (commented):
Looks great 🌟

lidimayra:
> We can change this layer to infer a value for certificate_of_authenticity if any coa_by_* fields are present, etc. I believe certificate_of_authenticity is required on the backend

I think this is handled in the back-end already.
If `coa_by_gallery` or `coa_by_authenticating_body` are passed, `certificate_of_authenticity` is automatically set to `true` [here](https://github.com/artsy/gravity/blob/75be0151187dfbb3d904a2054c73718d7cd90f54/app/models/concerns/artwork_condition.rb#L20-L22).

I guess the validation is the opposite: we're not able to just set `certificate_of_authenticity` to `true` unless any of the other 2 are enabled.

We still need to surface the `certificate_of_authenticity` field anyway since we want to allow them to set it to `false`.

So, on the client-side, I guess the following should already work once this PR is merged:

- No certificate of authenticy: `certificateOfAuthenticity: false`
- Issued by gallery: `coaByGallery: true`
- Issued by authenticating body: `coaByAuthenticatingBody: true`
- Issued by authenticating body and gallery: `coaByAuthenticatingBody:, coaByGallery: true`

Without the need to pass `certificate_of_authenticity` for any of the last 3 scenarios. But let's double check if this is really the case.

lidimayra (dismissed):
Nice!!!

jpotts244:
Ah ok @lidimayra maybe we can adjust the `BatchEditBuilderService` to match the way an organic artwork is treated. It is looking like we _expect_ `certificate_of_authenticity` to be present as a top level field and then check the sub booleans of `coa_*`

```ruby
# BatchEditBuilderService

def certificate_of_authenticity
      return {} unless attributes.has_key?(:certificate_of_authenticity)

      coa = string_to_boolean(attributes[:certificate_of_authenticity])
      coa_attributes = {certificate_of_authenticity: coa}

      if coa
        coa_by_gallery = attributes[:coa_by_gallery]
        coa_by_authenticating_body = attributes[:coa_by_authenticating_body]

        unless coa_by_gallery || coa_by_authenticating_body
          raise_or_log_error "coa_by_gallery or coa_by_authenticating_body is required to set certificate_of_authenticity to true"
        end
        coa_attributes[:coa_by_gallery] = string_to_boolean(coa_by_gallery) if coa_by_gallery

        coa_attributes[:coa_by_authenticating_body] = string_to_boolean(coa_by_authenticating_body) if coa_by_authenticating_body
      end

      coa_attributes
    end
```

([ref](https://github.com/artsy/gravity/blob/main/app/services/partner_artworks/batch_edit_builder_service.rb#L10-L29))

lidimayra:
@jpotts244 Ahhh true!! Yes, agreed, I guess we could just adapt the builder to handle each one of those fields in isolation, so the client wouldn't be forced to pass the boolean one in case any of the other ones is passed

jpotts244:
Cool! can do that on the side here today

jpotts244:
I guess Volt also does some logic on its side to set the sub flags to `false` if the user selects 'I have no certificate of authenticity'

and then sends all 3 to the API

so many fields to keep track of 🕷️ I wonder if we can make some of this easier on the front end by moving some of this to the backend. I'll ping you guys in slack! going to test some theories for a bit


Screenshot 2025-08-13 at 10 48 05 AM

PR #7008, opened 2025-08-13
feat(onyx-1874): NWFY - create view frequency penalty experiment

nickskalkin:
We open a new experiment for the [view frequency penalty](https://github.com/artsy/zenith/pull/546). You can find an Unleash feature flag [here](https://unleash.artsy.net/projects/default/features/onyx_nwfy-view-frequency-penalty-experiment).

### DIFF
--- a/src/lib/featureFlags.ts
+++ b/src/lib/featureFlags.ts
@@ -15,8 +15,8 @@ const FEATURE_FLAGS_LIST = [
   "onyx_enable-quick-links-v2",
   "onyx_enable-home-view-auction-segmentation",
   "onyx_enable-quick-links-price-budget",
-  "onyx_nwfy-at-risk-gallery-boost-experiment",
   "onyx_based_on_your_saves_home_view_section",
+  "onyx_nwfy-view-frequency-penalty-experiment",
 ] as const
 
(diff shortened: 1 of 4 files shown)


nickskalkin:
@anandaroop I can imagine we'll run more NWFY A vs C experiments in the future. Do you think it makes sense to create a generic experiment in Unleash (and here) called something like `nwfy_ab-test`? That way, we wouldn't need to modify Metaphysics code every time we add a new test

anandaroop:
Yes, I had the same thought. It's very repetitive to delete and re-create the Unleash toggles as well as the custom experiment code in Eigen.

One small point against that though — having clearly named and distinct experiments in Unleash gives us a better event log / audit trail e.g. [here](https://unleash.artsy.net/projects/default/features/onyx_nwfy-price-affinity-test/logs?limit=25&feature=IS%3Aonyx_nwfy-price-affinity-test). We have occasionally referred to those logs to confirm when an experiment was enabled/disabled/etc. Not sure if that's worth the extra hassle, but I'd be up for discussing with the team.

anandaroop:
The change in this PR is actually smaller and smoother than what I had done previously (removing and then re-creating the code). I guess my 2¢ is that I'm ok with this little bit of friction in exchange for the clearer audit trail, but it's not a strong opinion.

PR #7009, opened 2025-08-13
feat: Add title to artwork bulk metadata mutation

olerichter00:
Addresses [AMBER-1877]

## Description

Add title to artwork bulk metadata mutation.


lidimayra (approved):
⭐

PR #7002, opened 2025-08-12
fix(serialization): Fix array serialization for complex nested objects in mutations

damassi:
### Description

While working on https://github.com/artsy/metaphysics/pull/7001, noticed a fundamental limitation in sending data to gravity, in that we could only handle simple serialization, so anything more complex like 
```
foo: [
  { position: 0 },
  { position: 1 }
]
```

would fail with a 400. 

The fix was simple: use the `indices` format instead of `brackets` [format](https://github.com/ljharb/qs?tab=readme-ov-file#stringifying) which ultimately end up deserializing back into the same JSON structure. Example:

- ids[]=13&ids[]=27 → {ids: ["13", "27"]}
- ids[0]=13&ids[1]=27 → {ids: ["13", "27"]}

Once this fix was in, I could successfully send this mutation:

```graphql
mutation {
  updateArtworkImportRowImages(
    input: {
      artworkImportID: "648e4d88-2ddf-4500-a295-b67760a4e5f6"
      images: [
        { id: "1bd4407a-90d6-4972-8cc0-c805a2dc9c31", position: 3 }
        { id: "a5458866-d25b-4cab-9745-44607abad42d", position: 2 }
        { id: "bb874e9c-b33e-47d7-8613-668dcb180359", position: 1 }
      ]
    }
  ) {
    updateArtworkImportRowImagesOrError {
      ... on UpdateArtworkImportRowImagesSuccess {
        artworkImportID
      }
      
      ... on UpdateArtworkImportRowImagesFailure {
        mutationError {
          error
          detail
          message
          statusCode
        }
      }
    }
  }
}
```

Tagging a few folks because this is a fairly sensitive change that is low in our stack. 

Issues of this kind (arrays of various shapes being sent over the wire) has come up a few times over the years and its always confusing; this should fix it so that we can now send any shape of JSON and solve the issue for good.

PR #7015, opened 2025-08-15
feat(imports): Add batch matchArtworkImportImages mutation

damassi:
⚠️ This wont actually work until https://github.com/artsy/metaphysics/pull/7002 is merged, so lets hold on merging this til then!

### Description

Adds a new `matchArtworkImportImages` mutation that allows matching multiple images in a single request, corresponding to the new `match_images` endpoint added in https://github.com/artsy/gravity/pull/19251.

This provides a more efficient way to match multiple images at once rather than sending individual mutations for each image.

### Example

```graphql
mutation {
  matchArtworkImportImages(
    input: {
      artworkImportID: "artwork-import-1"
      images: [
        {
          fileName: "cat.jpg"
          s3Key: "/some/path/cat.jpg"
          s3Bucket: "someBucket"
        }
        {
          fileName: "dog.jpg"
          s3Key: "/some/path/dog.jpg"
          s3Bucket: "someBucket"
        }
      ]
      rowID: "row-1"
    }
  ) {
    matchArtworkImportImagesOrError {
      ... on MatchArtworkImportImagesSuccess {
        success
      }
    }
  }
}
```

cc @artsy/amber-devs

### DIFF
--- a/src/schema/v2/ArtworkImport/matchArtworkImportImagesMutation.ts
+++ b/src/schema/v2/ArtworkImport/matchArtworkImportImagesMutation.ts
@@ -1,0 +1,129 @@
+import {
+  GraphQLString,
+  GraphQLObjectType,
+  GraphQLUnionType,
+  GraphQLNonNull,
+  GraphQLList,
+  GraphQLInputObjectType,
+  GraphQLBoolean,
+} from "graphql"
+import { mutationWithClientMutationId } from "graphql-relay"
+import { ResolverContext } from "types/graphql"
+import {
+  formatGravityError,
+  GravityMutationErrorType,
+} from "lib/gravityErrorHandler"
+import { ArtworkImportType } from "./artworkImport"
+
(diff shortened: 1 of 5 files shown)


src/schema/v2/ArtworkImport/matchArtworkImportImagesMutation.ts
```diff
@@ -0,0 +1,129 @@
...
+  },
+  outputFields: {
+    matchArtworkImportImagesOrError: {
+      type: ResponseOrErrorType,
+      resolve: (result) => result,
+    },
+  },
+  mutateAndGetPayload: async (
+    { artworkImportID, images, rowID },
+    { artworkImportMatchImagesLoader }
```
jpotts244, line 100: Once we add the new loader to [Gravity.ts](https://github.com/artsy/metaphysics/blob/main/src/lib/loaders/loaders_with_authentication/gravity.ts) should be good to go with the new endpoint

PR #7024, opened 2025-08-20
feat(onyx-1883): INTERNAL_AUTOSUGGEST mode for matchConnection

nickskalkin:
Adds `INTERNAL_AUTOSUGGEST` search mode. When specified, it calls the `/match/suggest/internal` endpoint, created to support admin panel needs (e.g., search by slug, ID, etc.).

Related PRs:
- https://github.com/artsy/gravity/pull/19267
- https://github.com/artsy/gravity/pull/19274

### DIFF
--- a/src/schema/v2/Match.ts
+++ b/src/schema/v2/Match.ts
@@ -26,6 +26,7 @@ import { SearchMode } from "./search"
 import { SearchEntity } from "./search/SearchEntity"
 import { ShowType } from "./show"
 import { TagType } from "./tag"
+import { compact } from "lodash"
 
 const MODELS = {
   Article: { loader: "articleLoader", type: ArticleType },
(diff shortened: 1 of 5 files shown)


src/schema/v2/Match.ts
```diff
@@ -89,7 +98,10 @@ export const MatchConnection: GraphQLFieldConfig<void, ResolverContext> = {
 
     const results = await Promise.all(
       body.map(async ({ id, label }) => {
-        const loader = loaders[MODELS[label].loader]
+        const loader = loaders[MODELS[label]?.loader]
```
nickskalkin, line 101: Without these changes `matchConnection` fails when search returns an entity which type is not in the [MODELS array](https://github.com/artsy/metaphysics/blob/516dfb87e42aad5a0776aa3143bda46b5645cb39/src/schema/v2/Match.ts#L30-L41). For example, such request fails:

```
{
  matchConnection(term: "Trolleys", entities: [], mode: AUTOSUGGEST) {
    edges {
      node {
        ... on Artist {
          name
        }
      }
    }
  }
}
```

because Trolleys returns artist series, which are not supported by Match.

dblandin: I'm surprised this hasn't caused an issue before. Do we currently only use matchConnection when we're using a non-empty / non-default entities argument?

Can we add a spec to exercise this updated behavior?

dblandin (approved):
Overall looks good to me!

One suggestion might be to split the `fix` on the result processing – handling an unsupported result type – from the `feat` introducing the additional search mode. Might be useful to have those as separate commits when merged into main. It also feels like we should be able to write a test to cover this scenario.

src/schema/v2/Match.ts
nickskalkin: > Do we currently only use matchConnection when we're using a non-empty / non-default entities argument

apparently we don't use it with `entities: []`, that's why this issue was not noticed before. In any case, I think it's worth fixing. I'm extracted this change + tests to a separate PR like you suggested https://github.com/artsy/metaphysics/pull/7026

nickskalkin:
thanks @dblandin, I've created two separate PRs instead of this one:

- https://github.com/artsy/metaphysics/pull/7026
- https://github.com/artsy/metaphysics/pull/7027

dblandin:
Sorry - didn't mean to create a bunch of busy work for you! I was imagining two separate commits in this PR and then rebase merging but separate PRs works as well 👍🏼

PR #7029, opened 2025-08-21
chore(EMI-1996): stop pushing schema to Pulse

fladson:
After the work to remove Artemis from Pulse, we don't need to sync the schema there anymore.

@artsy/emerald-devs


MrSltun (approved):

fladson:
> Maybe we should have a follow-up after this to remove the schema from Pulse?

Yes, that's my plan 👍

fladson:
This is failing on CI, I've raised a question to Sapphire to see how we can change the env value - https://artsy.slack.com/archives/CA8SANW3W/p1755784672536339

PR #7032, opened 2025-08-21
feat: Thread through support for batch editing of international shipping fee

jpotts244:
Adds support for `internationalShippingFeeCents` in the batch update flows

Associated gravity changes here: https://github.com/artsy/gravity/pull/19285

cc @artsy/amber-devs


olerichter00 (approved):
🌟 🌟 🌟

PR #7037, opened 2025-08-22
fix: param/body construction on createNewGeminiAssetLoader

dblandin:
We were passing acl as an query param on the URL path but also providing acl in the request body.

This commit is focused on fixing this one instance but it's possible that an [earlier PR][#7002] to fix array serialization in mutations and POST request bodies introduced incompatibility with combining both query params in the URL path and a request body at the same time.

[#7002]: https://github.com/artsy/metaphysics/pull/7002

cc/ @joeyAghion

### DIFF
--- a/src/lib/loaders/loaders_without_authentication/gemini.ts
+++ b/src/lib/loaders/loaders_without_authentication/gemini.ts
@@ -11,7 +11,7 @@ export default () => ({
   // The outer function is so that we can pass params from the schema,
   // into the gemini api options.
   createNewGeminiAssetLoader: ({ name, acl }) =>
-    geminiUncachedLoader(`uploads/new.json?acl=${acl}`, {
+    geminiUncachedLoader("uploads/new.json", {
       acl,
       headers: {
         Authorization: "Basic " + toBase64(name + ":"),


damassi:
@dblandin 

> it's possible that an https://github.com/artsy/metaphysics/pull/7002 to fix array serialization in mutations and POST request bodies introduced incompatibility with combining both query params in the URL path and a request body at the same time

Can you verify this in a test?

dblandin:
> Can you verify this in a test?

@damassi Yep – will tag you in the PR.

dblandin:
I won't have time to get to the spec today but if helpful, you can easily reproduce the expected vs problematic response with the following GraphQL mutation:

```graphql
mutation {
  requestCredentialsForAssetUpload(
    input: { acl: "private", name: "convection-staging" }
  ) {
    asset {
      policyDocument {
        conditions {
          acl
        }
      }
    }
  }
}
```

You'll receive a mismatched `public-read` ACL before this fix and the expected `private` ACL now. Will pick up this spec (and any other desired changes to the fetch / request logic) on Monday.

damassi:
Yah this totally makes sense. We were hand coding the args vs using the normal object flow to and it sent things a bit haywire.

PR #7043, opened 2025-08-25
fix: updated price visibility implementation in bulk update mutation to align with backend

lilyfromseattle:
I didn't look closely enough at Lidi's gravity PR and instead followed the existing pattern for single artworks which ended up not working for exact price when I tried hooking up the front end. Fixing now.

### DIFF
--- a/src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
+++ b/src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
@@ -35,6 +35,7 @@ interface Input {
     coaByGallery?: boolean
     coaByAuthenticatingBody?: boolean
     conditionDescription?: string
+    displayPriceRange?: boolean
     domesticShippingFeeCents?: number
     ecommerce: boolean
     exactPrice?: boolean
@@ -46,7 +47,6 @@ interface Input {
     medium?: string
     offer: boolean
     priceAdjustment?: number
-    displayPriceRange?: boolean
     priceHidden?: boolean
     priceListed?: number
     provenance?: string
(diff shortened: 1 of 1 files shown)


olerichter00 (approved):
🌟

src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
```diff
@@ -45,6 +45,7 @@ interface Input {
     medium?: string
     offer: boolean
     priceAdjustment?: number
+    displayPriceRange?: boolean
```
lidimayra: Minor: could we keep this one on the same line it was placed before these changes? It was easier to spot it when alphabetically sorted.

src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
```diff
@@ -141,13 +142,17 @@ const BulkUpdateArtworksMetadataInput = new GraphQLInputObjectType({
...
+      description: "Set artwork price visibility to price range",
     },
     priceHidden: {
       type: GraphQLBoolean,
-      description: "Show/Hide the price of an artwork",
+      description: "Set artwork price visibility to price on request",
+    },
+    priceListed: {
+      type: GraphQLFloat,
+      description: "The price for the artworks",
```
lidimayra: Same here, can we keep the alphabetical sorting we had before these changes?

src/schema/v2/partner/BulkOperation/bulkUpdateArtworksMetadataMutation.ts
```diff
@@ -301,6 +306,7 @@ export const bulkUpdateArtworksMetadataMutation = mutationWithClientMutationId<
         medium: metadata.medium,
         offer: metadata.offer,
         price_adjustment: metadata.priceAdjustment,
+        display_price_range: metadata.displayPriceRange,
```
lidimayra, line 314: would also be nice to keep this one where it was before (line 292)

PR #7054, opened 2025-09-02
feat(DIAM-49): Add filtersForArtworksConnection to discoveryCategoryConnection query

araujobarret:
Related to [DIAM-49]

This PR adds a way of coexisting curated collections and normal artworks with a filter(case of price "collections", under 500 etc). On Force that's hardcoded, that's why it's easy to make this distinction.

Major changes:
- filtersForArtworksConnection returning the filters values + title
- filtersForArtworksConnection#artworksConnection returning the corresponding artworks 
- added href to discoveryCategoryConnection

```graphql
discoveryCategoryConnection(slug: "collect-by-price") {
		category
		href
		filtersForArtworksConnection(first: 3) {
			totalCount
			edges {
				node {
					href
					title
						edges {
							node {
								title
							}
						}
					}
				}
			}
		}
	}
}
```

cc @artsy/diamond-devs 

[DIA-49]: https://artsyproduct.atlassian.net/browse/DIA-49?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DIAM-49]: https://artsyproduct.atlassian.net/browse/DIAM-49?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/schema/v2/__tests__/discoveryCategoryConnection.test.ts
+++ b/src/schema/v2/__tests__/discoveryCategoryConnection.test.ts
@@ -1,28 +1,28 @@
 import { runQuery } from "schema/v2/test/utils"
 
-describe("discoveryCategoryConnection", () => {
-  const mockFilterArtworksLoader = jest.fn().mockResolvedValue({
-    hits: [
-      {
-        _id: "artwork-1",
-        title: "Test Artwork 1",
-        slug: "test-artwork-1",
-        id: "test-artwork-1",
-      },
-      {
-        _id: "artwork-2",
-        title: "Test Artwork 2",
-        slug: "test-artwork-2",
-        id: "test-artwork-2",
-      },
-    ],
-    aggregations: {
-      total: {
-        value: 2,
-      },
+const mockFilterArtworksLoader = jest.fn().mockResolvedValue({
+  hits: [
+    {
+      _id: "artwork-1",
+      title: "Test Artwork 1",
+      slug: "test-artwork-1",
+      id: "test-artwork-1",
     },
-  })
+    {
+      _id: "artwork-2",
+      title: "Test Artwork 2",
+      slug: "test-artwork-2",
+      id: "test-artwork-2",
+    },
+  ],
+  aggregations: {
+    total: {
+      value: 2,
+    },
+  },
+})
 
+describe("discoveryCategoryConnection", () => {
   beforeEach(() => {
     mockFilterArtworksLoader.mockClear()
   })
@@ -60,7 +60,12 @@ describe("discoveryCategoryConnection", () => {
 
   it("returns artworkConnection for categories with filters", async () => {
     const context = {
-      filterArtworksLoader: mockFilterArtworksLoader,
+      unauthenticatedLoaders: {
+        filterArtworksLoader: mockFilterArtworksLoader,
+      },
+      authenticatedLoaders: {
+        filterArtworksLoader: mockFilterArtworksLoader,
+      },
     }
 
     const query = `
@@ -69,13 +74,11 @@ describe("discoveryCategoryConnection", () => {
           title
           category
           filtersForArtworksConnection(first: 10) {
-            totalCount
             edges {
               node {
                 href
                 title
                 artworksConnection(first: 5) {
-                  totalCount
                   edges {
                     node {
                       slug
@@ -100,13 +103,16 @@ describe("discoveryCategoryConnection", () => {
     expect(firstFilter.href).toBe("/collect?price_range=*-500")
     expect(firstFilter.title).toBe("Art under $500")
     expect(firstFilter.artworksConnection.edges).toHaveLength(2)
-    expect(firstFilter.artworksConnection.totalCount).toBe(2)
-    expect(firstFilter.artworksConnection.totalCount).toBe(2)
   })
 
   it("returns empty artworkConnection for categories without filters", async () => {
     const context = {
-      filterArtworksLoader: mockFilterArtworksLoader,
+      unauthenticatedLoaders: {
+        filterArtworksLoader: mockFilterArtworksLoader,
+      },
+      authenticatedLoaders: {
+        filterArtworksLoader: mockFilterArtworksLoader,
+      },
     }
 
     const query = `
@@ -222,7 +228,12 @@ describe("discoveryCategoryConnection", () => {
 
   it("can query filtersForArtworksConnection via Node interface", async () => {
     const context = {
-      filterArtworksLoader: mockFilterArtworksLoader,
+      unauthenticatedLoaders: {
+        filterArtworksLoader: mockFilterArtworksLoader,
+      },
+      authenticatedLoaders: {
+        filterArtworksLoader: mockFilterArtworksLoader,
+      },
     }
 
     const query = `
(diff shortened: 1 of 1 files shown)


src/lib/marketingCollectionCategories.ts
```diff
@@ -78,12 +116,24 @@ export const marketingCollectionCategories = {
...
+      price_range: [
+        { title: "Art under $500", price_range: "*-500" },
+        { title: "Art under $1000", price_range: "501-1000" },
+        { title: "Art under $2500", price_range: "1001-2500" },
+        { title: "Art under $5000", price_range: "2501-5000" },
+        { title: "Art under $10000", price_range: "5001-10000" },
+        { title: "Art under $25000", price_range: "10001-25000" },
+        { title: "Art above $25000", price_range: "25001-*" },
+      ],
+    },
```
egdbear: **[question]** Do you think we could simplify this by just adding another array field to the existing static data, which would only be used for filtering?

For example:

```js
Collect by Price: {
  id: "Collect by Price",
  slug: "collect-by-price",
  title: "Price",
  imageUrl: "https://files.artsy.net/images/collections-price-category.jpeg",
  // we could also have a marketing collection category type, e.g. "CollectCard", "MarketingCollectionCard"
  sortedCollectionSlugs: [...], // no breaking changes
  collectQueryFilters: [ // maybe a better name here?
    { title: "Art under $500", price_range: "*-500" },
    { title: "Art under $1000", price_range: "501-1000" },
    { title: "Art under $2500", price_range: "1001-2500" },
    { title: "Art under $5000", price_range: "2501-5000" },
    { title: "Art under $10000", price_range: "5001-10000" },
    { title: "Art under $25000", price_range: "10001-25000" },
    { title: "Art above $25000", price_range: "25001-*" }
  ]
}
```

Then, on Eigen, we could decide whether to render a `marketingCollection` or a `collect` query renderer based on the field name. The nice part is this could be easily feature-flagged.

Obviously, it would also be nice to rename this file (and related types) so it better reflects the mix between filters (collect) and marketing collections.

araujobarret: will do that! The only stronger opinion is about the name, collect is not a model per se, so we should call it as it is, `artworkFilters`.

araujobarret: changed it here https://github.com/artsy/metaphysics/pull/7054/commits/6540f7da295aaa01530c2279b3b701a27fd0ff6e

araujobarret: Gonna merge it if there are no blockers, can come later and make changes before merging the work from Eigen.

src/schema/v2/discoveryCategoriesConnection.ts
```diff
@@ -4,13 +4,89 @@ import {
...
+      type: GraphQLNonNull(GraphQLString),
+      description: "The href for this filtered connection",
+      resolve: ({ href }) => href,
+    },
+    title: {
+      type: GraphQLNonNull(GraphQLString),
+      description: "The display title for this filtered connection",
+      resolve: ({ title }) => title,
+    },
+    artworksConnection: {
```
egdbear: question: could we just use filterArtworkConnection here instead? Something like this maybe? [src/schema/v2/show.ts](https://github.com/artsy/metaphysics/blob/669e23f0940a520faa5b731506c945e339265e92/src/schema/v2/show.ts#L435)

src/schema/v2/discoveryCategoriesConnection.ts
```diff
@@ -45,10 +131,55 @@ export const DiscoveryCategoryType = new GraphQLObjectType<
...
+    href: {
+      type: GraphQLNonNull(GraphQLString),
+      description: "The href of the category",
+    },
     title: {
       type: GraphQLNonNull(GraphQLString),
       description: "The display title of the category",
     },
+    filtersForArtworksConnection: {
+      type: FiltersForArtworksConnectionType,
```
egdbear, line 99: [question]: Do we need a full connection and pagination for the filter list? Could this just be a list of 7 items?

src/schema/v2/discoveryCategoriesConnection.ts
araujobarret: will give it a shot!

src/schema/v2/discoveryCategoriesConnection.ts
araujobarret: I think we do need, we must avoid big fetches, the clients should be able to control that. In Eigen we're going to do a smart paginated fetch(2 steps) to avoid a big fetch on a screen that makes it slow(it is already)

src/schema/v2/discoveryCategoriesConnection.ts
araujobarret: it looks way better, good call 👍  https://github.com/artsy/metaphysics/pull/7054/commits/4fa1a70766c02cbeb9a42f2c41574ab9298c8448

ArtsyPrivateDanger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Warnings</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  The V2 schema in this PR has breaking changes with Force. Remember to update the Force schema if necessary.

Enum value 'DUPLICATE_IMAGE_FILENAMES' was removed from enum 'ArtworkImportError'
Enum value 'ARTSY_EXPRESS' was removed from enum 'FulfillmentOptionInputEnum'
Enum value 'ARTSY_STANDARD' was removed from enum 'FulfillmentOptionInputEnum'
Enum value 'ARTSY_WHITE_GLOVE' was removed from enum 'FulfillmentOptionInputEnum'
Enum value 'ARTSY_EXPRESS' was removed from enum 'FulfillmentOptionTypeEnum'
Enum value 'ARTSY_STANDARD' was removed from enum 'FulfillmentOptionTypeEnum'
Enum value 'ARTSY_WHITE_GLOVE' was removed from enum 'FulfillmentOptionTypeEnum'
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0695257dff7fcbde0eaeb1bef0fbbc6a5ae8dbac
</p>

araujobarret:
@egdbear are we good to go? 😬

_schemaV2.graphql
```diff
@@ -11054,13 +11147,30 @@ type DiscoveryCategoriesConnectionEdge {
 }
 
 # A category for discovering and browsing art
-type DiscoveryCategory {
+type DiscoveryCategory implements Node {
```
egdbear, line 11150: [question]: Is this a breaking change?

araujobarret: Usually not, I double-checked connecting Eigen with the older schema to my local server(with these changes), and no error pops up, I believe we're good to go. will keep an eye on Sentry anyways.

PR #7058, opened 2025-09-02
feat(EMI-2793): fix rich phone number region code

rquartararo:
If `regionCode` is provided from the client, use this code directly instead of getting the region code from the phone number.

### DIFF
--- a/src/schema/v2/order/__tests__/MeOrder.test.ts
+++ b/src/schema/v2/order/__tests__/MeOrder.test.ts
@@ -635,7 +635,7 @@ describe("Me", () => {
       })
 
       it("returns phoneNumber with rich values with a valid region code", async () => {
-        orderJson.buyer_phone_number_country_code = "us"
+        orderJson.buyer_phone_number_country_code = "ca"
         orderJson.buyer_phone_number = "7738675309"
         context = {
           meLoader: jest.fn().mockResolvedValue({ id: "me-id" }),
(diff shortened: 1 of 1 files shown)


src/schema/v2/order/__tests__/MeOrder.test.ts
```diff
@@ -635,7 +635,7 @@ describe("Me", () => {
       })
 
       it("returns phoneNumber with rich values with a valid region code", async () => {
-        orderJson.buyer_phone_number_country_code = "us"
```
chr-tatu, line 638: great idea to change to "ca"!

src/schema/v2/phoneNumber.ts
```diff
@@ -101,7 +102,10 @@ export const PhoneNumberType: GraphQLObjectType<
     regionCode: {
       description: "Two-letter region code (ISO 3166-1 alpha-2)",
       type: GraphQLString,
-      resolve: ({ parsedPhone, phoneUtil }) => {
+      resolve: ({ regionCode, parsedPhone, phoneUtil }) => {
+        if (regionCode) {
+          return regionCode.toLowerCase()
```
chr-tatu, line 107: is `toLowerCase()` really needed here?

rquartararo: You're right, we should always receive the string from the client in lowercase. I will remove!

rquartararo: Actually, I'm seeing some previously written tests where `phoneNumberCountryCode` is capitalized, which makes me think there are some instances where we do need to transform it (Google libphonenumber is uppercase). I will keep it for backwards compatibility.

PR #7053, opened 2025-08-29
chore: AMBER-1779 - adding more free text fields

Mitch-Henson:
Solves [AMBER-1779]

Connects the fields that were added in https://github.com/artsy/gravity/pull/19316

cc: @artsy/amber-devs 

[AMBER-1779]: https://artsyproduct.atlassian.net/browse/AMBER-1779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7059, opened 2025-09-03
chore: AMBER-1779 - adding bibliography

Mitch-Henson:
This was forgotten in https://github.com/artsy/metaphysics/pull/7053 as it was added in https://github.com/artsy/gravity/pull/19316

cc: @artsy/amber-devs


Mitch-Henson:
Merging to get unblocked with some Volt changes

PR #7064, opened 2025-09-04
chore(EMI-2767): use Impulse unread by partner

leamotta:
https://artsyproduct.atlassian.net/browse/EMI-2767

Rely on Impulse to say if the conversation has been read by the Partner or not, instead of doing the logic again in MP.

Should be merged **_AFTER_** the Impulse migration is successfully run https://github.com/artsy/impulse/pull/951/


lidimayra (approved):
Niice! ⭐

PR #7071, opened 2025-09-05
test: force schema mismatch, danger should leave comment

amonkhouse:
checking to see if danger leaves a comment...

PR #7065, opened 2025-09-04
chore: remove last stiched type from gravity

egdbear:
### Description
This PR removes the final stitched UserAddress type from the Gravity-stitched schema in Metaphysics and defaults behavior to the resolvers. It also removes the feature flag and any conditional logic that previously toggled stitching on/off.

### What changed
Removed `USE_UNSTITCHED_USER_ADDRESS` env flag from src/config.ts and package.json scripts.
- Stop extending Me.addressConnection and UserAddress.id via stitching.
- Remove stitched resolvers and extension SDL for UserAddress.
- Removed duplicated and excluded type/mutation lists that were previously gated by the flag. 

/cc @artsy/diamond-devs

PR #7072, opened 2025-09-05
feat: add github workflow to compare force schema

amonkhouse:
Adds reusable workflow to run `compareForceSchema` checks via danger. The intent is to replace the current Peril checks.

Looking at [peril-settings](https://github.com/artsy/peril-settings/blob/main/peril.settings.json#L21-L22) it seems we are only running the `compareForceSchema` checks and not `dangerfile.ts`. Please let me know if I'm mistaken.

I set `fail-on-errors` to false, based off behaviours I could see in other PRs. For example, [this PR](https://github.com/artsy/metaphysics/pull/7065) failed the comparison but the check was marked as successful (apologies if this changes by the time you read this!).

**Testing**
[Here](https://github.com/artsy/metaphysics/pull/7071) is an example of a message being left when there's a breaking change.

### DIFF
--- a/.github/workflows/run-danger.yml
+++ b/.github/workflows/run-danger.yml
@@ -9,6 +9,6 @@ jobs:
     secrets:
       danger-token: ${{ secrets.DANGER_TOKEN }}
     with:
-      dangerfile: "peril/compareForceSchema.ts"
+      dangerfile: "danger/compareForceSchema.ts"
       install-from-caller: true
       fail-on-errors: false
--- a/.github/workflows/run-danger.yml
+++ b/.github/workflows/run-danger.yml
@@ -1,0 +1,14 @@
+name: ☢️ Run Danger
+
+on:
+  pull_request:
+
+jobs:
+  run-danger:
+    uses: artsy/duchamp/.github/workflows/run-danger.yml@main
+    secrets:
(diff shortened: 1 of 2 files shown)


.github/workflows/run-danger.yml
```diff
@@ -0,0 +1,14 @@
...
+on:
+  pull_request:
+
+jobs:
+  run-danger:
+    uses: artsy/duchamp/.github/workflows/run-danger.yml@main
+    secrets:
+      danger-token: ${{ secrets.DANGER_TOKEN }}
+    with:
+      dangerfile: "peril/compareForceSchema.ts"
```
dblandin: **Small suggestion (non-blocking)**: maybe we could change this directory to `danger/` while we're making this change?

amonkhouse: oop good spot! I'll make that change

amonkhouse: To avoid having to make some short-term changes to `peril-settings` I'm actually going to duplicate that file for now.

amonkhouse: Done, I'll wipe the **peril** folder when we turn off Peril. Having another version of the file in the right place will make this migration easier as I won't have to make sure I rename this folder and turn off Peril at the same time.

PR #7086, opened 2025-09-11
feat(EMI-2423): add support for artsy shipping options

fladson:
Expanding the Fulfillment options to match what we have at https://github.com/artsy/exchange/blob/main/app/services/fulfillment_option_service.rb#L2

@artsy/emerald-devs


src/schema/v2/order/types/sharedOrderTypes.ts
```diff
@@ -127,6 +127,9 @@ const FulfillmentOptionTypeEnum = new GraphQLEnumType({
     INTERNATIONAL_FLAT: { value: "INTERNATIONAL_FLAT" },
     PICKUP: { value: "PICKUP" },
     SHIPPING_TBD: { value: "SHIPPING_TBD" },
+    ARTSY_STANDARD: { value: "ARTSY_STANDARD" },
```
chr-tatu, line 130: Don't we have to add all of these? https://github.com/artsy/exchange/blob/5363df3abdc4e20d6907cd35d8cf1e9246cfa747/app/models/shipping_quote.rb#L19-L33

Just a question - I'm less familiar with arta.

fladson: I am basing that on the code execution for the new checkout, which is relying on `FulfillmentOptionService::SHIPPING_RATE_TYPES`, but we may need to add those in the future.

PR #7089, opened 2025-09-11
fix(EMI-2813): update me mutation phone number display format and region code

rquartararo:
https://artsyproduct.atlassian.net/browse/EMI-2813

- Fixes the user profile update mutation so that when a user provides their phone number with the rich phone input that contains a country code (like "ca" for Canada), the system now returns the correct region code (not "us").
- Added a `phoneNumberParced` field for saved addresses containing parsed phone number information, allowing us to query `display` phone number format for a users saved addresses.

### DIFF
--- a/src/schema/v2/me/userAddress/userAddress.ts
+++ b/src/schema/v2/me/userAddress/userAddress.ts
@@ -7,6 +7,7 @@ import {
 } from "graphql"
 import { IDFields } from "../../object_identification"
 import { ResolverContext } from "types/graphql"
+import { PhoneNumberType, resolvePhoneNumber } from "../../phoneNumber"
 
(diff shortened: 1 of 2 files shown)


src/schema/v2/me/userAddress/userAddress.ts
```diff
@@ -53,6 +54,15 @@ export const UserAddressType = new GraphQLObjectType<any, ResolverContext>({
       description: "Phone number",
       resolve: ({ phone_number }) => phone_number,
     },
+    phoneNumberParsed: {
```
rquartararo, line 57: @chr-tatu I added this as a separate field so we do not create any breaking changes in force. This will allow us to receive `display` phone numbers and tell if they are valid or not for saved addresses.

PR #7094, opened 2025-09-14
chore: add phoneNumberParsed to stitched user address

rquartararo:
Adds `phoneNumberParsed` to the stitched user address, allowing the client to query if user phone numbers are valid and get their `display` format.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -25015,6 +25015,9 @@ type UserAddress {
   # Phone number country code
   phoneNumberCountryCode: String
 
+  # Phone number with parsing and validation details
+  phoneNumberParsed: PhoneNumberType
+
   # Postal Code
   postalCode: String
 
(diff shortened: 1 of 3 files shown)


egdbear:
👋 
[question]: not sure I fully get it, why do we need the stitched version/resolver here?

rquartararo:
@egdbear Honestly I'm kinda confused as well. I added `phoneNumberParsed` in https://github.com/artsy/metaphysics/pull/7089, but the graghql schema did not update, which I assumed was due to `USE_UNSTITCHED_USER_ADDRESS=false` and when I added it to the stitched version, it updated. Is there a better way to add a new feild to the user address?

egdbear (approved):
> @egdbear Honestly I'm kinda confused as well. I added `phoneNumberParsed` in #7089, but the graghql schema did not update, which I assumed was due to `USE_UNSTITCHED_USER_ADDRESS=false` and when I added it to the stitched version, it updated. Is there a better way to add a new feild to the user address?

Right! Sounds like everything is working as expected! [This PR](https://github.com/artsy/metaphysics/pull/7065
) should already be merged, and once that’s in, things should work. I don’t want to block this, so as soon as this PR is merged, I’ll go ahead and merge the next one.

PR #7104, opened 2025-09-18
feat: add summaryCheckmarks field to CollectorProfileType

oxaudo:
Thinks this is the best place to put this logic of selecting up to 3 most applicable summary checkmarks for a collector profile and artwork of interest. 

@starsirius - not exactly sure how we'll use it in impulse though as we'll need to make a separate call to MP for this I believe? Maybe we can pair on that part when you get a chance?

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -5365,6 +5365,12 @@ type CollectorProfileType implements Node {
   savedArtworksCount: Int!
   selfReportedPurchases: String
 
+  # Up to three checkmark strings describing the collector in relation to the artwork/partner.
+  summaryCheckmarks(
+    # This can be specified, and is injected in a conversation context for convenience.
+    artworkID: String
+  ): [String]
+
   # An artwork-specific paragraph describing the collector.
   summaryParagraph(
     # This can be specified, and is injected in a conversation context for convenience.
@@ -14147,6 +14153,12 @@ type InquirerCollectorProfile {
   ): String
   savedArtworksCount: Int!
   selfReportedPurchases: String
+
+  # Up to three checkmark strings describing the collector in relation to the artwork/partner.
+  summaryCheckmarks(
+    # This can be specified, and is injected in a conversation context for convenience.
+    artworkID: String
+  ): [String]
 
   # An artwork-specific paragraph describing the collector.
   summaryParagraph(
--- a/src/schema/v2/CollectorProfile/collectorProfile.ts
+++ b/src/schema/v2/CollectorProfile/collectorProfile.ts
@@ -249,6 +249,91 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
       return paragraph
     },
   },
+  summaryCheckmarks: {
+    type: new GraphQLList(GraphQLString),
+    description:
+      "Up to three checkmark strings describing the collector in relation to the artwork/partner.",
+    args: {
+      artworkID: {
+        type: GraphQLString,
+        description:
+          "This can be specified, and is injected in a conversation context for convenience.",
+      },
+    },
+    resolve: async (
+      { id: collector_profile_id, artworkID },
+      { artworkID: artworkIDFromArgs },
+      { collectorProfileSummaryLoader }
+    ) => {
+      if (!collectorProfileSummaryLoader) {
+        throw new Error("You must be signed in to perform this action.")
+      }
+
+      const { raw_attributes = {} } = await collectorProfileSummaryLoader({
+        artwork_id: artworkIDFromArgs || artworkID,
+        collector_profile_id,
+      })
+
+      const results: string[] = []
+      const checks = [
+        {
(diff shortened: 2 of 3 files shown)


src/schema/v2/CollectorProfile/collectorProfile.ts
```diff
@@ -249,6 +249,91 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
...
+      let i = 0
+      while (results.length < 3 && i < checks.length) {
+        if (checks[i].flag) {
+          results.push(checks[i].text)
+        }
+        i++
+      }
+
+      // Recent sign up status (last in priority). If not recent => Active user; if recent => New user
+      if (results.length < 3) {
```
oxaudo, line 325: Not sure if this should only be inserted if there is no other checkmarks or if there is less than 3. Need to verify with Cami and Vivi - but either way super small change.

oxaudo: @starsirius - refactored some stuff here to prepend User info in case we have less than 3 attributes returned. So the logic is as specked in Figma after all.

src/schema/v2/CollectorProfile/collectorProfile.ts
```diff
@@ -249,6 +249,90 @@ export const CollectorProfileFields: GraphQLFieldConfigMap<
       return paragraph
     },
   },
+  summaryAttributes: {
+    type: new GraphQLList(GraphQLString),
```
mzikherman, line 253: I wonder if this should be structured more, vs. an array of strings. This almost feels like 'badges' - what if the UI/design changes and instead of this exact string being needed we represent this differently, or the text itself needs to change based on the client, etc.

Maybe an example of this kind of more structured data is artist insights? - https://github.com/artsy/metaphysics/blob/8f6c0a43eb9c87e5c4e7d6a642f8ea9882f7fe60/src/schema/v2/artist/insights.ts#L28

So we'd have a first-class `CollectorProfileInsights` or `CollectorProfileSummary` type that this would be a (non-null) list of?

_schemaV2.graphql
```diff
@@ -5365,6 +5365,12 @@ type CollectorProfileType implements Node {
   savedArtworksCount: Int!
   selfReportedPurchases: String
 
+  # Up to three checkmark strings describing the collector in relation to the artwork/partner.
+  summaryAttributes(
+    # This can be specified, and is injected in a conversation context for convenience.
+    artworkID: String
+  ): [String]
```
starsirius, line 5372: Asked a [question](https://artsy.slack.com/archives/CP9P4KR35/p1758224720311189) re how Impulse as a trusted app accesses MP for multiple upstream APIs (in our case, Exchange and Gravity).

Before we convert all queries to MP, we can start by querying MP for this field specifically. There seems to be a few options:

First one is through conversations, like how CMS queries for summary paragraph:

```graphql
query {
  conversation(id: "123") {
    collectorResume {
      collectorProfile {
        summaryParagraph
      }
    }
  }
}
```

The query comes back to Impulse itself for conversations (which feels a bit weird), and we might need to update Impulse to accept the queries from trusted apps.

Second one is through the root collector profile, and I think it's also doable.

```graphql
query {
  collectorProfile(userID: "abc") {
    summaryParagraph(artworkID: "def")
  }
}
```

I might need to dig deeper but it _seems_ the queries from the [existing Metaphysics client](https://github.com/artsy/impulse/blob/e120c19be41a8c540ae6406082e4b66b7274cab1/app/clients/metaphysics.rb) are _accepted_ by Gravity, even tho the [Xapp-Token was for Exchange](https://github.com/artsy/impulse/blob/e120c19be41a8c540ae6406082e4b66b7274cab1/app/clients/metaphysics.rb#L17). If not, I think we can still update the client to send proper Xapp-Token for Gravity somehow.

src/schema/v2/CollectorProfile/collectorProfile.ts
oxaudo: @mzikherman - very valid. My thinking here was that sounds like we are actively rethinking what/how this info is going to be presented to partners. And for now here is a `paragraph` that essentially the same data but in the 'sentence' form already exists on this 'collector profile type' so I went with the field parallel to the paragraph one and as for now we literally need just 'ordered array of strings' - just returning that. But ultimately - agree - more structured ` CollectorProfileSummary` or even something that specifies that it's `PartnerFacingCollectorProfile` summary probably makes sense. Just not sure we know enough now to figure out that should be included there.

oxaudo: @mzikherman - think it's ok if we go with this simple version for now and figure out more structured data with the next follow up?

mzikherman: Definitely! Might be a nice cleanup when we're ready.

PR #7118, opened 2025-09-24
fix(DIAM-91): show is displaying different/incorrect dates (using mobile web Artsy)

lilyfromseattle:
The type of this PR is: FIX

This PR solves [DIAM-91](https://artsyproduct.atlassian.net/browse/DIAM-91?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

I first tried to solve this in force but it was ultimately a metaphysics issue

@artsy/diamond-devs
Description

Steps to reproduce:
Pick any show from the [/shows](https://www.artsy.net/shows) page
Navigate to the show, and then to its partner page
Refresh the partner page (← important because we start with a full ssr page load here)
You will prob see the show in question, listed here as a current or upcoming show
Note the start/end dates for the show — they are likely correct
Now click on the View More link to navigate back to the show (← this is a client-side navigation)
Note the start/end dates for the show — they are likely incorrect, one day too early
Refresh the show page (← another full ssr page)
Note the start/end dates for the show — they are likely correct again

Can reproduce:
Yes

URL Affected:
https://www.artsy.net/show/bowery-gallery-rachel-siporin-bathers

Anything else we should know?

Reporter:

Lake <[lake.barrett@artsymail.com>](mailto:lake.barrett@artsymail.com)

https://artsy.slack.com/archives/C07PRTJSD6G/p1756405561749169

Before:
https://github.com/user-attachments/assets/9f0ed9da-2a9a-4bf2-bb2e-f2dfa6dc6e71

After:
https://github.com/user-attachments/assets/dd95b733-8cef-4460-8316-b6300a1e7e8c

[DIAM-91]: https://artsyproduct.atlassian.net/browse/DIAM-91?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -22222,9 +22222,6 @@ type Show implements EntityWithFilterArtworksConnectionInterface & Node {
   exhibitionPeriod(
     # Formatting option to apply to exhibition period
     format: ExhibitionPeriodFormat = LONG
-
-    # A tz database time zone, otherwise falls back to "X-TIMEZONE" header. See http://www.iana.org/time-zones, https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
-    timezone: String
   ): String
 
   # If the show is in a Fair, then that fair
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -22222,6 +22222,9 @@ type Show implements EntityWithFilterArtworksConnectionInterface & Node {
   exhibitionPeriod(
     # Formatting option to apply to exhibition period
     format: ExhibitionPeriodFormat = LONG
+
+    # A tz database time zone, otherwise falls back to "X-TIMEZONE" header. See http://www.iana.org/time-zones, https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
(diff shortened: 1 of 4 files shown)


anandaroop (commented):
:ceiling-cat: This looks promising but I'd feel better about the test if it better reflected the upstream data coming from Gravity.

https://github.com/artsy/metaphysics/blob/1e1494fab60c68fc60f8ad138e96bedb61654483/src/schema/v2/__tests__/show.test.js#L17-L18

Note how that is all mocked as if timestamps are at noon, whereas CMS/Gravity round these to midnight:
Screenshot 2025-09-24 at 4 52 36 PM

If you adjust the mock data to match, does the updated logic still pass?

(The test code is obv much older than your fix, but it might be a factor in why this went undetected for so long.)

PR #7120, opened 2025-09-25
feat(genomebot): enable impression tracking on two more home view artwork rails

anandaroop:
In order to have useful metrics for the genomebot rollout we want to start tracking item impressions for the Eigen home view rails that rely on genome data.

Since Eigen [already requests](https://github.com/artsy/eigen/blob/d3a75dae627a70c472c12ab6bcd7648391afba23/src/app/Scenes/HomeView/Sections/HomeViewSectionArtworks.tsx#L155) the `trackItemImpressions` field in its artwork section component, all that's needed is to set that flag to `true` for the sections in question.


nickskalkin:
Confirmed that it works on staging. Push schema changes step fails though: https://app.circleci.com/pipelines/github/artsy/metaphysics/28959/workflows/251ab894-932c-45e7-849f-4c3df7aa8f63/jobs/58859

PR #7124, opened 2025-09-29
chore: add collector profile endpoints to non-authenticated loaders

ansor4:
### Description

Part of the flow that will support [EMI-2804](https://artsyproduct.atlassian.net/browse/EMI-2804)

#### Changes
- Adds `collector_profiles` and `collector_profile_summary` endpoints to `loaders_without_authentication`.

Allows for requests for `collectorProfile` and `summaryAttributes` to be called without `X-ACCESS-TOKEN`. Instead, we'll pass `X-XAPP-TOKEN` and gate on gravity's side.

Screenshot 2025-09-29 at 10 25 58 AM
Screenshot 2025-09-29 at 10 26 26 AM

[EMI-2804]: https://artsyproduct.atlassian.net/browse/EMI-2804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/loaders/loaders_without_authentication/gravity.ts
+++ b/src/lib/loaders/loaders_without_authentication/gravity.ts
@@ -76,6 +76,12 @@ export default (opts) => {
       { headers: true }
     ),
     collectionLoader: gravityLoader((id) => `collection/${id}`),
+    collectorProfilesLoader: gravityLoader(
+      "collector_profiles",
+      {},
+      { headers: true }
+    ),
+    collectorProfileSummaryLoader: gravityLoader("collector_profile_summary"),
     createInvoicePaymentLoader: gravityLoader(
       (id) => `invoice/${id}/payment`,
       {},


src/lib/loaders/loaders_without_authentication/gravity.ts
```diff
@@ -76,6 +76,12 @@ export default (opts) => {
       { headers: true }
     ),
     collectionLoader: gravityLoader((id) => `collection/${id}`),
+    collectorProfilesLoader: gravityLoader(
```
mzikherman, line 79: One of the few occurrences where we add this to the _without_ authentication to accomplish the task!

(usually it's b/c we're not personalizing or authorizing properly and respecting the current user, so the fix is typically to add the same-named loader to the _with_ authentication file! This of course makes sense b/c the use-case is a trusted app and not current user)

mzikherman (approved):

src/lib/loaders/loaders_without_authentication/gravity.ts
ansor4: yeah it felt....weird to add it here, thanks for the sanity check!

PR #7130, opened 2025-09-30
fix: Adds artlogic, artcloud, and batch upload to import source types

jpotts244:
fix: Adds `artlogic`, `artcloud`, and `BATCH_UPLOAD` to import source types
Resolves a portion of [AMBER-2032](https://artsyproduct.atlassian.net/browse/AMBER-2032)

This is needed as Volt relies on `ArtworkImportSource` Enum type for import source related filtering

Examples

## Before
Note `importSource` is `null` despite having import sources in the DB and thus hiding the Import source filter 🐛 

https://github.com/user-attachments/assets/717a6eaf-1f36-4b98-ac70-0b66e8cfe97d

## After
Note `importSource` value is present and filtering works as expected ✅ 

https://github.com/user-attachments/assets/282e3052-d079-4fd2-bd15-cea6782f0219

cc @artsy/amber-devs 

[AMBER-2032]: https://artsyproduct.atlassian.net/browse/AMBER-2032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -3449,6 +3449,9 @@ type ArtworkImportRowImage {
 }
 
 enum ArtworkImportSource {
+  ARTCLOUD
+  ARTLOGIC
+  BATCH_UPLOAD
   CONVECTION
   MY_COLLECTION
 }
--- a/src/schema/v2/artwork/index.ts
+++ b/src/schema/v2/artwork/index.ts
@@ -112,6 +112,9 @@ const has_multiple_editions = (edition_sets) => {
 const IMPORT_SOURCES = {
   CONVECTION: { value: "convection" },
   MY_COLLECTION: { value: "my collection" },
+  ARTLOGIC: { value: "artlogic" },
+  ARTCLOUD: { value: "artcloud" },
+  BATCH_UPLOAD: { value: "BATCH_UPLOAD" },
 } as const
 
 const ARTIST_IN_HIGH_DEMAND_RANK = 9


Mitch-Henson:
Wanted to highlight this recent Gravity PR though: https://github.com/artsy/gravity/pull/19395 as it may add new types of `import_source` values. The ACE team adds in some values sometimes, e.g. `ACE - batch - vip` so it would end up being `BATCH_UPLOAD: ACE - batch - vip`, amongst other potential values. Not sure if you want to show these as filterable values, but wanted to let you know in any case!

lidimayra (approved):
Niiice!! ⭐ 

And good shout from @Mitch-Henson on the other import source types.
I'd say it's fine to not take those into account for the filter dropdown.

The first time @olerichter00 worked on this filter (https://github.com/artsy/volt/pull/8974), we also found thousands of records in Redshift containing other import source values (such as `Salesforce`). The decision by then was to focus only on the source types that bring value to partners. The list here matches the same definition [added to Volt](https://github.com/artsy/volt/blob/3003608cc8ba18c95463e73268bbd48888b9af5c/app/javascript/components/ArtworkFilter/components/ArtworkFilter/filters/ArtworkFilterImportSource.tsx#L13-L34) by then, so LGTM!

jpotts244:
Nice thanks for the context @Mitch-Henson 👀 , good thing to keep in mind

PR #7137, opened 2025-10-03
chore: upgrade to yarn4

ovasdi:
This upgrades to yarn4 following [force](https://github.com/artsy/force/pull/16096/files), [volt](https://github.com/artsy/volt/pull/9905/files) examples and also this [upgrade playbook](https://www.notion.so/artsy/Upgrading-to-yarn-4-and-security-upgrades-for-repos-using-npm-281cab0764a080cfab5df82b902f7287)


Dockerfile
ovasdi, line 1: Problem: Yarn v4 removed several CLI flags and changed how production
  dependencies are handled.

  Changes:
  - Added config files: `.yarnrc.yml` and `.yarn/` directory now copied to
  builder stage (required for Yarn 4 to function)
  - `--frozen-lockfile` → `--immutable`: Yarn 4's new flag for CI installs that
  ensures lockfile isn't modified
  - Removed deprecated flags: `--production`, `--quiet`, `--force`
  - New production deps approach: Instead of `yarn install --production`, now
  uses `yarn workspaces focus --all --production` to create production-only
  node_modules _after_ the build stage
  - Simplified copy: Changed from `/opt/node_modules.prod` to
  `/app/node_modules` since we no longer need to preserve separate dev/prod
  modules during multi-stage build

src/config.ts
ovasdi, line 1: This is to address CI failures when running `ensure-schema-update` job.

Problem: Schema dumping requires building the GraphQL schema (to serialize it), which [loads](https://github.com/artsy/metaphysics/blob/bb3bd90a628306fbd2be1c30bc0a2b60c26e025d/scripts/dump-schema.ts#L12) `src/config.ts` via that import, which validates that `mustHave` ENV vars are present. But schema dumping never actually uses those credentials - it's just introspecting type definitions. CI failed because credentials like GRAVITY_ID, GRAVITY_SECRET, etc. weren't available. This has been [silently failing](https://app.circleci.com/pipelines/github/artsy/metaphysics/29039/workflows/730cc1b0-d821-4a3b-9ce6-53f6babf0c3d/jobs/59006?invite=true#step-109-186_79) in CI (without causing a job failure) since secrets migration and while not totally clear it might be because yarn v4 changed how script failures manifest and how exit codes are exposed to parent processes.

After confirming the issue was missing vars, solution:
  - Added SKIP_CONFIG_VALIDATION env var check in src/config.ts
  - Injected SKIP_CONFIG_VALIDATION: "true" in
  scripts/dump-staging-schema.js
  - Config validation now skipped when dumping schema, but still enforced
  for actual server startup

Why this approach: Schema generation is purely about type definitions and doesn't need runtime API credentials. Instead of adding a context in CI or complex logic just skip the check.

package.json
```diff
@@ -6,20 +6,20 @@
...
-    "build:lib": "babel src --copy-files --extensions '.ts,.js' --ignore src/test,src/integration,src/**/__tests__ -s inline -d build/src",
+    "build:lib": "babel src --copy-files --extensions '.ts,.js' --ignore 'src/test,src/integration,src/**/__tests__' -s inline -d build/src",
     "build:remote-schemas": "cp src/data/*.graphql build/src/data/",
     "build": "yarn build:lib && yarn build:index && yarn build:fixtures && yarn build:remote-schemas",
     "ci": "yarn test",
     "deploy-cloudflare-workers:staging": "yarn wrangler deploy --keep-vars --config=workers/caching/metaphysics-cdn.toml --env staging",
     "deploy-cloudflare-workers": "yarn wrangler deploy --keep-vars --config=workers/caching/metaphysics-cdn.toml --env production",
     "dev": "DEBUG=info,warn,error babel-node --extensions '.ts,.js' --inspect index.js",
     "dump-schema": "babel-node --extensions '.ts,.js' ./scripts/dump-schema.ts",
-    "dump:local": "yarn dump-schema _schemaV2.graphql & wait",
```
ovasdi: Not sure why but after yarn upgrade this was [failing](https://app.circleci.com/pipelines/github/artsy/metaphysics/29047/workflows/fbd08572-75e6-43dc-8d74-fe60f07c48f7/jobs/59019?invite=true#step-109-337_23). Decided to just remove the `& wait`

src/config.ts
ovasdi, line 1: This is to address CI failures when running `ensure-schema-update` job.

Problem: Schema dumping requires building the GraphQL schema (to serialize it), which [loads](https://github.com/artsy/metaphysics/blob/bb3bd90a628306fbd2be1c30bc0a2b60c26e025d/scripts/dump-schema.ts#L12) `src/config.ts` via that import, which validates that `mustHave` ENV vars are present. But schema dumping never actually uses those credentials - it's just introspecting type definitions. CI failed because credentials like GRAVITY_ID, GRAVITY_SECRET, etc. weren't available. This has been [silently failing](https://app.circleci.com/pipelines/github/artsy/metaphysics/29039/workflows/730cc1b0-d821-4a3b-9ce6-53f6babf0c3d/jobs/59006?invite=true#step-109-186_79) in CI (without causing a job failure) since secrets migration and while not totally clear it might be because yarn v4 changed how script failures manifest and how exit codes are exposed to parent processes.

After confirming the issue was missing vars, solution:
  - Added SKIP_CONFIG_VALIDATION env var check in src/config.ts
  - Injected SKIP_CONFIG_VALIDATION: "true" in
  scripts/dump-staging-schema.js
  - Config validation now skipped when dumping schema, but still enforced
  for actual server startup

Why this approach: Schema generation is purely about type definitions and doesn't need runtime API credentials. Instead of adding a context in CI or complex logic just skip the check.

package.json
```diff
@@ -6,24 +6,24 @@
...
     "deploy-cloudflare-workers": "yarn wrangler deploy --keep-vars --config=workers/caching/metaphysics-cdn.toml --env production",
     "dev": "DEBUG=info,warn,error babel-node --extensions '.ts,.js' --inspect index.js",
     "dump-schema": "babel-node --extensions '.ts,.js' ./scripts/dump-schema.ts",
-    "dump:local": "USE_UNSTITCHED_ACCEPT_PARTNER_AGREEMENT=false yarn dump-schema _schemaV2.graphql & wait",
+    "dump:local": "USE_UNSTITCHED_ACCEPT_PARTNER_AGREEMENT=false yarn dump-schema _schemaV2.graphql",
     "dump:staging": "node scripts/dump-staging-schema.js",
     "lint": "eslint . --ext ts",
     "lint:fix": "eslint . --fix --ext ts",
-    "prepare": "patch-package",
+    "postinstall": "patch-package",
```
ovasdi, line 26: This was necessary in order to apply `patch-package` following `yarn install`, otherwise many type errors will surface.
It appears that `prepare` was removed in yarn v2+ and while the [docs](https://yarnpkg.com/advanced/lifecycle-scripts#postinstall) recommend against using `postinstall` lifecycle script, maybe its OK to keep this for now. 🤔 

Yarn has native [package patching](https://yarnpkg.com/features/patching) so perhaps that is a worthy follow-up, to migrate away from using `postinstall` to apply patches and to yarn native package patching.

package.json
ovasdi: Had to do this a second time after merging main into this branch. https://github.com/artsy/metaphysics/pull/7137/commits/f02ea902eaf940d457c2b277513d12b36a6a89ed

ovasdi:
@brainbicycle wanted to ask you for review of this in case you have some insight or feedback seeing as you've done a yarn upgrade yourself.

.yarn/releases/yarn-4.10.3.cjs
mc-jones, line 1: I'm not super familiar with the yarn 4 architecture, so maybe a silly question, but should we just `.gitignore` the entire `.yarn` directory?

.yarnrc.yml
```diff
@@ -0,0 +1,9 @@
+nodeLinker: node-modules
+
+npmMinimalAgeGate: 1440 # 1 day in minutes (24 * 60)
```
mc-jones, line 3: Any specific reason for this value? i.e. why not 2 days or 7?

Dockerfile
```diff
@@ -12,20 +12,23 @@ RUN apk --no-cache --quiet add \
...
+  yarn cache clean
 
 # Copy application code
 COPY  . ./
 
 RUN yarn build
 
+# Create production-only node_modules
+RUN rm -rf node_modules && \
+  yarn workspaces focus --all --production
```
mc-jones, line 35: Will stripping out the dev dependencies for `builder-base` create issues for local dev, which [uses](https://github.com/artsy/metaphysics/blob/1a976f3012e0a24fad9ebfda81af36526201d6c3/hokusai/templates/docker-compose-service-dev.yml.j2#L3) this image?

mc-jones:
A few small Qs, but largely LGTM. What ended up being the type error issue?

.yarn/releases/yarn-4.10.3.cjs
damassi: This is by design; they add this for runtime consistency, after git clone -- this ensures, for certain, that all developers are running the same version of yarn.

ovasdi: @damassi do you know if that also applies to CI, unless a different version is specified via CI config that is. I see [here](https://app.circleci.com/pipelines/github/artsy/metaphysics/29590/workflows/4d717848-2a45-435b-bcda-c81be1759282/jobs/60115?invite=true#step-105-102_23) that CI appears to have just picked it up.

damassi: if its in the docker image, it should be good I think! (i never needed to think about this in force or volt, for example)

.yarnrc.yml
ovasdi: I copied the value from [eigen](https://github.com/artsy/eigen/blob/7b94e626277593d9450f80da62ed5e94fef9c04c/.yarnrc.yml#L3) and energy.

Dockerfile
ovasdi: I don't know if and what issue it would create as I didn't test that specifically but you did identify a regression from the old approach in that now dev image would end up with prod only `node_modules`. I will fix this to ensure that doesn't happen. Thanks!

ovasdi: Take a look at https://github.com/artsy/metaphysics/pull/7137/commits/14ff847cb45cc8a7954a853833907efc7efa84ef
I ended up adding a new stage, seemed like the neatest approach. What do you think?

ovasdi:
> What ended up being the type error issue?

@mc-jones the type errors I came to you about were a result of `prepare` lifecycle script getting removed in modern yarn versions (related [comment](https://github.com/artsy/metaphysics/pull/7137#discussion_r2494641908)) and thus `patch-package` not running. I over relied on claude to drive this one and it was happy to try and fix all the issues that patching resolves. So after discussing with you I decided to pause yarn4 and just disable running scripts under yarn1. Once I disabled the scripts and saw that the same type errors surfaced I was able to identify that `patch-package` was now not getting applied, thats when it occurred to me that perhaps it is what also happened under yarn 4, and it did, just for a different reason.

Side note that under yarn4 applied patch output during `yarn install` is not displayed as it was under yarn1. You must add `--inline-builds` option to see it.

.yarn/releases/yarn-4.10.3.cjs
mc-jones: Interesting yarn docs for suggestions on what to ignore and what the different sub-directories do [here](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored).

ArtsyPrivateDanger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Warnings</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>The V2 schema in this PR has breaking changes with Force. Remember to update the Force schema if necessary.

Type 'OfferExchangeError' was removed
Type 'OfferMutationError' was removed
Type 'OfferMutationResponse' was removed
Type 'OfferMutationSuccess' was removed
Type 'createBuyerOfferInput' was removed
Type 'createBuyerOfferPayload' was removed
Type 'updateBuyerOfferInput' was removed
Type 'updateBuyerOfferPayload' was removed
Field 'createBuyerOffer' was removed from object type 'Mutation'
Field 'updateBuyerOffer' was removed from object type 'Mutation'
Field 'order' was removed from object type 'Offer'</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 74a2a53f606ce0eba440ef5e619df8b67a00bd27
</p>

.yarn/releases/yarn-4.10.3.cjs
ovasdi: I followed the recommendations:
- [.gitignore (non zero-install)](https://github.com/artsy/metaphysics/pull/7137/commits/40219f86a073af742521be061c72e9c7a7527a72)
- [.gitattributes](https://github.com/artsy/metaphysics/pull/7137/commits/74a2a53f606ce0eba440ef5e619df8b67a00bd27)

ovasdi:
I am surprised to see these warnings https://github.com/artsy/metaphysics/pull/7137#issuecomment-3493042170 and https://github.com/artsy/metaphysics/pull/7137#issuecomment-3493043406 seeing as this PR did not modify the schema.

.yarn/releases/yarn-4.10.3.cjs
brainbicycle: fwiw we tried at first without this in eigen and it complicated our lives and then ended up adding it. https://github.com/nodejs/corepack in ci if you are fighting with versions in most cases you can just do corepack enable (or it will work automatically) and get the right version of yarn rather than having to manually install and specify a version.

ovasdi:
> I am surprised to see these warnings [#7137 (comment)](https://github.com/artsy/metaphysics/pull/7137#issuecomment-3493042170) and [#7137 (comment)](https://github.com/artsy/metaphysics/pull/7137#issuecomment-3493043406) seeing as this PR did not modify the schema.

OK I think I get why these warnings suddenly appeared. Its because schema in `main` advanced.

src/config.ts
brainbicycle: nice find 😬

brainbicycle (approved):
this looks good to me! props!

Dockerfile
mc-jones: I'm thinking we probably can just consolidate the `builder-production` and `production` images since I don't think the former would be used except for the yarn install. But LMK if I'm missing something. Happy to chat through too. 

```dockerfile
# ---------------------------------------------------------
# Release image
# ---------------------------------------------------------
#
# Release stage. This stage creates the final docker image that will be
# released. It contains only production dependencies and artifacts.
#
FROM node:22.5-alpine AS production

# moved these into this image and 
# deleted the `COPY --chown=deploy:deploy --from=builder-production /app/node_modules ./node_modules` below
RUN rm -rf node_modules && \
  yarn workspaces focus --all --production

WORKDIR /app

RUN apk --no-cache --quiet add \
  bash \
  dumb-init \
  git \
  && adduser -D -g '' deploy

RUN chown deploy:deploy $(pwd)

# Switch to deploy user
USER deploy

COPY --chown=deploy:deploy --from=builder-production /app/.circleci ./.circleci
COPY --chown=deploy:deploy --from=builder-production /app/build ./build
COPY --chown=deploy:deploy --from=builder-production /app/src/data ./src/data
COPY --chown=deploy:deploy --from=builder-production /app/scripts/load_secrets_and_run.sh load_secrets_and_run.sh

ENTRYPOINT ["/usr/bin/dumb-init", "./load_secrets_and_run.sh"]

CMD ["node", "--heapsnapshot-signal=SIGUSR2", "build/index.js"]
```

mc-jones (approved):
Thanks Ozzie!

Dockerfile
ovasdi: - so the `rm -rf node_modules && \` could be removed because there is no modules to delete at that point considering the `node:...` base
- we would also need to copy dependency manifests (along with .yarn) to production stage in order to run `yarn workspaces...`. so those files would need to get copied in each stage, right?
```
FROM node:22.5-alpine AS production
...
COPY --chown=deploy:deploy package.json yarn.lock .yarnrc.yml ./
COPY --chown=deploy:deploy .yarn ./.yarn
COPY --chown=deploy:deploy patches ./patches
...
```

I did briefly consider two stage approach but when I realized id have to COPY manifests twice, I opted for the intermediary stage. Feels like a pretty small difference, one where `production` stage is a bit faster and doesn't require additional COPYs and one where `production` stage is a bit slower but more self-contained 🤔

PR #7555, opened 2026-03-24
feat: add paginationResolverLazy for lazy totalCount requests

egdbear:
## Description

Adds `paginationResolverLazy` similar to `paginationResolver`, but with optional `totalCount`.

When `totalCount` is not provided, `hasNextPage` is inferred based on whether a full page of results is returned, avoiding the need for a count query.

---

## Context

This lays the foundation for skipping `total_count=true` on Gravity requests when clients don’t request `totalCount`.

Currently, every connection query triggers a MongoDB aggregation count on Gravity (uncached for authenticated users), even when the total count is not needed. This change helps reduce unnecessary load and improve performance.

---

## Behavior

- If `totalCount` **is provided**:
  - Behaves identically to `paginationResolver`

- If `totalCount` **is not provided**:
  - `hasNextPage` is inferred from result size
  - `pageCursors` returns `null`
  - No count query is required

---

## Tradeoffs

- Example with `first: 10` results (e.g., requesting 10 items and receiving exactly 10), the client will make one additional empty request before stopping

---

Follow-up PRs will incrementally adopt this, starting with the highest-traffic endpoints. 

Historical conversation https://github.com/artsy/metaphysics/issues/866


src/schema/v2/fields/pagination.ts
```diff
@@ -190,48 +190,66 @@ export function connectionWithCursorInfo(
...
-  page,
-  size,
-  body,
-  args,
-  resolveNode,
-}: {
-  /** Args object containing either `page/size` or `first/after` */
+interface PaginationParams<T> {
   args: ConnectionArguments
-  /** Returned from loader */
```
mzikherman, line 204: #minior these comments were kinda useful!

src/schema/v2/fields/pagination.ts
```diff
@@ -190,48 +190,66 @@ export function connectionWithCursorInfo(
...
 }
 
+export const paginationResolverLazy = <T>(
+  params: PaginationParams<T> & { totalCount?: number }
+) => {
+  const { totalCount, offset, page, size, body, args, resolveNode } = params
+
+  // When totalCount is unavailable, infer from result size:
+  // full page → assume more exist, partial page → last page
+  const arrayLength =
```
mzikherman, line 245: doube-nested ternary is a bit complex, maybe

``` ts
const hasMore = body.length === size
const arrayLength = totalCount ?? (hasMore ? offset + size + 1 : offset + body.length)
```

is better.

src/schema/v2/fields/pagination.ts
```diff
@@ -190,48 +190,66 @@ export function connectionWithCursorInfo(
...
+    totalCount != null
+      ? totalCount
+      : body.length === size
+      ? offset + size + 1
+      : offset + body.length
+
+  return {
+    totalCount: totalCount ?? null,
+    pageCursors:
+      totalCount != null ? createPageCursors({ page, size }, totalCount) : null,
```
mzikherman: page cursors shouldn't be nullable based on the typing in this file - my guess (speculation) is that many use cases for clients fetching connections - even if they aren't using `totalCount` - might be for both `pageInfo` (for `hasNextPage` and `endCursor` - inf. scroll pagination) and/or `pageCursors` (windowed pagination on web).

It looks like the `pageInfo` stuff works with the `arrayLength` (of course the caveat is if the total result set can be evenly split into exact pages), but what about the web use case of `pageCursors`?

For the POC, I would probably add support for that and even switch a connection (maybe one used on both Eigen/Force) to use this new resolver?

(for the web update, you would need to fetch total count if `pageCursors` is being selected - and can even throw in `totalCount` for now as well although we may not even need some of those on clients vs. `edges` existence - so it does mean the skipping of total count savings may only apply to Eigen/inf scroll style requests)

mzikherman (commented):
Biggest comment is to build it out a bit more - support `pageCursors` and `totalCount` being queried for, which I believe _would_ result in `total_count: true` needing to make it to the API request. (which I think also means the savings from this PR may more apply to Eigen/inf scroll style requests where `pageInfo` is what's being used, on web we typically _do_ use `pageCursors` for windowed pagination).

egdbear:
Added `needsTotalCount(info)` utility fn which checks the client's selection set, when `totalCount` or `pageCursors` is requested, we fetch the count from Gravity as before, otherwise we skip it. When the count isn't fetched, `pageCursors` returns a safe current page fallback instead of fake page numbers from synthetic data. 

Also migrated `savedArtworks` as the POC behind an `ENABLE_LAZY_TOTAL_COUNT` flag, and confirmed that there are no breaking changes in Eigen or Force.

egdbear:
https://github.com/artsy/metaphysics/issues/866

egdbear:
Will change this asap and use this against artworks/artists resolvers.

PR #7584, opened 2026-04-10
feat: catalog artwork documents

leamotta:
Support catalog artwork documents as seen in https://github.com/artsy/gravity/pull/19990

@artsy/amber-devs


lidimayra (approved):
🌟

MrSltun (approved):
🌟

PR #7590, opened 2026-04-10
chore: Include errorMessage among InstagramPost fields

lidimayra:
I just realised we're not returning potential error messages for Instagram posts.

It's not essential, since for user feedback we'll rely on the data coming from websockets rather than the mutation. Still, this is useful data for us (devs) to be able to fetch directly from graphl queries when testing / debugging issues, so it makes sense to have it exposed.

cc @artsy/amber-devs


olerichter00 (approved):
Great idea! 🙏

PR #7615, opened 2026-04-22
chore(EMI-3191): expose shipping radius

rquartararo:
Exposing `shipping_radius` to determine if we show "Additional processing times may vary by destination" on the client.

### DIFF
--- a/_schemaV2.graphql
+++ b/_schemaV2.graphql
@@ -19905,6 +19905,9 @@ type Order {
   # Display short version of order's artwork location
   shippingOrigin: String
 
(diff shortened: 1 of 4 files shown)


rquartararo:
@starsirius the updated [design](https://www.figma.com/design/kaBdUapJIWs9YL9ATlPDzw/Checkout-Redesign?node-id=10446-108024&m=dev) does not show any difference in timeframe copy between arta shipping types so this change is all we should need

erikdstock (approved):

PR #7617, opened 2026-04-23
chore: add a preview of a bulk update operation to the schema

mzikherman:
Based on https://github.com/artsy/gravity/pull/20059 - adds the 'preview' of a bulk update operation to the schema, currently returning the counts of artworks that will be edited/skipped in a selection.


mzikherman:
@mc-jones will merge this one in too.

